PR #3, opened 2025-10-09
Let the agent have the final say about which Catalog is used

jacobsimionato:
This tweaks the way catalog negotiation works between server and client in order to:
- Give the server the final say about which catalog to use in the case where there are multiple compatible catalogs shared between server and client
- Let the server use a different catalog in each surface if it wants to

## Before
1. Server advertises catalogs A, B, C and ability to accept a dynamic schema
2a. Client supports catalogs B, C, D, and it chooses catalog C by sending clientUiCapabilities message with a single Catalog
2b. Alternatively, client can choose to send a dynamic schema e.g. for a dev workflow

## After

1. Client advertises well-known catalogs B, C, D via clientUiCapabilities message, and dynamic catalog E, which it assigns a URI to
2a. Server sends beginRendering message which chooses catalog C, then sticks to that in all surface updates
2b. Server sends beginRendering message which specifies catalog E (the dynamic one)


docs/proposals/v0_8.md
```diff
@@ -287,28 +291,55 @@ The client sends messages to the server to report user actions, declare capabili
...
-          "description": "A URI pointing to a predefined component catalog schema."
-        },
-        "dynamicCatalog": {
-          "$ref": "https://raw.githubusercontent.com/google/a2ui/refs/heads/main/schemas/v0.1/catalog.json"
+        "catalogs": {
+          "type": "array",
+          "description": "An array of component catalogs supported by the client.",
+          "items": {
+            "type": "object",
+            "description": "A description of a supported catalog. Exactly one of 'catalogUri' or 'dynamicCatalog' must be provided.",
```
dmandar, line 302: why does it have to be only one? Can a client use a default catalog for one surface (genUI) and a custom one for another?

PR #27, opened 2025-10-17
docs: Update protocol docs and schema files

gspencergoog:
Update the protocol design doc to match the latest schema.

Update the schema files, eliminating duplicates and synchronizing with the docs.


README.md
```diff
@@ -1,19 +1,19 @@
-# A2UI (Generative UI Language Format) Protocol
+# A2UI (Agent to UI) Protocol
```
holtskinner, line 1: Should this be Agent2UI to match Agent2Agent?

docs/a2ui_protocol.md
```diff
@@ -158,29 +159,28 @@ A2UI's component model is designed for flexibility, separating the protocol from
...
           }
         }
       },
       {
         "id": "another-component-id",
-        "componentProperties": { ... }
+        "component": { ... }
       }
     ]
   }
```
holtskinner, line 185: For all of these JSON imports, if they are just being pulled in from the `specification.json` files, I'd recommend using mkdocs snippets to prevent issues with syncing.

See this example in the A2A Protocol doc https://github.com/a2aproject/A2A/blob/a2de7981cadeaa5197bee56cbd6ab7b2c5da2541/docs/specification.md?plain=1#L7

docs/a2ui_protocol.md
```diff
@@ -414,13 +433,21 @@ This message provides a feedback mechanism for the server. It is sent when the c
 1.  **Component Definition** (from `updateSurface`):
 
     ```json
+    {
```
holtskinner, line 435: This section is technically not valid json. These should be contained within an array or parent object, or they should be two separate blocks. This is why there are red highlights in the diff.

gspencergoog: Yes, it was meant to illustrate a fragment of what goes into `updateSurface` (indicated by the "from `updateSurface`:" part above), so I didn't include the surrounding object. I'll add some more context so that's more obvious.

README.md
gspencergoog: I think it's better to use English to define the acronym. Defining it as Agent2Agent sounds like we're defining an acronym with another acronym or codename.

docs/a2ui_protocol.md
gspencergoog: Okay, thanks, that's an enhancement I can work on in another PR, since this one is already long.

PR #31, opened 2025-10-20
Initial commit for a2ui_restaurant_finder and a generic a2UI extension.

dmandar:


a2a_samples/a2ui_extension/spec/spec.md
```diff
@@ -0,0 +1,51 @@
...
+The URI of this extension is https://github.com/google/a2ui/a2a_samples/a2ui_extension/spec/spec.md.
+
+This is the only URI accepted for this extension.
+
+## Core Concepts
+The A2UI extension is built on three main concepts:
+
+Surfaces: A "Surface" is a distinct, controllable region of the client's UI. The spec uses a surfaceId to direct updates to specific surfaces (e.g., a main content area, a side panel, or a new chat bubble). This allows a single agent stream to manage multiple UI areas independently.
+
+Catalogs: The a2ui extension is component-agnostic. All UI components (e.g., Text, Row, Button) and their stylings are defined in a separate Catalog Definition Schema. This allows clients and servers to negotiate which catalog to use.
```
jacobsimionato: Nit: I think we should use the term Catalog Definition document here. It is a document that conforms to the Catalog Definition Schema.

a2a_samples/a2ui_restaurant_finder/tools.py
```diff
@@ -0,0 +1,885 @@
...
+            logger.error(
+                f'  - Error: restaurant_data.json not found at {file_path}'
+            )
+        except json.JSONDecodeError:
+            logger.error(f'  - Error: Failed to decode JSON from {file_path}')
+
+    return json.dumps(items)
+
+
+def get_ui_instructions(base_url: str) -> str:
```
jacobsimionato: Why are you exposing this context via a tool that requires an additional model inference? Can you model the actual output as a tool call and provide the schema? Or provide the schema as a structured output schema for the final response? Or just inline this into the overall system prompt?

dmandar: Yes, I had this inline in the overall system prompt earlier, but got fb from team members to have this as separate tools for now. We can move things around, but this is cleaner to start with (also not noticeable perceived e2e latency so far).
I think having a single tool that returns the data and schema might be a good middleground. I will be playing around with this, but want to get in something that works first.

a2a_samples/a2ui_extension/spec/spec.md
dmandar: Sure, done.

a2a_samples/a2ui_restaurant_finder/tools.py
gspencergoog: I agree with Jacob, this seems like an odd way to structure it.  If you're not going to use structured output, then just putting this into the system prompt seems more straightforward. You can keep it as a separate string and interpolate it in order to keep the implementations separate, if that's the goal.

If you are going to use structured output, then it either needs to be set as the structured output schema, or needs to be the parameter schema for a tool.

Is the LLM in "forced tool calling" mode?  Because if not, it may just decide not to call this tool, whereas having it in the system prompt means it will see it regardless.

I know this is only an example, but that means people will copy this pattern.

dmandar: I updated to have the UI sent in the prompt. We will keep the data fetch as a tool call. PTAL.

PR #42, opened 2025-10-23
Align apps with schema

gspencergoog:
This aligns the restaurant finder server, the web restaurant app, and the schema in the spec so that they all match.

I had trouble getting things to run/build because there seemed to be missing config files, so I generated those and switched to pnpm for the web app.  It all can be built and run now.


a2a_samples/a2ui_extension/src/a2ui_ext/__init__.py
```diff
@@ -23,7 +23,7 @@ def agent_extension(self) -> AgentExtension:
             description="Provides a declarative a2ui UI JSON structure in messages.",
             params={
                 "supportedSchemas": [
-                    "https://raw.githubusercontent.com/google/a2ui/refs/heads/main/schemas/v0.1/standard_catalog.json"
+                    "https://raw.githubusercontent.com/google/A2UI/refs/heads/main/specification/json/server_to_client.json"
```
dmandar, line 26: can we also update this in the a2a extension? spec.md

a2a_samples/a2ui_restaurant_finder/prompt_builder.py
```diff
@@ -641,23 +637,37 @@
           "description": "An array of data entries. Each entry must contain a 'key' and exactly one corresponding typed 'value*' property.",
           "items": {
             "type": "object",
-            "description": "A single data entry. Exactly one 'value_' property should be provided alongside the key.",
+            "description": "A single data entry. Exactly one 'value*' property should be provided alongside the key.",
             "properties": {
               "key": {
                 "type": "string",
                 "description": "The key for this data entry."
```
jacobsimionato, line 644: I know this is an issue with the schema source of truth, but we should be clear on the key format, and how it handles lists of objects, e.g. how do you define an object like:
```json
{
  "books": [
    {
      "title": "Dune",
      "author": "Frank Herbert"
    },
    {
      "title": "1984",
      "author": "George Orwell"
    },
    {
      "title": "The Hobbit",
      "author": "J.R.R. Tolkien"
    }
  ]
}
```

paullewis:
This changes rather a lot in the web code, and not all of it correctly. I am happy to do a pass on that myself if we are happy that the spec is up-to-date.

paullewis:
Also I don't want to switch to pnpm if possible. It's great and all, but it's not the de facto choice for web people.

paullewis:
@gspencergoog I've aligned the web with the checked-in spec here: https://github.com/google/A2UI/pull/44

a2a_samples/a2ui_extension/src/a2ui_ext/__init__.py
gspencergoog: OKay, done, but note that the links don't actually work (they get 404) until the repo is public.  There's a token I could add to the URL now to make them work, but I figured it was better to have the final version.

a2a_samples/a2ui_restaurant_finder/prompt_builder.py
gspencergoog: You're right, there's no good way to define that right now. The closest you could get would be a list of titles:

```json
{
  "/book_titles": {
    "valueList" : [{"valueString": "Dune"}, {"valueString":"1984"}, {"valueString": "The Hobbit"}]
  }
}
```

You could define individual paths, e.g. "/books/dune/title", "/books/dune/author", and use those as valus for a card, and then collect the cards in a list by id.

I agree we want to support your scenario, though.

gspencergoog:
>  This changes rather a lot in the web code, and not all of it correctly. I am happy to do a pass on that myself if we are happy that the spec is up-to-date.

Sorry, I just needed something that I could test the restaurant finder changes against. The spec is up to date. I reverted all the web changes I made in this PR.

When I tried your updated web code with this updated restaurant finder just now, I got this, so it probably needs one more look:

```
Uncaught (in promise) Error: Invalid data; expected Button
    at #buildNodeRecursive (model-processor.ts:587:17)
    at model-processor.ts:671:35
    at Array.map (<anonymous>)
    at #resolvePropertyValue (model-processor.ts:670:35)
    at #buildNodeRecursive (model-processor.ts:447:61)
    at model-processor.ts:671:35
    at Array.map (<anonymous>)
    at #resolvePropertyValue (model-processor.ts:670:35)
    at #buildNodeRecursive (model-processor.ts:447:61)
    at #resolvePropertyValue (model-processor.ts:663:38)
```

> Also I don't want to switch to pnpm if possible. It's great and all, but it's not the de facto choice for web people.

No worries, I just had trouble getting it to run with `npm`, so I switched it to something I knew.  I'm not a web person, but most of my node experience is with GenKit, and I use `pnpm` there.

PR #53, opened 2025-10-24
[web] Fix button rendering

crisbeto:
The button text is now rendered as a child component, rather than a string label.

### DIFF
--- a/web/lib/src/0.8/ui/text.ts
+++ b/web/lib/src/0.8/ui/text.ts
@@ -62,7 +62,7 @@ export class Text extends Root {
           this.surfaceId ?? A2UIModelProcessor.DEFAULT_SURFACE_ID
         );
 
-        if (textValue === null) {
+        if (textValue === null || textValue === undefined) {
           return html`(empty)`;
         }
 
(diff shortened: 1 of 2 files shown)


web/lib/src/0.8/ui/text.ts
```diff
@@ -62,7 +62,7 @@ export class Text extends Root {
           this.surfaceId ?? A2UIModelProcessor.DEFAULT_SURFACE_ID
         );
 
-        if (textValue === null) {
+        if (textValue == null) {
```
paullewis: Can we make this more explicit, i.e., `if (textValue === null || textValue === '')`? Right now it could be ambiguous as to whether we forgot to threequals or that it was an intentional loose comparison.

web/lib/src/0.8/types/types.ts
```diff
@@ -393,6 +392,11 @@ export interface ResolvedColumn {
   alignment?: "start" | "center" | "end" | "stretch";
 }
 
+export interface ResolvedButton {
+  child: AnyComponentNode;
+  action: Action;
```
paullewis:
```
export interface ResolvedButton {
  child: AnyComponentNode;
  action: Button["action"];
}
```

Let's bring the base Action through

PR #62, opened 2025-10-26
Add a new sample for a contact lookup agent

dmandar:


a2a_samples/a2ui_contact_lookup/pyproject.toml
gspencergoog, line 1: I think that the OSS Licensing script would flag this as needing copyright headers: toml is "source code" in it's mind.

a2a_samples/pyproject.toml
gspencergoog, line 1: This also needs a header.

a2a_samples/a2ui_contact_lookup/pyproject.toml
dmandar: Well we already have one for the restaurant agent which doesnt have a header. 
https://github.com/google/A2UI/blob/main/a2a_samples/a2ui_restaurant_finder/pyproject.toml
Given this, am hoping we don't need one. Can add one tomorrow if we see a need.

a2a_samples/pyproject.toml
dmandar: same as above.

PR #68, opened 2025-10-27
[spec] Add weight and Icon

paullewis:


specification/json/server_to_client.json
```diff
@@ -53,6 +53,10 @@
                 "type": "string",
                 "description": "The unique identifier for this component."
               },
+              "weight": {
+                "type": "number",
+                "description": "The relative weight of this component within a Row or Column. Note: this may ONLY be set when the component is a direct descendant of a Row or Column."
```
gspencergoog: Maybe a short description of how the weights are added and distributed and what they affect?  There's got to be a concise way to say this, but without guidance, it might try weird things (e.g. negative numbers?)

FYI, the entire algorithm that Flutter uses is described here: https://api.flutter.dev/flutter/widgets/Flex-class.html

Maybe "Weights of all siblings are added and their relative sizes allocated from the containers size in the axis direction. For instance, if one component has a weight of 1, and another 2, then the second component will be allocated twice the size in the parent container's axis direction. Not specifying the weight allows the component to be given its intrinsic size."

paullewis: Well, elsewhere we reference other CSS properties, so we could do the same here with flex-grow: https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow

paullewis: I've added a reference to `flex-grow` in the same fashion as `object-fit` from `Image`

gspencergoog: Let's go with that, and if we have problems we can elaborate.

PR #70, opened 2025-10-27
[a2a] Update restaurant example to use weights

paullewis:


a2a_samples/a2ui_restaurant_finder/prompt_builder.py
```diff
@@ -69,6 +69,10 @@
                 "type": "string",
                 "description": "The unique identifier for this component."
               },
+              "weight": {
```
dmandar, line 72: lgtm, although i guess it also needs to be in the

dmandar:
FYI @gspencergoog @jacobsimionato

PR #84, opened 2025-10-29
feat: Add primary button support

gspencergoog:
# Description

This PR introduces support for a primary attribute on Buttons in the A2UI schema and client implementations.

This change allows the server to designate a button as the primary action in a group, enabling clients to style it differently and improve user experience.

Key Changes:

Schema Update (server_to_client.json):
- Added an optional primary boolean attribute to the Button component definition.
A2A Samples:
- Updated a2ui_examples.py in a2ui_contact_lookup to demonstrate the usage of the primary attribute.
- Updated prompt_builder.py in both a2ui_contact_lookup and a2ui_restaurant_finder to include the primary attribute in the embedded schema.
- Added the primary attribute to the schema in a2ui_schema.py.

### DIFF
--- a/web/contact/index.html
+++ b/web/contact/index.html
@@ -195,11 +195,6 @@
         width: 100svw;
         height: 100svh;
       }
-      button.is-primary {
-        background-color: var(--p-40);
-        color: var(--p-100);
-        border: 1px solid var(--p-30);
-      }
     </style>
   </head>
   <body>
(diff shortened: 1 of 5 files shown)


gspencergoog:
This adds the primary attribute to Button in the server side, but I wasn't able to get the web implementation I had to actually work correctly, since I have no idea what I'm doing.  I'll leave the implementation to @paullewis for that part.

PR #85, opened 2025-10-29
Contact sample: Show a simple message when buttons are clicked.

dmandar:


gspencergoog (approved):
Is there anything I need to do for the client side to support this?  Seems like it's all server-side, so probably not, but I thought I'd ask.

PR #90, opened 2025-10-29
[web] Fix model processor

paullewis:
It was anticipating `Array` not `Map`, so now we support both.

### DIFF
--- a/web/lib/src/0.8/data/model-processor.ts
+++ b/web/lib/src/0.8/data/model-processor.ts
@@ -448,7 +448,9 @@ export class A2UIModelProcessor implements ModelProcessor {
     visited: Set<string>,
     dataContextPath: string
   ): AnyComponentNode | null {
-    const baseComponentId = componentId.split(":").at(0) ?? "";
+    const idParts = componentId.split(":");
+    const baseComponentId = idParts[0] ?? "";
+    const idSuffix = idParts.length > 1 ? `:${idParts.slice(1).join(":")}` : "";
     const { components } = surface;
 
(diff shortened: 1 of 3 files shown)


web/lib/src/0.8/data/model-processor.ts
```diff
@@ -743,6 +760,26 @@ export class A2UIModelProcessor implements ModelProcessor {
           });
         }
 
+        // Handle Map data.
+        const mapCtor = this.#mapCtor;
+        if (data instanceof mapCtor) {
+          return Array.from(data.keys()).map((key) => {
```
crisbeto, line 766: nit: Array.from has a second parameter that's a mapping function so you can do it in a single call, e.g. `Array.from(data.keys(), key => ...)`.

web/lib/src/0.8/data/model-processor.ts
```diff
@@ -698,19 +701,30 @@ export class A2UIModelProcessor implements ModelProcessor {
...
     visited: Set<string>,
-    dataContextPath: string
+    dataContextPath: string,
+    idSuffix = ""
   ): ResolvedValue {
     // 1. If it's a string that matches a component ID, build that node.
     if (typeof value === "string" && surface.components.has(value)) {
-      return this.#buildNodeRecursive(value, surface, visited, dataContextPath);
+      return this.#buildNodeRecursive(
+        `${value}${idSuffix}`,
```
crisbeto, line 710: Would it make sense to construct this inside `buildNodeRecursive` so we don't have to repeat it?

PR #94, opened 2025-10-29
Add time off agent sample

wrenj:
Adds demo time off agent


wrenj:
Getting basic agent in. Both screens render fine in a2ui. Will move to component catalog next.

PR #98, opened 2025-10-31
Add icon enum values to specify allowed icons

gspencergoog:
# Description

This specifies the allowed icons for the Icon component.  There are probably better ways to do icons in the future, but for now there are some challenges with the way things are:

1) Icon fonts are pretty huge unless you can subset them.  But... you can't subset them if you don't know what icons will be in use.
2) Specifying all 8,000+ Material icon names in the spec is impractical, so I picked the top 50 or so. These *should* take care of most common use cases, and for the others people can serve PNGs and use images, or we can add an SVG component later and somehow get the LLM to generate the SVG directly (maybe look it up with a tool?).

Anyhow, for now this is a good stopgap.

I also converted the a2ui_protocol.md doc to use mkdocs and to generate the Mermaid diagrams on the fly.

### DIFF
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -134,8 +134,8 @@ markdown_extensions:
 plugins:
   - search
   - macros
-  - include-markdown
-  - mermaid2
+  # - include-markdown
+  # - mermaid2
   # - llmstxt:
   #     full_output: llms-full.txt
   #     sections:


dmandar:
is there a way to specify a custom icon?

gspencergoog:
> is there a way to specify a custom icon?

Yes, you can use a PNG (in an Image component).

PR #99, opened 2025-10-31
Use the schema from the canonical place

gspencergoog:
# Description

This uses the schema from `specification/json/server_to_client.json` in the examples and web code so that they all share the same schema and we don't have several copies sitting around.

Also, updated the contact example to use icons that exist (or will exist once https://github.com/google/A2UI/pull/98 lands) in the schema enum for `Icon`.


a2a_samples/a2ui_contact_lookup/a2ui_examples.py
```diff
@@ -72,12 +72,12 @@
       { "id": "description_text_1", "component": { "Text": { "text": { "path": "title"} } } } ,
       { "id": "description_text_2", "component": { "Text": { "text": { "path": "team"} } } } ,
       { "id": "description_column", "component": { "Column": { "children": { "explicitList": ["user_heading", "description_text_1", "description_text_2"]} , "alignment": "center"} } } ,
-      { "id": "calendar_icon", "component": { "Icon": { "name": { "literalString": "calendar_today"} } } } ,
+      { "id": "calendar_icon", "component": { "Icon": { "name": { "literalString": "calendarToday"} } } } ,
```
paullewis, line 75: This was set to the font string here: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:calendar_today:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=calendar+to&icon.size=24&icon.color=%231a1a1a

Does it still work to use camelCase?

gspencergoog: No, it doesn't: there are now a fixed set of icons, so there will need to be a mapping anyhow.  Although I guess we could use snake_case and make a snake_case to snake_case mapping. I used camel case because it matched the naming convention for enums, which is what the icon value is now.

There are also some icons which don't map 1:1 because they don't match each other (e.g. lock and lock_open: one is rendered as an outline and the other is solid), so I'm mapping them to the matching version. (lock -> lock_outlined, and lock_open as-is).  There are a couple of others too, because icon naming is inconsistent, and I wanted to have the LLM be able to reason about on vs off more easily (e.g. instead of notifications and notifications_none, or favorite and favorite_outline, I used notifications and `notificationsOff`, and`favoriteOff` and `visibilityOff`).

gspencergoog: Sorry, I forgot that you maybe haven't seen this one yet: here's some more context about the Icons: https://github.com/google/A2UI/pull/98

gspencergoog:
@paullewis @dmandar Okay if I land this?

gspencergoog:
Abandoning this for now, because it's somewhat out of date.

PR #100, opened 2025-10-31
Add time off agent, includes tool approach and using a2a agent executor

wrenj:


dmandar (approved):
Can you add a screenshot?

wrenj:
this is an old one let me revert it

PR #103, opened 2025-11-03
docs: Add Waitlist Form to the homepage and move docs to private directory

holtskinner:
Temporary move of the spec doc to private directory for initial launch.


docs_private/a2ui_protocol.md
gspencergoog, line 1: Why are we hiding this?  This is the protocol description that is being kept up to date with the schemas, etc. It should be the canonical document for the protocol.

holtskinner: Correct, this is for the initial launch where we will only have a sign up form publicly visible. We will make the protocol docs public at a later date. Requested by @mvakoc

gspencergoog: Ahh, OK.  So this only affects the github pages site.

PR #115, opened 2025-11-11
Move specification to v0.8 folder and centralize JSON schemas, specification docs, and associated eval

jacobsimionato:
- Move `specification/json` -> `specification/v0.8/json` (Rationale: We need to have multiple versions coexist soon)
- Move `eval` -> `specification/v0.8/eval` (Rationale: The eval logic is somewhat customized per version and it's simpler to branch than support multiple versions)
- Move `docs_private/a2ui_protocol.md` -> `specification/0.8/docs/a2ui_protocol.md` (Rationale: The formal specification doc is version-specific)

Update references to these docs from elsewhere.

### DIFF
--- a/web/lib/src/0.8/schemas/server_to_client.json
+++ b/web/lib/src/0.8/schemas/server_to_client.json
@@ -136,10 +136,60 @@
                     "properties": {
                       "name": {
                         "type": "object",
-                        "description": "The name of the icon to display. This can be a literal string ('literal') or a reference to a value in the data model ('path', e.g. '/icon/name').",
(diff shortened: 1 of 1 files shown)


web/lib/src/0.8/schemas/server_to_client.json
gspencergoog, line 1: This update might require some updates to the web code for it to render icons and buttons properly.

jacobsimionato: True. I tried restaurant finder and contact card, and they seem fine. The good thing is that the new fields are basically additive, so I think the renderer can ignore them if necessary. I don't want to exclude this file, because it's essentially generated so that would leave the repo in an inconsistent state, which I think it was in before this change.

PR #1, opened 2025-09-28
Update the documents and schemas to a2ui v0.8

jacobsimionato:

PR #119, opened 2025-11-12
Refactor v0.8 JSON schemas to reflect the generic schema vs the schema with the standard catalog embedded

jacobsimionato:
This change splits apart the different schemas for A2UI v0.8, back to structure that was established in https://github.com/google/A2UI/pull/1/files, but was later simplified, losing some of the information.

The changes are:
- Make `server_to_client.json` generic to not represent any particular catalog. It now uses `additionalProperties` so Gemini can't inference against it directly, but this is fine, because Gemini can inference against `server_to_client_with_standard_catalog.json`. Formally, the A2UI protocol supports standard and custom catalogs, so the official schema supports the superset of both, and therefore no specific component list.
- Add `server_to_client_with_standard_catalog.json`, which is the same as the previous `server_to_client.json` and has the standard catalog baked in.
- Add a 'catalog_description_schema.json` which describes a specific catalog including both its components and styling options.

Update docs and references to the above. Things should keep working, because the demo apps point to `server_to_client_with_standard_catalog.json` instead of just `server_to_client.json` right now.

### DIFF
--- a/specification/0.8/json/standard_catalog_definition.json
+++ b/specification/0.8/json/standard_catalog_definition.json
@@ -2,10 +2,12 @@
   "components": {
     "Text": {
       "type": "object",
+      "additionalProperties": false,
       "properties": {
         "text": {
           "type": "object",
           "description": "The text content to display. This can be a literal string or a reference to a value in the data model ('path', e.g., '/doc/title'). While simple Markdown formatting is supported (i.e. without HTML, images, or links), utilizing dedicated UI components is generally preferred for a richer and more structured presentation.",
+          "additionalProperties": false,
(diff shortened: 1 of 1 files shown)


specification/0.8/docs/a2ui_protocol.md
```diff
@@ -10,6 +10,7 @@ A Specification for a JSONL-Based, Streaming UI Protocol
 
 Created: Sep 19, 2025
 Updated: Nov 10, 2025
+Updated: Nov 10, 2025
```
gspencergoog: This is a duplicate

PR #122, opened 2025-11-13
Proposed Schema evals

gspencergoog:


specification/0.9/json/common_types.json
```diff
@@ -0,0 +1,94 @@
...
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "path": {
+          "type": "string"
+        }
+      },
+      "oneOf": [{ "required": ["literalArray"] }, { "required": ["path"] }]
+    },
+    "componentProperties": {
```
jacobsimionato, line 59: does this need a special subschema? It's only used in one place, right?

jacobsimionato: I think we should have a childProperty one though.

specification/0.9/json/server_to_client.json
```diff
@@ -0,0 +1,88 @@
...
+          "type": "string",
+          "description": "The unique identifier for the UI surface this data model update applies to."
+        },
+        "path": {
+          "type": "string",
+          "description": "An optional path to a location within the data model (e.g., '/user/name'). If omitted, or set to '/', the entire data model will be replaced."
+        },
+        "contents": {
+          "type": "object",
+          "description": "The data to be updated in the data model. This can be any valid JSON object.",
```
jacobsimionato, line 70: So much nicer!!

specification/0.9/json/server_to_client.json
```diff
@@ -0,0 +1,88 @@
...
+          "description": "The unique identifier for the UI surface to be rendered."
+        },
+        "root": {
+          "type": "string",
+          "description": "The ID of the root component to render."
+        },
+        "styles": {
+          "type": "object",
+          "description": "Styling information for the UI.",
+          "properties": {
```
jacobsimionato, line 22: Can we follow the pattern in v0.8 where this is defined in the catalog document

specification/0.9/json/server_to_client.json
```diff
@@ -0,0 +1,88 @@
...
+        "surfaceId": {
+          "type": "string",
+          "description": "The unique identifier for the UI surface to be updated. If you are adding a new surface this *must* be a new, unique identified that has never been used for any existing surfaces shown."
+        },
+        "components": {
+          "type": "array",
+          "description": "A list containing all UI components for the surface.",
+          "minItems": 1,
+          "items": {
+            "$ref": "component_catalog.json#/$defs/anyComponent"
```
jacobsimionato, line 50: Can we have this not reference specific components and just be a generic object?

gspencergoog: Referencing a specific component keeps it from making them up.  We can try it and see what the evals look like, though.

specification/0.9/json/server_to_client.json
gspencergoog: Yes, but I worry about types not matching (as we've seen in the other schema).  E.g. when it gives an integer for a text field.

specification/0.9/json/server_to_client.json
gspencergoog: Sure.

PR #124, opened 2025-11-20
Update Angular A2UI renderer to support custom components

suyangw-g:
Also fix the broken demos due to recent changes in json files

### DIFF
--- a/angular/projects/lib/src/lib/catalog/default.ts
+++ b/angular/projects/lib/src/lib/catalog/default.ts
@@ -15,6 +15,7 @@
  */
 
 import { inputBinding } from '@angular/core';
+import { v0_8 } from '@a2ui/web-lib';
 import { Catalog } from '../rendering/catalog';
 import { Row } from './row';
 import { Column } from './column';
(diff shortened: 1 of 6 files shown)


angular/projects/lib/src/lib/catalog/default.ts
```diff
@@ -23,117 +24,165 @@ import { Text } from './text';
 export const DEFAULT_CATALOG: Catalog = {
   Row: {
     type: () => Row,
-    bindings: ({ properties }) => [
-      inputBinding('alignment', () => properties.alignment ?? 'stretch'),
-      inputBinding('distribution', () => properties.distribution ?? 'start'),
-    ],
+    bindings: (node) => {
+      const properties = (node as v0_8.Types.RowNode).properties;
```
crisbeto, line 28: Could we avoid the cast here by typing the `node` parameter above as `v0_8.Types.RowNode`?

suyangw-g: It doesn't seem to be allowed in Typescript.

```
In TypeScript, function parameters are contravariant, meaning you cannot assign a function that accepts a narrower type (like RowNode) to a slot that expects a wider type (like AnyComponentNode).
```

PR #128, opened 2025-11-20
Refactor Angular ModelProcessor to decouple from A2UIClient

suyangw-g:
This change refactors the Angular implementation to match the architecture used in the Lit implementation, separating state management from network requests.

Key changes:
- Remove `A2UIClient` and `makeRequest` from `ModelProcessor`. The processor is now a pure state machine.
- Update `DynamicComponent` to emit events rather than triggering requests directly.
- Update `contact` and `restaurant` applications to act as orchestrators: they now listen for events, handle network calls via `A2UIClient`, and update the `ModelProcessor` with results.
- Rename `client.ts` to `types.ts` in the data library.

Benefits:
- Improves separation of concerns.
- Ensures architectural consistency across Lit and Angular platforms.
- Makes the codebase easier to maintain and test.


angular/projects/lib/src/lib/data/processor.ts
```diff
@@ -14,15 +14,19 @@
  limitations under the License.
  */
 
-import { inject, Injectable, signal } from '@angular/core';
 import { v0_8 } from '@a2ui/web-lib';
-import { A2UIClient } from './client';
+import { Injectable } from '@angular/core';
+import { firstValueFrom, Subject } from 'rxjs';
```
crisbeto, line 19: We're generally trying to move away from rxjs whenever we can so I was avoiding it here too. Maybe we can implement a simple pub/sub system here instead?

suyangw-g: I prefer addressing this comment in a separate cl.

What's the recommended way by Angular team or you to replace the usage of Subject? Promise.withResolvers, callback functions, or sth else?

PR #130, opened 2025-11-21
Add a draft for the protocol docs.

gspencergoog:
This adds the protocol document for the 0.9 proposed spec.

### DIFF
--- a/specification/0.9/docs/a2ui_protocol.md
+++ b/specification/0.9/docs/a2ui_protocol.md
@@ -250,6 +250,8 @@ Container components (like `Row`, `Column`, `List`, and `Card`) have properties 
 
 This model allows the server to send component definitions in any order, as long as all necessary components are present before rendering is triggered.
 
+There must be exactly one component with the ID `root` in the component tree, acting as the root of the component tree. Until that component is defined, other component updates will have no visible effect, and they will be buffered until a root component is defined. Once a root component is defined, the client is responsible for rendering the tree in the best way possible based on the available data, skipping invalid references.
+
 ```mermaid
 flowchart TD
     subgraph "Server Stream"


specification/0.9/docs/a2ui_extension_specification.md
jacobsimionato, line 1: Maybe we can remove this file for now and check with @dmandar about whether it's strictly necessary. I was hoping that `a2ui_protocol.md` might be enough by itself, to avoid duplication. Also, this version of `a2ui_extension_specification.md` branched from v0.8 could do with some formatting improvements, e.g. more use of headings, bold etc.

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+**Status:** Draft
+**Created:** Nov 20, 2025
+
+# A2UI (Agent to UI) Protocol v0.9
+
+A Specification for a JSONL-Based, Streaming UI Protocol
+
+## Introduction
+
+The A2UI Protocol is designed for dynamically rendering user interfaces from a stream of JSON objects in a JSONL stream sent from an A2A server. Its core philosophy emphasizes a clean separation of UI structure and application data, enabling progressive rendering as the client processes each message.
```
jacobsimionato: At the protocol level, I don't think we need to talk about JSON-L at all, because we won't actually be sending JSON-L - just a stream of A2A messages over SSE. Each A2A message is a discrete piece of data, as far as the protocol is concerned.

Inside the agent server, we may well create these messages by having the LLM generate a JSON-L stream, and incrementally parse that stream, wrap each entry in an A2A message, and send it to the client. But that's an implementation detail of the inference library and not part of the protocol. In these protocol docs, we could mention that the protocol is designed to incrementally update a UI based on a sequence of A2A messages that can be delivered over streaming transport such as SSE. I think you describe that second part really well already below.

WDYT?

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+The protocol is designed to be "prompt-first," meaning its structure is intended to be generated directly by a Large Language Model (LLM) as part of a prompt. This approach allows for a more extensive and expressive component catalog because the schema is not constrained by the limited primitives of structured output formats. Instead, validation and error correction occur after the JSON is generated.
+
+Communication occurs via a stream of JSON objects. The client parses each object as a distinct message and incrementally builds or updates the UI. The server-to-client protocol defines four message types:
+
+- `createSurface`: Signals the client to prepare a new UI area (a "surface") for rendering.
+- `surfaceUpdate`: Provides a list of component definitions to be added to or updated in a specific surface.
+- `dataModelUpdate`: Provides new data to be inserted into or to replace a surface's data model.
+- `deleteSurface`: Explicitly removes a surface and its contents from the UI.
+
+## Design Philosophy: Prompt-First Generation
```
jacobsimionato: I think it would be good to clearly separate the description of the v0.9 protocol, from the comparison with v0.8. Most readers of this doc (especially LLMs) will not have heard of v0.8, so we want to make it clear any comparison is like a changelog section which you can ignore, and which we can just delete from v0.10 etc.

So how about renaming this "Comparison with v0.8" or "Changes in v0.9"

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+Version 0.9 of the A2UI protocol represents a philosophical shift from previous versions. While v0.8 was optimized for LLMs that support structured output, v0.9 is designed to be embedded directly within a model's prompt. The LLM is then asked to produce JSON that matches the provided examples and schema descriptions.
+
+This "prompt-first" approach offers several advantages:
+
+1.  **Richer Schema:** The protocol is no longer limited by the constraints of structured output formats. This allows for more complex, expressive, and comprehensive component catalogs.
+2.  **Improved Generative Performance:** By providing the schema and examples in the prompt, the LLM has more context, leading to more accurate and reliable JSON generation.
+3.  **Post-Generation Validation:** The responsibility for strict schema adherence is shifted to a post-generation validation step. This allows for more robust error handling and correction, as the system can identify discrepancies and even attempt to fix them before rendering, and not be constrained by the limited primitives and complexity constraints of structured output formats.
```
jacobsimionato: I don't think this "allows for more robust error handling and correction". Maybe it "relies on" such handling?

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+# A2UI (Agent to UI) Protocol v0.9
+
+A Specification for a JSONL-Based, Streaming UI Protocol
+
+## Introduction
+
+The A2UI Protocol is designed for dynamically rendering user interfaces from a stream of JSON objects in a JSONL stream sent from an A2A server. Its core philosophy emphasizes a clean separation of UI structure and application data, enabling progressive rendering as the client processes each message.
+
+The protocol is designed to be "prompt-first," meaning its structure is intended to be generated directly by a Large Language Model (LLM) as part of a prompt. This approach allows for a more extensive and expressive component catalog because the schema is not constrained by the limited primitives of structured output formats. Instead, validation and error correction occur after the JSON is generated.
```
jacobsimionato: I don't think "It's structure is intended to be generated directly by a Large Language Model (LLM) as part of a prompt" is quite correct.

Maybe just drop this section?

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+3.  **Post-Generation Validation:** The responsibility for strict schema adherence is shifted to a post-generation validation step. This allows for more robust error handling and correction, as the system can identify discrepancies and even attempt to fix them before rendering, and not be constrained by the limited primitives and complexity constraints of structured output formats.
+4.  **Modularity:** The schema can be refactored into separate, more manageable components (e.g., [`common_types.json`], [`standard_catalog_definition.json`], [`server_to_client.json`]), improving maintainability and modularity.
+
+## Protocol Overview & Data Flow
+
+The A2UI protocol uses a unidirectional stream of JSON messages from the server to the client to describe and update the UI. The client consumes this stream, builds the UI, and renders it. User interactions are handled separately, typically by sending events to a different endpoint, which may in turn trigger new messages on the UI stream.
+
+1.  **Create Surface:** The server sends a `createSurface` message to instruct the client to initialize a new, independent UI region.
```
jacobsimionato, line 50: Maybe label this as an "Example sequence of events", to be clear that it doesn't have to be exactly like this, e.g. data model update can occur after surfaceUpdate.

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+    Client-->>-Server: (UI is updated)
+
+    Server->>+Client: 6. deleteSurface(surfaceId: "main")
+    Note right of Client: Client removes the UI for the "main" surface
+    Client-->>-Server: (UI is gone)
+```
+
+## Message Structure
+
+The protocol defines four primary message types in the [`server_to_client.json`] schema:
```
jacobsimionato: This duplicates the section near the top. Let's merge them and be clear on detail. E.g. for surfaceUpdate, define that *must* refer to a surface that already exists, either because it's known to the agent already, or it has been created with createSurface. Same with data model update.

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+### The Envelope (`server_to_client.json`)
+
+The [`server_to_client.json`] schema is the top-level entry point. Every line streamed by the server must validate against this schema. It handles the message dispatching.
+
+### The Catalog (`standard_catalog_definition.json`)
+
+The [`standard_catalog_definition.json`] schema contains the definitions for all specific UI components (e.g., `Text`, `Button`, `Row`). By separating this from the envelope, developers can easily swap in custom catalogs (e.g., `material_catalog.json` or `cupertino_catalog.json`) without rewriting the core protocol parser.
+
+An A2UI message is a JSON object that must contain exactly one of the following keys: `createSurface`, `surfaceUpdate`, `dataModelUpdate`, or `deleteSurface`. The key indicates the type of message.
```
jacobsimionato:
```suggestion
An A2UI server-to-client message is a JSON object that must contain exactly one of the following keys: `createSurface`, `surfaceUpdate`, `dataModelUpdate`, or `deleteSurface`. The key indicates the type of message.
```

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+The [`common_types.json`] schema defines reusable primitives used throughout the protocol.
+
+- **`stringOrPath` / `numberOrPath` / `booleanOrPath`**: The core of the data binding system. Any property that can be bound to data is defined as an object that accepts either a literal value OR a `path` string (JSON pointer).
+- **`childrenProperty`**: Defines how containers hold children. It supports:
+
+  - `explicitList`: A static array of component IDs.
+  - `template`: A dynamic generator that creates children based on a list in the data model.
+
+### The Envelope (`server_to_client.json`)
```
jacobsimionato:
```suggestion
### Server to client message structure - "The envelope"
```

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Introduction
+
+The A2UI Protocol is designed for dynamically rendering user interfaces from a stream of JSON objects in a JSONL stream sent from an A2A server. Its core philosophy emphasizes a clean separation of UI structure and application data, enabling progressive rendering as the client processes each message.
+
+The protocol is designed to be "prompt-first," meaning its structure is intended to be generated directly by a Large Language Model (LLM) as part of a prompt. This approach allows for a more extensive and expressive component catalog because the schema is not constrained by the limited primitives of structured output formats. Instead, validation and error correction occur after the JSON is generated.
+
+Communication occurs via a stream of JSON objects. The client parses each object as a distinct message and incrementally builds or updates the UI. The server-to-client protocol defines four message types:
+
+- `createSurface`: Signals the client to prepare a new UI area (a "surface") for rendering.
```
jacobsimionato:
```suggestion
- `createSurface`: Declares and configures a new UI area for rendering.
```

I think it's hard to know what "prepare" means practically

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+### The Catalog (`standard_catalog_definition.json`)
+
+The [`standard_catalog_definition.json`] schema contains the definitions for all specific UI components (e.g., `Text`, `Button`, `Row`). By separating this from the envelope, developers can easily swap in custom catalogs (e.g., `material_catalog.json` or `cupertino_catalog.json`) without rewriting the core protocol parser.
+
+An A2UI message is a JSON object that must contain exactly one of the following keys: `createSurface`, `surfaceUpdate`, `dataModelUpdate`, or `deleteSurface`. The key indicates the type of message.
+
+## The Protocol Schema Messages
+
+These are the messages that make up each line in the protocol stream.
```
jacobsimionato:
```suggestion
These are the messages that make up each message in the protocol stream.
```

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+### `createSurface`
+
+This message instructs the client to initialize a new surface, which is a designated area for rendering a UI. When this message is received, the client should prepare to receive subsequent `surfaceUpdate` messages for the same `surfaceId`.
```
jacobsimionato:
```suggestion
This message instructs the client to initialize a new surface, which is a designated area for rendering a UI. When this message is received, the client should prepare to receive subsequent `surfaceUpdate` and `dataModelUpdate` messages for the same `surfaceId`.
```

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+**Properties:**
+
+- `surfaceId` (string, required): A unique identifier for the UI surface to be rendered.
+- `theme` (object, optional): Theming information for the UI, such as font and primary color.
```
jacobsimionato:
```suggestion
- `theme` (object, optional): Theming information for the UI, such as primary color.
```

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+**Example:**
+
+```json
+{
+  "createSurface": {
+    "surfaceId": "user_profile_card",
+    "theme": {
+      "font": "Open Sans",
```
jacobsimionato:
```suggestion
```

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+      "primaryColor": "#007bff"
+    }
+  }
+}
+```
+
+### `surfaceUpdate`
```
jacobsimionato, line 132: Maybe explain this can only be sent for a surface that already exists, because it's known to exist by the agent or has been created.

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+These are the messages that make up each line in the protocol stream.
+
+### `createSurface`
+
+This message instructs the client to initialize a new surface, which is a designated area for rendering a UI. When this message is received, the client should prepare to receive subsequent `surfaceUpdate` messages for the same `surfaceId`.
+
+**Properties:**
+
+- `surfaceId` (string, required): A unique identifier for the UI surface to be rendered.
```
jacobsimionato, line 116: Maybe explain this is typically a name with meaning e.g. "trip_options_panel" and it has to be unique within the client's Gen UI context.

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+{
+  "deleteSurface": {
+    "surfaceId": "user_profile_card"
+  }
+}
+```
+
+## Example Stream
+
+The following example demonstrates a complete interaction to render a Contact Form as a JSONL stream.
```
jacobsimionato:
```suggestion
The following example demonstrates a complete interaction to render a Contact Form, expressed as a JSONL stream.
```

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+| **Divider**        | A horizontal or vertical dividing line.                 |
+| **Modal**          | A dialog that appears over the main content.            |
+| **Button**         | A clickable button that dispatches an action.           |
+| **CheckBox**       | A checkbox with a label and a boolean value.            |
+| **TextField**      | A field for user text input.                            |
+| **DateTimeInput**  | An input for date and/or time.                          |
+| **MultipleChoice** | A component for selecting one or more options.          |
+| **Slider**         | A slider for selecting a numeric value within a range.  |
+
+### Data Binding
```
jacobsimionato, line 303: Can we also talk about templates here, and also about relative vs absolute paths (if we want that concept?)

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+
+- `name` (string, required): The name of the action.
+- `surfaceId` (string, required): The ID of the surface where the action originated.
+- `sourceComponentId` (string, required): The ID of the component that triggered the action.
+- `timestamp` (string, required): An ISO 8601 timestamp.
+- `context` (object, required): A JSON object containing any context provided in the component's `action` property.
+
+### `clientUiCapabilities`
+
+This message informs the server about the client's UI rendering capabilities, such as the component catalog it supports.
```
jacobsimionato: Maybe remove this section for now, and I'll readd it later in a different way, seeing as we removed this message type.

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,368 @@
...
+## Message Structure
+
+The protocol defines four primary message types in the [`server_to_client.json`] schema:
+
+1.  **`surfaceUpdate`**: Transmits the component tree structure.
+2.  **`dataModelUpdate`**: Transmits the data state.
+3.  **`createSurface`**: Signals the client to prepare a new UI area for rendering. (Replaces `beginRendering` from v0.8).
+4.  **`deleteSurface`**: Removes a surface.
+
+## The Protocol Schemas
```
jacobsimionato, line 79: How about also having a section that describes what a custom catalog definition is, and how to write one?

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Sure, how about I just talk about it being a stream.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Hmm. Okay. I wanted to mention this because it highlights that the protocol isn't really designed to fit in to structured output. I'll see if I can reword/reduce it.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Sure.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: I added a paragraph on this.  It feels kind of "too simple", but I think in this document, which is concerned with the protocol itself, that is sufficient, since custom catalog items are mostly an implementation detail.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: I kept your idea of `createSurface` being required before referencing a surface id.  We might still remove the `createSurface` message altogether, but I realized that one benefit of having it is that the client can validate that a surface ID hasn't been hallucinated.

~~Actually, it's not required for it to refer to a surface that already exists.  If a surface is referred to before a `createSurface` with that ID is sent, then it is just held until the `createSurface` message arrives.  This allows the LLM to control how progressive the UI is: it can stream the UI as it builds it by specifying `createSurface` first, or wait until it is all there by specifying `createSurface` after sending the corresponding `surfaceUpdate` messages.~~

specification/0.9/docs/a2ui_protocol.md
gspencergoog: I just removed it.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Done.

specification/0.9/docs/a2ui_extension_specification.md
gspencergoog: Okay, I'll remove it for now.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Done.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Done.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Yes, I'll add that.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Removed

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -0,0 +1,364 @@
...
+
+The set of available UI components and their properties is defined in a **Component Catalog**. The standard catalog is defined in [`standard_catalog_definition.json`]. This allows for different clients to support different sets of components, including custom ones. The server must generate `surfaceUpdate` messages that conform to the component catalog understood by the client.
+
+### UI Composition: The Adjacency List Model
+
+The A2UI protocol defines the UI as a flat list of components. The tree structure is built implicitly using ID references. This is known as an adjacency list model.
+
+Container components (like `Row`, `Column`, `List`, and `Card`) have properties that reference the `id` of their child component(s). The client is responsible for storing all components in a map (e.g., `Map<String, Component>`) and recreating the tree structure at render time.
+
+This model allows the server to send component definitions in any order, as long as all necessary components are present before rendering is triggered.
```
jacobsimionato, line 251: Now we have removed beginRendering, maybe we just say that if the root node is present, then the renderer is responsible for rendering the tree as best as possible based on the available data, skipping invalid references? Or we just leave that undefined for now? I think we'll have to think carefully about it when we do streamed rendering, but maybe we can do a v0.9.1. which purely adds clarifications on renderer behavior without touching the schema.

specification/0.9/eval/.gitignore
```diff
@@ -7,7 +7,7 @@
 
 # testing
 /coverage
-/results
+#/results
```
jacobsimionato: Undo?

specification/0.9/eval/src/index.ts
jacobsimionato, line 1: Are these changes intended to be included?

gspencergoog: No, I'll move them to another PR.  I wanted to add more test cases.

specification/0.9/eval/.gitignore
gspencergoog: Oh, I know what happened here.  I was commenting it out so that Antigravity could access them.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Good idea.  I added that, and also mentioned that there is only one "root" component, and that it always has the ID "root". 

One thing that occurs to me is that it is currently hard to clear the buffer of orphaned components.  If I send a bunch of components, they'll all get buffered until the root arrives.  Once it arrives, then it's still possible to have things which are not yet connected to the tree in the buffer (children of a column where the column hasn't yet been defined, for instance).  There's not really a good time to clear out those orphans, since we don't know when things are "done".  For now, I don't think this is a big deal, since we're mostly talking about small UIs with limited lifespans, but if A2UI matures into something where it's building whole apps, then there will need to be a garbage collection phase.

Issue #135, opened 2025-11-24
Properly document the revised, metadata-based catalog negotiation mechanism in the v0.8 spec


jacobsimionato:
Let's formalize this based on the discussions we had.

# Coding agent prompt
I want to update the specification in v0.8 of the A2UI protocol to change the way that "custom catalogs" are negotiated.

You must update the specification/0.8 documents based on this. First, just make a detailed plan of what to update and wait for my approval.

When updating the docs, consider:
- Include separate .json schema document files with key schemas e.g. the one for the agent card and for the a2uiClientCapabilities message.
- Try to make it very clear how the A2UI protocol relates to the A2A protocol, for example exactly where each custom piece of data should be provided within A2A's message types. Examples should likely show the A2A context to make this clear.
- We should include a specific section in the specification to explain this whole mechanism and the rationale behind it, but *also* make sure that every other part of the specification, schemas, documentation is consistent with this.

### Summary

There is a concept of a standard catalog. Each A2UI extension version has an associated standard catalog.

The client defines which catalog is used to generate and update UI surfaces.

Each client can only choose a single catalog to use across all its surfaces.

The client must indicate which catalog to use in *every* request to the server. The server does not remember which catalog to use across requests.

Catalogs are identified via URIs, but neither client nor server should fetch the catalog definition from those URIs at runtime - they are just stable identifiers for catalogs. The URI associated with a catalog should conventionally point to a JSON catalog definition document for that catalog.

Catalogs are immutable. If a catalog is modified, it should be given a new URI.

A catalog definition defines the contract between server and client for what kinds of UIs can be represented. A catalog is defined by a catalog definition. The catalog definition contains a list of components and the global theming options which can be applied across an entire surface.

### Resources

Make sure you read *everything* in @specification/v0.8 before solving this task.

## Server catalog support advertisement
The server advertises the catalogs that is has available in its agent card, something like this:

```
{
  "name": "Restaurant finder",
  "capabilities": {
    "extensions": [
      {
        "uri": "https://a2ui.org/ext/a2a-ui/v0.8",
        "description": "Ability to render A2UI",
        "required": false,
        "params": {
          "supportedCatalogUris": [
            "https://github.com/a2ui/v0_8/default_catalog.json",
            "https://internal.google.com/a2ui/v0.8/myCustomCatalog.json"
          ],
          "acceptsInlineCatalog": True
        }
      }
    ]
  }
}
```

[...]

PR #136, opened 2025-11-24
Update v0.8 spec with updated catalog negotiation style

jacobsimionato:
Summary of changes:
- Replace clientUiCapabilities message type with a similar metadata parameter which is included in *every* message, rather than just once.
- Resolve ambiguities in the schema, e.g. whether custom components are additive or not.
- Add JSON schemas for key metadata fields etc.
- Rename "dynamic catalogs" to "inline catalogs".

Fixes https://github.com/google/A2UI/issues/135


specification/0.8/docs/a2ui_protocol.md
```diff
@@ -84,19 +83,140 @@ The central philosophy of A2UI is the decoupling of three key elements:
...
-### 1.2. The JSONL Stream: The Unit of Communication
+### 1.2. Catalog Negotiation
+
+A central concept in A2UI is the **Catalog**, which defines the contract between the server and client for what UI components can be rendered. Each client must select a single catalog to use for all UI surfaces, and it must inform the server of this choice with every request.
+
+#### 1.2.1. Principles of Catalog Negotiation
+
+-   **Statelessness**: The server does not remember which catalog the client is using across different requests. The client must specify its chosen catalog in every A2A message it sends.
+-   **Client-Driven**: The client is in control of which catalog is used.
+-   **Immutability**: Catalogs are identified by URIs and are considered immutable. If a catalog definition changes, it must be given a new URI.
```
wrenj: I would make them backward compatible, only breaking changes need a new version.

I was thinking if Moma keeps adding styles to their catalog that default to None, then there is no harm in just publishing them to the same URL. Client with older versions compiled in will just use it. It's going to be a pain if every time you fix a typo in the description you have to make a new URL.

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -84,19 +83,140 @@ The central philosophy of A2UI is the decoupling of three key elements:
...
 
-### 1.2. The JSONL Stream: The Unit of Communication
+### 1.2. Catalog Negotiation
+
+A central concept in A2UI is the **Catalog**, which defines the contract between the server and client for what UI components can be rendered. Each client must select a single catalog to use for all UI surfaces, and it must inform the server of this choice with every request.
+
+#### 1.2.1. Principles of Catalog Negotiation
+
+-   **Statelessness**: The server does not remember which catalog the client is using across different requests. The client must specify its chosen catalog in every A2A message it sends.
+-   **Client-Driven**: The client is in control of which catalog is used.
```
wrenj: Fine to publish for now, but heads up I'm working on a proposal to change this since it doesn't work with orchestrators where the expert agents could be using different catalogs (eg, one agent using default and the other using moma catalog with the canvas element).

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -84,19 +83,140 @@ The central philosophy of A2UI is the decoupling of three key elements:
...
+-   **Immutability**: Catalogs are identified by URIs and are considered immutable. If a catalog definition changes, it must be given a new URI.
+-   **URIs as Identifiers**: Catalog URIs (e.g., `https://a2ui.org/catalogs/standard/0.8`) are stable identifiers. Neither the client nor the server should fetch catalog definitions from these URIs at runtime.
```
wrenj: I'd add some blurb about security and fetching at runtime. 

Neither the client nor the server should fetch catalog definitions from these URIs at runtime for improved security.

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: SGTM. I updated wording around this.

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: I updated this based on your proposal.

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: I updated this to use catalogId rather than catalogUri to avoid this confusion.

jacobsimionato:
Hey, I overhauled this based on your new proposal. PTAL!

specification/0.8/docs/a2ui_extension_specification.md
```diff
@@ -24,26 +24,31 @@ Catalog Definition Schema: A standard format for defining a library of component
...
+  "required": false,
+  "params": {
+    "acceptsInlineCatalogs": true
+  }
+}
+```
 
-params.acceptsDynamicSchemas: (OPTIONAL) A boolean indicating if the agent can accept a clientUiCapabilities message containing a dynamicCatalog. If omitted, this defaults to false.
+#### Parameter Definitions
+- `params.acceptsInlineCatalogs`: (OPTIONAL) A boolean indicating if the agent can accept an `inlineCatalogs` array in the client's `a2uiClientCapabilities`. If omitted, this defaults to `false`.
```
wrenj, line 54: single inline catalog right?

specification/0.8/docs/a2ui_extension_specification.md
```diff
@@ -24,26 +24,31 @@ Catalog Definition Schema: A standard format for defining a library of component
...
 
-Client Capability Declaration
-The client-to-server spec includes a clientUiCapabilities message. If a client wishes to use a specific catalog (other than the server's default) or provide its own dynamic catalog, it MUST send this message after the connection is established and before the first user prompt.
+Client Capabilities
+The client sends its capabilities to the server in an `a2uiClientCapabilities` object. This object is **not** sent as a standalone message, but is instead included in the `metadata` field of every A2A `Message` sent from the client to the server.
```
wrenj: The standard component catalog will be used as the default if a2uiClientCapabilities is not present in message.metadata (note: agents could not support the standard catalog)

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -62,13 +62,12 @@ Communication occurs via a JSON Lines (JSONL) stream. The client parses each lin
 
 - `surfaceUpdate`: Provides a list of component definitions to be added to or updated in a specific UI area called a "surface."
 - `dataModelUpdate`: Provides new data to be inserted into or to replace a surface's data model. Each surface has its own data model.
-- `beginRendering`: Signals to the client that it has enough information to perform the initial render, specifying the ID of the root component.
+- `beginRendering`: Signals to the client that it has enough information to perform the initial render, specifying the ID of the root component and, optionally, the component catalog to use.
```
wrenj, line 65: I'd require it in begin render, or you could default it to the standard one

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -167,13 +166,100 @@ The following is a complete, minimal example of a JSONL stream that renders a us
...
+    ]
+  }
+}
+```
+
+#### 2. Client Declares Supported Catalogs
+
+In **every** message sent to the server, the client includes an `a2uiClientCapabilities` object within the A2A `Message` metadata. This object informs the server of all catalogs the client can render.
+
+- `supportedCatalogIds` (array of strings, required): A list of identifiers for all pre-defined catalogs the client supports. The client MUST always include the standard catalog ID here.
```
wrenj: contents of these catalogs are expected to be compiled into the agent and not downloaded at runtime for predictability, security, etc

specification/0.8/docs/a2ui_extension_specification.md
wrenj: i see multiple below

whats the usecase you are envisioning here? not saying having an array is a problem just curious if it solves any problems not solved by a single inline catalog?

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -167,13 +166,100 @@ The following is a complete, minimal example of a JSONL stream that renders a us
...
+- `inlineCatalogs` (array of objects, optional): An array of full Catalog Definition Documents. This allows a client to provide custom, on-the-fly catalogs. This should only be provided if the server has advertised `acceptsInlineCatalogs: true`.
+
+**Example A2A Message with Client Capabilities:**
+```json
+{
+  "metadata": {
+    "a2uiClientCapabilities": {
+      "supportedCatalogIds": [
+        "a2ui.org:standard_catalog_0_8_0",
```
wrenj: I'm not opposed to using uris here, it could help debugging

I do like the flexible approach with ids though, eg google could embed the rev number google3/learning/agents/moma_catalog@12342322

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -167,13 +166,100 @@ The following is a complete, minimal example of a JSONL stream that renders a us
...
+        "my-company.com:custom-reporting-catalog-1.2"
+      ],
+      "inlineCatalogs": [
+        {
+          "catalogId": "my-company.com:temp-signature-pad-catalog",
```
wrenj: nice

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -167,13 +166,100 @@ The following is a complete, minimal example of a JSONL stream that renders a us
...
+    }
+  }
+}
+```
+
+#### 3. Server Chooses Catalog and Renders
+
+The server receives the client's capabilities and chooses a catalog to use for a specific UI surface. The server specifies its choice in the `beginRendering` message using the `catalogId` field.
+
+- `catalogId` (string, optional): The identifier of the chosen catalog. This ID must be one of the `supportedCatalogIds` or the `catalogId` from one of the `inlineCatalogs` provided by the client.
```
wrenj, line 253: should we require this? in your example if its not present, what does the client do? just assume the default catalog?

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -167,13 +166,100 @@ The following is a complete, minimal example of a JSONL stream that renders a us
...
+
+If the `catalogId` is omitted, the client **MUST** default to the standard catalog for the protocol version (`a2ui.org:standard_catalog_0_8_0`).
```
wrenj: is this helpful versus requiring it?

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -167,13 +166,100 @@ The following is a complete, minimal example of a JSONL stream that renders a us
...
+```json
+{
+  "beginRendering": {
+    "surfaceId": "unique-surface-1",
+    "catalogId": "my-company.com:custom-reporting-catalog-1.2",
+    "root": "root-component-id"
+  }
+}
+```
+Each surface can use a different catalog, providing a high degree of flexibility.
```
wrenj: , particularly in multi-agent systems where different agents may be using different catalogs.

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -167,13 +166,100 @@ The following is a complete, minimal example of a JSONL stream that renders a us
...
 
 #### Schemas for Developers
 
-When building an agent, it is recommended to use a resolved schema that includes the specific component catalog you are targeting (e.g., `server_to_client_with_standard_catalog.json`). This provides the LLM with a strict definition of all available components and their properties, leading to more reliable UI generation. The generic `server_to_client.json` is the abstract wire protocol, while the resolved schema is the concrete tool for generation.
+When building an agent, it is recommended to use a resolved schema that includes the specific component catalog you are targeting (e.g., a custom schema combining `server_to_client.json` with your `my-company.com:custom-reporting-catalog-1.2` definition). This provides the LLM with a strict definition of all available components and their properties, leading to more reliable UI generation. The generic `server_to_client.json` is the abstract wire protocol, while the resolved schema is the concrete tool for generation.
```
wrenj: add some description of how to build the final catalog 

heres the logic i used
https://github.com/google/A2UI/blob/main/a2a_agents/python/adk/samples/rizzcharts/component_catalog_builder.py#L80

specification/0.8/docs/a2ui_extension_specification.md
gspencergoog: I can totally see an ecosystem of inline catalogs evolving, available to client developers to include in their clients.  You could concatenate them, I suppose, but why not have a list?

specification/0.8/docs/a2ui_protocol.md
gspencergoog: Yeah, URIs are good because they can be self documenting, even if only used as identifiers. (e.g. https://json-schema.org/draft/2020-12/schema )

specification/0.8/docs/a2ui_extension_specification.md
wrenj: the agent picks a single catalog to use, the inline catalogs aren't concatenated, so the client would say here are 3 inline catalogs pick one which I don't understand

dmandar: My take 1. InlineCatalogs hopefully are only used in non-prod environments. The server agent permitting which catalog(s) in its agent card is the primary mechanism we should push for. 
2. For inline catalogs, one can imagine client developers keeping separate catalog files for each of their components. If we support that, then both the server and client need to have a long list of catalogs to manage. The client could collate components off of each catalog and send the list inline.

specification/0.8/docs/a2ui_protocol.md
dmandar: Why is the agent card option not sufficient here?

specification/0.8/docs/a2ui_protocol.md
dmandar: +1 to URIs.

specification/0.8/docs/a2ui_protocol.md
```diff
@@ -167,13 +166,105 @@ The following is a complete, minimal example of a JSONL stream that renders a us
...
+    "prompt": {
+      "text": "Find me a good restaurant"
+    }
+  }
+}
+```
+
+#### 3. Server Chooses Catalog and Renders
+
+The server receives the client's capabilities and chooses a catalog to use for a specific UI surface. The server specifies its choice in the `beginRendering` message using the `catalogId` field.
```
dmandar, line 251: i'm wondering if there should be a createSurface to be sent for each surface creation.

specification/0.8/docs/a2ui_extension_specification.md
jacobsimionato: I put multiple inline catalogs here because I figured it's consistent with advertising multiple well-known catalogs which an agent can pick from. If a client supports 5 different catalogs, shouldn't it also be able to advertise them all inline? Or should it have to pick only one of them to advertise inline, and advertise the rest only via ID?

Thinking about it a bit more, I can see how this is not especially likely to be used in practice, because how will agents know which catalog to pick? But I kind of still want to leave it here, because it doesn't add much complexity to the API, and I think it might even make the client SDKs simpler - you will just have a boolean like "advertiseCatalogsInline" rather than a setting which specific catalog of your many catalogs you'd like to advertise inline.

I agree with Mandar that this is mostly a mechanism for local development, so we don't have to think too hard about it.

How strongly does everyone feel? I'll remove this if there is consensus that it's overly complex!

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: @wrenj  Re requiring this: I want to add it as an optional parameter to beginRender in v0.8 and default to standard catalog, to make this change non-breaking. Otherwise all our agents theoretically need updating right away. In v0.9, let's make this required and explicit. I filed https://github.com/google/A2UI/issues/193 for this.

@dmandar Re the switch I've made here to having the client advertise multiple catalogs and have the server choose, instead of the inverse - see https://docs.google.com/document/d/1SppqNwHytVIUp7o8x5py5Z7AIiMdz-pzgC0SWh9qfkE/edit?tab=t.lwpqe4lgroyz#bookmark=id.p3dk9n24f9ja which was prompted by James's request above. I think it's important that the agent choose the catalog, because in a multi-agent system, the client may not know which subagent is going to reply, so it can't know which catalog to choose.

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: There is a rename of beginRendering to createSurface proposed in v0.9 right now. But let's not add this churn to v0.8.

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: I want to add it as an optional parameter to beginRender in v0.8 and default to standard catalog, to make this change non-breaking. Otherwise all our agents theoretically need updating right away. Filed https://github.com/google/A2UI/issues/193 to make it required for v0.9.

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: As above - non breaking for now.

specification/0.8/docs/a2ui_extension_specification.md
gspencergoog: I like it as a ~~list.~~ dictionary.

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: Updated.

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: I kept the name as 'supportedCatalogIds', but switched to suggesting URIs as a convention, and updated examples. Does that sound okay?

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: Added.

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: Added some example logic.

specification/0.8/docs/a2ui_protocol.md
wrenj: sgtm

specification/0.8/docs/a2ui_protocol.md
wrenj: tbh breaking changes are fine, nobody is really using in 1P, and updating the agent samples is easy, not sure about for other teams you work with

specification/0.8/docs/a2ui_extension_specification.md
wrenj: multiple inline catalogs sgtm

its not a list though its a dictionary right? of catalogId: inlineCatalogJson ?

gspencergoog: Yes, you're right, it's a dict.

jacobsimionato: The catalog_description_schema contains catalogId already, which I think makes sense - a catalog should declare it's Id. Now, inlineCatalogs references this schema. So I think it makes sense to just have an array of these, and each of them has an ID in it, rather than redundantly including the ID as a key as well. Yes, there could be duplicate Ids, and that would be an error.

WDYT?

specification/0.8/docs/a2ui_protocol.md
jacobsimionato: Hey that's fair. I guess I want to avoid creating more work trying to get all the renderers in sync for v0.8 too!

PR #142, opened 2025-11-24
Move Lit examples into their own lit/ directory

ava-cassiopeia:
An attempt at organizing the code a bit to distinguish shared web code from framework-specific web code. This is _some_ of the steps toward the final repo structure, hopefully putting us in a good position to update to the final structure once we finalize it.

`tests/` is top level in `web/` right now, which will probably eventually be folded into `lib/`, since it only seems to test `lib/` anyway.

This is setup for future work in which I'll try to extract shared code from the Angular and Lit code into a shared library.

**Reviewer hint:** The key code to review is the `package.json` files, which is where I made some small tweaks to refs with the move.

### DIFF
--- a/.github/workflows/editor_build.yml
+++ b/.github/workflows/editor_build.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
 
 jobs:
-  build-and-test:
+  build:
     runs-on: ubuntu-latest
 
     steps:


web/lib/src/0.8/ui/root.ts
```diff
@@ -33,7 +33,7 @@ import { StringValue } from "../types/primitives.js";
 import { AnyComponentNode, SurfaceID } from "../types/types";
 import { Theme } from "../types/types.js";
```
zbirkenbuel: Why the two different import statements from the same file?  AnyComponentNode, SurfaceID, and Theme are all in the same file.  Assuming you're trying to standardize on importing from .js files

ava-cassiopeia: Oop, good point I didn't catch that; fixed.

ava-cassiopeia:
I'm going to move ahead with this PR so that it doesn't end up becoming a merge conflict nightmare, but I'm happy to follow up on any further organization steps once we finalize the doc.

PR #145, opened 2025-11-24
Move web tests into lib (which is what they test)

ava-cassiopeia:
Moves the web test file into `web/lib/`, because it only tests the web lib's functionality to begin with and doesn't make sense as a separate project.

In the process I switched the type-checking strictness to true for the test file, which caught a typing error: it _seems_ like we want to have recursive value maps (the test implies this, anyway) so I updated the type to reflect that. However, that broke some other typecasting happening in `model-processor.ts`. I took a stab at resolving it, but would love some insight on the right way to approach this from someone who's been on this project longer.


web/lib/src/0.8/types/types.ts
```diff
@@ -263,19 +263,17 @@ export interface SurfaceUpdateMessage {
...
+}
+
+// ValueMap is a type of DataObject for passing to the data model.
+export type ValueMap = DataObject & {
+  key: string;
+  /** May be JSON */
+  valueString?: string;
+  valueNumber?: number;
+  valueBoolean?: boolean;
+  valueMap?: ValueMap[];
```
zbirkenbuel, line 276: Do you care that this can now be recursive to any depth where previously it wouldn't have been?

ava-cassiopeia: I _believe_ the existing tests imply that's a desired behavior (cc: @paullewis), but I can come back and fix this in a future PR if I'm wrong.

PR #154, opened 2025-11-26
Add 3P Chat Canvas working with A2A and A2UI to the angular sample projects folder. 

suyangw-g:
Changes include a shared Chat Canvas component and its demo app using the component.

Clients may use the ChatCanvas demo as a playground to test the A2UI integration in their A2A agents, with limitations in the A2A Parts that can be rendered by default.

Documentations are included for how to use the component and how to run the demo app.


angular/angular.json
```diff
@@ -190,6 +190,152 @@
           }
         }
       }
+    },
+    "a2a-chat-canvas": {
```
crisbeto, line 277: Why are we setting this up as another library? Are we planning on reusing it for other demos or publishing it to npm?

angular/projects/a2a-chat-canvas/src/lib/a2a-renderer/a2a-renderer.html
```diff
@@ -0,0 +1,25 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+@if (componentClassResource.hasValue()) {
+  @let componentClass = componentClassResource.value();
+
+  @if (componentClass) {
+    <ng-container
```
crisbeto, line 21: I was doing something similar when I was setting up the `a2ui-renderer`. It was a bit problematic, because:
1. All the components will get retained in the bundle, rather than only loading the relevant ones.
2. This adds an extra DOM node around each component which makes it more annoying to style.

angular/angular.json
suyangw-g: The primary reason for structuring this as a shared component is acceleration. We want to provide customer clients with a 'drop-in' Chat Canvas so they can stand up an A2UI-enabled agentic chat app application quickly for experimentation, rather than building the UI from scratch.

regarding the roadmap:

NPM: We are architecting it so it can be published to npm in the future to simplify distribution, though we aren't doing that immediately.

Playground: The demo app is intended to serve as a playground for 3P clients (and potentially for internal integrations like Gemini for Enterprise) to test their agents.

As adoption grows, we plan to treat this as a standard library with full unit test coverage and versioning."

angular/projects/a2a-chat-canvas/src/lib/a2a-renderer/a2a-renderer.html
suyangw-g: 1. If you look at one of the component used here angular/projects/a2a-chat-canvas/src/lib/a2a-renderer/catalog/default-text-part/renderer-config.ts, I thought they have already been set up to be lazy-loaded.
2. pushed a new commit to make a2a-renderer removed from the layout tree

PR #167, opened 2025-12-01
Decouple Lit UI component catalog from the core library.

sugoi-yuzuru:
Before the change, the coupling caused conflicting pointers to ui components causing Angular apps rendered on browsers to get confused whether to render the angular compnent of the Lit component. By decoupling the library, this change effectively omits the custom components defined for the Lit app to creep into the Angular app's compilation.

A new file `web/lib/src/0.8/core.ts` is introduced as the core library for Angular apps to reference. The content includes everything that `web/lib/src/0.8/index.ts` contained EXCEPT for the UI package which includes the Lit component catalog. Importing these UI components into the Angular app can cause conflict between the Angular components and Lit components causing the UI rendering to break. 

All components in the Angular is fixed to reference the core library specifically instead of the index library which includes the UI.

### DIFF
--- a/angular/projects/a2a-chat-canvas/src/lib/a2a-renderer/catalog/a2ui-data-part/a2ui-data-part.ts
+++ b/angular/projects/a2a-chat-canvas/src/lib/a2a-renderer/catalog/a2ui-data-part/a2ui-data-part.ts
@@ -20,7 +20,7 @@ import { ChatService } from '@a2a_chat_canvas/services/chat-service';
 import { UiMessageContent } from '@a2a_chat_canvas/types/ui-message';
 import { isA2aDataPart } from '@a2a_chat_canvas/utils/type-guards';
 import { Surface } from '@a2ui/angular';
-import { v0_8 } from '@a2ui/web-lib';
+import * as v0_8 from '@a2ui/web-lib/0.8';
 import { Component, computed, inject, input } from '@angular/core';
 
 /**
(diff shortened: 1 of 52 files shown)


angular/projects/a2a-chat-canvas/src/lib/a2a-renderer/catalog/a2ui-data-part/a2ui-data-part.ts
```diff
@@ -20,7 +20,7 @@ import { ChatService } from '@a2a_chat_canvas/services/chat-service';
 import { UiMessageContent } from '@a2a_chat_canvas/types/ui-message';
 import { isA2aDataPart } from '@a2a_chat_canvas/utils/type-guards';
 import { Surface } from '@a2ui/angular';
-import { v0_8 } from '@a2ui/web-lib';
+import * as v0_8 from '@a2ui/web-lib/core';
```
crisbeto: I think we should rework it so that the import path is `@a2ui/web-lib/0.8`. That way bundlers will be able to tree shake the versions not used by the current app, if we were to publish this to npm.

sugoi-yuzuru: Thank you!
I made a small adjustment to make the versioning clearer in the node package.

PR #169, opened 2025-12-01
Show programmatic orchestrator routing for userAction based on surfaceId

wrenj:


a2a_agents/python/adk/samples/orchestrator/agent.py
```diff
@@ -37,12 +42,42 @@ class OrchestratorAgent:
...
+    @classmethod
+    async def programmtically_route_user_action_to_subagent(
+        cls,
+        callback_context: CallbackContext,
+        llm_request: LlmRequest,
+    ) -> LlmResponse:
+        if (
+            llm_request.contents
+            and (last_content := llm_request.contents[-1]).parts
+            and (a2a_part := part_converters.convert_genai_part_to_a2a_part(last_content.parts[-1]))
```
jonsharkey, line 54: Is the userAction always the last part?

wrenj: yeah it comes in as the only part. if the client batches multiple user actions together this will break since we only do one transfer_to_agent

PR #189, opened 2025-12-02
Simplify property specification 

gspencergoog:
# Description

This simplifies the property specification now that we're no longer afraid of `oneOf` in the schema.

- Instead of needing `literalString`, `literalNumber`, etc., we just use the actual values.
- Instead of `explicitList` in the children lists, we just specify the list of IDs.

It's more readable, and I'm getting better eval results from it, even on Flash Lite.


specification/0.9/eval/src/validator.ts
```diff
@@ -64,7 +64,7 @@ export function validateSchema(
   if (data.updateComponents) {
     validateSurfaceUpdate(data.updateComponents, errors);
   } else if (data.updateDataModel) {
-    validateDataModelUpdate(data.updateDataModel, errors);
+    validateupdateDataModel(data.updateDataModel, errors);
```
jacobsimionato:
```suggestion
    validateUpdateDataModel(data.updateDataModel, errors);
```

specification/0.9/json/standard_catalog_definition.json
```diff
@@ -356,15 +358,20 @@
...
                     "description": "Defines the value to be included in the context as either a literal value or a path to a data model value (e.g. '/user/name').",
-                    "properties": {
-                      "path": { "type": "string" },
-                      "literalString": { "type": "string" },
-                      "literalNumber": { "type": "number" },
-                      "literalBoolean": { "type": "boolean" }
-                    },
-                    "additionalProperties": false
+                    "oneOf": [
+                      { "type": "string" },
```
jacobsimionato, line 363: I wonder if we should try to support an arbitrary static context object here, so that LLMs can bake any structured data into the action context here. Could it work to set additionalProperties to true on the object, and then renderer logic can do something like:
- If value.type == object AND value.properties.contains("path")
     => Do data model lookup
- Else
-   => Include literal value

I dunno, maybe that contract is too unclear

I think what you propose here is great and a strict improvement though, because we didn't have this previously, so let's merge and consider future improvements.

gspencergoog: Yeah, I worry that the LLM will inadvertently generate an object that has a "path" attribute that has nothing to do with data binding. (e.g. some kind of directions widget where "path" describes how to get somewhere).

PR #192, opened 2025-12-03
Add LLM-evaluated evals for 0.9

gspencergoog:
# Description

This adds a clone of the eval script that instead of using algorithmic evaluations, uses an LLM to evaluate the results.

I think I'm seeing better evaluations now: the algorithmic evaluations are subject to problems with the expectations not being flexible enough to take into account things like the LLM adding an extra label or slightly different construction to the output, and didn't verify a lot of things.  This does, although it suffers the opposite problem of hallucinations in the evaluation step, or the evaluator being capricious and making up requirements.  I've tuned the evaluation prompt some, but it could probably use more.  I'm seeing about an 80-90% success rate on Gemini Flash, and when I look at the non-schema failures, they are mostly small failures (e.g. "the output uses a Column when the prompt requested a Row").

I kept this as a separate script for now, but I could just replace the existing evaluator with it if you think it's good enough.

### DIFF
--- a/specification/0.9/eval/src/evaluation_flow.ts
+++ b/specification/0.9/eval/src/evaluation_flow.ts
@@ -42,6 +42,7 @@ export const evaluationFlow = ai.defineFlow(
           })
         )
         .optional(),
+      evalPrompt: z.string().optional(),
     }),
   },
   async ({ originalPrompt, generatedOutput, evalModel, schemas }) => {
@@ -178,6 +179,7 @@ Return a JSON object with the following schema:
         pass: result.pass,
         reason: result.reason || "No reason provided",
         issues: result.issues || [],
+        evalPrompt: evalPrompt,
       };
     } catch (e: any) {
       logger.error(`Error during evaluation: ${e}`);
--- a/specification/0.9/eval/src/evaluator.ts
+++ b/specification/0.9/eval/src/evaluator.ts
@@ -140,6 +140,7 @@ export class Evaluator {
       pass: boolean;
       reason: string;
       issues?: { issue: string; severity: IssueSeverity }[];
+      evalPrompt?: string;
     },
     overallSeverity?: IssueSeverity
   ) {
@@ -160,5 +161,15 @@ export class Evaluator {
       ),
       yaml.dump({ ...evaluationResult, overallSeverity })
     );
+
+    if (evaluationResult.evalPrompt) {
+      fs.writeFileSync(
+        path.join(
+          detailsDir,
+          `${result.prompt.name}.${result.runNumber}.eval_prompt.txt`
+        ),
+        evaluationResult.evalPrompt
+      );
+    }
   }
 }
--- a/specification/0.9/eval/src/index.ts
+++ b/specification/0.9/eval/src/index.ts
@@ -274,7 +274,7 @@ async function main() {
     .option("eval-model", {
       type: "string",
       description: "Model to use for evaluation",
-      default: "gemini-2.5-flash-lite",
+      default: "gemini-2.5-flash",
       choices: modelsToTest.map((m) => m.name),
     })
     .option("clean-results", {
--- a/specification/0.9/eval/src/types.ts
+++ b/specification/0.9/eval/src/types.ts
@@ -27,5 +27,6 @@ export interface EvaluatedResult extends ValidatedResult {
     reason: string;
     issues?: { issue: string; severity: IssueSeverity }[];
     overallSeverity?: IssueSeverity;
+    evalPrompt?: string;
   };
 }
--- a/specification/0.9/json/common_types.json
+++ b/specification/0.9/json/common_types.json
@@ -68,7 +68,7 @@
     },
     "weight": {
       "type": "number",
-      "description": "The relative weight of this component within a Row or Column. This corresponds to the CSS 'flex-grow' property. Note: this may ONLY be set when the component is a direct descendant of a Row or Column."
+      "description": "The relative weight of this component within a Row or Column. This is similar to the CSS 'flex-grow' property."
     },
     "contextValue": {
       "description": "A value that can be a string, number, boolean, or a path to a value.",


specification/0.9/eval_llm/src/evaluation_flow.ts
```diff
@@ -0,0 +1,165 @@
...
+
+- You can be lenient in your evaluation for URLs, as the generated output may use a placeholder URL for images and icons.
+- If label text is similar but not exact, you can still pass the test as long as the meaning is the same. (e.g. "Cancel" vs "Cancel Order")
+- If the generated output is missing a component that is specified in the user request, it is required to exist in the output in order to pass the test. If it is not specified, it is not required.
+- If the request is vague about the contents of a label or other property, you can still pass the test as long as it can be construed as matching the intent.
+
+Return a YAML object with the following structure:
+
+\`\`\`yaml
+pass: boolean  # Whether the generated UI meets the requirements
```
jacobsimionato: Why not structured output?

And perhaps this could be a list of failures, to make it easier to understand if there are multiple issues? We could even define an enum to get the LLM to categorise each issue it finds, with an "other" option as an escape hatch.

gspencergoog:
> I'd be comfortable replacing the existing tool if you think this is more helpful, and if it's clear enough in the results the difference between validation failures and evaluation failures. That way, we can take the evaluation failures with a pinch of salt. It might be nice to try to break each of them down further into categories, so you can do a big eval run and see "3 missing content issues, 2 invalid JSON, 4 invalid component references".

I already do break out the schema validation errors vs the LLM evaluation failures, but I have eliminated the matchers we had before, so the schema evaluation is strictly the schema (well, I added a check for the "root" component ID, but that is special). I eliminated the matchers because they were generating false failures anyhow, so I'm not sure adding them back in would help.

I could easily add an inference that summarizes the failures, since I already have failure descriptions computed at that point.

specification/0.9/eval_llm/src/evaluation_flow.ts
gspencergoog: Sure, that's a good idea.  I did have a JSON schema here, but I found that it was harder to read the JSON output than it was YAML.  But we can do JSON output here and just save the files as YAML anyhow (which I'm doing anyhow).

specification/0.9/eval_llm/src/prompts.ts
```diff
@@ -0,0 +1,382 @@
...
+  {
+    name: "deleteSurface",
+    description: "A DeleteSurface message to remove a UI surface.",
+    promptText: `Generate a JSON message containing a deleteSurface for the surface 'dashboard-surface-1'.`,
+  },
+  {
+    name: "dogBreedGenerator",
+    description:
+      "A prompt to generate a UI for a dog breed information and generator tool.",
+    promptText: `First, generate a 'createSurface' message with surfaceId 'main'. Then, generate a 'updateComponents' message to describe the following UI:
```
jacobsimionato: How about having an "evaluationNotes" parameter here, which can include a list of points to consider in evaluation, which is passed to the LLM-as-a-judge. I imagine having that lever could help us de-noise the evals if certain scenarios are being misinterpreted by the judge? I guess we can just add it later if we need it!

jacobsimionato:
> > I'd be comfortable replacing the existing tool if you think this is more helpful, and if it's clear enough in the results the difference between validation failures and evaluation failures. That way, we can take the evaluation failures with a pinch of salt. It might be nice to try to break each of them down further into categories, so you can do a big eval run and see "3 missing content issues, 2 invalid JSON, 4 invalid component references".
> 
> I already do break out the schema validation errors vs the LLM evaluation failures, but I have eliminated the matchers we had before, so the schema evaluation is strictly the schema (well, I added a check for the "root" component ID, but that is special). I eliminated the matchers because they were generating false failures anyhow, so I'm not sure adding them back in would help.
> 
> I could easily add an inference that summarizes the failures, since I already have failure descriptions computed at that point.

This makes sense - I'm happy to replace the old evals if you feel this is generally working better. Yep, I agree we were getting false negatives from the matchers because they were not flexible enough.

Re categorizing errors - I wonder if it could help to get the LLM judge to output errors according to category. That might help us to interpret the results if there are certain classes of error that have more ambiguity and therefore false positives/negatives etc. I'm not sure because I haven't seen the output, but I remember you said you're seeing like 10-20% eval failure which is way more than before, so I assume there are false negatives here. I think we should try to get back to 95%+ pass rate for validation+evaluation before ratifying v0.9, but we can do that incrementally.

gspencergoog:
> Re categorizing errors - I wonder if it could help to get the LLM judge to output errors according to category. That might help us to interpret the results if there are certain classes of error that have more ambiguity and therefore false positives/negatives etc. I'm not sure because I haven't seen the output, but I remember you said you're seeing like 10-20% eval failure which is way more than before, so I assume there are false negatives here. I think we should try to get back to 95%+ pass rate for validation+evaluation before ratifying v0.9, but we can do that incrementally.

Yes, there are definitely false negatives.  A lot of times it's just the LLM being pedantic and the resulting UI would look fine.

I'm going to add an LLM summary of the failures, including categorizing the failures into rough groups. I'm also going to add a "severity" field to the evaluation so that it can rate the severity of the issues it finds so we can rank them as "minor, major, and critical" and then tally those up.

I think I will swap the existing eval out for this one.  I don't think the other one scales as well, and it's harder to maintain, even if this one is a bit fuzzier.

specification/0.9/eval_llm/src/prompts.ts
gspencergoog: Nice idea.  I'll add that!

gspencergoog:
An example of the new output:

# Evaluation Summary

## Model: gemini-2.5-flash

| Prompt Name                              | Avg Latency (ms)     | Schema Fail     | Eval Fail       | Minor           | Significant     | Critical        |
|------------------------------------------|----------------------|-----------------|-----------------|-----------------|-----------------|-----------------|
| animalKingdomExplorer                    | 16321                |                 |                 |                 |                 |                 |
| calendarEventCreator                     | 4315                 |                 | 1 / 1           |                 | 1               |                 |
| chatRoom                                 | 4260                 | 1 / 1           |                 |                 |                 |                 |
| checkoutPage                             | 4079                 |                 |                 |                 |                 |                 |
| cinemaSeatSelection                      | 5256                 |                 | 1 / 1           | 1               |                 |                 |
| contactCard                              | 3334                 |                 |                 |                 |                 |                 |
| courseSyllabus                           | 4157                 |                 |                 |                 |                 |                 |
| dashboard                                | 3823                 |                 |                 |                 |                 |                 |
| deleteSurface                            | 1273                 |                 |                 |                 |                 |                 |
| dogBreedGenerator                        | 7188                 |                 |                 |                 |                 |                 |
| eCommerceProductPage                     | 5246                 |                 |                 |                 |                 |                 |
| fileBrowser                              | 3804                 |                 |                 |                 |                 |                 |
| fitnessTracker                           | 6847                 |                 | 1 / 1           |                 | 1               |                 |
| flashcardApp                             | 4334                 |                 |                 |                 |                 |                 |
| flightBooker                             | 3806                 |                 |                 |                 |                 |                 |
| hotelSearchResults                       | 5141                 |                 |                 |                 |                 |                 |
| interactiveDashboard                     | 5054                 |                 |                 |                 |                 |                 |
| jobApplication                           | 3507                 |                 |                 |                 |                 |                 |
| kanbanBoard                              | 4489                 |                 |                 |                 |                 |                 |
| loginForm                                | 3110                 |                 |                 |                 |                 |                 |
| musicPlayer                              | 3992                 |                 |                 |                 |                 |                 |
| nestedDataBinding                        | 5777                 |                 | 1 / 1           |                 |                 | 1               |
| newsAggregator                           | 5392                 |                 |                 |                 |                 |                 |
| notificationCenter                       | 3579                 |                 |                 |                 |                 |                 |
| photoEditor                              | 3310                 |                 |                 |                 |                 |                 |
| podcastEpisode                           | 4116                 |                 |                 |                 |                 |                 |
| productGallery                           | 3115                 |                 | 1 / 1           |                 |                 | 1               |
| productGalleryData                       | 1729                 |                 |                 |                 |                 |                 |
| profileEditor                            | 3952                 |                 |                 |                 |                 |                 |
| recipeCard                               | 4553                 |                 |                 |                 |                 |                 |
| restaurantMenu                           | 5793                 |                 |                 |                 |                 |                 |
| settingsPage                             | 4938                 |                 |                 |                 |                 |                 |
| simpleCalculator                         | 9586                 |                 |                 |                 |                 |                 |
| smartHome                                | 3106                 |                 |                 |                 |                 |                 |
| socialMediaPost                          | 4199                 |                 |                 |                 |                 |                 |
| stockWatchlist                           | 4219                 |                 |                 |                 |                 |                 |
| surveyForm                               | 4538                 |                 |                 |                 |                 |                 |
| travelItinerary                          | 9134                 |                 | 1 / 1           |                 | 1               |                 |
| triviaQuiz                               | 3307                 |                 | 1 / 1           | 1               |                 | 1               |
| updateDataModel                          | 1685                 |                 |                 |                 |                 |                 |
| videoCallInterface                       | 3497                 |                 |                 |                 |                 |                 |
| weatherForecast                          | 7263                 |                 |                 |                 |                 |                 |

**Total successful runs:** 34 / 42 (81.0% success)

### Failure Analysis

## Analysis of `gemini-2.5-flash` Failures

### Error Types

*   **Schema Validation:** 1
*   **Evaluation Failure:** 7

### Failure Patterns

*   **Incorrect property usage/missing properties:** The model often misinterprets how properties should be applied or fails to include required properties.
    *   *Example:* In `chatRoom`, there are numerous "must NOT have additional properties" and "must have required property" errors.
    *   *Example:* In `productGallery`, `productId` was expected as a literal string but was a path.
    *   *Example:* In `travelItinerary`, `CheckBox` components had empty `label` properties instead of activity descriptions.
*   **Initialization/Data Model Misinterpretations:** The model struggles with correctly initializing components or updating the data model as requested.
    *   *Example:* In `calendarEventCreator`, `DateTimeInput` components were not initialized with empty strings in `updateDataModel`.
*   **Structural/Format Issues:** The model occasionally produces malformed output or misunderstands compositional structures.
    *   *Example:* In `triviaQuiz`, the output contained two separate JSON objects instead of a single array.
    *   *Example:* In `triviaQuiz`, the `Button`'s `child` property was misinterpreted to *be* the text instead of *containing* it.
*   **Missing elements/messages:** The model can omit entire requested elements or messages.
    *   *Example:* In `nestedDataBinding`, a requested `updateDataModel` message was entirely missing.
*   **UI Layout/Styling nuances:** The model sometimes misses subtle UI layout or styling requests.
    *   *Example:* In `fitnessTracker`, the `Slider` was missing a visible label.
    *   *Example:* In `cinemaSeatSelection`, the 'Screen' text component was not explicitly centered.

### Summary

The `gemini-2.5-flash` model, with 34 successes and 8 failures, demonstrates a generally strong performance. However, the failures indicate a recurring challenge with adhering to specific schema requirements, particularly regarding property definitions and the distinction between literal values and data model paths. The model also occasionally struggles with correctly initializing component values, producing properly formatted multi-object JSON outputs, and interpreting implicit UI layout requests such as labels for sliders or explicit centering for text. These issues point towards areas where the model could improve its understanding of component catalog specifications and the nuances of user interface construction.

---

## Overall Summary

- **Total tool failures:** 0 / 42
- **Number of runs with any failure (tool error, validation, or eval):** 8 / 42 (81.0% success)
- **Severity Breakdown:**
  - **Minor:** 2
  - **Significant:** 3
  - **Critical (Eval):** 3
  - **Critical (Schema):** 0
- **Mean Latency:** 4765 ms
- **Median Latency:** 4209 ms
- **Models with at least one failure:** gemini-2.5-flash

Issue #193, opened 2025-12-03
Ensure custom catalog negotiation API is adapted and updated for v0.9


jacobsimionato:
We are specifying catalog negotiation in v0.8 - see https://github.com/google/A2UI/pull/136

In v0.9, we should make sure this is copied across with some potential changes:
- Wherever the catalog ID is specified by the server, let's make it required and explicit, rather than optional with a default to the standard catalog
- Consider where it should be specified.
  - Option 1: Put it on createSurface alongside "theme", require every surface to use exactly one catalog
  - Option 2: Remove "theme", remove createSurface completely, put catalogId on surfaceUpdate (or even on each component?) and allow components from different catalogs to intermingle together on a surface

jacobsimionato:
CC @gspencergoog

gspencergoog:
I want to go with option 2, and add catalogId to componentUpdate.

gspencergoog:
I worry a little that each componentUpdate could have a different catalog ID, but I guess as long as the client supports them all they might play well together.

jacobsimionato:
Lol "might" play well together. Yeah, bad interactions are also a fear of mine, but I figure now that we're getting rid of the agent theming (hopefully!), there is a possibility. I'm a little inspired by Yegor's idea of an ecosystem of catalog extensions for charting etc. Previously I pushed against that (motivated by worries about the theming complexity), but if we can eliminate themes, maybe it would actually be awesome.

Issue #170, opened 2025-12-01
Add new fields to the clientToServer "error" message to support custom events use case


jacobsimionato:
See internal doc: https://docs.google.com/document/d/1uAp0lx4eYVF3QdIz2dAogjpzAOuLgAyvNU7CYMnzS1E/edit?tab=t.0#bookmark=id.jfqtw2t18he7

Draft proposal from the doc (see **doc for source of truth**!!):

```json
{
  "title": "Error",
  "description": "Reports a client-side error.",
  "type": "object",
  "properties": {
    "code": {
      "title": "Code",
      "description": "A stable, machine-readable string representing the error type (e.g., 'INVALID_ARGUMENT', 'NOT_FOUND', 'INTERNAL_ERROR').",
      "type": "string"
    },
    "message": {
      "title": "Message",
      "description": "A short description of the error. The message SHOULD be limited to a concise single sentence.",
      "type": "string"
    },
    "data": {
      "title": "Data",
      "description": "Additional information about the error. The value of this member is defined by the sender (e.g. detailed error information, nested errors etc.).",
      "type": "object",
      "additionalProperties": true
    },
    "surfaceId": {
      "type": "string",
      "description": "A unique identifier for the UI surface relating to the event."
    },
  },
  "required": [
    "code",
    "message"
  ]
}
```

jacobsimionato:
CC @wrenj

jacobsimionato:
CC @gspencergoog

PR #194, opened 2025-12-03
Refine A2UI 0.9 Protocol: Remove `createSurface`, standardize bindings, and update schemas

gspencergoog:
# Description

This PR introduces significant refinements to the A2UI 0.9 protocol, focusing on simplifying the surface lifecycle, standardizing data binding paths, and optimizing component properties. These changes are **BREAKING** and require updates to both clients and servers.

### Protocol Lifecycle
-   **`createSurface` Removed**: The `createSurface` message type has been completely removed from the protocol. Clients must now handle surface creation implicitly or via other means when receiving an `updateComponents` message.
-   **`updateComponents` Updates**:
    -   **Required `catalogId`**: The `updateComponents` message now requires a `catalogId` property (URI string) to identify the component catalog definition used for the update.

### Schema & Components
-   **Data Binding Standardization**:
    -   Renamed `dataBinding` to `path` in `common_types.json` and `childrenProperty` templates. This unifies the syntax for referencing data model paths across the protocol.
-   **Button Component**:
    -   **Context Refactor**: The `context` property in `Button` actions is now a standard JSON object (map) instead of an array of key-value pairs. This improves token efficiency and developer ergonomics.
-   **TextField Component**:
    -   **Property Renamed**: `textFieldType` has been renamed to `usageHint` to align with the naming convention used in `Text` and `Image` components.

### Client-to-Server Messages
-   **Error Format**:
    -   **Renamed `type` to `code`**: In the `ValidationFailed` error object, the `type` property has been renamed to `code` and must be `"VALIDATION_FAILED"`.
    -   **Added `surfaceId`**: The `ValidationFailed` error now requires a `surfaceId` property.
    -   **Updated `path`**: The `path` property in `ValidationFailed` errors is now explicitly defined as a JSON Pointer (e.g., `/components/0/props/text`).

### Protocol & Schemas (`json/`)
-   **`server_to_client.json`**:
    -   Removed `createSurface` definition.
    -   Added `catalogId` to `updateComponents`.
-   **`client_to_server.json`**:
    -   Renamed `type` to `code` in `ValidationFailed` error.
    -   Added `surfaceId` to `ValidationFailed` error.
    -   Updated `path` description to specify JSON Pointer.
-   **`common_types.json`**:
    -   Renamed `dataBinding` to `path`.
-   **`standard_catalog_definition.json`**:
    -   Updated `Button` and `TextField` definitions as described above.

### Documentation (`docs/`)
-   **`a2ui_protocol.md`**:
    -   Updated the "Protocol Overview" to remove `createSurface`.
    -   Updated message definitions and examples to reflect all schema changes.
    -   Added a note clarifying the new surface lifecycle expectations.
    -   Updated the "Example Stream" and Mermaid diagrams.
    -   Updated the "Standard Validation Error Format" section to match the new `client_to_server.json` schema.

### Evaluation Tools (`eval/`)
-   **`validator.ts`**:
    -   Removed `createSurface` validation logic.
    -   Added validation for `catalogId` in `updateComponents`.
    -   Updated validation logic to support the new `path` property and `Button` context structure.
    -   Integrated `ajv-formats` to support URI format validation.
-   **`prompts.ts` & `flows.ts`**:
    -   Updated all prompts to generate `updateComponents` messages with `catalogId`.
    -   Removed instructions to generate `createSurface`.
    -   Updated system prompts to enforce the new protocol rules.

## Verification
-   Ran `pnpm run eval` with `gemini-2.5-flash` and `gemini-2.5-flash-lite` to verify that the updated prompts and validator work correctly.
-   Verified that the generated JSON matches the new schema requirements (e.g., `catalogId` presence, correct component properties).

Fixes https://github.com/google/A2UI/issues/193
Fixes https://github.com/google/A2UI/issues/170

### DIFF
--- a/specification/0.9/json/client_to_server.json
+++ b/specification/0.9/json/client_to_server.json
@@ -79,36 +79,49 @@
       ]
     },
     "error": {
-      "description": "Reports a client-side error. The content is flexible, but 'ValidationFailed' errors have a strict structure.",
+      "description": "Reports a client-side error.",
       "oneOf": [
         {
           "type": "object",
           "title": "Validation Failed Error",
           "properties": {
-            "type": {
-              "const": "ValidationFailed"
+            "code": {
+              "const": "VALIDATION_FAILED"
+            },
+            "surfaceId": {
+              "type": "string",
+              "description": "The id of the surface where the error occurred."
             },
             "path": {
               "type": "string",
-              "description": "The path to the invalid field (e.g. 'components[0].props.text')."
+              "description": "The JSON pointer to the field that failed validation (e.g. '/components/0/props/text')."
             },
             "message": {
               "type": "string",
-              "description": "A description of why validation failed."
+              "description": "A short one-sentence description of why validation failed."
             }
           },
-          "required": ["type", "path", "message"]
+          "required": ["code", "path", "message", "surfaceId"]
         },
         {
           "type": "object",
           "title": "Generic Error",
           "properties": {
-            "type": {
+            "code": {
               "not": {
-                "const": "ValidationFailed"
+                "const": "VALIDATION_FAILED"
               }
+            },
+            "message": {
+              "type": "string",
+              "description": "A short one-sentence description of why the error occurred."
+            },
+            "surfaceId": {
+              "type": "string",
+              "description": "The id of the surface where the error occurred."
             }
           },
+          "required": ["code", "surfaceId", "message"],
           "additionalProperties": true
         }
       ]
--- a/specification/0.9/docs/a2ui_protocol.md
+++ b/specification/0.9/docs/a2ui_protocol.md
@@ -423,17 +423,19 @@ This loop allows for a high degree of flexibility and robustness, as the system 
 
 If validation fails, the client (or the system acting on behalf of the client) should send an `error` message back to the LLM. To ensure the LLM can understand and correct the error, use the following standard format within the `error` message payload:
 
-- `type` (string, required): Must be `"ValidationFailed"`.
-- `path` (string, required): A JSONPath or dot-notation string indicating the location of the error (e.g., `components[2].props.text`).
-- `message` (string, required): A human-readable description of the validation error.
+- `code` (string, required): Must be `"VALIDATION_FAILED"`.
+- `surfaceId` (string, required): The ID of the surface where the error occurred.
+- `path` (string, required): The JSON pointer to the field that failed validation (e.g. `/components/0/props/text`).
+- `message` (string, required): A short one-sentence description of why validation failed.
 
 **Example Error Message:**
 
 ```json
 {
   "error": {
-    "type": "ValidationFailed",
-    "path": "components[2].props.text",
+    "code": "VALIDATION_FAILED",
+    "surfaceId": "user_profile_card",
+    "path": "/components/0/props/text",
     "message": "Expected stringOrPath, got integer"
   }
 }


specification/0.9/json/client_to_server.json
```diff
@@ -79,9 +79,39 @@
...
+            }
+          },
+          "required": ["type", "path", "message"]
+        },
+        {
+          "type": "object",
+          "title": "Generic Error",
+          "properties": {
+            "type": {
+              "not": {
```
jacobsimionato, line 111: Ha, this is so cool!

specification/0.9/json/client_to_server.json
```diff
@@ -79,9 +79,39 @@
...
+            "message": {
+              "type": "string",
+              "description": "A description of why validation failed."
+            }
+          },
+          "required": ["type", "path", "message"]
+        },
+        {
+          "type": "object",
+          "title": "Generic Error",
```
jacobsimionato, line 108: Could we add code, message, data, surfaceId as proposed in https://github.com/google/A2UI/issues/170 ?

I wonder if, given this structure, it makes sense to identify validation errors using a convention within that structure e.g. a specially documented code or something? I don't have a super strong feeling on this. I guess it maybe depends on whether the errors are expected to just be reported and monitored (only vague structure needed) or be used for error correction etc (strong contract needed). I figure error detection and fixing on the server side can be implemented entirely on the server side, so hopefully these are just used for reporting.

specification/0.9/json/server_to_client.json
```diff
@@ -4,27 +4,6 @@
   "description": "Describes a JSON payload for an A2UI (Agent to UI) message, which is used to dynamically construct and update user interfaces.",
   "type": "object",
   "oneOf": [
-    {
-      "properties": {
```
jacobsimionato, line 8: So great to see this disappear!

specification/0.9/json/standard_catalog_definition.json
```diff
@@ -416,7 +396,7 @@
           "$ref": "common_types.json#/$defs/stringOrPath",
           "description": "The value of the text field."
         },
-        "textFieldType": {
+        "usageHint": {
           "type": "string",
           "description": "The type of input field to display.",
           "enum": ["date", "longText", "number", "shortText", "obscured"]
```
jacobsimionato, line 402: Random thought: Why do we need "date" type here when we have a DateTimeInput below?

specification/0.9/json/common_types.json
```diff
@@ -79,14 +93,17 @@
...
         {
           "type": "object",
-          "description": "A template for generating a dynamic list of children from a data model list. `componentId` is the component to use as a template, and `dataBinding` is the path to the map of components in the data model. Values in the map will define the list of children.",
+          "description": "A template for generating a dynamic list of children from a data model list. The `componentId` is the component to use as a template.",
           "properties": {
             "componentId": {
               "$ref": "#/$defs/id"
             },
-            "dataBinding": { "type": "string" }
+            "path": {
```
jacobsimionato, line 101: I wonder if this should be a shared schema, just to centralize the description of "path". E.g. I see booleanOrPath and stringArrayOrPath don't have any description of the path field above.

Maybe this is taking things too far though, with too much indirection. Maybe we just have a prompt fragment that describes some overall A2UI concepts like this.

specification/0.9/json/client_to_server.json
gspencergoog: > Could we add code, message, data, surfaceId as proposed in https://github.com/google/A2UI/issues/170 ?

Sure!

I think that an LLM will be involved in pretty much all error correction, so as long as the error condition is clear it should be able to fix the error.  I'm not clear on how much original context is needed to fix the errors though, but I guess that's agent dependent? Probably the orchestrator will need to send the error back to the original agent to handle and fix.

specification/0.9/json/standard_catalog_definition.json
gspencergoog: Yeah, I guess so, although I think maybe this is mean to be a "dd.mm..yyyy" formatted input field, and the other could be more of a date picker.  It does seem kind of redundant.

specification/0.9/json/common_types.json
gspencergoog: I think it's OK to keep the paths in each definition.  Too many levels of indirection might affect quality.

specification/0.9/json/client_to_server.json
gspencergoog: Okay, I added in surfaceId, message and changed "type" to "code".  I didn't add data because the generic error has additionalProperties:true on it and so they can add any data they want to.

PR #197, opened 2025-12-04
Update typescript style

yuxuanzeng:
Remove accessor
Replace # with private
explicit override


jacobsimionato:
Sorry this never made it in. It looks like the issues were fixed in other PRs.

PR #203, opened 2025-12-04
create generic avatar icon

wrenj:


angular/package-lock.json
```diff
@@ -515,6 +515,7 @@
       "version": "20.2.14",
```
jonsharkey, line 515: Please revert this file.

PR #216, opened 2025-12-05
Create Python build GH workflow

ava-cassiopeia:
**WIP**


ava-cassiopeia:
Gonna abandon this for now because of some of the other Python build issues we've discovered.

PR #226, opened 2025-12-05
Reintroduce createSurface, add evolution guide

gspencergoog:
feat!: Refactor protocol with createSurface, flattened props, and data ops

This PR introduces major structural changes to the A2UI protocol to improve efficiency, clarity, and extensibility.

### 🚀 Features

-   **Reintroduced `createSurface` Message**: Introduced a dedicated `createSurface` message to initialize a surface with a specific `catalogId`. This separates surface creation from content updates.
-   **Flattened Component Structure**: Removed the `props` wrapper object. Component type (`component`) and properties (e.g., `text`, `children`) are now top-level fields alongside `id` and `weight`. This reduces JSON nesting and token usage.
-   **Operation-Based Data Updates**: Updated `updateDataModel` to use an `op` (e.g., `replace`) and `value` structure, aligning with standard patch patterns.

### 💥 Breaking Changes

-   **Protocol**: The `updateComponents` message no longer accepts `catalogId`. It must be preceded by a `createSurface` message.
-   **Protocol**: The component structure is flattened. `props` is removed.
    -   *Old*: `{"id": "1", "props": {"component": "Text", "text": "Hello"}}`
    -   *New*: `{"id": "1", "component": "Text", "text": "Hello"}`
-   **Protocol**: `updateDataModel` `contents` property is replaced by `op` and `value`.
-   **Protocol**: `clientUiCapabilities` message removed from client-to-server schema.

### 📝 Documentation

-   Updated `specification/0.9/docs/a2ui_protocol.md` to reflect the new schema and message flows.
-   Added `specification/0.9/docs/evolution_guide.md` to document the protocol's evolution and changes.

### 🛠️ Evaluation & Tooling

-   Updated `validator.ts`, `prompts.ts`, and flow scripts to support the new protocol version.
-   Refactored JSON schemas (`server_to_client.json`, `client_to_server.json`, `standard_catalog_definition.json`) to enforce the new structure.


specification/0.9/docs/a2ui_protocol.md
```diff
@@ -103,21 +104,35 @@ Custom catalogs can be used to define additional UI components or modify the beh
 
 ## Envelope Message Structure
 
-The envelope defines three primary message types, and every message streamed by the server must be a JSON object containing exactly one of the following keys: `updateComponents`, `updateDataModel`, or `deleteSurface`. The key indicates the type of message, and these are the messages that make up each message in the protocol stream.
+The envelope defines four primary message types, and every message streamed by the server must be a JSON object containing exactly one of the following keys: `createSurface`, `updateComponents`, `updateDataModel`, or `deleteSurface`. The key indicates the type of message, and these are the messages that make up each message in the protocol stream.
+
+### `createSurface`
 
+This message signals the client to create a new surface and begin rendering it. When this message is sent, the client will expect (or should have already received) `updateComponents` and/or `updateDataModel` messages for the same `surfaceId` that define the component tree. One of the components in one of the components lists MUST have an `id` of `root` to serve as the root of the component tree.
```
jacobsimionato: How about we define that create surface is always sent before update components, just to avoid confusion about lifecycle, and detect the surface naming collisions mentioned before. We can have both messages in the right order in a single Part to support atomic updates.

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -103,21 +104,35 @@ Custom catalogs can be used to define additional UI components or modify the beh
...
+
+**Properties:**
+
+- `surfaceId` (string, required): The unique identifier for the UI surface to be rendered.
+- `catalogId` (string, required): A string that uniquely identifies the component catalog used for this surface. It is recommended to prefix this with an internet domain that you own, to avoid conflicts (e.g., `mycompany.com:somecatalog`).
```
jacobsimionato: Update the example domain to use a URI - i updated 0.8 to do this based on feedback.

specification/0.9/docs/a2ui_protocol.md
```diff
@@ -162,7 +176,8 @@ This message is used to send or update the data that populates the UI components
 
 - `surfaceId` (string, required): The unique identifier for the UI surface this data model update applies to.
 - `path` (string, optional): A path to a specific location within the data model (e.g., `/user/name`). If omitted or set to `/`, the entire data model for the surface will be replaced.
-- `contents` (object, required): The data to be updated in the data model. This can be any valid JSON object.
+- `op` (string, optional): The operation to perform on the data model. Must be 'replace' or 'delete'. If omitted, defaults to 'replace'.
+- `value` (object, required): The data to be updated in the data model. This can be any valid JSON object.
```
jacobsimionato: Should this be optional? I assume not required for deletion?

specification/0.9/docs/evolution_guide.md
```diff
@@ -0,0 +1,281 @@
+# A2UI Protocol Evolution Guide: v0.8 to v0.9
+
+This document serves as a comprehensive guide to the changes between A2UI version 0.8 and version 0.9. It details the shifts in philosophy, architecture, and implementation, providing a reference for stakeholders and developers migrating between versions.
+
+## 1. Executive Summary
+
+**Version 0.9 represents a fundamental philosophical shift from "Structured Output First" to "Prompt First."**
+
+- **v0.8** was designed to be generated by an LLM using Structured Output above all else, optimized for LLMs that support strict JSON mode or function calling (which is also a form of Structured Output). It relied on deep nesting and specific wrapper structures that were definable in the limited shema features available for structured output but often confusing for an LLM to generate.
+- **v0.9** is designed to be **embedded directly in an LLM's system prompt**. The schema is refactored to be more human-readable and "token-efficient" for the model to understand. It prioritizes patterns that LLMs naturally excel at (like standard JSON objects for maps) over strict structured output-friendly structures (like arrays of key-value pairs).
```
jacobsimionato, line 10: I’m personally still in favour of highlighting the role of schema adapters in inference, because right now, even with templating, everyone is too stuck on generating the schemas directly. I know v0.9 actually evaluates remarkably well, but there are still good reasons to omit parts of the schema or do templating etc, so I personally think we should frame it as a simplified and elegant API that can be generated directly.

specification/0.9/docs/evolution_guide.md
```diff
@@ -0,0 +1,281 @@
...
+
+**v0.9 (`updateDataModel`)**:
+
+- **Renamed**: `dataModelUpdate` -> `updateDataModel`.
+- **Standard JSON**: The `value` property is now a standard **JSON object**.
+- **Op**: The `op` property is added to allow for more complex updates (e.g., `replace`, `remove`).
+- **Structure**: `{ "name": "Alice" }`
+- **Reason**: LLMs are trained to generate JSON objects. Forcing them to generate an "adjacency list" representation of a map was inefficient and error-prone.
+
+### 4.3. Client Capabilities
```
jacobsimionato: Maybe remove this because the plan is to basically keep it the same, I assume?

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Ahh, yes, I did this too, just missed this one.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Okay, done.

specification/0.9/docs/a2ui_protocol.md
gspencergoog: Yes, not allowed for delete, required for the others.

specification/0.9/docs/evolution_guide.md
gspencergoog: > I personally think we should frame it as a simplified and elegant API that can be generated directly.

Isn't that what I'm doing?  The 0.9 spec doesn't mention schema adapters for inference because we haven't defined how that works and where they would be inserted (and who runs them).  Once we do, then this schema becomes the one that is generated by the schema adapters, I would think, since it's concerned with being generatable by the models.  Doesn't that make it the one that is a "simplified and elegant API that can be generated directly"?

There are two kinds of adapters we have talked about and need to design: schema adapters which modify/subset the schema that is sent to the LLM at inference time, and protocol adapters which take the output from an LLM and standardize it (programmatically) for sending down the wire.  We should define those.

specification/0.9/docs/evolution_guide.md
gspencergoog: Yes, if they've been removed from the 0.8 spec.  I was waiting for your PR to land to finish this section, actually. (https://github.com/google/A2UI/pull/136)

PR #236, opened 2025-12-08
Update Angular setup

crisbeto:
Makes the following changes for the Angular examples:
* Adds `usageHint` support to `Text` and `Image`.
* Removes `Heading` since it's not used anymore.
* Fixes a typo in the `Text` component for Lit.
* Updates to Angular 21.

Screenshot of the restaurant finder after the changes:
image


angular/projects/lib/src/lib/catalog/image.ts
```diff
@@ -36,16 +36,27 @@ import * as v0_8 from '@a2ui/web-lib/0.8';
...
+  readonly usageHint = input.required<v0_8.Types.ResolvedImage['usageHint'] | null>();
+
   protected readonly resolvedUrl = computed(() => this.resolvePrimitive(this.url()));
+
+  protected classes = computed(() => {
+    const usageHint = this.usageHint();
+
+    return {
+      ...this.theme.components.Image.all,
+      ...(usageHint ? this.theme.components.Image[usageHint] : {}),
```
jacobsimionato: Does this reference actual themes for all the usage hints e.g. to make avatar images small etc? I couldn't see that in this PR. Is it in the underlying "web" shared library?

crisbeto: It's merging in the classes that should apply to all images with the classes that only apply to the specific usage hint. It was already in the Lit implementation, I just copied it over.

PR #242, opened 2025-12-09
Refactor: Universal App Shell, Theming, Profile Pic Fix, and Shell Re…

dmandar:
…name

Refactors the client application into a Universal App Shell (app.ts) capable of dynamically loading configurations for different agents (Restaurant, Contacts).

Universal Shell: Replaced restaurant.ts with app.ts and added AppConfig support via configs/.
Theming: Implemented ThemeManager for global style injection and renamed elegant.css to styles.css (See doc in PR)
.
Core Components: Updated 
Image component in renderers/lit to support fit (object-fit) and usageHint for better styling control.
Python Agents: Refactored restaurant_finder and contact_lookup to use external a2ui_examples.py for cleaner prompt generation.
Fixes: Resolved oversized profile pictures in Contact Lookup by implementing usageHint="avatar" and fit="cover".


samples/client/lit/shell/THEMING.md
```diff
@@ -0,0 +1,85 @@
...
+*   **When to touch**: Rarely. Only if you need to change the fundamental layout behavior of a component.
+
+### 2. **Theme Layer (`styles.css`)**
+*   **Role**: Visual Design System.
+*   **What it does**: Defines the "look and feel" using standard CSS. It handles:
+    *   Glassmorphism effects (`backdrop-filter`)
+    *   Shadows and elevations
+    *   Typography scales and weights (using **Outfit** font)
+    *   Animations (hover states, transitions)
+*   **Key Mechanism**: It consumes **CSS Variables** (e.g., `var(--primary-color)`, `var(--bg-gradient)`) instead of hardcoding colors.
```
jacobsimionato, line 21: Is there a list of these variables somewhere, so it's clear what levers you can control?

samples/client/lit/shell/configs/contacts.ts
```diff
@@ -0,0 +1,29 @@
...
+    'Verifying information...',
+    'Just a moment...'
+  ],
+  serverUrl: 'http://localhost:10003',
+  theme: {
+    // Contacts Theme (Teal/Green/Emerald)
+    // Overriding the primary colors to give it a distinct look
+    '--primary-color': 'light-dark(#10b981, #34d399)', // Emerald 500/400
+    '--primary-gradient': 'linear-gradient(135deg, light-dark(#10b981, #34d399) 0%, light-dark(#059669, #059669) 100%)',
+    '--button-text': '#ffffff',
```
jacobsimionato, line 20: How does this interact with the use of primary vs secondary buttons? I would expect secondary buttons may be light colored, and so this white text may not be visible on them?

samples/client/lit/shell/configs/restaurant.ts
```diff
@@ -0,0 +1,22 @@
...
+  heroImageDark: '/hero-dark.png',
+  placeholder: 'Top 5 Chinese restaurants in New York.',
+  loadingText: [
+    'Finding the best spots for you...',
+    'Checking reviews...',
+    'Looking for open tables...',
+    'Almost there...'
+  ],
+  serverUrl: 'http://localhost:10002',
+  theme: {
```
jacobsimionato, line 16: Could it potentially work to just use the same theme for all agents, to emphasize that they're interchangeable? Maybe the agents themselves include hero images etc in their layouts, to help identify their content?

samples/client/lit/shell/configs/restaurant.ts
```diff
@@ -0,0 +1,22 @@
+import { AppConfig } from './types.js';
+
+export const config: AppConfig = {
+  key: 'restaurant',
+  title: 'Restaurant Finder',
+  heroImage: '/hero.png',
+  heroImageDark: '/hero-dark.png',
+  placeholder: 'Top 5 Chinese restaurants in New York.',
+  loadingText: [
```
jacobsimionato, line 9: I wonder if in the future we could build an agent-driven loading experience, e.g. have the agent stream back layouts containing messages like this while it's thinking.

samples/client/lit/shell/app.ts
```diff
@@ -0,0 +1,745 @@
...
+
+  @state()
+  accessor #isDark = false;
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    // Load config from URL
+    const urlParams = new URLSearchParams(window.location.search);
+    const appKey = urlParams.get('app') || 'restaurant';
```
jacobsimionato, line 244: In order to really emphasize that a single client codebase can talk to many agents, what if we had a menu at the top of the UI, where you can choose from the known demos, or else even type in your own URL and use some custom server, with a default theme etc?

I put a UX mock for something like this here: https://docs.google.com/document/d/1y6JLGp-1CTKQfT7CdnSVVjPN3qhl9-Ga3V3JpHQHNR0/edit?tab=t.0#bookmark=kix.aw09l4rlm8

samples/client/lit/shell/configs/contacts.ts
jacobsimionato: These kinds of interactions are why I think it would be helpful to list out the variables somewhere and what impact they have.

ava-cassiopeia:
The title itself implies that there are multiple discrete changes in this PR. Can you break it up so it's a little easier to review and follow?

samples/client/lit/shell/THEMING.md
dmandar: I added them to styling.md

samples/client/lit/shell/configs/contacts.ts
dmandar: Good catch! Currently, the A2UI Button component doesn't support variants (primary/secondary)—every button gets the primary-gradient background, so white text is always safe.

When we add secondary button support to the schema/component, we'll definitely need to introduce a separate variable (e.g., --button-secondary-text) to handle contrast on lighter backgrounds.

samples/client/lit/shell/configs/restaurant.ts
dmandar: I would rather we create another example to use as a starting point if people wish. This is pretty modular now so devs could experiment with different themes easily.

samples/client/lit/shell/configs/restaurant.ts
dmandar: yep!

samples/client/lit/shell/app.ts
dmandar: We could ..just that i didn't want to complicate the app in case devs just use this as a starting point.

dmandar:
Screenshot 2025-12-09 at 1 34 51 PM
Screenshot 2025-12-09 at 1 34 44 PM

PR #245, opened 2025-12-09
Add workflow to verify Python example builds

ava-cassiopeia:
This should "lock in" (har har) the Python changes I made so that theoretically they can replicably be rebuilt on other folks' machines.

Also includes a typo fix as part of testing that the workflow will trigger correctly.

### DIFF
--- a/.github/workflows/java_build_and_test.yml
+++ b/.github/workflows/java_build_and_test.yml
@@ -1,0 +1,34 @@
+name: Java sample build and test
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'a2a_agents/java/**'
+  pull_request:
+    paths:
+      - 'a2a_agents/java/**'
+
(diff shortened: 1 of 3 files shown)


.github/workflows/editor_build.yml
```diff
@@ -3,7 +3,11 @@ name: Editor build
 on:
   push:
     branches: [ main ]
+    paths-ignore:
+      - 'a2a_agents/python/adk/samples/**'
```
jacobsimionato, line 7: Why do we have to ignore this folder in all the workflows? It isn't referenced below, so I thought nothing would be built from here anyway?

ava-cassiopeia: Nah, workflows run on all of the conditions listed here (eg. all pushes to main and all PRs), they don't consider the other dependent files below.

PR #247, opened 2025-12-09
Add workflow file for Java example

ava-cassiopeia:
I guess I'm just adding workflow files to everything, don't mind me doing some janitor work 🤷🏻‍♀️ 

Also updates the Java version in the maven package to 21, because that's a more recent LTS and what most people will be running.


jacobsimionato:
10/10 great java!

PR #256, opened 2025-12-10
Add support for server side styling override for logoUrl, primaryColo…

dmandar:
…r and font.

Also add back logoUrl to 0.8 schema

### DIFF
--- a/a2a_agents/python/adk/samples/contact_lookup/a2ui_schema.py
+++ b/a2a_agents/python/adk/samples/contact_lookup/a2ui_schema.py
@@ -45,7 +45,11 @@ A2UI_SCHEMA = r'''
               "type": "string",
               "description": "The primary UI color as a hexadecimal code (e.g., '#00BFFF').",
               "pattern": "^#[0-9a-fA-F]{6}$"
-            }
+            },
+            "logoUrl": {
+              "type": "string",
+              "description": "The URL of the brand logo."
+            },
           }
         }
       },
--- a/samples/client/lit/shell/app.ts
+++ b/samples/client/lit/shell/app.ts
@@ -694,9 +694,50 @@ export class A2UILayoutEditor extends SignalWatcher(LitElement) {
   async #sendAndProcessMessage(request) {
     const messages = await this.#sendMessage(request);
 
+    // Apply server-side styles if present in beginRendering
+    for (const msg of messages) {
+      if ('beginRendering' in msg && msg.beginRendering.styles) {
+        this.#applyServerStyles(msg.beginRendering.styles);
+      }
+    }
+
     this.#lastMessages = messages;
     this.#processor.clearSurfaces();
     this.#processor.processMessages(messages);
+  }
+
+  #applyServerStyles(styles: any) {
+    const root = document.documentElement;
+
+    // 1. Primary Color
+    if (styles.primaryColor) {
+      root.style.setProperty('--primary-color', styles.primaryColor);
+      // Generate a simple dim variant for now (could be more sophisticated)
+      root.style.setProperty('--p-60', styles.primaryColor);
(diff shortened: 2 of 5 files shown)


dmandar:
Screenshot 2025-12-10 at 9 16 49 AM
Screenshot 2025-12-10 at 9 18 07 AM

samples/client/lit/shell/theme/styles.css
```diff
@@ -11,7 +11,7 @@
 
 /* Global overrides */
 * {
-  font-family: 'Outfit', sans-serif !important;
+  font-family: var(--bb-font-family, 'Outfit', sans-serif) !important;
```
ava-cassiopeia, line 14: Whoa, I don't think I've encountered this file before. Why these overrides as opposed to setting this in `html {}` or `body {}`?

samples/client/lit/shell/app.ts
```diff
@@ -694,11 +694,52 @@ export class A2UILayoutEditor extends SignalWatcher(LitElement) {
   async #sendAndProcessMessage(request) {
     const messages = await this.#sendMessage(request);
 
+    // Apply server-side styles if present in beginRendering
+    for (const msg of messages) {
+      if ('beginRendering' in msg && msg.beginRendering.styles) {
```
ava-cassiopeia, line 699: **NIT:** I think you can use a nullish coalescing operator here.

```suggestion
      if (msg.beginRendering?.styles) {
```

samples/client/lit/shell/app.ts
```diff
@@ -694,11 +694,52 @@ export class A2UILayoutEditor extends SignalWatcher(LitElement) {
...
     this.#processor.clearSurfaces();
     this.#processor.processMessages(messages);
   }
 
+  #applyServerStyles(styles: any) {
+    const root = document.documentElement;
+
+    // 1. Primary Color
+    if (styles.primaryColor) {
+      root.style.setProperty('--primary-color', styles.primaryColor);
```
ava-cassiopeia, line 714: Are we validating these styles anywhere? `primaryColor` could be like truly any value as far as I can tell.

samples/client/lit/shell/app.ts
```diff
@@ -694,11 +694,52 @@ export class A2UILayoutEditor extends SignalWatcher(LitElement) {
...
+
+
+
+    // 3. Logo / Hero Image
+    if (styles.logoUrl) {
+      // Update the config dynamically to show the new hero image
+      this.config = {
+        ...this.config,
+        heroImage: styles.logoUrl,
+        heroImageDark: styles.logoUrl, // Use same for dark mode unless specified otherwise
```
ava-cassiopeia, line 738: I don't see a way to specify otherwise, right? Should we say we're just going to set this for both until we have a way to specify that?

a2a_agents/python/adk/samples/contact_lookup/a2ui_schema.py
```diff
@@ -45,7 +45,11 @@
               "type": "string",
               "description": "The primary UI color as a hexadecimal code (e.g., '#00BFFF').",
               "pattern": "^#[0-9a-fA-F]{6}$"
-            }
+            },
```
gspencergoog, line 48: If you add this here, please also update the specification/0.8/a2ui_protocol.md and the JSON schema in specification/0.8/json. That should be the source of truth for the schema.  

In fact, this code should probably be reading those schemas directly and not defining a copy that is likely to get out of date.

samples/client/lit/shell/theme/styles.css
dmandar: this is going to be rewired by Paul..hopefully this PR will not be needed then..

PR #260, opened 2025-12-11
[web] Update shell styles

paullewis:
This brings the styling in line with the pre-existing theme machinery. This applies to both the restaurant and contacts examples for both light & dark themes.


renderers/lit/src/0.8/ui/text.ts
```diff
@@ -109,10 +118,21 @@ export class Text extends Root {
...
+    let additionalStyles: Record<string, string> = {};
+    const styles = this.theme.additionalStyles?.Text;
+    if (styles) {
+      if ("h1" in styles) {
+        const hint = this.usageHint ?? "body";
+        additionalStyles = styles[hint] as Record<string, string>;
+      } else {
+        additionalStyles = styles;
+      }
+    }
```
ava-cassiopeia, line 130: **NIT:** this logic smells to me like it would be better served as its own private method for readability, as you could improve the readability by using guard clauses instead of nested conditionals; something like:

```typescript
#getAdditionalStyles(...): Record<string, string> {
  const styles = this.theme.additionalStyles?.Text;
  if (!styles) return {};

  // etc.
}
```

Feel free to do this in a followup PR if you want.

renderers/lit/src/0.8/ui/surface.ts
```diff
@@ -75,17 +74,19 @@ export class Surface extends Root {
       for (const [key, value] of Object.entries(this.surface.styles)) {
         switch (key) {
           case "primaryColor": {
-            for (let i = 0; i <= 100; i++) {
-              styles[`--p-${i}`] = `color-mix(in srgb, ${value} ${
-                100 - i
-              }%, #fff ${i}%)`;
-            }
+            // Ignored for now.
```
ava-cassiopeia, line 77: **NIT:** Is it worth clarifying why this is being ignored? Here and below.

PR #270, opened 2025-12-12
[web] Bring back LLM-driven colors & fonts

paullewis:
cc @dmandar


renderers/lit/src/0.8/ui/surface.ts
```diff
@@ -73,26 +73,41 @@ export class Surface extends Root {
     if (this.surface?.styles) {
       for (const [key, value] of Object.entries(this.surface.styles)) {
         switch (key) {
+          // Here we generate a palette from the singular primary color received
+          // from the surface data. We will want the values to range from
+          // 0 <= x <= 100, where 0 = back, 100 = white, and 50 = the primary
```
ava-cassiopeia, line 78:
```suggestion
          // 0 <= x <= 100, where 0 = black, 100 = white, and 50 = the primary
```

Issue #281, opened 2025-12-15
Remove "date" type from TextField seeing as we have `DateTimeInput` instead


jacobsimionato:
I think it adds confusion to have both supported.

There is still validationRegexp on TextField, so people can use this if they want.

PR #282, opened 2025-12-15
Remove `date` from `TextField` component to remove duplication relative to `DateTimeInput`

jacobsimionato:
I think it adds confusion to have this seeing as we already have `DateTimeInput`.

There is still `validationRegexp` on `TextField`, so people can use this if they want.

Fixes https://github.com/google/A2UI/issues/281


gspencergoog (approved):
Makes sense

PR #283, opened 2025-12-15
refactor: extract a2ui_schema

zeroasterisk:
cp from sample contact_finder folder


jacobsimionato:
LGTM - this is better I think. What we really want though is for the Python code to load the .json file directly from the `specification/0.8/json` folder, so we have one source of truth.

dmandar:
thx ..yeah this was done in contact sample i think but i forgot to pull over

PR #294, opened 2025-12-16
Fix README client setup instructions

hadarge:
## Summary
- Add missing steps to install and build the Lit renderer before running the shell client

## Problem
Following the current README instructions for running the client results in a build failure:
```
tsc: command not found
```

This happens because the `renderers/lit` directory needs its dependencies installed separately before the shell client can build.

## Solution
Updated the "Run the Client (Frontend)" section to include the necessary steps:
1. Navigate to `renderers/lit`
2. Run `npm install` and `npm run build`
3. Then proceed to the shell client setup


jacobsimionato (approved):
Thank you!!

PR #297, opened 2025-12-16
chore: add .gitattributes for line ending normalization

SliverKeigo:
- Set default text files to use LF line endings
- Explicitly configure shell scripts and Makefiles
- Mark binary files to prevent Git normalization
- Improve cross-platform collaboration experience


SliverKeigo:
Hi  @jacobsimionato 👋

Just a gentle ping in case this got lost in notifications.

It's a simple .gitattributes addition - should be a quick review!

Would appreciate if someone could take a look when convenient.

Thanks! 🙏

gspencergoog:
What's the motivation for this change?  Git automatically handles line endings by default, so forcing it to use LF on all platforms seems problematic for people using editors on Windows.

SliverKeigo:
Totally fair point, Git can handle line endings automatically. The catch is that without a repo-level rule, the actual behavior still ends up depending on each person’s local config (especially core.autocrlf) and sometimes their editor/tooling. So it usually works, but it’s easy to get occasional CRLF/LF churn in diffs.

This .gitattributes is just meant to make the repo’s behavior deterministic and keep diffs/CI predictable. If we have Windows-native files that truly need CRLF (e.g. .bat/.cmd), I’m happy to add explicit exceptions:
```
*.bat text eol=crlf
*.cmd text eol=crlf
```

gspencergoog:
Sorry, I don't think we need to do this.  This hasn't historically been a problem (there are currently no files with CR/LF endings in the repo), and I don't want to add another configuration file that needs to be maintained.

Thanks for the PR, I do appreciate that you were interested in contributing, but I'm going to close this.

Issue #299, opened 2025-12-16
[BUG] npm install in quick start guide broken


Tehsmash:
While walking through the quickstart guide I was unable to `npm install` the lit sample code.

```
$ npm install
npm warn deprecated node-domexception@1.0.0: Use your platform's native DOMException instead
npm error code E401
npm error Incorrect or missing password.
npm error If you were trying to login, change your password, create an
npm error authentication token or enable two-factor authentication then
npm error that means you likely typed your password in incorrectly.
npm error Please try again, or recover your password at:
npm error   https://www.npmjs.com/forgot
npm error
npm error If you were doing some other operation then your saved credentials are
npm error probably out of date. To correct this please try logging in again with:
npm error   npm login
npm error A complete log of this run can be found in: /Users/sambetts/.npm/_logs/2025-12-16T10_44_01_722Z-debug-0.log
```

Error from the debug log:

```
447 verbose stack HttpErrorAuthUnknown: Unable to authenticate, need: Basic realm="https://us-npm.pkg.dev"
447 verbose stack     at /opt/homebrew/Cellar/node@22/22.20.0_1/lib/node_modules/npm/node_modules/npm-registry-fetch/lib/check-response.js:88:17
447 verbose stack     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
448 verbose statusCode 401
449 verbose pkgid yargs-parser@https://us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/yargs-parser/-/yargs-parser-21.1.1.tgz
450 error code E401
451 error Incorrect or missing password.
452 error If you were trying to login, change your password, create an
452 error authentication token or enable two-factor authentication then
452 error that means you likely typed your password in incorrectly.
452 error Please try again, or recover your password at:
452 error   https://www.npmjs.com/forgot
452 error
452 error If you were doing some other operation then your saved credentials are
452 error probably out of date. To correct this please try logging in again with:
452 error   npm login
```

I double checked that I have no `~/.npmrc` file configured to rule out that there was some invalid credentials stored.

As a debug setup I removed `package-lock.json`, after that `npm install` was successful. I think the `package-lock.json` might have repos which require login stored in it, specifically `https://us-npm.pkg.dev`. 

This is a snippet of the lock file diff after I successfully installed:

```
     "node_modules/yargs-parser": {
       "version": "21.1.1",
[...]

PR #306, opened 2025-12-16
Update lit sample npm registry

brycewcole:
Fixes https://github.com/google/A2UI/issues/299

Regenerate package-lock.json to remove us-npm.pkg.dev (Google Artifact Registry) URLs for public packages, which were causing npm install to fail without authenticated GAR access.


dmandar:
Thanks for your PR!

PR #311, opened 2025-12-17
doc: fixing A2UI logo display in specs

didier-durand:
Hi,

Logo for A2UI doesn't display correctly in specification page. This PR tries to fix it for 0.9 in light mode.

Please, validate this change (without merging) and I am happy to extend the PR to dark mode as well for both v0.8 and v0.9

Cheers,

Didier


zeroasterisk:
Thanks for fixing the logo display, @didier-durand! 🙏

The specification directory structure has since changed — the files now live in `specification/v0_9/docs/` rather than `specification/0.9/docs/`, and the docs were significantly restructured. The current spec header already has the logo path updated. This PR is no longer applicable due to the path changes. Closing as stale.

PR #318, opened 2025-12-17
fix(lit): use cross-platform script for copy-spec to support Windows

jeevanjoseph03:
## Problem
The `renderers/lit` build script `copy-spec` relied on Unix-specific commands (`mkdir -p` and `cp`), causing the build to fail on Windows environments.

## Solution
Replaced the shell command with a new cross-platform Node.js script `scripts/copy-schemas.js` that uses the `fs` module to securely copy the schema files.

## Verification
- Verified that `npm run copy-spec` runs successfully on Windows 11.
- Verified that `npm run build` completes without errors.
- Confirmed that the Restaurant Finder demo now starts and runs correctly.

### DIFF
--- a/renderers/lit/package.json
+++ b/renderers/lit/package.json
@@ -30,7 +30,7 @@
   },
   "wireit": {
     "copy-spec": {
-      "command": "mkdir -p src/0.8/schemas && cp ../../specification/0.8/json/*.json src/0.8/schemas",
+      "command": "node scripts/copy-schemas.js",
       "files": [
         "../../specification/0.8/json/*.json"
       ],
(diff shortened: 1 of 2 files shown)


zeroasterisk:
Thanks for the Windows cross-platform fix, @jeevanjoseph03! 🙏

The Lit renderer build system has been significantly refactored since this PR was opened. The `copy-spec.js` script is no longer part of the Lit build pipeline — the package.json now uses `wireit` for builds without the separate spec copy step. This PR is no longer applicable. Closing as stale — your work helped us think through cross-platform compatibility.

PR #321, opened 2025-12-17
docs: automate syncing of v0.9 schemas to hosted site (#232)

### LINKED ISSUE #232: Upload v0.9 schemas to URLs at a2ui.org that they reference
In client_to_agent.json etc, we have a  line like `"$id": "https://a2ui.org/specification/0.9/client_to_agent.json",`.

Let's make sure we actually host this schema at this URL on our site!

Same for the other schemas.

kaifcodec:
## Description
This PR automates the synchronization of v0.9 JSON schemas into the documentation build process. Previously, changes to the specification were not reflected on the hosted site unless manually moved.

### Changes
* **Workflow Triggers**: Added `specification/**` to the watch list so the docs rebuild whenever a schema is updated.
* **Auto-Sync Step**: Added a build step to create `docs/specification/0.9` and copy the latest `.json` schemas from the source directory before the MkDocs build.

Fixes #232

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,6 +23,7 @@ on:
       - "requirements-docs.txt"
       - "mkdocs.yml"
       - "docs/**"
+      - "specification/**"
   pull_request:
     branches:
       - main
@@ -31,6 +32,7 @@ on:
       - "requirements-docs.txt"
       - "mkdocs.yml"
       - "docs/**"
+      - "specification/**"
 
 jobs:
   build_and_deploy:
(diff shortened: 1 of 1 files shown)


zeroasterisk:
Thanks for this automation improvement, @kaifcodec! Keeping schemas synced to the hosted docs automatically is exactly the right approach. 🙏

This has since been implemented in the docs workflow — the current `.github/workflows/docs.yml` already copies the entire specification directory to docs and flattens JSON files. The implementation also handles both v0_8 and v0_9 schemas. Closing as superseded.

kaifcodec:
Thanks for the update, @zeroasterisk! Glad to see the automation is now 
handled natively, the flattening for clean URLs is a great touch.

Just a quick heads-up: it looks like `specification/**` is currently 
missing from the `on: paths:` triggers in `docs.yml`. Without that 
path included, a change to a schema file alone won't trigger the 
documentation rebuild/deploy.

Happy to open a quick PR to add that trigger if it was overlooked!

PR #329, opened 2025-12-17
Add package.json for A2UI CLI tools

cypherdavy:


jacobsimionato:
Hi @cypherdavy ! Thanks for looking into a CLI tool for A2UI. I see you've sent few small PRs incrementally adding things here. We'd be excited to have a CLI tool, but I'm wondering if you could develop it to an MVP state where we can try it out and see the code together in context, and then send it for review as a single PR with some working functionality, rather than reviewing pieces file-by-file?

cypherdavy:
Sure!!

jacobsimionato:
Thank you for the time it took to write this contribution! Closing this here, but I left an explanation in https://github.com/google/A2UI/pull/339.

PR #342, opened 2025-12-18
fix(lit): correctly bind TextField type property to enable number inputs

SliverKeigo:
## Description

This PR fixes an issue where the `TextField` component in the Lit renderer was ignoring the `type` property (e.g., `"number"`) from the A2UI JSON payload.

## Problem

A property name mismatch between the renderer and component caused the input type to be ignored:

| Location | Property Name | Code |
|----------|---------------|------|
| `root.ts` (Sender) | `.type` | `.type=${node.properties.type}` |
| `text-field.ts` (Receiver) | `inputType` | `accessor inputType` |

**Result:** `<input type="number">` rendered as `<input type="text">`, allowing non-numeric input in number fields.

## Solution

1. **Lit Component (`text-field.ts`)**: Renamed `inputType` → `textFieldType` to avoid conflicts with the component's class type identifier.

2. **Root Renderer (`root.ts`)**: Updated binding to `.textFieldType=${node.properties.type}`.

3. **Samples (`a2ui_examples.py`)**: Added explicit `"type": "shortText"` and `"type": "number"` usage for clarity.

## Verification

- ✅ `textFieldType` correctly receives values from JSON payload
- ✅ HTML input renders with `type="number"` when specified

---

Hi @jacobsimionato 👋

Could you please take a look or help arrange a review?

Thanks! 🙏

### DIFF
--- a/renderers/lit/src/0.8/ui/root.ts
+++ b/renderers/lit/src/0.8/ui/root.ts
@@ -420,7 +420,7 @@ export class Root extends SignalWatcher(LitElement) {
             .dataContextPath=${node.dataContextPath}
             .label=${node.properties.label}
             .text=${node.properties.text}
-            .type=${node.properties.type}
+            .textFieldType=${node.properties.type}
             .validationRegexp=${node.properties.validationRegexp}
             .enableCustomElements=${this.enableCustomElements}
           ></a2ui-textfield>`;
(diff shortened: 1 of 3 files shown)


zeroasterisk:
Thanks for this fix, @SliverKeigo! The `.type` property binding for TextField was indeed wrong. 🙏

This has been addressed in the codebase — the TextField component now uses `textFieldType` as the property name (reflected in PR [#806](https://github.com/google/A2UI/pull/806) and the underlying web_core refactors). The fix was implemented differently than your PR but achieves the same result. Closing as superseded — thanks for identifying the issue!

PR #345, opened 2025-12-18
feat: Clarify NYC-only dataset and add no results message

roiyz:
This pull request is for improving the user experience for running restaurants finder example code by clarifying that it's just based on a local NYC restaurants dataset, as well as improving the results text when no restaurant is found. 
This also demonstrates how the llm model chooses a different renedring path when no results are found. 

Verified both for positive and negative scenarios. 

The new message to the user looks like the following:
image


zeroasterisk:
@laxmi-genai after your review of the sample work, please take a look at this one.

zeroasterisk:
Thanks for this improvement, @roiyz! Adding clear messaging when no results are found is great UX. 🙏

The restaurant finder sample has undergone significant refactoring since this PR was opened (the sample moved to a new location and the prompt system was rewritten). This PR would need a complete rebase to be mergeable, and the original files it targets have changed substantially. Closing as stale — if you're interested in contributing this improvement, a fresh PR against the current codebase would be the way to go!

PR #274, opened 2025-12-12
Add A2UI-Composer

maxkorp:
Adds an app called A2UI-Composer which generates a2ui widgets. Available hosted at a2ui-editor.ag-ui.com.

Note the hosted ui is slightly out of sync since it is deploying from a different repository, will clean up once this is all live as discussed with @zeroasterisk 

composer

https://github.com/user-attachments/assets/08d297e5-e38f-45ba-9da4-0bcefe366dcb

PR #365, opened 2025-12-19
Add A2UI composer

maxkorp:
Original PR here: https://github.com/google/A2UI/pull/274
Unsure why it closed. Had to make an entirely new fork to remake a new PR

Adds an app called A2UI-Composer which generates a2ui widgets. Available hosted at a2ui-editor.ag-ui.com.

Note the hosted ui is slightly out of sync since it is deploying from a different repository, will clean up once this is all live as discussed with @zeroasterisk 

composer

https://github.com/user-attachments/assets/08d297e5-e38f-45ba-9da4-0bcefe366dcb


a2ui-composer/public/file.svg
ava-cassiopeia, line 1: Where did you get these icons? Did you create them yourself, or do you have copyright rights to them?

a2ui-composer/src/app/api/copilotkit/a2ui-prompt.ts
```diff
@@ -0,0 +1,270 @@
+// A2UI Component Catalog and Protocol Schema for LLM prompt
+// Based on A2UI Protocol v0.8
+
+export const A2UI_COMPONENT_CATALOG = {
```
jacobsimionato, line 4: Not blocking: It would be awesome if we could pull this file from the specification/ folder using some build step, or depend on the TS version of the inference libraries that @nan-yu  is working on.
E.g. https://github.com/google/A2UI/blob/main/specification/v0_8/json/standard_catalog_definition.json

a2ui-composer/src/data/gallery/account-balance.ts
jacobsimionato, line 1: Not blocking: Eventually I'd love to see this set of samples extracted out into a shared location like `specification/v0.8/sample_data/templates/` so that different tools (e.g. gallery applications using different renderers) could reuse the same sample data.

To do this well, I think we need a well-defined template format

a2ui-composer/src/types/widget.ts
jacobsimionato, line 1: Not blocking:

I propose we call this concept a "Template" to make it clear that it's a higher level concept than Component.

I'm wondering if you can work with @nan-yu to formalize a template schema that can be supported by the composer and inference libraries, so you can create a template file at dev time, then drop it into your server project to get the LLM to fill out the template at runtime.

Things a template probably needs:
- List of A2UI messages to define template
- Defined way to populate template, e.g. replace strings with certain syntax, or just drop params into the data model at the root level via an updateDataModel message.
- Definition of template params including description of what each of them means, to help LLMs populate templates at runtime
- Example data (to allow the template to be rendered in a gallery view for dev purposes, or so it can be used for the "example" style of inference)

We can refine all this later, though!

jacobsimionato:
I merged this PR so that we can continue to iterate on the main branch collaboratively. @ava-cassiopeia and my comments still stand though, and we'd love help from the AG-UI team to address these.

PR #369, opened 2025-12-20
[fix] set encoding to utf-8 in samples' tool.py while open sample data file

zhujie2006:
##Problem##
The A2UI sample running in a Windows shell can get this, while using a different codec
image

##Solution##
Set the encoding parameter while opening the xxx_data.json file

### DIFF
--- a/samples/agent/adk/contact_lookup/tools.py
+++ b/samples/agent/adk/contact_lookup/tools.py
@@ -34,7 +34,7 @@ def get_contact_info(name: str, tool_context: ToolContext, department: str = "")
     try:
         script_dir = os.path.dirname(__file__)
         file_path = os.path.join(script_dir, "contact_data.json")
-        with open(file_path) as f:
+        with open(file_path, encoding="utf-8") as f:
             contact_data_str = f.read()
             if base_url := tool_context.state.get("base_url"):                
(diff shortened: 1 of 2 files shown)


triceviola24-sys:
http://Graphic Web Consultant&GoogleLLC 
http://vio-ado-7140-655444-78569265

bckim9489:
Thanks for the fix - this resolves the Windows UnicodeDecodeError nicely 👍  

I was able to reproduce the original issue on Windows 11 (Python + uv):
`UnicodeDecodeError: 'cp949' codec can't decode ...` when reading `restaurant_data.json`.

With this change (`encoding="utf-8"`), the sample agent starts successfully and the demo flow works end-to-end on Windows.

As a small, optional follow-up idea (non-blocking):
would you consider using `pathlib` for building the file path and reading the data?
It can slightly improve readability and keep the path handling fully platform-independent.

For example:

```python
from pathlib import Path

file_path = Path(__file__).with_name("restaurant_data.json")
restaurant_data_str = file_path.read_text(encoding="utf-8")
```

Totally optional - the current fix already solves the problem well.
Just wanted to share this as a possible improvement.

jacobsimionato:
I'm sorry we didn't get to review this helpful PR in reasonable time!

The sample you fixed has since been deleted, so I'll close this now.

PR #374, opened 2025-12-22
Make Lit copy-spec cross-platform

RobinsonBeato:
## Summary
The Lit renderer build currently relies on `mkdir -p` and `cp`, which fail on Windows. This replaces that step with a small Node script so `npm run build` works cross-platform.

## Changes
- Replace the `copy-spec` command with a Node script.
- Add `scripts/copy-spec.mjs` to copy spec JSON files into `src/0.8/schemas`.

## Why this matters
Windows contributors currently hit a hard build failure during `copy-spec`. This makes the build portable across OSes without changing outputs.

## Testing
- `cd renderers/lit`
- `npm install`
- `npm run build`
- - Tested on Windows (PowerShell)
image

### DIFF
--- a/renderers/lit/package.json
+++ b/renderers/lit/package.json
@@ -30,7 +30,7 @@
   },
   "wireit": {
     "copy-spec": {
-      "command": "mkdir -p src/0.8/schemas && cp ../../specification/0.8/json/*.json src/0.8/schemas",
+      "command": "node scripts/copy-spec.mjs",
       "files": [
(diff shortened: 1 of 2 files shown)


zeroasterisk:
Thanks for the Windows cross-platform fix, @RobinsonBeato! 🙏

The Lit renderer build system has since been refactored and the `copy-spec.js` script is no longer part of the build pipeline. This PR is no longer applicable. Closing as stale.

PR #377, opened 2025-12-22
fix(lit): Pass enableCustomElements from Surface to Root

samgoodgame:
Surface accepts `enableCustomElements` but wasn't passing it to the child `<a2ui-root>`. This prevents custom components from rendering when using the Surface wrapper.

All other container components (List, Card, Column, Row, etc.) already pass this property—Surface was the only one missing it.


zeroasterisk:
This is used in PR #378 and extracted out here as a simpler review step.

PR #387, opened 2025-12-24
Fix: Angular contacts lookup example

### LINKED ISSUE #386: [BUG] (angular) - Contacts list demo has several issues
While testing the contacts list demo, I found several issues. Listing them down and will raise a PR fixing all of them at once.

1. Incorrect URL for the server (should be 10003 instead of 10002)

Image

2. The "Message" button doesn't work for sending a message to the contact. Results in a blank screen

3. The contact's links (based on JSON data) are not "linkified". I.e. we can't click and open them

4. Usage of icons are broken (e.g. HTML ends up with `locationOn` instead of `location-on`)

Image

Also, making the prompt a bit dynamic to include correct icons of URL, location, calendar, etc

Image

5. This is more of an enhancement, but following and messaging a contact should result in a card/UI which shows the name of the person followed, and perhaps a back button. I've fixed that locally. Will include in the PR

Image

--

I will raise a PR for all of these.

AhsanAyaz:
fix: enhance icon handling, expand contact data, and refine agent action context and UI for contact lookup sample

Fixes #386

### Contact Card
2025-12-24-204728_hyprshot

### Upon following contact

2025-12-24-204756_hyprshot

### Messaging contact

image

### Contact Card

2025-12-24-202741_hyprshot

### Following contact

2025-12-24-202805_hyprshot

### Messaging contact

2025-12-24-202859_hyprshot

### DIFF
--- a/samples/agent/adk/contact_lookup/a2ui_schema.py
+++ b/samples/agent/adk/contact_lookup/a2ui_schema.py
@@ -182,7 +182,7 @@ A2UI_SCHEMA = r'''
                               "help",
                               "home",
                               "info",
-                              "link",
+
                               "locationOn",
                               "lock",
                               "lockOpen",
--- a/samples/agent/adk/contact_lookup/prompt_builder.py
+++ b/samples/agent/adk/contact_lookup/prompt_builder.py
@@ -69,7 +69,7 @@ def get_ui_prompt(base_url: str, examples: str) -> str:
     -   **For finding contacts (e.g., "Who is Alex Jordan?"):**
         a.  You MUST call the `get_contact_info` tool.
         b.  If the tool returns a **single contact**, you MUST use the `CONTACT_CARD_EXAMPLE` template. Populate the `dataModelUpdate.contents` with the contact's details. If additional important fields (like 'favorite_framework' or 'meetupPlace') are present, you MUST add a new 'Row' to the 'info_rows_column' (matching the structure of existing info rows).
-            - Use the 'link' icon if the value is a URL.
+
             - Use the 'calendar_today' icon if the value represents a date, time, or schedule.
             - Use the 'location_on' icon if the value represents a location or place.
             - Otherwise, use the 'star' icon.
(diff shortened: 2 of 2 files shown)


samples/agent/adk/contact_lookup/a2ui_schema.py
```diff
@@ -182,6 +182,7 @@
                               "help",
                               "home",
                               "info",
+                              "link",
```
jacobsimionato: We can't add this here, because it is inconsistent with the specification in https://github.com/google/A2UI/blob/main/specification/0.8/json/standard_catalog_definition.json

Can we make do without the link icon?

It's a pretty reasonable feature request to add support for this in a future spec version though!!

jacobsimionato (changes requested):

samples/agent/adk/contact_lookup/a2ui_schema.py
AhsanAyaz: Thanks for sharing the reason here.
I can update the code to remove it for now to keep the PR clean. Makes sense.

AhsanAyaz: @jacobsimionato , @crisbeto I've updated the PR to remove the usage of the `link` icon

jacobsimionato: Thanks! I'll let @crisbeto review from here because he knows the Angular codebase much better than me

renderers/angular/src/lib/catalog/icon.ts
```diff
@@ -40,5 +40,8 @@ import { Primitives } from '@a2ui/lit/0.8';
 })
 export class Icon extends DynamicComponent {
   readonly name = input.required<Primitives.StringValue | null>();
-  protected readonly resolvedName = computed(() => this.resolvePrimitive(this.name()));
+  protected readonly resolvedName = computed(() => {
+    const name = this.resolvePrimitive(this.name());
+    return name ? name.replace(/[A-Z]/g, (letter) => `_${letter.toLowerCase()}`) : null;
```
crisbeto, line 45: This seems off to me. The runtime shouldn't be modifying the icon name, but rather the agent should return the right name to begin with.

AhsanAyaz: I can look into why that happens in depth. However, it wasn't returning the right name. It was returning `locationOn` instead of `location-on`, for example.

zeroasterisk:
Thanks for the Angular contacts fix and the linkify improvement, @AhsanAyaz! These are solid practical fixes. 🙏

The Angular renderer has been significantly restructured since this PR was opened — the component files moved to `renderers/angular/src/v0_8/components/` with a different architecture, and the icon component has been rewritten. This PR would need a complete rewrite to apply to the current codebase. Closing as stale. If these issues still exist in the current renderer, a fresh PR would be very welcome!

PR #405, opened 2025-12-30
update key names, show supported catalogs in agent card

wrenj:


samples/client/angular/package-lock.json
sugoi-yuzuru, line 1: Can I confirm that this change is intentional and needed?

AFAIU, they do not seem to be concerning. Yet, it seems irrelevant to the scope of the PR.

PR #410, opened 2025-12-30
test(renderers/angular): add vitest browser testing

43081j:
There's no real test setup in the angular renderer project yet.

I was roaming around the codebase and figured I'd push this PR. It has the following:

- Drop karma/jasmine
- Switch to `build:unit-test`
- Use `chromium` so component tests run in a real browser
- Add missing rxjs dependency
- add 2 basic test files

Note i haven't touched angular in a long time, so a bunch of the tests may not be best practice. If there are better ways to do any of this, do let me know so i can update and learn.

It seems like we should also have some shared test utils file somewhere for things like the mock theme. Where would be sensible to put that?

Similarly, this `waitForSelector` function is a shame. I needed it because the default catalog uses dynamic imports and `whenStable` isn't aware of these. `whenStable` completes long before the dynamic import, but i can't seem to find anything built-in we can wait for other than polling for the DOM.

**If you planned on having a different test setup**, feel free to throw this PR away! It has been good to learn the codebase either way.

### DIFF
--- a/renderers/angular/package-lock.json
+++ b/renderers/angular/package-lock.json
@@ -10,6 +10,7 @@
       "dependencies": {
         "@a2ui/lit": "file:../lit",
         "markdown-it": "^14.1.0",
+        "rxjs": "^7.8.0",
         "tslib": "^2.3.0"
       },
       "devDependencies": {
--- a/renderers/angular/package.json
+++ b/renderers/angular/package.json
@@ -7,6 +7,7 @@
   "dependencies": {
     "@a2ui/lit": "file:../lit",
     "markdown-it": "^14.1.0",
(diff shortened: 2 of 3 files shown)


zeroasterisk:
Thanks for adding vitest browser testing to the Angular renderer, @43081j! Better testing infrastructure is always valuable. 🙏

The Angular renderer has been completely restructured with v0.9 work, and the test setup has been updated as part of [#878](https://github.com/google/A2UI/pull/878) (Angular v0.9 Renderer). The `@angular/build:unit-test` builder is now in the config. Closing as superseded — your approach was right and it happened to land via a different PR!

PR #396, opened 2025-12-28
feat: Enable Rizzcharts sample and enhance client message handling

doitundoit:
### Summary

This PR enables the rizzcharts sample agent with improved map support and significantly hardens the Lit client's message handling capabilities. It addresses issues with MIME type definitions and improves the client's ability to handle various data formats.

#### Rizzcharts Agent

Map Support: Updated `agent.py` and `map.json` to support map images, allowing for either Google Maps API integration or fallback placeholders.

#### Lit Client Improvements

Robust Message Handling:
- Implemented a fallback mechanism to retrieve agent messages from history when needed.
- Added support for parsing string-encoded JSON data parts.

Configuration:
- Fixed the A2UI MIME type constant to `application/json+a2ui`.
- Added debug logging (`console.debug`) for better traceability.

Demo Setup: 
- Added rizzcharts demo scripts and configuration to `samples/client/lit/package.json`.

### Verification
-  Python: Verified syntax and build configuration for rizzcharts agent.
- TypeScript: Successfully built the lit shell client (npm run build:tsc).

PR #419, opened 2026-01-03
feat: improve Lit client robustness for message handling

doitundoit:
## Summary
This PR improves the Lit client's robustness when handling agent response messages.

## Changes
- **History fallback**: Added fallback mechanism to retrieve agent messages from history when `status.message.parts` is empty
- **String-encoded JSON handling**: Parse string-encoded JSON data parts that some agents may return
- **Debug logging**: Added `console.debug` statements for better traceability during development

### Affected file:
- `samples/client/lit/shell/client.ts`

## Verification
- TypeScript build passes (`npm run build:tsc`)

## Related
Split from #396 as requested by reviewer to keep commit scope narrow and targeted.

### DIFF
--- a/samples/client/lit/shell/client.ts
+++ b/samples/client/lit/shell/client.ts
@@ -97,46 +97,59 @@ export class A2UIClient {
     }
 
     const result = (response as SendMessageSuccessResponse).result as Task;
+    
+    // Log the full response for debugging purposes
     console.debug("Full Server Response Result:", JSON.stringify(result, null, 2));
 
-    let responseParts = result.status.message?.parts;
+    if (result.kind === "task") {
+      let responseParts = result.status.message?.parts;
 
-    // Fallback: If no parts in status.message, check the last agent message in history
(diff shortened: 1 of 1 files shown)


paullewis (approved):
Nit: might be worth adding a bit of an explanation to this but approving all the same.

doitundoit:
> Nit: might be worth adding a bit of an explanation to this but approving all the same.

Hi, thanks for the feedback! I've added concise comments explaining the rationale behind the history fallback and JSON parsing logic for better clarity.

zeroasterisk:
Thanks for improving the Lit shell client robustness, @doitundoit! The history-based fallback for session resumption is a good idea. 🙏

The Lit shell client has been significantly updated since this PR was opened, and the A2A SDK response format has evolved as well. This PR would need a substantial rebase against the current codebase. Closing as stale — but the insight about handling empty status parts via history fallback is worth filing as a focused issue if you find it's still needed!

PR #420, opened 2026-01-03
feat: Add map image support to Rizzcharts with local placeholder

doitundoit:
## Summary
This PR adds map image support to the Rizzcharts sample agent, using a local placeholder image instead of relying on external services.

## Changes
- **New asset**: Added `map-placeholder.png` to `samples/client/lit/shell/public/`
- **Updated map.json**: Added `map-image` component using local `/map-placeholder.png` path
- **Updated agent.py**: Added map image instructions that:
  - Use local placeholder by default
  - Support Google Maps Static API when `GOOGLEMAPS_API_KEY` env var is set

### Affected files:
- `samples/client/lit/shell/public/map-placeholder.png` [NEW]
- `samples/agent/adk/rizzcharts/examples/standard_catalog/map.json`
- `samples/agent/adk/rizzcharts/agent.py`

## Verification
- Python syntax verified
- JSON syntax verified

## Related
Addresses feedback from #396 - reviewer requested avoiding external placeholder services like placehold.co.


samples/agent/adk/rizzcharts/agent.py
```diff
@@ -72,13 +72,27 @@ def get_instructions(cls, readonly_context: ReadonlyContext) -> str:
         else:
             raise ValueError(f"Unsupported catalog uri: {catalog_uri if catalog_uri else 'None'}")
 
+        # Check for Google API Key
+        googlemaps_api_key = os.getenv("GOOGLEMAPS_API_KEY")
+        if googlemaps_api_key:
+            map_image_instruction = f"""
```
sugoi-yuzuru, line 78: Could you explain the need for delegating the LLM to decide whether to use Google Maps or a fall-back image? 
Can we update the client side code to decide this?

One thing that catches my eye here is that the code will embed a GoogleMaps API Key inside the prompt which is preferably kept private as possible.

doitundoit: Hi, @sugoi-yuzuru 

I agree that embedding the API key in the prompt is a security concern and the decision should move to the client.

To enable client-side decision making, I'm considering adding a new GoogleMap component. The server would send map data (coordinates, pins), and the client would decide how to render based on API key availability. This fully addresses your concerns but requires adding a new component and extending the catalog.

Would you be okay with this approach?

zeroasterisk:
Thanks for adding Google Maps Static API support to RizzCharts, @doitundoit! The map visualization feature with optional API key fallback is a nice enhancement. 🙏

The RizzCharts agent has been significantly modified since this PR was opened (removed and re-added in March 2026, plus the parser was refactored in [#780](https://github.com/google/A2UI/pull/780)), and this PR now has merge conflicts. It would need a rebase to be mergeable. Closing as stale — if you're still interested in contributing this feature, a fresh PR against the current codebase would be welcome!

PR #434, opened 2026-01-05
Fixes to uv.lock indexes

jonsharkey:
1. pyproject.toml files need to be configured to use a non-private repository. Some were already configured, but now they all are.
2. One of those was for the tests. I needed to regenerate the uv.lock file to use the public repository.
3. uv.lock files for the individual examples are not used. They all use `samples/agent/adk/uv.lock`. I deleted these.


jonsharkey:
> Thanks! This looks great. Do we still need a CI step to make sure nothing creeps in again in the future? (Ideally one that tries to run `uv sync` (or another command) with 3P-only credentials.)

I'll submit this now and look into a CI step for another PR.

PR #490, opened 2026-01-14
Remove data syncing from protocol

gspencergoog:
# Description

This removes the CRDT data synchronization from the protocol.  It was decided that this is relatively complex, and lacks a clear use case.  It has inherent value, but without a use case driving it, we felt it was underspecified.

Instead, to meet the need of having the client data model be available to the agent, we added an optional data broadcasting property to surfaces.  When this property is true, the client will append the entire data model to each A2A message as part of the metadata for the message.

## Summary of Changes

This pull request significantly refactors the data synchronization strategy within the protocol. The previous, intricate CRDT-based approach, which involved detailed conflict resolution and version tracking, has been replaced with a more straightforward data broadcasting model. This change aims to reduce protocol complexity and implementation overhead by ensuring the server receives the complete, current client data model upon user interaction, thereby eliminating the need for distributed state management and complex change tracking on the client side.

### Highlights

* **CRDT Data Synchronization Removal**: The complex CRDT (Conflict-free Replicated Data Type) based data synchronization mechanism, including Hybrid Logical Clocks (HLC) and Fractional Indexing, has been entirely removed from the protocol due to its complexity and lack of a clear use case.
* **New Data Broadcasting Mechanism**: A simpler, optional data broadcasting feature has been introduced. Surfaces can now be configured with a `broadcastDataModel` property, which, when true, causes the client to append the entire data model to the metadata of every A2A message sent to the server.
* **Protocol Simplification**: The protocol has been simplified by removing the `watchDataModel` message type and the `dataModelChanged` client-to-server message. Data model updates from the server to the client (`updateDataModel`) no longer require `actorId`, `hlc`, or `versions` fields, streamlining the update process.

<details>
<summary><b>Changelog</b></summary>

* **specification/v0_9/docs/a2ui_protocol.md**
    * Removed the `watchDataModel` message type from the protocol specification.
    * Updated the count of primary message types from five to four in the Envelope Message Structure section.
    * Added a new optional `broadcastDataModel` boolean property to the `createSurface` message, allowing clients to send their full data model with A2A messages.
    * Removed extensive documentation sections detailing CRDT, Hybrid Logical Clocks (HLC), Fractional Indexing, JSON Pointers, and Tombstones.
    * Revised the 'Data Model Updates: Synchronization and Convergence' section to describe the new broadcasting mechanism and server-to-client update semantics.
    * Updated the example stream for `updateDataModel` to remove `actorId`, `hlc`, and `versions` fields, reflecting the simplified schema.
    * Corrected a duplicate `deleteSurface` line in the message type list.
    * Clarified that deletion in data models should use `null` instead of `undefined`.
    * Updated the reference for client capabilities schema from `a2ui_client_capabilities_schema.json` to `a2ui_client_capabilities.json` and added a reference to `a2ui_data_broadcast.json`.
* **specification/v0_9/docs/evolution_guide.md**
    * Added a 'Data Synchronization' row to the summary table, highlighting the shift from implicit synchronization in v0.8.1 to explicit broadcasting in v0.9.
    * Introduced a new section '5.4. Data Model Broadcasting' to explain the rationale and mechanics of the new broadcasting approach, emphasizing its mechanical simplicity and removal of complex client-side tracking.
* **specification/v0_9/json/a2ui_client_capabilities_schema.json**
    * File renamed to `specification/v0_9/json/a2ui_client_capabilities.json`.
    * Added an `$id` property to the schema definition.
* **specification/v0_9/json/a2ui_data_broadcast.json**
    * Added a new JSON schema file defining the structure for broadcasting A2UI surface data models within A2A message metadata.
* **specification/v0_9/json/client_to_server.json**
    * Removed the `dataModelChanged` message type definition.
    * Removed `dataModelChanged` from the `oneOf` array, simplifying the valid client-to-server message types.
* **specification/v0_9/json/common_types.json**
    * Removed the `HlcString` and `VersionVector` definitions.
    * Modified the `DataUpdate` schema by removing `hlc` and `pos` properties.
    * Updated the description for the `value` property in `DataUpdate` to explicitly state that deletion is represented by setting it to `null`.
* **specification/v0_9/json/server_to_client.json**
    * Removed `WatchDataModelMessage` from the list of allowed server-to-client message types.
    * Added the `broadcastDataModel` boolean property to the `CreateSurfaceMessage` schema.
    * Removed `actorId` and `versions` properties from the `UpdateDataModelMessage` schema and updated its `required` fields.
    * Removed the `WatchDataModelMessage` definition.
* **specification/v0_9/test/cases/client_messages.json**
    * Removed the test case for the `dataModelChanged` message, aligning with its removal from the protocol.
* **specification/v0_9/test/cases/contact_form_example.jsonl**
    * Updated the `updateDataModel` entry to conform to the new schema, using an `updates` array and omitting `actorId`, `hlc`, and `versions`.
* **specification/v0_9/test/cases/contact_form_example_test.json**
    * Updated the `updateDataModel` test case to reflect the new schema, specifically using an `updates` array for data model changes.
</details>


specification/v0_9/docs/evolution_guide.md
```diff
@@ -222,10 +222,23 @@ Specifying an unknown surfaceId will cause an error. It is recommended that clie
...
+
+### 5.4. Data Synchronization
+
+**v0.8.1:**
+
+- Data synchronization was implicit and relied on ad-hoc mechanisms.
+
+**v0.9:**
+
+- **Explicit Broadcasting**: `createSurface` introduced `broadcastDataModel` (boolean).
```
jacobsimionato, line 238: I like this name 'broadcastDataModel'

specification/v0_9/json/server_to_client.json
```diff
@@ -75,17 +77,24 @@
...
+            "path": {
+              "type": "string",
+              "description": "An optional path to a location within the data model (e.g., '/user/name'). If omitted, or set to '/', refers to the entire data model."
             },
-            "versions": {
-              "description": "The version vector of the data model describing the state of the data model before the updates were applied.",
-              "$ref": "common_types.json#/$defs/VersionVector"
+            "value": {
+              "description": "The data to be updated in the data model. If present, the value at 'path' is replaced (or created). If omitted, the key at 'path' is removed.",
+              "type": [
```
jacobsimionato: This allows for any JSON type other than 'integer'. Is that deliberate to not allow integer? I assume integer gets coerced to number nicely by TS clients etc, so should we just omit the "type" declaration here so this can be *any* JSON value?

specification/v0_9/docs/evolution_guide.md
```diff
@@ -222,10 +222,23 @@ Specifying an unknown surfaceId will cause an error. It is recommended that clie
...
+- **Single-Path Updates**: Server pushes updates via `updateDataModel` using simple `path`/`value` pairs (no CRDTs or vector clocks).
+- **Broadcasting**: When `broadcastDataModel` is true, the client includes the full data model in every A2A message metadata.
+- **Reason**: Simplifies the protocol by removing complex conflict resolution (CRDTs) while ensuring the server always has the latest state when it needs to act.
```
jacobsimionato: Can remove this and mention of CRDTs to avoid confusion, seeing as this will be used by agents to upgrade code from v0.8 to v0.9

jacobsimionato (approved):

specification/v0_9/json/server_to_client.json
gspencergoog: Leaving out `int` was actually an oversight.  Removed the type field, because I forgot why I added it. There was a reason, but I can't think of what it was now.

specification/v0_9/docs/evolution_guide.md
gspencergoog: Yeah, this was Gemini leaking its task focus into the document.  Gone.

specification/v0_9/docs/evolution_guide.md
```diff
@@ -222,10 +222,22 @@ Specifying an unknown surfaceId will cause an error. It is recommended that clie
...
+
+**v0.8.1:**
+
+- Data synchronization was implicit and relied on ad-hoc mechanisms.
+
+**v0.9:**
+
+- **Explicit Broadcasting**: `createSurface` introduced `broadcastDataModel` (boolean).
+- **Single-Path Updates**: Server pushes updates via `updateDataModel` using simple `path`/`value` pairs.
+- **Broadcasting**: When `broadcastDataModel` is true, the client includes the full data model in every A2A message metadata.
```
wrenj, line 240: Add some blurb

- **Broadcasting**: When `broadcastDataModel` is true, the client includes the full data model for relevant surfaces in every A2A message metadata. Relevant surfaces are defined as surfaces that created the A2A message (eg, userAction from a button click), those that are active or viewable by the usable, and possibly recent ones in chat. Clients have some discretion on determining which surfaces are relevant; the intention is to prevent the A2A message from growing unbounded.

specification/v0_9/json/a2ui_data_broadcast.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://a2ui.dev/specification/v0_9/a2ui_data_broadcast.json",
+  "title": "A2UI Data Broadcast Schema",
+  "description": "Schema for broadcasting A2UI surface data models in A2A message metadata. This object should be placed in the `a2uiDataBroadcast` field of the metadata.",
```
wrenj, line 5: for broadcasting A2UI surfaces data from client to server

wrenj, line 5: explain the key is the surface id?

specification/v0_9/docs/evolution_guide.md
gspencergoog: I think that is too vague.  Agent implementers would not be able to rely on which data is available, and maybe their use case requires it all.

Maybe `broadcastDataModel` should have more configuration to say what is broadcast?  Then at least it is definitive as to what will be broadcast.  We could make it be an enum ("all", "onlyVisible", "onlyModified", etc.), or it could contain an object that has keys that are surface IDs, and values that are a list of paths within the data models to broadcast?

I understand the goal is to limit how large it grows.  Maybe it should have a size limit? But then how to decide what to truncate...  Hmm.  This feels like it needs more thought.

wrenj: eh I think less is more, we can leave it as is. If its an issue Gemini Enterprise can just say don't send more than 10 surfaces worth of datamodels or something, happy to just leave it out

PR #491, opened 2026-01-15
Move weight property to standard catalog

jacobsimionato:
This PR moves the `weight` property from `common_types.json` to `standard_catalog.json`.

**Rationale:**
- Make the `weight` property specific to the `standard_catalog` rather than a core primitive.
- Establish the pattern that A2UI is not opinionated about specific component properties.
- Ensure catalogs and renderers are free to include whatever common properties they want without being bound by the core protocol.


jacobsimionato:

gspencergoog (approved):

PR #497, opened 2026-01-16
v0.9 Spec Update: DateTimeInput Refinements and openUrl Function

jacobsimionato:
This PR updates the A2UI v0.9 specification with the following changes:

- **DateTimeInput**: Removed the `outputFormat` property as it was underspecified and better handled at the presentation layer. Added `min` and `max` properties to support date/time range constraints using ISO 8601 strings.
- **openUrl**: Added a new `openUrl` function to the standard catalog to facilitate opening URLs (e.g., for links).
- **Schema Updates**: Updated `common_types.json` and `a2ui_client_capabilities.json` to include `void` as a valid function return type, supporting the new `openUrl` function.
- **Generalized Actions**: Generalized the `action` concept by moving the definition to `common_types.json`. It now supports:
    - Server-side actions via an `event` wrapper.
    - Local client-side actions via a `functionCall` wrapper.
    - Updated `Button` component to use this new shared `Action` type.

### Examples

**Server Action (Event):**
```json
"action": {
  "event": {
    "name": "submit_form",
    "context": {
      "itemId": "123"
    }
  }
}
```

**Local Action (Function Call):**
```json
"action": {
  "functionCall": {
    "call": "openUrl",
    "args": {
      "url": "${/url}"
    }
  }
}
```


wrenj (approved):

gspencergoog:
> * **openUrl**: Added a new `openUrl` function to the standard catalog to facilitate opening URLs (e.g., for links).

This is a good addition, but I'm wondering if we need to also include some hints about how to use it (i.e. with buttons): I don't want the LLM to go "Oooh, links! I know how to do Markdown links, so I'll add those to all my text components!".  I mean, unless we want to also allow links in Markdown.  I think if we do, that will require more complex configuration of the Text widget to specify the functions that the links activate, which might add a lot of complexity that people would have to duplicate on any custom text component.

specification/v0_9/json/common_types.json
```diff
@@ -211,7 +211,7 @@
         "returnType": {
           "type": "string",
           "description": "The expected return type of the function call.",
-          "enum": ["string", "number", "boolean", "array", "object", "any"],
+          "enum": ["string", "number", "boolean", "array", "object", "any", "void"],
```
gspencergoog, line 150: Why is "any" not sufficient for this?  I mean, void is more explicit, and I like that, but it isn't necessary.

specification/v0_9/json/common_types.json
```diff
@@ -317,6 +317,46 @@
...
+          "required": ["name"],
+          "additionalProperties": false
+        },
+        {
+          "type": "object",
+          "description": "Executes a local client-side function.",
+          "properties": {
+            "function": {
+              "type": "string",
+              "description": "A string specifying the local function to trigger (e.g., 'openUrl(${/url})'). This string can represent nested function calls and data model references."
```
gspencergoog: If we're going to do function calls like this here, then we should probably make the  `FunctionCall` type consistent with this.  `FunctionCall` should just have one argument "call" that is this formatted string. This will require the same parsing and change registration as the string format function will.

Or we should make this consistent and just replace this entire definition with a `FunctionCall` definition, which would mean that the server action needs to have a discriminator field.

specification/v0_9/json/standard_catalog.json
```diff
@@ -597,9 +584,13 @@
               "type": "boolean",
               "description": "If true, allows the user to select a time."
             },
-            "outputFormat": {
-              "type": "string",
-              "description": "The desired format for the output string after a date or time is selected."
+            "min": {
```
gspencergoog, line 588: I would add regex constraints to these that force them to be data/time strings in UTC time zone:

e.g. `^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:\\.\\d{1,9})?Z$`

specification/v0_9/json/common_types.json
gspencergoog: One advantage of the second option is that it can be validated purely with JSON.

specification/v0_9/json/common_types.json
jacobsimionato: I worry that not having a way to express a void return type explicitly creates confusion about expectations for LLMs and humans. E.g. imagine a catalog has a function `any loadBooks()` to trigger a "fire and forget" async load which doesn't actually return any data synchronously, and an "array getBooks()" function which returns books from the local database. I could imagine an agent building a layout that attempts to use `loadBooks()` as a parameter value to load data seeing as it may return something, while really it should use `getBooks()`. Having the void return type makes it super clear what the contract is.

I don't actually feel that strongly though - I'd be okay with 'any' if you feel strongly.

specification/v0_9/json/common_types.json
jacobsimionato: Great point. I actually kind of missed that FunctionCall was already defined, sorry.

I used it here - is this what you were imagining?

What do you think about the discriminator pattern to differentiate between functionCall (which is super clear, but adds a layer of nesting to the output) vs the "oneOf" style (more token efficient but less clear)?

specification/v0_9/json/standard_catalog.json
jacobsimionato: Good idea. It looks like JSON schema has something built-in for this, so I added it. It's a bit messy because these are actually DynamicStrings, but maybe it's fine.

jacobsimionato:
> > * **openUrl**: Added a new `openUrl` function to the standard catalog to facilitate opening URLs (e.g., for links).
> 
> This is a good addition, but I'm wsdondering if dso also allow links in Markdown. I think if we do, that will require more complex configuration of the Tesdfsdfxt widget to specify the functions that the links activate, which might add a lot of complexity that people would have to duplicate on any custom text component.

Re markdown links: I actually think we should support them, and we sort of do implicitly already in the spec seeing as we support markdown. I don't think they need to have any relationship with the openUrl function I added here - renderers should just make markdown links work as best as they can, the same as they already do for plain non-A2UI markdown content.

Re helping the LLM: Yes, that's fair! Do you think it should block this PR? I kind of want to just get the spec right here, and then we can figure out prompting strategies later. But also, this comment is also a warning sign that the spec is getting too complex. Do you think this is good complexity to add now? Or too much? Personally, I think it's a natural use of the Function concept you added and there is 1P customer demand which motivates me to do it now.

specification/v0_9/json/common_types.json
gspencergoog: I don't feel that strongly.  Let's keep "void".

specification/v0_9/json/standard_catalog.json
gspencergoog: Yes, I think that's fine.

specification/v0_9/json/common_types.json
gspencergoog: Yes, this looks good.  I think the discriminator pattern is clearer.

PR #500, opened 2026-01-16
Generalize A2UI Actions (v0.9)

jacobsimionato:
Generalizes the concept of A2UI actions by moving the `action` schema from the `Button` component to `common_types.json` as a shared `Action` type.

This new `Action` type supports:
1.  The existing server-side action structure (object with `name` and `context`).
2.  A new local action capability, specified via a `function` parameter (string or object).

Example of local action usage: `"action": "openUrl(${/url})"`

This allows components other than Button to potentially define actions in the future, and enables local client-side behavior.

### DIFF
--- a/specification/v0_9/docs/a2ui_protocol.md
+++ b/specification/v0_9/docs/a2ui_protocol.md
@@ -270,6 +270,53 @@ flowchart TD
 
 ```
 
+### Defining Actions
+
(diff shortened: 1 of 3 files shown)


jacobsimionato:
Dropping in favour of https://github.com/google/A2UI/pull/497

Issue #207, opened 2025-12-05
Consider adding "borderless" or "fullBleed" or "plain" usageHint on "Button"


jacobsimionato:
There are sometimes buttons (defining "button" loosely) where the visual appearance of the button is entirely the child component. For example, text you can click that triggers some action, or an icon, or an image-defined button. Obviously, this has a compromises around splash effects etc, but it is a pattern.

If you try to make an icon button right now, it often looks weird because the icon isn't properly centered etc.

It might be nice to instead have a usageHint that can remove all button appearance other than the child, so the button wrapper is used purely to add tap detection and define the action.

<img width="644" height="395" alt="Image" src="https://github.com/user-attachments/assets/b367edf4-914c-4737-b648-26e88d4ae28c" />

jacobsimionato:
CC @gspencergoog

jacobsimionato:
Addressed by https://github.com/google/A2UI/pull/501

(edited :-D)

jacobsimionato:
Addressed by

PR #501, opened 2026-01-16
Update v0.9 Spec: Add borderless Button variant alongside Primary

jacobsimionato:
This PR updates the v0.9 specification to align the `Button` component with other components by replacing the boolean `primary` property with a `variant` property (supporting 'primary' and 'borderless').


jacobsimionato:
Addresses https://github.com/google/A2UI/issues/207

PR #504, opened 2026-01-16
Proposal with simple interpolation

wrenj:
Proposal using just simple client functions instead of interpolation syntax

### DIFF
--- a/specification/v0_9/docs/a2ui_protocol.md
+++ b/specification/v0_9/docs/a2ui_protocol.md
@@ -270,9 +270,9 @@ flowchart TD
 
 ```
 
-## Data Model Representation: Binding, Scope, and Interpolation
+## Data Model Representation: Binding and Scope
 
-This section describes how UI components **represent** and reference data from the Data Model. A2UI relies on a strictly defined relationship between the UI structure (Components) and the state (Data Model), defining the mechanics of path resolution, variable scope during iteration, and interpolation.
+This section describes how UI components **represent** and reference data from the Data Model. A2UI relies on a strictly defined relationship between the UI structure (Components) and the state (Data Model), defining the mechanics of path resolution, variable scope during iteration.
 
 ### Path Resolution & Scope
 
@@ -341,22 +341,6 @@ When a container component (such as `Column`, `Row`, or `List`) utilizes the **T
   // "/company" is Absolute. Resolves to "Acme Corp" globally.
 }
 ```
-
-#### Client-Side Functions
-
-Results of client-side functions can be interpolated. Function calls are identified by the presence of parentheses `()`.
-
-- `${now()}`: A function with no arguments.
-- `${formatDate(${/currentDate}, 'yyyy-MM-dd')}`: A function with positional arguments.
-
-Arguments can be **Literals** (quoted strings, numbers, or booleans), or **Nested Expressions**.
-
-#### Nested Interpolation
-
-Expressions can be nested using additional `${...}` wrappers inside an outer expression to make bindings explicit or to chain function calls.
-
-- **Explicit Binding**: `${formatDate(${/currentDate}, 'yyyy-MM-dd')}`
-- **Nested Functions**: `${upper(${now()})}`
 
 #### Type Conversion
 
@@ -568,6 +552,8 @@ The [`standard_catalog.json`] provides the baseline set of components and functi
 
 ### Functions
 
+#### Validation Functions
+
 | Function          | Description                                                              |
 | :---------------- | :----------------------------------------------------------------------- |
 | **required**      | Checks that the value is not null, undefined, or empty.                  |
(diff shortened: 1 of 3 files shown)


specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -568,61 +552,47 @@ The [`standard_catalog.json`] provides the baseline set of components and functi
...
 {
-  "id": "user_welcome",
+  "id": "receipt",
   "component": "Text",
   "text": {
-    "call": "string_format",
+    "call": "formatString",
     "args": {
-      "value": "Hello, ${/user/firstName}! Welcome back to ${/appName}."
+      "template": "Bought ${quantity} ${item} on ${date} for ${price}.",
```
gspencergoog, line 584: I see what you're going for here, but it seems like it's a lot more verbose than the previous design, and still requires parsing/validating of the template to interpolate the values, and escaping of the template markers.  I realize that it's a simpler parser, but it's not that much simpler (and that's not really where we want to optimize complexity).

An equivalent `string_format` argument would be:

```
"Bought ${formatNumber(${/transaction/count}, 0)} ${pluralize(${/transaction/count}, 'apple', 'apples')}  on ${formatDate(${/transaction/timestamp}, 'MMM d, yyyy')} for ${formatCurrency{${/transaction/amount}, 'USD')}"
```

To me, this seems easier/smaller to generate, since it doesn't require the extra argument names, "call", and "args" objects and eliminates much of the nesting. It has only marginally harder parsing complexity (which we'll put into a reusable SDK function anyhow).

specification/v0_9/json/standard_catalog.json
```diff
@@ -716,20 +716,107 @@
...
+            "description": "A map of named values to be interpolated into the template. Keys must match the placeholders in the template.",
+            "additionalProperties": true
           }
-        ],
-        "unevaluatedProperties": false
+        }
+      }
+    },
+    {
+      "name": "formatNumber",
```
jacobsimionato, line 739: I prefer this lower-magic version of string interpolation function calling overall. But I still wonder if there is some middle ground where you need to explicitly pass arguments to stringFormat, but these specific formatting options for certain types are still built into the formatString function rather than separated out, which makes the "call sites" super complex.

E.g. trying to take inspiration from Python:
```
{
  "id": "receipt",
  "component": "Text",
  "text": {
    "call": "formatString",
    "args": {
      "template": "Bought ${quantity} ${item} on ${date:%m %d, %Y} for ${price.2f}.",
      "args": {
        "quantity": {path: "/transaction/count"},
        "item": {path: "/transaction/name"},
        "date": {path: "/transaction/timestamp"},
        "price": {path: "/transaction/amount"},
      }
    }
  }
}
```

At least this way, the core spec and data flow is still pretty simple and synchronous, but the standard catalog just happens to have a slightly magical function that can understand some fancy formatting placeholders.

jacobsimionato (approved):

specification/v0_9/json/standard_catalog.json
wrenj: That was my original proposal, but I thought we wanted to do the dart / calling functions approach for string formatting. I'd actually prefer just a single string_format function if that what you prefer. @gspencergoog thoughts?

derekdon:
This direction makes more sense to me than the original proposal. Moving formatting out of inline string templates avoids drifting into eval-like territory, reduces brittleness, and aligns better with the trust-boundary goals of the spec. Pre-formatting values before interpolation is a cleaner and more predictable model.

I was going to suggest something similar conceptually, though I wonder if this could be taken one step further by separating concerns more explicitly. Instead of expanding a generic functions map, it may be cleaner to introduce a dedicated formatters (or format) catalog specifically for interpolation.

That catalog could expose a predefined set of formatters (date, number, currency, pluralization, etc.) that operate over resolved values, rather than requiring the model or spec author to reason about low-level formatting details. In most cases these could defer to standard platform APIs and inferred context, for example locale detection:

```
Intl.NumberFormat().format(number)
Intl.NumberFormat("de-DE").format(number)
Intl.DateTimeFormat(locale, options).format(date)
```

This keeps formatting deterministic, testable, and implementation-owned, rather than encoded in templates or model output. Users could still register custom formatters where needed, but the common cases would be handled consistently out of the box.

I also think this structure makes it easier to support full i18n down the line. Once the template itself is treated as an i18n resource, with values passed in as already-formatted tokens, localization becomes a much more natural extension of the system rather than a retrofit.

Overall, the move from:

`Today is ${formatDate(${/timestamp}, 'yyyy-MM-dd')}`

to:

`Today is ${date}`

is the right abstraction boundary IMO. The remaining question is mostly about where those formatting primitives live and how strongly we want to constrain them for safety, portability, and localization.

wrenj:
@derekdon Thanks for the comment. Our thinking was to put something super basic in here so its easy for open source renderers to implement and let clients add more advanced formatting like i18n in custom catalogs. We had looked at just wrapping MessageFormat syntax https://messageformat.github.io/messageformat/guide/ but didn't want each renderer to have to import the message format library. Based on Jacob's comment above and to keep this simple I think I'll just make this a simple string_format function with some basic python style formatters.

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -342,22 +342,6 @@ When a container component (such as `Column`, `Row`, or `List`) utilizes the **T
 }
 ```
 
-#### Client-Side Functions
```
jacobsimionato, line 345: @gspencergoog this defines a DSL to call functions from within string templates. I think if we do decide to support a syntax like this, we should move this specification to the "standard catalog" section and make it specific to the `string_format` function etc. So maybe `string_format` happens to use some magical DSL and be able to refer to the data model and other client-side functions directly (we will expose the necessary rendering framework APIs to make this feasible to implement). A different catalog can invent some other magical DSL. But the core A2UI specification is more purely JSON-based.

specification/v0_9/json/standard_catalog.json
gspencergoog: I talked this over with Jacob, and I still think that I'd rather use the more "modern" version of string interpolation that just takes a single string:

- It it just simpler to construct if the LLM needs to construct one string instead of a string and also a separate set of nested args.
- They both have equivalent functionality, and don't provide any extra validation, since any of these options need to have validation that involves validating the template or string against the args anyhow.
- Having a fixed set of built in formatting options is not as flexible in the long run, although I guess there's nothing saying we couldn't have some built in ones along with the ability to interpolate function outputs.
- Also, with a single string there's no chance of having a mismatch between the number of arguments and the substitution points.

So, how about this compromise: We keep the args in the string, but include some basic formatters that don't need function calls?

Like so:

```json
{
  "id": "receipt",
  "component": "Text",
  "text": {
    "call": "formatString",
    "args": {
      "template": "Bought ${/transaction/quantity} ${/transaction/name} on ${/transaction/timestamp:%m %d, %Y} for ${/transaction/amount:.2f}.",
    }
  }
}
```

In the end, I think the best solution will be whatever generates the best results from the various LLMs, and we kind of need to run an implementation through evals to know that definitively.

jacobsimionato: Just wanted to add: if we remove mention of the DSL above from the core A2UI specification, and just include it in the definition for the stringFormat function in the standard catalog, then developers using custom catalogs would be completely free to implement your exact proposal, @wrenj . 

So people using the standard catalog get a convenient string format syntax that is (hopefully!) simple enough for humans and LLMs to write directly. But Catalog creators have the hooks they need to innovate with other approaches, and they are not tethered to the DSL we are using in the standard catalog.

Re the actual DSL (sorry, not sure if it technically counts as a language), maybe we should try to stick more closely to the Python fstring format, e.g. by using `{/dataModelReference}` instead of `${/dataModelReference}`, so the LLM can leverage its knowledge there. Though this might encourage it to hallucinate python code and formatting options. Just seems confusing to be quite close to Python, but then arbitrarily make some different design decisions.

specification/v0_9/docs/a2ui_protocol.md
gspencergoog: Yeah, this section should be removed.  I agree that it should only be part of the string_format function's definition, and not general for all strings. (I thought I had already removed that when I added  `string_format`).

specification/v0_9/json/standard_catalog.json
gspencergoog: > Re the actual DSL (sorry, not sure if it technically counts as a language), maybe we should try to stick more closely to the Python fstring format,

It was intended to be close to TypeScript, JavaScript, Dart, Bash, Kotlin, PHP and Groovy.  I figured that was a larger learning base than Python and Rust.

jacobsimionato: > It was intended to be close to TypeScript, JavaScript, Dart, Bash, Kotlin, PHP and Groovy. I figured that was a larger learning base than Python and Rust.

Oh that makes perfect sense. My language knowledge is not as broad as yours :-D. I was referring to the colon syntax in `{/transaction/amount:.2f}` which I suggested based on Python, but probably its the same as other languages.

Overall, I don't feel strongly other than borrowing familiar syntax ideas, which we already have!

specification/v0_9/json/standard_catalog.json
```diff
@@ -716,20 +716,107 @@
...
+        "type": "object",
+        "required": [ "value" ],
+        "properties": {
+          "value": {
+            "$ref": "common_types.json#/$defs/DynamicNumber",
+            "description": "The monetary amount."
+          },
+          "currencyCode": {
+            "$ref": "common_types.json#/$defs/DynamicString",
+            "description": "Optional. The ISO 4217 currency code (e.g., 'USD', 'EUR'). If omitted, uses the device's default currency."
```
gspencergoog, line 775: How does the LLM know the device's default currency to be able to put in the relevant value?

wrenj:
chatted offline with greg, we'll keep the syntax as is and make sure any custom function can access the interpolation functionality as well as hook up databinding, probably using a callcontext that can resolve data model values or functions

specification/v0_9/json/standard_catalog.json
```diff
@@ -716,20 +716,107 @@
...
+          "pattern": {
+            "$ref": "common_types.json#/$defs/DynamicString",
+            "description": "A Unicode TR35 date pattern string.\n\nToken Reference:\n- Year: 'yy' (26), 'yyyy' (2026)\n- Month: 'M' (1), 'MM' (01), 'MMM' (Jan), 'MMMM' (January)\n- Day: 'd' (1), 'dd' (01), 'E' (Tue), 'EEEE' (Tuesday)\n- Hour (12h): 'h' (1-12), 'hh' (01-12) - requires 'a' for AM/PM\n- Hour (24h): 'H' (0-23), 'HH' (00-23) - Military Time\n- Minute: 'mm' (00-59)\n- Second: 'ss' (00-59)\n- Period: 'a' (AM/PM)\n\nExamples:\n- 'MMM dd, yyyy' -> 'Jan 16, 2026'\n- 'HH:mm' -> '14:30' (Military)\n- 'h:mm a' -> '2:30 PM'\n- 'EEEE, d MMMM' -> 'Friday, 16 January'"
+          }
+        }
+      }
+    },
+    {
+      "name": "pluralize",
+      "description": "Returns the singular string if the count is 1, otherwise returns the plural string.",
```
gspencergoog, line 801: Note that this definition is only sufficient for English text. Other languages will need more complex rules, and more arguments.

PR #510, opened 2026-01-20
Simplify server-side validation of A2UI messages by adding ChildId to explicitly mark component references

jacobsimionato:
This PR introduces a dedicated `ChildId` type to the `common_types.json` schema and updates the standard catalog and documentation to use it. It also completely removes the redundant `id` type definition, replacing all references with `ChildId`.

### Rationale

In A2UI v0.9, validating referential integrity (ensuring that a component's `child` or `trigger` property points to a valid ID in the same surface) currently relies on heuristics. A validator has to "guess" which fields represent IDs based on naming conventions (e.g., checking if a property is named `child`, `children`, `content`, or `trigger`).

This approach is brittle, especially for **Custom Catalogs**, where a developer might introduce a new component (e.g., `SplitView`) with properties like `leftPane` and `rightPane`. A generic validator would not know these are structural links that need to be checked.

### The Fix

We introduce a semantic marker type in `common_types.json`:

```json
"ChildId": {
  "type": "string",
  "description": "The unique identifier for a component, used for both definitions and references within the same surface."
}
```

We then update the standard catalog, the definition of `ChildList`, and `ComponentCommon` to explicitly `$ref` this type. This ensures that every `id` property and every property that refers to an `id` is explicitly typed.

### Validator Logic (Hypothetical)

With this change, a validator can be implemented deterministically without hardcoded property names. It simply inspects the resolved schema for each property:

```python
# Pseudocode for a robust validator
def validate_component(component, schema):
    for prop, value in component.items():
        prop_schema = resolve_schema(prop, schema)
        
        # Check if this property is structurally defined as an ID Reference
        if prop_schema[''] == 'common_types.json#//ChildId':
            if value not in known_surface_ids:
                raise ValidationError(f"Dangling reference: Component '{value}' not found.")
```

This ensures that any component in any catalog (standard or custom) that uses these types will automatically benefit from referential integrity checks.

### DIFF
--- a/specification/v0_9/docs/a2ui_protocol.md
+++ b/specification/v0_9/docs/a2ui_protocol.md
@@ -88,10 +88,10 @@ The [`common_types.json`] schema defines reusable primitives used throughout the
 - **`DynamicString` / `DynamicNumber` / `DynamicBoolean` / `DynamicStringList`**: The core of the data binding system. Any property that can be bound to data is defined as a `Dynamic*` type. It accepts either a literal value, a `path` string ([JSON Pointer]), or a `FunctionCall` (function call).
 - **`ChildList`**: Defines how containers hold children. It supports:
 
-  - `array`: A static array of string component IDs.
+  - `array`: A static array of `ComponentId` component references.
   - `object`: A template for generating children from a data binding list (requires a template `componentId` and a data binding `path`).
 
-- **`ChildId`**: A reference to the unique ID of another component within the same surface.
+- **`ComponentId`**: A reference to the unique ID of another component within the same surface.
 
 ### Server to Client Message Structure: The Envelope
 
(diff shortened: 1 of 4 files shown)


specification/v0_9/json/common_types.json
```diff
@@ -31,7 +35,7 @@
           "description": "A template for generating a dynamic list of children from a data model list. The `componentId` is the component to use as a template.",
           "properties": {
             "componentId": {
-              "$ref": "#/$defs/id"
```
gspencergoog, line 34: Shouldn't this PR remove the `id` type, since it's no longer used?  The`id` was meant to serve a similar purpose, with perhaps a too-generic name.

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -103,6 +103,19 @@ The [`standard_catalog.json`] schema contains the definitions for all specific U
 
 Custom catalogs can be used to define additional UI components or modify the behavior of existing components. To use a custom catalog, simply include it in the prompt in place of the standard catalog. It should have the same form as the standard catalog, and use common elements in the [`common_types.json`] schema.
 
+### Validator Compliance & Custom Catalogs
+
+To ensure that automated validators can verify the integrity of your UI tree (checking that parents reference existing children), custom catalogs MUST adhere to the following strict typing rules:
+
+1.  **Single Child References:** Any property that holds the ID of another component MUST use the `ComponentIdRef` type defined in `common_types.json`.
```
nan-yu: Should we be more specific to name it `childIdRef`?

specification/v0_9/json/common_types.json
jacobsimionato: Oh yes! Ummm sorry I was vibecoding too freely and missed that there was an ID type already defined right above my new one.

specification/v0_9/docs/a2ui_protocol.md
jacobsimionato: How about just `ComponentId`?

I like the idea of including "Ref" in the name, to imply that it's a validated reference. But I think it's even neater if we can share the same type between declarations and references, in which case a generic name makes more sense.

I think when we write a validator, we can have special cased logic so that it knows that the root `id` parameter of each Component is a declaration (seeing as this is part of the core A2UI specification), but any other parameters with type ComponentId must be references.

nan-yu: Sounds good to me. Thanks!

PR #512, opened 2026-01-20
refactor: Separate out Typescript classes that are shared between Ang…

sugoi-yuzuru:
This PR introduces a new package `@a2ui/web_core` which may need to be released before introducing dependencies to it from `@a2ui/angular` and `@a2ui/lit`.

# Description

Separate out Typescript classes that are shared between Angular & Lit Renderer from the `@a2ui/lit` package into `@a2ui/web_core`.

This change will allow the `@a2ui/angular` to depend on `@a2ui/web_core` instead of the `@a2ui/lit` package and reduce unnecessary code dependencies. This reduced scope of dependency will also reduce the risk of component tags collision where catalog components defined in Angular space conflicting the catalog components defined in Lit space.

## Tested sample applications 
 
 - [x] Lit - Contact : https://screencast.googleplex.com/cast/NTAwNDk3ODkxNjY4Nzg3Mnw2OGQ4ZTZlYi1iMQ
 - [x] Angular - Contact :  https://screencast.googleplex.com/cast/NTE5Mjg5OTIyMTI1ODI0MHw1ODFhNDZlNi02MQ
 - [x] Angular - Restaurant : https://screencast.googleplex.com/cast/NjY0MTA0MDE4OTU1NDY4OHw2OTljYjQ2MC0zNg
 - [x] Angular - Rizzcharts : https://screencast.googleplex.com/cast/NDY0NzI2ODMyMDgwNDg2NHw2NTg4MGJiMy05Ng
 - [x] Angular - Orchestrator : https://screencast.googleplex.com/cast/NTUwNzAyMjM5MTg2OTQ0MHw4MzRiZDE5NC0wYw
 - [x] Angular - Gallery : https://screencast.googleplex.com/cast/NjE4NjU3OTQ5OTQxNzYwMHxlMTI4NWQ4Yy04Zg
 - [x]  Tools - Editor : https://screencast.googleplex.com/cast/NTkyMzIzNzk0MTczOTUyMHw1ZjQ3OGJkMi0xYg
 - [x]  Tools - Inspector : https://screencast.googleplex.com/cast/NjYxMTk5NzYyMDY5OTEzNnxiOTcyN2Y3YS1mNw

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md


ava-cassiopeia:
I am intending to review this, but due to the quantity of files, it may take me a second to make it through this (even though the changes are mostly fairly straightforward).

renderers/lit/src/0.8/ui/root.ts
```diff
@@ -28,13 +28,16 @@ import {
...
-import { Theme, AnyComponentNode, SurfaceID } from "../types/types.js";
+import { A2uiMessageProcessor, Primitives, Types } from "@a2ui/web_core";
 import { themeContext } from "./context/theme.js";
 import { structuralStyles } from "./styles.js";
 import { componentRegistry } from "./component-registry.js";
 
+type AnyComponentNode = Types.AnyComponentNode;
+type SurfaceID = Types.SurfaceID;
+type Theme = Types.Theme;
+type StringValue = Primitives.StringValue;
```
ava-cassiopeia: Could you just import these types directly from `@a2ui/web_core/types/types` or something instead of renaming them like this?

renderers/lit/src/0.8/ui/text-field.ts
```diff
@@ -17,24 +17,22 @@
 import { html, css, nothing } from "lit";
 import { customElement, property } from "lit/decorators.js";
 import { Root } from "./root.js";
-import { StringValue } from "../types/primitives.js";
+import { Primitives, Types, A2uiMessageProcessor } from "@a2ui/web_core";
```
ava-cassiopeia: Same as my comment above, do you think it's more readable to include the `Primitives` / `Types` namespaces?

renderers/lit/prepare-publish.mjs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// 4. Adjust Paths for Dist
+// The built files are in dist/src/, but when we publish dist/, they become src/
+function adjustPath(p) {
+  if (p && p.startsWith('./dist/')) {
+    return './' + p.substring(7); // Remove ./dist/
+  }
+  return p;
+}
```
ava-cassiopeia: **NIT:** Can this function live in a different file or at the bottom of the page, so it's not in the middle of the logic?

renderers/lit/prepare-publish.mjs
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+// 5. Write to dist/package.json
+writeFileSync(join(distDir, 'package.json'), JSON.stringify(litPkg, null, 2));
+
+// 6. Copy README and LICENSE if they exist
+['README.md', 'LICENSE'].forEach(file => {
+  const src = join(dirname, file);
+  if (existsSync(src)) {
```
ava-cassiopeia: Should we throw if these don't exist? We expect (and want) them to, right?

samples/client/angular/angular.json
```diff
@@ -443,8 +443,8 @@
               "budgets": [
                 {
                   "type": "initial",
-                  "maximumWarning": "500kB",
-                  "maximumError": "1MB"
+                  "maximumWarning": "1MB",
+                  "maximumError": "2MB"
```
ava-cassiopeia: This is a big jump, why did the bundle size get so much larger?

sugoi-yuzuru: Thanks for checking. 
I was able to revert this change. However, orchestrator is currently at 900KB+ so we are just below the threshold. 

I noticed that I had forgotten to remove reference to Lit renderer from the sample app which caused an increase in the size. But this has been cleaned up now and we are back to below the original threshold.

renderers/lit/src/0.8/ui/root.ts
sugoi-yuzuru: I can inline the namespace'd accessor. Will that be preferable?

I'm inclined to keep the namespaces so that each file can limit the scope of the elements that they want to import from `@a2ui/web_core`.

renderers/lit/src/0.8/ui/text-field.ts
sugoi-yuzuru: Readability wise, I feel there are pros and cons. `@a2ui/web_core` concepts are more recognizable and their groupings become more obvious inline, while making the code a bit verbose. 

However, my larger motivation was to offer individual files granular control over which part of the `@a2ui/web_core` library it wants to import. With some of the generic naming we have, I noticed that wholesale importing can lead to type-name conflicts.

renderers/lit/prepare-publish.mjs
sugoi-yuzuru: Done.

renderers/lit/prepare-publish.mjs
sugoi-yuzuru: Done.

renderers/lit/src/0.8/ui/root.ts
ava-cassiopeia: Yeah that would be fine, I think if we're leaning into the namespaces, then we should use them consistently.

renderers/lit/src/0.8/ui/text-field.ts
ava-cassiopeia: That's fair, I think we can leave this as-is for now, noting my comment about about consistently using the namespaces. Deep tree shaking should be able to sort those out, and they add context.

renderers/lit/src/0.8/ui/root.ts
sugoi-yuzuru: I have revised the way `@a2ui/web_core` exports smaller packages to remove the need for namespacing.

renderers/lit/src/0.8/ui/text-field.ts
sugoi-yuzuru: Chatted offline, and made updates to enable package-specific import in the following fashion. 

```
import { StringValue } from "@a2ui/web_core/types/primitives";
```

renderers/web_core/src/v0_8/schemas/a2ui_client_capabilities_schema.json
```diff
@@ -0,0 +1,23 @@
+{
```
wrenj, line 1: approved

however in a followon CL can you find a way to use relative paths or data packages so we don't duplicate these files?

PR #517, opened 2026-01-21
Remove String Interpolation section

gspencergoog:
# Description

This removes the description of general string interpolation and leaves only the `string_format` function that can do string interpolation.  This allows clients to use their own interpolation, and simplifies implementation of a basic renderer that doesn't use the standard catalog.

### DIFF
--- a/specification/v0_9/docs/evolution_guide.md
+++ b/specification/v0_9/docs/evolution_guide.md
@@ -229,7 +229,7 @@ Specifying an unknown surfaceId will cause an error. It is recommended that clie
 - **String Formatting**: Introduced the `string_format` function, which supports `${expression}` syntax for interpolation.
 - **Unified Expression Language**: Allows embedding JSON Pointer paths (absolute and relative) and client-side function calls directly within the format string.
 - **Nesting**: Supports recursive nesting of expressions (e.g., `${formatDate(${/timestamp}, 'yyyy-MM-dd')}`).
-- **Restriction**: Note that string interpolation `${...}` is **ONLY** supported within the `string_format` function. It is not supported in other string properties to strictly separate data binding definitions from static content.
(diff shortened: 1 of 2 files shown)


specification/v0_9/docs/evolution_guide.md
```diff
@@ -229,6 +229,7 @@ Specifying an unknown surfaceId will cause an error. It is recommended that clie
 - **String Formatting**: Introduced the `string_format` function, which supports `${expression}` syntax for interpolation.
 - **Unified Expression Language**: Allows embedding JSON Pointer paths (absolute and relative) and client-side function calls directly within the format string.
 - **Nesting**: Supports recursive nesting of expressions (e.g., `${formatDate(${/timestamp}, 'yyyy-MM-dd')}`).
+- **Restriction**: Note that string interpolation `${...}` is **ONLY** supported within the `string_format` function. It is not supported in other string properties to strictly separate data binding definitions from static content.
```
wrenj: is **ONLY** supported with the string_format function and in client functions that explicitly implement string interpolation

wrenj (approved):

specification/v0_9/docs/evolution_guide.md
jacobsimionato: I think that other client functions that happen to implement string interpolation using the same syntax as `string_format` are out of scope for this doc though, right?

jacobsimionato (approved):

specification/v0_9/docs/evolution_guide.md
gspencergoog: > I think that other client functions that happen to implement string interpolation using the same syntax as string_format are out of scope for this doc though, right?

Yes, I think so.  Let's just remove mention of other client functions. When I was adding that, I was thinking "but what if they don't implement *this* syntax".

PR #524, opened 2026-01-21
add accessibility attributes to components in standard catalog

wrenj:
Please double check. Gemini did implementation with some tweaks

### DIFF
--- a/specification/v0_9/json/common_types.json
+++ b/specification/v0_9/json/common_types.json
@@ -16,7 +16,7 @@
           "$ref": "#/$defs/DynamicString",
           "description": "Overrides the content description. Use this to describe icons or provide extra context for buttons."
         },
-        "hint": {
+        "description": {
           "$ref": "#/$defs/DynamicString",
           "description": "Describes the result of performing an action on the element (e.g., 'Double tap to submit')."
         },


gspencergoog:
The Gemmi review comments do seem relevant.

specification/v0_9/json/common_types.json
```diff
@@ -8,11 +8,72 @@
       "type": "string",
       "description": "The unique identifier for a component, used for both definitions and references within the same surface."
     },
+    "AccessibilityAttributes": {
```
gspencergoog, line 11: This is comprehensive, but it will add a lot to the complexity and size of the schema.  We may need to consider this when we look at filtering of the schema in the SDKs for the renderers.  No changes needed here, though.

gspencergoog (commented):
How are accessibility state changes reflected back to the agent?  Does it have its own event mechanism and context?

wrenj:
> How are accessibility state changes reflected back to the agent? Does it have its own event mechanism and context?

Only way for agent to learn about updates would be to bind them to the datamodel.

I would be fine cutting accessibility state and saying these need to be managed by the components themselves. eg, the input can say if its invalid or not based on validation. WDYT?

wrenj:
> The Gemmi review comments do seem relevant.

removed the altTexts

wrenj:
> > How are accessibility state changes reflected back to the agent? Does it have its own event mechanism and context?
> 
> Only way for agent to learn about updates would be to bind them to the datamodel.
> 
> I would be fine cutting accessibility state and saying these need to be managed by the components themselves. eg, the input can say if its invalid or not based on validation. WDYT?

okay removed the accessibility states, ptal now

Issue #235, opened 2025-12-07
Define component garbage collection behavior and make it easy to implement


jacobsimionato:
An agent can continually update components on the client, and ideally we would remove unused components, to avoid leaking memory and retaining data that shouldn't be retained.

In v0.8, you could theoretically implement client-side garbage collection, but it would need to be implemented at the specific catalog level and could not be implemented at the framework level. Framework level would be better. The issue is that to garbage collect, you need to understand the graph structure, e.g. the parent-child relationship between components. In the catalog, there was no real structure around what a child property looks like e.g. it's typically just a string field called `child`, but could also be called `imageChild` or `headerContent` etc.

I propose that in v0.9, we define that the client should run garbage collection on the tree after every A2UI Part is processed (which could perform arbitrary atomic operations - see https://github.com/google/A2UI/issues/234). That way, a server can do stuff like atomically restructure the tree to add a new layer without triggering garbage collection. When garbage collection runs, it should remove all components which are not reachable from the root node.

I'm less certain on exactly how to identify orphaned components to garbage collect. I think it should ideally be simple, and not require an entire renderer to be implemented to do this. That way, it's easy for us to do interesting things like have a server-side A2UI accumulator which parses incremental A2UI updates to maintain a mirror of the client state, to use as context etc. Ideally, this could be implemented in a way that is generic to any Catalog.

### Option (A) - child/children name suffix convention
- Require that every single child field ends with the suffix "Child"
- Require every array of children ends with suffix "Children"
- Parser logic in the client scrapes the graph to find child and children IDs to identify orphaned elements

### Option (B) - catalog-specific renderer implementation **[status quo]**

The client renderer marks components as "used" as it renderers. At the end of the render cycles, we can then identify and remove unliked components.

### Option (C) - formal schema annotation -  **[my favourite]** 

In the Catalog definition, add some identifying flag on child and Children properties. A parser can read the Catalog definition schema to figure out what properties are actually child and Children properties, regardless of name, and then parse the tree.

In v0.9, we formally have a [childrenProperty schema](https://github.com/google/A2UI/blob/e14d693b16ccda6362a8254bd3cc1cf156184ee6/specification/0.9/json/common_types.json#L87), so perhaps client logic could use that to find children for multiple child components. We could also add a child property schema. Then the parser could perhaps find references to these schemas in the catalog definition.

jacobsimionato:
@gspencergoog

gspencergoog:
[...]

PR #528, opened 2026-01-22
feat: Support component deletion via component: null

jacobsimionato:
## Description

This PR introduces a mechanism to explicitly delete components from a surface within the `updateComponents` message.

### Changes

- **Schema Update**: The `server_to_client.json` schema for `updateComponents` has been modified. The `components` list now accepts a 'tombstone' object in addition to standard component definitions.
- **Tombstone Format**: A tombstone is defined as an object with the target component's `id` and the `component` property set explicitly to `null`.
- **Documentation**: Updated `a2ui_protocol.md` to document this new behavior.
- **Testing**: Added `component_deletion.json` to the test suite and updated the evaluation framework to support this feature.

### Rationale

Previously, removing a component typically required re-sending the entire parent structure or using `deleteSurface` and recreating it, which is inefficient and can disrupt the user experience. This change allows for granular updates where specific components can be removed while others are added or updated in the same atomic message.

### Sample Code

To delete a component with ID `loading-spinner` and simultaneously add a new `status-text` component:

```json
{
  "updateComponents": {
    "surfaceId": "main",
    "components": [
      {
        "id": "loading-spinner",
        "component": null
      },
      {
        "id": "status-text",
        "component": "Text",
        "text": "Loaded Successfully"
      }
    ]
  }
}
```


jacobsimionato:
This addresses https://github.com/google/A2UI/issues/235

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -173,9 +179,8 @@ This message provides a list of UI components to be added to or updated within a
         "text": "John Doe"
       },
       {
-        "id": "user_title",
-        "component": "Text",
-        "text": "Software Engineer"
+        "id": "old_component_id",
+        "component": null
```
wrenj: why wouldn't it work to just exclude component?

{
  "id": "old_component_id"
}

updateDataModel overwrites instead of merging, and it treats null as remove key, so it should have the same effect, no?

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -148,12 +148,18 @@ This message signals the client to create a new surface and begin rendering it.
 
 ### `updateComponents`
 
-This message provides a list of UI components to be added to or updated within a specific surface. The components are provided as a flat list, and their relationships are defined by ID references in an adjacency list. This message may only be sent to a surface that has already been created. Note that components may reference children or data bindings that do not yet exist; clients should handle this gracefully by rendering placeholders (progressive rendering).
+This message provides a list of UI components to be added, updated, or removed within a specific surface. The components are provided as a flat list, and their relationships are defined by ID references in an adjacency list. This message may only be sent to a surface that has already been created.
```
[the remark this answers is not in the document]
gspencergoog: Yes, we should keep the part describing progressive rendering.

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -185,19 +141,25 @@ This message signals the client to create a new surface and begin rendering it.
...
+To add or update a component, provide the full component definition object. If a component with the same `id` already exists, it will be replaced.
+
+**Deleting Components:**
+To delete a component, provide an object with the component's `id` and set the `component` property to `null`. This removes the component from the client's memory.
 
 **Properties:**
 
 - `surfaceId` (string, required): The unique identifier for the UI surface to be updated. This is typically a name with meaning (e.g. "user_profile_card"), and it has to be unique within the context of the GenUI session.
-- `components` (array, required): A list of component objects. The components are provided as a flat list, and their relationships are defined by ID references in an adjacency list.
+- `components` (array, required): A list of component objects (for updates) or tombstone objects (for deletions).
```
gspencergoog, line 162: What about the adjacency list description?  I think we should keep that.

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -173,9 +179,8 @@ This message provides a list of UI components to be added to or updated within a
         "text": "John Doe"
       },
       {
-        "id": "user_title",
-        "component": "Text",
-        "text": "Software Engineer"
+        "id": "old_component_id",
+        "component": null
       }
```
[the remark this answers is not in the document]
gspencergoog: This is a valid issue don't remove user_title: it's referenced.

gspencergoog:
How do we expect this to be used?  Will we be prompting the LLM to do garbage collection?  Wouldn't it be more effective to just have a message from the LLM to the client that it believes all relevant components are being referenced, and the client is free to do garbage collection? I have doubts that the LLM will be able (or have the information to) keep track of all component references.

specification/v0_9/docs/a2ui_protocol.md
jacobsimionato: Oh yes, let's reinstate that sorry!

specification/v0_9/docs/a2ui_protocol.md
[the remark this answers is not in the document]
jacobsimionato: Sounds good. Can you update the example so that it actions this feedback by reinstating user_title, then adding a second json block that includes messages to update the root component to remove the reference to user_title, and then delete user_title?

specification/v0_9/docs/a2ui_protocol.md
jacobsimionato: Oh that's a great idea! Let's make the "component" property optional, but include description for that property that makes it clear that it should only be omitted to indicate deleting an existing component. I think I should run an eval on that too to make sure that removing the "required" aspect does confuse LLMs too much.

jacobsimionato:

PR #542, opened 2026-01-23
Add React renderer

lukasmoschitz:
# React Renderer for A2UI

## Summary

Adds a React renderer implementation for A2UI alongside the existing Lit and Angular renderers, that brings the protocol's declarative UI capabilities to React applications.

## Approach

- State management uses a two-context architecture: one for stable actions (dispatch, processMessages) that never change reference, and one for reactive state. This prevents unnecessary re-renders when only actions are needed. All components are wrapped with `React.memo()` for additional performance optimization.
- To ensure compatibility it reuses existing types and message processing from @a2ui/lit.
- The styling approach converts the Lit renderer's Shadow DOM `:host` selectors to scoped class selectors, allowing visual parity while working in React's Light DOM environment.

## Components

| Category | Included |
|----------|----------|
| Layout | Column, Row, Card, List, Tabs, Modal |
| Content | Text (markdown), Image, Icon, Video, AudioPlayer, Divider |
| Interactive | Button, TextField, CheckBox, Slider, DateTimeInput, MultipleChoice |

## Core Modules

- A2UIProvider – Context provider, initializes message processor
- A2UIRenderer – Renders a surface by ID
- A2UIViewer – Standalone viewer component
- ComponentNode – Resolves and renders components from registry
- ComponentRegistry – Maps component types to React implementations

## Demo images

restaurant

shopping

purchase

### DIFF
--- a/samples/client/react/shell/src/main.tsx
+++ b/samples/client/react/shell/src/main.tsx
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import React from 'react';
 import ReactDOM from 'react-dom/client';
 import { App } from './App';
--- a/samples/client/react/shell/index.html
+++ b/samples/client/react/shell/index.html
@@ -1,3 +1,19 @@
+<!--
+ Copyright 2026 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
 <!DOCTYPE html>
 <html lang="en">
   <head>
--- a/samples/client/react/shell/src/App.css
+++ b/samples/client/react/shell/src/App.css
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 /*
  * Shell styles - exact port of Lit shell's static styles (app.ts)
  * Transformation: :host → .shell, nested CSS → flat CSS
--- a/samples/client/react/shell/src/App.tsx
+++ b/samples/client/react/shell/src/App.tsx
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import { useState, useEffect, useCallback, useMemo, useRef, FormEvent } from 'react';
 import {
   A2UIProvider,
--- a/samples/client/react/shell/src/client.ts
+++ b/samples/client/react/shell/src/client.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import { Part, SendMessageSuccessResponse, Task } from '@a2a-js/sdk';
 import { A2AClient } from '@a2a-js/sdk/client';
 import type { Types } from '@a2ui/react';
--- a/samples/client/react/shell/src/configs/index.ts
+++ b/samples/client/react/shell/src/configs/index.ts
@@ -1,2 +1,18 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 export * from './types';
 export { config as restaurantConfig } from './restaurant';
--- a/samples/client/react/shell/src/configs/restaurant.ts
+++ b/samples/client/react/shell/src/configs/restaurant.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import { AppConfig } from './types';
 
 export const config: AppConfig = {
--- a/samples/client/react/shell/src/configs/types.ts
+++ b/samples/client/react/shell/src/configs/types.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import type { Theme } from '@a2ui/react';
 
 /**
--- a/samples/client/react/shell/src/mock/index.ts
+++ b/samples/client/react/shell/src/mock/index.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 export {
   createRestaurantListMessages,
   createBookingFormMessages,
--- a/samples/client/react/shell/src/mock/restaurantMessages.ts
+++ b/samples/client/react/shell/src/mock/restaurantMessages.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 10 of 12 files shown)


renderers/react/src/components/content/Text.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const expected = ['h1', 'h2', 'h3', 'h4', 'h5', 'caption', 'body'];
+  return expected.some((v) => v in styles);
+}
+
+/**
+ * Markdown-it instance for rendering markdown text.
+ * Uses synchronous import to ensure availability at first render (matches Lit renderer).
+ *
+ * Configuration matches Lit's markdown directive (uses MarkdownIt defaults):
+ * - html: false (default) - Security: disable raw HTML
```
ditman, line 47: MarkdownIt doesn't perform any sanitization on the HTML output of its `render` method.

This component should use [DOMPurify](https://www.npmjs.com/package/dompurify) or similar on the rendered markdown as it is passed into `dangerouslySetInnerHTML`.

(This ensures that this component doesn't open an XSS attack because of a future (mis)configuration of the MarkdownIt renderer, see [this](https://talk.commonmark.org/t/how-can-i-safely-render-user-submitted-markdown-without-allowing-raw-html/5058/2), for example)

Alternatively, this renderer may be better off using a different markdown renderer, like [react-markdown](https://www.npmjs.com/package/react-markdown)? I understand this might not be desirable for visual parity, however.

We're also looking at injecting the ability to render markdown from the outside, so this might be not be a blocker for this PR :)

renderers/react/src/components/content/Text.tsx
```diff
@@ -0,0 +1,171 @@
...
+    if (!classString) continue;
+
+    // Create regex to match opening tags (handles self-closing and regular)
+    const tagRegex = new RegExp(`<${element}(?=\\s|>|/>)`, 'gi');
+    replacements.push([tagRegex, `<${element} class="${classString}"`]);
+  }
+
+  let result = html;
+  for (const [regex, replacement] of replacements) {
+    result = result.replace(regex, replacement);
```
ditman: If I'm understanding this method correctly, it's looking at the rendered markdown and applying classnames to certain tag names through regexes.

This looks quite brittle and a pain to maintain! There's an example in the `markdown-it` on how to apply CSS classnames to elements through the renderer rules, here:

* https://github.com/markdown-it/markdown-it/blob/master/docs/examples/renderer_rules.md#to-add-a-default-css-class-to-an-element

This is a similar approach to what the Lit renderer is doing here:

* https://github.com/google/A2UI/blob/9c35418439b99dc742ccb3187267655b27f01b9f/renderers/lit/src/0.8/ui/directives/markdown.ts#L103-L117

I'd suggest this step to be done by leveraging the existing markdown-it API, rather than doing string manipulations.

renderers/react/src/styles/index.ts
```diff
@@ -0,0 +1,465 @@
...
+ *
+ * Transformation rules:
+ *   :host          → .a2ui-surface .a2ui-{component}
+ *   section        → .a2ui-surface .a2ui-{component} section
+ *   ::slotted(*)   → .a2ui-surface .a2ui-{component} section > *
+ */
+export const componentSpecificStyles: string = `
+/* =========================================================================
+ * Card (from Lit card.ts static styles)
+ * ========================================================================= */
```
ditman: Why are these comments part of the output string?

ditman (commented):
Thanks for the PR! I left some comments about the handling of markdown in the Text widget on my first quick look to this, but more knowledgeable people should actually dive deeper than that :)

There's some brittleness by grabbing the structural styles directly from the lit renderer, but then maintaining a list of the classnames to be added to each markdown element in this repository; I wonder if some of those might make sense to be moved to a shared package somehow (this is not for this PR, of course)

ditman:
I'm actually working on decoupling the Markdown rendering from the lit and angular renderers here, and I'm planning to do the same to the React renderer once it lands:
https://github.com/google/A2UI/pull/595, still a ways to go.

gspencergoog (commented):
This looks reasonable to me, but I'm not familiar with React, so I'm going to abstain from approving anything here.

renderers/react/src/components/content/Text.tsx
lukasmoschitz: Hi @ditman, thank's for your input!

The current setup has html: false explicitly set on the MarkdownIt instance, which escapes raw HTML in the input before it hits the DOM — so the misconfiguration risk is already guarded at the renderer level.

We'd rather hold off on DOMPurify until the injectable renderer lands, since that design naturally moves sanitization to the caller and we'd just be removing it again shortly after. Happy to track it as a follow-up issue in the meantime!

On react-markdown — agreed it's cleaner, but swapping it out would break visual parity with the Lit renderer and is out of scope for this PR.

renderers/react/src/components/content/Text.tsx
lukasmoschitz: Thank's for the suggestion! We've replaced applyMarkdownTheme with markdown-it's renderer rules API, using token.attrJoin() on _open tokens before rendering and cleaning up afterward.
This should now mirror the Lit renderer's #applyTagClassMap pattern.

renderers/react/src/styles/index.ts
lukasmoschitz: Fair point, they were developer notes explaining the transformation from Lit to React
-> I moved them into the JSDoc above.

renderers/react/src/core/A2UIProvider.tsx
```diff
@@ -0,0 +1,210 @@
...
+/**
+ * @deprecated Use useA2UIContext instead. This alias exists for backward compatibility only.
+ */
+export const useA2UIStore = useA2UIContext;
+
+/**
+ * @deprecated This selector pattern does not provide performance benefits with React Context.
+ * Components will re-render on any context change regardless of what you select.
+ * Use useA2UIContext() or useA2UI() directly instead.
+ *
```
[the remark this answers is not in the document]
lukasmoschitz: No backward compatibility concern - removed both.

renderers/react/src/core/A2UIViewer.tsx
```diff
@@ -0,0 +1,217 @@
...
+  // Generate a stable surface ID based on the definition
+  const baseId = useId();
+  const surfaceId = useMemo(() => {
+    const definitionKey = `${root}-${JSON.stringify(components)}`;
+    let hash = 0;
+    for (let i = 0; i < definitionKey.length; i++) {
+      const char = definitionKey.charCodeAt(i);
+      hash = (hash << 5) - hash + char;
+      hash = hash & hash;
+    }
```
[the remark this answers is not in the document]
lukasmoschitz: Minor cosmetic improvement - removed dead line and applied suggestion

renderers/react/src/components/content/Text.tsx
```diff
@@ -0,0 +1,157 @@
...
+  h4: Record<string, string>;
+  h5: Record<string, string>;
+  body: Record<string, string>;
+  caption: Record<string, string>;
+}
+
+function isHintedStyles(styles: unknown): styles is HintedStyles {
+  if (typeof styles !== 'object' || !styles || Array.isArray(styles)) return false;
+  const expected = ['h1', 'h2', 'h3', 'h4', 'h5', 'caption', 'body'];
+  return expected.some((v) => v in styles);
```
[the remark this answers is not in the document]
lukasmoschitz: The HintedStyles interface has been modified to make its content optional.

samples/client/react/shell/src/main.tsx
```diff
@@ -0,0 +1,17 @@
...
+import ReactDOM from 'react-dom/client';
+import { initializeDefaultCatalog } from '@a2ui/react';
+import { injectStyles } from '@a2ui/react/styles';
+import { App } from './App';
+
+// Initialize the default component catalog
+initializeDefaultCatalog();
+
+// Inject A2UI structural CSS (required for theme utility classes)
+injectStyles();
```
Ezreal09: Hello, I suggest automatically injecting styles into A2UIProvider. Manually injecting styles is awkward and prone to omissions.

renderers/react/src/components/content/Text.tsx
ditman: Sounds good! We can follow later with using the injectable renderer (it just landed and will get manually published shortly!)

renderers/react/visual-parity/lit/src/main.ts
```diff
@@ -0,0 +1,214 @@
...
+import { allFixtures, type FixtureName, type ComponentFixture } from '../../fixtures';
+import { getTheme, themeNames, type ThemeName } from '../../fixtures/themes';
+
+// Themed surface component that provides theme context directly
+// This matches the pattern from @copilotkit/a2ui-renderer's themed-a2ui-surface
+// Key insight: @lit/context doesn't propagate through <slot>, so we must
+// provide the theme directly on a component that renders a2ui-surface as a child
+@customElement('themed-a2ui-surface')
+class ThemedA2UISurface extends LitElement {
+  @provide({ context: UI.Context.themeContext })
```
ditman, line 32: My PR extracting the Markdown renderer to a separate package has landed and will probably break this visual-parity test (for the Text widget (?)).

I think the solution would be to inject the Markdown renderer from `@a2ui/markdown-it` like here:

* I think I've seen some parity tests for the Text widget. My Markdown PR just landed and in that PR I stripped the default Markdown rendering from the Lit renderer.

Functionality can be restored by injecting the @a2ui/markdown-it renderer, like here:

* https://github.com/google/A2UI/blob/72d98949305f3a5ebee4801e176ab67b206b32b9/samples/client/lit/shell/app.ts#L61-L62

@jacobsimionato should be publishing the package to NPM shortly so you can take a standard NPM dependency on it, instead of a `path:` one!

renderers/react/visual-parity/types/pixelmatch.d.ts
```diff
@@ -0,0 +1,22 @@
+declare module 'pixelmatch' {
```
ditman: Why define this here instead of taking a dev dependency on https://www.npmjs.com/package/@types/pixelmatch ?

renderers/react/package.json
```diff
@@ -0,0 +1,96 @@
...
+    "test:watch": "vitest",
+    "typecheck": "tsc --noEmit",
+    "lint": "eslint src",
+    "lint:fix": "eslint src --fix",
+    "format": "prettier --write \"src/**/*.{ts,tsx}\"",
+    "format:check": "prettier --check \"src/**/*.{ts,tsx}\"",
+    "clean": "rm -rf dist"
+  },
+  "dependencies": {
+    "@a2ui/lit": "file:../lit",
```
ditman: I left a remark above about the unbundling of the markdown renderer in the lit renderer.

If I understood correctly, we're only taking a dependency on the lit renderer to reuse styles (for visual parity).

I think in that case, it'd make sense to pin the @a2ui/lit dependency to a very specific version of the package so changes that would considered "minor" in lit (certainly the CSS structure should be an implementation detail) won't break this one.

Also, this package won't be publishable until this is a proper npm dependency, right?

samples/client/react/shell/public/hero-dark.png
ditman, line 1: I also published a slightly different version of this file (removing the dark "borders" around part of the image) to the lit renderer that might break visual parity tests!

.github/CODEOWNERS
```diff
@@ -0,0 +1,35 @@
...
+# Agents
+/a2a_agents/ @nan-yu @dmandar
+
+# Documentation
+/docs/ @zeroasterisk @dmandar @gspencer @jacobsimionato
+
+# Renderers
+/renderers/angular/ @ditman @ava-cassiopeia  @crisbeto
+/renderers/lit/ @ditman @ava-cassiopeia @paullewis
+/renderers/web_core/ @ditman @ava-cassiopeia
```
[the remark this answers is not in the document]
ditman: Good bot!

renderers/react/src/components/content/Text.tsx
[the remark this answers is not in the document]
ava-cassiopeia: LGTM, resolving this comment.

renderers/react/src/core/A2UIProvider.tsx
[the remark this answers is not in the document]
yourkarma6788: THE QUESTION IS, WHO TF CREATED ALL THESE ILLEGAL APIS THAT I DID NOT AUTHORIZE. STOP VIOLATING MY RIGHTS!!!!!!!!!!

.github/CODEOWNERS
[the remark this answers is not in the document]
lukasmoschitz: updated CODEOWNERS to reflect ownership

renderers/react/visual-parity/lit/src/main.ts
lukasmoschitz: Done! I've added `@a2ui/markdown-it` as an NPM dependency and injected `renderMarkdown` via `UI.Context.markdown` on the `ThemedA2UISurface` component in the Lit visual-parity harness.

renderers/react/visual-parity/types/pixelmatch.d.ts
lukasmoschitz: Replaced with `@types/pixelmatch` as a dev dependency and removed the manual type declaration.

renderers/react/package.json
lukasmoschitz: I've actually used this opportunity to replace the `@a2ui/lit` dependency entirely with `@a2ui/web_core: ^0.8.0` (published on NPM) - therefore everything should be publishable now.

samples/client/react/shell/public/hero-dark.png
lukasmoschitz: I was actually wondering about that border when first running the sample demo, glad you fixed it
-> I updated the React samples to use the same image so visual parity is maintained.

samples/client/react/shell/src/main.tsx
lukasmoschitz: Fair point - I've moved style injection and catalog initialization into A2UIProvider so it happens automatically on first render. Both calls are idempotent, so multiple providers are safe. Users now just need:

```
<A2UIProvider>
  <A2UIRenderer surfaceId="main" />
</A2UIProvider>
```

No manual setup required anymore. Thank's for this input.

renderers/react/src/components/content/Icon.tsx
```diff
@@ -1,5 +1,5 @@
 import { memo } from 'react';
-import type { Types } from '@a2ui/lit/0.8';
+import type * as Types from '@a2ui/web_core/types/types';
```
ditman, line 18: These imports are so much better!

renderers/react/visual-parity/lit/src/main.ts
ditman: I changed a little bit the `renderMarkdown` method (made it async), this might break slightly until all the packages are at the right revisions. If using a NPM dependency, we should be fine for now! Once the new web_core and markdown-it packages update, we might need a dependency on markdown-it 0.0.2 which has the async implementation of the `renderMarkdown` function.

Apologies for this change, I realized too late that I need the renderMarkdown to be asynchronous for google's markdown rendering to work properly 🤦

See this:
* https://github.com/google/A2UI/pull/787

(Only dep changes would be needed, nothing on the actual code of the app)

renderers/react/visual-parity/package.json
```diff
@@ -25,6 +26,7 @@
   },
   "devDependencies": {
     "@playwright/test": "^1.58.1",
+    "@types/pixelmatch": "^5.2.6",
```
ditman, line 29: Thanks for taking this dependency!

renderers/react/package.json
```diff
@@ -48,7 +48,7 @@
     "clean": "rm -rf dist"
   },
   "dependencies": {
-    "@a2ui/lit": "file:../lit",
+    "@a2ui/web_core": "^0.8.0",
```
ditman, line 51: There's no dependency on @a2ui/lit anymore? sick!

samples/client/react/shell/public/hero-dark.png
ditman: Haha yeah, at first I thought there was some CSS oddity or artifact in the Lit renderer, then I noticed the image was wrong. I'm missing a dark dropshadow around the store, but it looks good enough. Thanks for taking the updated version!

ava-cassiopeia:
I'm going to take one more pass at this, but I think I'm pretty happy with the state of this, and it looks like pretty idiomatic React to me.

renderers/react/src/components/content/Icon.tsx
ava-cassiopeia: +1

renderers/react/src/styles/README.md
ava-cassiopeia, line 1: I appreciate this doc a lot - thanks for taking the time to put it together!

ava-cassiopeia:
Oh darn, looks like there's a merge conflict on the codeowners now. @lukasmoschitz can you sort that out so I don't have to create another PR on Copilotkit (lol)?

timoslanitis-aisera:
@lukasmoschitz shouldn’t this be published on npm? The README says `npm install @a2ui/react`, but I can’t find the package on npm registry

lukasmoschitz:
> @lukasmoschitz shouldn’t this be published on npm? The README says `npm install @a2ui/react`, but I can’t find the package on npm registry

Thanks @timoslanitis-aisera for the heads up. We still had a few issues to sort out, but the npm package should be available soon.

francox9:
@lukasmoschitz
Thanks for the good work! We have the same question for the availability on npm. Would you be able to have a timeline on when it would become available there?

ava-cassiopeia:
I'm planning on publishing this to NPM today (US time, sometime within the next 8 hours after this comment's timestamp, lol)

ava-cassiopeia:
Published: https://www.npmjs.com/package/@a2ui/react

PR #551, opened 2026-01-26
Refactor FunctionCall to enable full validation

wrenj:
Each function in the `functions` object fully specifies the function, including all validation requirements.

Prior to this change, the `length` function (for example) could have specified an invalid type for the min or max (e.g., "five" instead of "5"), or a negative value for `min` or `max`.

It is now possible to fully validate the parameters and return type of each function. A new set of tests has been created for all the existing functions.

Third parties can add their own functions by following the steps in the a2ui_custom_functions.md document.

# Description

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/specification/v0_9/docs/a2ui_custom_functions.md
+++ b/specification/v0_9/docs/a2ui_custom_functions.md
@@ -12,7 +12,7 @@ This guide demonstrates how to create a `custom_catalog.json` that adds a string
 Create a JSON Schema file (e.g., `custom_catalog.json`) that defines your
 function parameters.
 
-Use the `functions` property to define a map of function schemas, and a group in `$defs` to collect them.
+Use the `functions` property to define a map of function schemas.
 
 ```json
 {
@@ -67,31 +67,33 @@ Use the `functions` property to define a map of function schemas, and a group in
 }
 ```
 
-## 2. Combine with the Standard Schema
+## 2. Make the functions available 
 
-To use this custom catalog in your application, you must override the
-`FunctionCall` definition to include your new function definitions.
-
-Because `common_types.json` defines `FunctionCall` as a choice (`oneOf`), you
-simply create a new list of choices that includes both the standard and custom
-function groups.
-
+The `FunctionCall` definition refers to a [catalog-agnostic reference](a2ui_protocol.md#the-standard-catalog).
+In your catalog, you simply need to define the `anyFunctions` reference:
 ```json
 {
-  // Add to the same file as created above...
-  
-  // Import definitions from the standard schema
   "$defs": {
-    "FunctionCall": {
-      "description": "Invokes a standard OR custom function.",
+    "anyFunction": {
       "oneOf": [
-        // 1. Allow all standard functions
-        {
-          "$ref": "https://a2ui.dev/specification/v0_9/standard_catalog.json#/$defs/anyFunction"
-        },
-        // 2. Allow the custom functions
-        { "$ref": "#/functions/trim" },
-        { "$ref": "#/functions/getScreenResolution" }
+        {"$ref": "#/functions/trim"},
+        {"$ref": "#/functions/getScreenResolution"}
+      ]
+    }
+  }
+}
+```
+
+If you want to incorporate functions defined in the [`standard_catalog.json`],
+those can be added too:
+```json
+{
+  "$defs": {
+    "anyFunction": {
+      "oneOf": [
+        {"$ref": "#/functions/trim"},
+        {"$ref": "#/functions/getScreenResolution"},
+        {"$ref": "catalog.json#/$defs/anyFunction" }
       ]
     }
   }
@@ -100,7 +102,7 @@ function groups.
 
 ## How Validation Works
 
-When a `FunctionCall` is validated against this combined schema:
+When a `FunctionCall` is validated:
 
 1. **Discriminator Lookup:** The validator looks at the `call` property of the
    object.
--- a/specification/v0_9/json/common_types.json
+++ b/specification/v0_9/json/common_types.json
@@ -170,7 +170,7 @@
       },
       "required": ["call"],
       "oneOf": [
-        { "$ref": "standard_catalog.json#/$defs/anyFunction" }
+        { "$ref": "catalog.json#/$defs/anyFunction" }
       ]
     },
     "LogicExpression": {


specification/v0_9/docs/a2ui_custom_functions.md
```diff
@@ -0,0 +1,134 @@
...
+    "FunctionCall": {
+      "description": "Invokes a standard OR custom function.",
+      "oneOf": [
+        // 1. Allow all standard functions
+        {
+          "$ref": "https://a2ui.dev/specification/v0_9/standard_catalog.json#/$defs/StandardFunctions"
+        },
+        // 2. Allow the custom functions
+        {
+          "$ref": "https://example.com/schemas/custom_catalog.json#/$defs/CustomFunctions"
```
wrenj: just inline your functions here, no need for another file

here is an example

https://github.com/google/A2UI/blob/34c4909c81cf7ffe900c738fc6c9e46a4a225769/samples/agent/adk/rizzcharts/rizzcharts_catalog_definition.json

specification/v0_9/docs/a2ui_custom_functions.md
```diff
@@ -0,0 +1,134 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://example.com/schemas/my_app_schema.json",
+  // Import definitions from the standard schema
+  "$defs": {
+    "FunctionCall": {
+      "description": "Invokes a standard OR custom function.",
+      "oneOf": [
+        // 1. Allow all standard functions
+        {
+          "$ref": "https://a2ui.dev/specification/v0_9/standard_catalog.json#/$defs/StandardFunctions"
```
wrenj: its just functions in the standard catalog

specification/v0_9/json/standard_catalog.json
```diff
@@ -933,6 +1027,22 @@
       "discriminator": {
         "propertyName": "component"
       }
+    },
+    "StandardFunctions": {
```
wrenj: can we just call these functions

I'm trying to get away from standard SDK vs customizations thinking, since the standard componnts are just a starter set and we want to encourage people to override them or do their own thing, lmk if that makes sense

specification/v0_9/docs/a2ui_custom_functions.md
```diff
@@ -0,0 +1,134 @@
...
+        },
+        // 2. Allow the custom functions
+        {
+          "$ref": "https://example.com/schemas/custom_catalog.json#/$defs/CustomFunctions"
+        }
+
+        // 3. (Optional) Allow fallback for other functions if strict validation isn't desired
```
wrenj: lets get rid of this

the idea with a2ui is everything is strongly typed through json schema and the custom catalog, so we dont need a generic function, since you just extend the standard catalog with your custom functionality

specification/v0_9/json/standard_catalog.json
```diff
@@ -648,241 +648,335 @@
       "unevaluatedProperties": false
     }
   },
-  "functions": [
-    {
-      "name": "required",
+  "functions": {
```
wrenj, line 651: can't we just put these directly in functioncall now?

I'm a little foncused it seems like we have
Common->FunctionCall
Standard Catalog -> functions
ApplicationCatalog -> FunctionCall redefinition

Can we just do
StandardCatalog->FunctionCall

specification/v0_9/test/cases/function_catalog_validation.json
```diff
@@ -0,0 +1,702 @@
...
+      "description": "formatDate: Invalid pattern type (null)",
+      "valid": false,
+      "data": {
+        "updateComponents": {
+          "surfaceId": "test",
+          "components": [
+            {
+              "id": "txt1",
+              "component": "Text",
+              "text": {
```
wrenj, line 647: do these work on standard catalog or just the application specific catalog?

specification/v0_9/json/standard_catalog.json
```diff
@@ -648,241 +648,335 @@
...
-          { "$ref": "common_types.json#/$defs/DynamicValue" },
-          {
-            "type": "object",
-            "properties": {
-              "min": {
-                "type": "integer",
-                "minimum": 0,
-                "description": "The minimum allowed length."
+      "properties": {
+        "call": { "const": "length" },
```
wrenj, line 694: For consistency can we model these the exact same way as components? with the function name as the key instead of using "call"? Maybe dont need to do now, greg might have looked at this approach earlier not sure why he didn't use the same pattern.

 "components": {
    "Text": { ... }
    "Icon": { ... }
}

specification/v0_9/docs/a2ui_custom_functions.md
jgindin: `$defs/StandardFunctions` provides some abstraction. If someone (for whatever reason) wants to only include a subset of the functions referenced there, they could do something like:
```
"oneOf": [
  { "$ref": "https://a2ui.dev/specification/v0_9/standard_catalog.json#/functions/required" }, 
  { "$ref": "https://a2ui.dev/specification/v0_9/standard_catalog.json#/functions/regex" }, 
  ...,
  { "$ref": "#/functions/trim" },
  { "$ref": "#/functions/getScreenResolution" } 
]
```

@wrenj -- I could add this as an alternative in the document if that would make sense?

specification/v0_9/json/standard_catalog.json
jgindin: The issue with putting them inline in `FunctionCall` is that we would lose the extensibility. E.g., right now, we define `FunctionCall` and we tell it that these are the options (via the `"oneOf": [ { "$ref": "standard_catalog.json#/$defs/StandardFunctions" }]`. So a 3rd party can redefine `FunctionCall` and add more functions (as described in the `a2ui_custom_functions.md` file).

specification/v0_9/json/standard_catalog.json
jgindin: I guess it's hard to see in this diff, but that is how the functions are modeled:

```
"functions": {
    "regex": {
        "type": "object",
        "description": "Checks that the value matches a regular expression string.",
        . . . 
    },
}
```

specification/v0_9/test/cases/function_catalog_validation.json
jgindin: These tests work against the functions defined in the standard catalog. 

If a 3rd party wanted to add their own functions, and write their own tests, they would have to make some changes to the `run_tests.py`, but it should all work.

specification/v0_9/json/standard_catalog.json
```diff
@@ -648,241 +648,335 @@
...
+                  "description": "The default/fallback string (used for general plural cases)."
+                }
+              },
+              "required": [
+                "other"
+              ],
+              "additionalProperties": false
+            }
+          ]
+        }
```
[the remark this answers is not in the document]
jgindin: done

specification/v0_9/json/common_types.json
```diff
@@ -168,7 +168,10 @@
           "default": "boolean"
         }
       },
-      "required": ["call"]
+      "required": ["call"],
+      "oneOf": [
```
gspencergoog, line 172: If we just leave this `oneOf` out of `common_types.json`, then step 2 in the documentation is entirely unnecessary.  Couldn't you instead add a required block at the end of the list of functions in the catalog that requires at least one of the the functions, so you don't need the separate `Functions` definition?  I'd like to find a way where the catalog contains all of the validation needed, and we don't need to customize the common types for each client.

specification/v0_9/json/standard_catalog.json
jacobsimionato: Our approach to custom catalog requires that all our generic schemas are not specific to any catalog, and they refer to the specifics of the current catalog via some placeholder name. E.g. `catalog.json#/$defs/anyComponent` or maybe `catalog.json#/$defs/anyFunction`. I realised I think we made a mistake here currently because `server_to_client.json` currently refers to `standard_catalog.json#/$defs/anyComponent` which isn't quite right I think. Sending https://github.com/google/A2UI/pull/563 to propose fixing this.

By using `anyComponent`, we get a bit of JSON-level validation because we force the name to be valid, and the set of params to be valid.

I don't think we get the same value in this PR, because functions are intended to be more flexible in how they are instantiated, e.g. all the parameters can be dynamic, they're positional parameters instead of named parameters etc, so there just isn't as much validation that is possible. So given v0.9 is already agreed upon, I'd prefer we keep things as-is.

But if we do go in this direction, then I'd love to try to make functions work as similar to components, as James suggested. So perhaps this might involve jumping to named parameters etc.

PR #556, opened 2026-01-27
docs: Fix discrepancies between v0.8 spec/docs and schemas

### LINKED ISSUE #495: v0.8 reference contradicts json schemas
1. Standard catalog id
In schema: `a2ui.org:standard_catalog_0_8_0`
In protocol spec: `https://github.com/google/A2UI/blob/main/specification/v0_8/json/standard_catalog_definition.json`
2. Component name difference
In `v0_8/json/standard_catalog_definition.json`: `CheckBox`
In `docs/reference/components.md`: `Checkbox`
3. `docs/reference/components.md` uses Material icon names like `check_circle`, but v0.8 catalog `Icon.name` enum uses `check`, `error`, etc.

jacobsimionato:
This PR resolves several discrepancies between the A2UI v0.8 specification/documentation and the JSON schemas, as identified in #495.

Changes:
1.  **Standard Catalog ID:** Updated `specification/v0_8/docs/a2ui_protocol.md` to use the correct ID `a2ui.org:standard_catalog_0_8_0`, aligning with the schema.
2.  **Component Name:** Updated `docs/reference/components.md` to use `CheckBox` (CamelCase) in the JSON example, matching the schema definition.
3.  **Icon Names:** Updated the icon example in `docs/reference/components.md` to use `check` instead of `check_circle`. This adopts the conservative approach of using an icon that is strictly allowed by the current v0.8 schema enum.

Fixes #495

### DIFF
--- a/specification/v0_8/docs/a2ui_protocol.md
+++ b/specification/v0_8/docs/a2ui_protocol.md
@@ -272,7 +272,7 @@ In order to do the substitution, based on the standard `server_to_client_schema`
 
 ```py
 component_properties = custom_catalog_definition["components"]
-style_properties = custom_catalog_definition["components"]
+style_properties = custom_catalog_definition["styles"]
 resolved_schema = copy.deepcopy(server_to_client_schema)
 
 resolved_schema["properties"]["surfaceUpdate"]["properties"]["components"]["items"]["properties"]["component"]["properties"] = component_properties


specification/v0_8/docs/a2ui_protocol.md
gspencergoog, line 195: The bot is right.  This should be updated too.

jacobsimionato: Good catch. I realised there are more examples of the github version, so I standardized on that instead.

PR #568, opened 2026-01-28
fix(lit, angular): correct DateTimeInput month indexing [Issue #566]

### LINKED ISSUE #566: Bug: DateTimeInput generates invalid HTML date values (0-indexed month)
The [DateTimeInput](file:///c:/Users/USER/DEVELOPMENT_SPACE/ImportantProjects/A2UI/renderers/lit/src/0.8/ui/datetime-input.ts#26-199) component in both Lit and Angular renderers incorrectly formats the date string for the underlying `<input type="date">` element, using 0-indexed months (00-11) instead of the HTML-required 1-indexed format (01-12).

## Bug Location

-   **Lit**: [renderers/lit/src/0.8/ui/datetime-input.ts](renderers/lit/src/0.8/ui/datetime-input.ts)
-   **Angular**: [renderers/angular/src/lib/catalog/datetime-input.ts](renderers/angular/src/lib/catalog/datetime-input.ts)

## Bug Details 

When a date is passed to the component (e.g., January 15, 2025):

1.  The code calls `date.getMonth()`, which returns `0` for January.
2.  It constructs the string: `2025-00-15`.
3.  It sets this as the `value` attribute of the input: `<input type="date" value="2025-00-15">`.
4.  **Result**: Browsers reject `00` as an invalid month and render an empty input.

```typescript
// Current Code
const month = this.#padNumber(date.getMonth()); // Returns "00" for January
```

## Impact

-   **Functionality**: Users cannot see pre-filled dates in January (and potentially other months depending on browser strictness).
-   **Cross-Platform**: Affects both Lit and Angular implementations equally.

shantoislamdev:
## Summary

Fix `DateTimeInput` component in both Lit and Angular renderers to use 1-indexed months (`01-12`) instead of 0-indexed months (`00-11`) for the HTML date input value.
Fixes #566

## Problem

The `DateTimeInput` component incorrectly formatted the `value` attribute for `<input type="date">` by using `date.getMonth()` directly (which is 0-indexed).

**Impact:**
- Browsers reject partial or invalid dates like `2025-00-01`.
- The input field appears empty instead of showing the pre-filled date.
- Affects both Lit and Angular implementations.

## Solution

Updated the date formatting logic in both renderers:
1.   **Lit**: `renderers/lit/src/0.8/ui/datetime-input.ts` - Line: `137`-`138`
2.  **Angular**: `renderers/angular/src/lib/catalog/datetime-input.ts` - Line `97`-`98`

**Change:**
```typescript
- const month = this.#padNumber(date.getMonth());
+ const month = this.#padNumber(date.getMonth() + 1);
```

## Testing

- [x] All test passed ✅

## Checklist

- [x] I have signed the [Google CLA](https://cla.developers.google.com/)
- [x] My code follows the project's style guidelines
- [x] I have tested my changes locally


selamw1:
While this PR fixes the primary date formatting bug, it doesn’t yet address the parsing of time-only values. JavaScript's `Date` constructor is unreliable when parsing '`HH:MM`' strings (often resulting in `Invalid Date`).

To resolve this, we should add an early exit in `formatInputValue`. If the `inputType` is 'time' and the value is already a valid time string, we should return it directly to bypass the `Date` object logic.

**Suggested Changes (to be applied to both Lit and Angular components):**

```TypeScript
// If it's a time-only input, return the value directly to avoid unreliable Date parsing
if (this.getInputType() === 'time' && /^\d{2}:\d{2}(:\d{2})?$/.test(value)) {
  return value;
}

const date = value ? new Date(value) : null;
// ... rest of the logic
```

What are your thoughts on this approach?

zeroasterisk:
Thanks for this fix, @shantoislamdev! Fixing zero-based month indexing is important for date handling. 🙏

This bug was already resolved in [#477](https://github.com/google/A2UI/pull/477) (merged Feb 12, 2026) for both Lit and Angular renderers. Closing as superseded.

PR #572, opened 2026-01-28
Fix schema URLs, publish schema on a2ui.org

### LINKED ISSUE #232: Upload v0.9 schemas to URLs at a2ui.org that they reference

gspencergoog:
# Description

This fixes a bunch of places where we use "a2ui.dev" and actually mean a2ui.org. Also, updates all of the catalog IDs to be consistent and point to a2ui.org URLs.

In addition, it redirects the correct URLs on a2ui.org to the specification schema files on github.

Fixes #232


mkdocs.yaml
```diff
@@ -183,6 +183,17 @@ plugins:
   #     sections:
   #       "Specification":
   #         - a2ui_protocol.md
-  # - redirects:
-  #     redirect_maps:
-  #       "index.md": "a2ui_protocol.md"
+  - redirects:
```
jacobsimionato, line 186: Oh very neat!!

jacobsimionato (approved):
Fantastic!

PR #573, opened 2026-01-28
Revert #515 and modify stringFormat to take named args.

gspencergoog:
# Description

This reverts #515 and instead of making the args be positional, have them be named, modifying the stringFormat format string so that its function call syntax also includes names.  This is to open the door to better function call validation.

I also fixed some failures in the evals due to the catalog aliasing change.

cc @jgindin


specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -571,9 +569,9 @@ _Replace the entire data model:_
...
 
 - **Targeted Delivery**: The data model is sent exclusively to the server that created the surface. Data cannot leak to other agents or servers.
-- **Trigger**: Data is sent only when a client-to-server message is triggered (e.g., by a user action like a button click). Passive data changes (like typing in a text field) do not trigger a network request on their own; they simply update the local state, which will be sent with the next action.
-- **Payload**: The data model is included in the transport metadata, tagged by its `surfaceId`.
-- **Convergence**: The server treats the received data model as the current state of the client at the time of the action.
+- **Trigger:** Data is sent only when an A2A message is triggered (e.g., by a user action like a button click). Passive data changes (like typing in a text field) do not trigger a network request on their own; they simply update the local state, which will be sent with the next action.
```
jacobsimionato: Switch back to 'client-to-server' message, so that the language here is agnostic of transport.

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -589,15 +587,15 @@ Input components (like `TextField`, `CheckBox`) can define a list of checks. Eac
...
     "message": "Zip code is required"
   },
   {
     "call": "regex",
-    "args": [
-      { "path": "/formData/zip" },
-      "^[0-9]{5}$"
-    ],
+    "args": {
+      "value": { "path": "/formData/zip" },
```
jacobsimionato, line 596: I already like this more!

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -715,9 +717,9 @@ Values from the data model can be interpolated using their JSON Pointer path.
   "component": "Text",
   "text": {
     "call": "formatString",
-    "args": [
-      "Hello, ${/user/firstName}! Welcome back to ${/appName}."
-    ]
+    "args": {
+      "value": "Hello, ${/user/firstName}! Welcome back to ${/appName}."
```
jacobsimionato, line 721: Ahhhh this is actually fine. I forgot that we're not doing explicit parameter plumbing here anyway, so JSON validation of the arguments was already not possible.

specification/v0_9/docs/evolution_guide.md
```diff
@@ -228,7 +228,7 @@ Specifying an unknown surfaceId will cause an error. It is recommended that clie
 
 - **String Formatting**: Introduced the `formatString` function, which supports `${expression}` syntax for interpolation.
 - **Unified Expression Language**: Allows embedding JSON Pointer paths (absolute and relative) and client-side function calls directly within the format string.
-- **Nesting**: Supports recursive nesting of expressions (e.g., `${formatDate(${/timestamp}, 'yyyy-MM-dd')}`).
+- **Nesting**: Supports recursive nesting of expressions (e.g., `${formatDate(value:${/timestamp}, format:'yyyy-MM-dd')}`).
```
jacobsimionato: Syntax nit: How about a space after the colon, for readability? E.g. `${formatDate(value: ${/timestamp}, format: 'yyyy-MM-dd')}`)

specification/v0_9/eval/src/index.ts
jacobsimionato, line 1: Sorry that I missed this before!

specification/v0_9/json/common_types.json
```diff
@@ -116,24 +146,40 @@
...
-        { "$ref": "#/$defs/DataBinding" },
-        { "$ref": "#/$defs/LogicExpression" }
+        {
+          "type": "boolean"
+        },
+        {
+          "$ref": "#/$defs/DataBinding"
+        },
+        {
+          "$ref": "#/$defs/LogicExpression"
```
jacobsimionato, line 156: An idea: Can we remove the concept of LogicExpression and just put And/Or/Not in the standard catalog  as functions with two parameters? I don't feel strongly and I don't want to add more churn, but just an idea to try to reduce the number of concepts and simplify the renderers a little.

jacobsimionato (approved):

specification/v0_9/docs/evolution_guide.md
gspencergoog: Sure.  The actual parser should be permissive and allow spaces before and after the colon, though.

specification/v0_9/json/common_types.json
gspencergoog: We definitely could. Let's make some examples and see how the evals do on functions vs built-in, but let's do that in another PR.

specification/v0_9/docs/a2ui_protocol.md
gspencergoog: Good point.  Done.

specification/v0_9/json/standard_catalog.json
```diff
@@ -654,211 +654,215 @@
...
-      "description": "Performs string interpolation of data model values and other functions in the catalog functions list and returns the resulting string. The value string can contain interpolated expressions in the `${expression}` format. Supported expression types include: JSON Pointer paths to the data model (e.g., `${/absolute/path}` or `${relative/path}`), and client-side function calls (e.g., `${now()}`). Function arguments must be literals (quoted strings, numbers, booleans) or nested expressions (e.g., `${formatDate(${/currentDate}, 'MM-dd')}`). To include a literal `${` sequence, escape it as `\\${`.",
+      "description": "Performs string interpolation of data model values and other functions in the catalog functions list and returns the resulting string. The value string can contain interpolated expressions in the `${expression}` format. Supported expression types include: JSON Pointer paths to the data model (e.g., `${/absolute/path}` or `${relative/path}`), and client-side function calls (e.g., `${now()}`). Function arguments must be named (e.g., `${formatDate(value:${/currentDate}, format:'MM-dd')}`). To include a literal `${` sequence, escape it as `\\${`.",
```
wrenj, line 745: maybe clarify that you are using Dart syntax here

${} and func(arg_name:value) I think are both dart. It might be easier for the LLM to reason about, versus the LLM thinking it has a one shot example and a made up language. 

I do want to flag again (sorry to beat a dead horse here) that the LLM will likely perform best using MessageFormat which it probably trained on millions of examples of (particularly user interface code examples). I know MessageFormat adds an import, and it's a verbose syntax, but we would do it once and be done and not expect clients to build their own string formatters (eg, we provide no % formatter).  I think it will also help steer A2UI to not feeling google specific by using an open standard instead of basing interpolation on Dart.

@gspencergoog @jacobsimionato

PR #576, opened 2026-01-29
Arr

wrenj:


specification/v0_9/docs/a2ui_extension_specification.md
```diff
@@ -69,28 +69,45 @@ To identify a `DataPart` as containing A2UI data, it must have the following met
 
 - `mimeType`: `application/json+a2ui`
 
-The `data` field of the `DataPart` contains a **single** A2UI JSON message (e.g., `createSurface`, `updateComponents`, `action`). It MUST NOT be an array of messages.
+The `data` field of the `DataPart` contains a **list** of A2UI JSON messages (e.g., `createSurface`, `updateComponents`, `action`). It MUST be an array of messages.
 
-### Atomicity and multiple messages
+### Processing Rules
```
jacobsimionato, line 74: This document is really explaining how A2UI acts as an extension to A2A. But I think this content should actually be at the top level, and apply to all transports.

How about:
- In this doc, just mention that `data` contains a list of messages as defined in the Server-to-Client Message List Schema.
- Add some section to the main v0.9 specification doc like "Transporting and processing A2UI messages" which also explains the use of the list structure, and includes these notes on atomicity and rendering and error handling. 
- Rename this doc to a2a_extension_specification.md (optional)

Sorry I now I suggested to edit this specific doc, but thinking about it a bit more, I think a restructure is helpful to make the spec more transport agnostic. I guess the restructure is orthogonal so you can land this and I can restructure later if you like.

jacobsimionato (approved):

PR #584, opened 2026-01-30
Named parameter function validation

jgindin:
# Description

Ensure all functions (including `args` and `returnValue`) are properly verified. `FunctionCall` uses a `oneOf` and a reference to the "catalog agnostic" `anyFunction`, allowing customization of the functions available in the user's catalog.

`CheckRule` now uses a separate `condition` , rather than putting the conditional and the error message together in a single `allOf`.

Removed the `LogicalExpression` in favor of explicit `and`, `or`, and `not` functions.

The functions declared in `standard_catalog` are in a form which is more consistent with the components:
* Declared in dictionary form
* There is an `anyFunction` definition for easy reference

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md


specification/v0_9/json/common_types.json
```diff
@@ -371,4 +312,4 @@
       ]
     }
   }
-}
+}
```
gspencergoog: nit: trailing newline missing.

specification/v0_9/json/standard_catalog.json
```diff
@@ -648,245 +648,442 @@
...
+        "returnType": {
+          "const": "void"
+        }
+      },
+      "required": ["call", "args"],
+      "unevaluatedProperties": false
+    },
+    "now": {
+      "type": "object",
+      "description": "Returns the current client timestamp in ISO 8601 format.",
```
gspencergoog: This should be clear about whether it is in the local time zone or in UTC.  Typically, I'd prefer things to always be in UTC, but this seems like it wants to be in the local time zone.  This does have some privacy implications, since now we're leaking what time zone the user is in.

specification/v0_9/docs/a2ui_custom_functions.md
```diff
@@ -0,0 +1,120 @@
+# Extending A2UI with Custom Functions
+
+A2UI functions are designed to be extensible. Third-party developers can define
+their own function catalogs while preserving strict validation for the standard
+set.
```
jacobsimionato: It's been a point of confusion for the A2UI project whether catalogs are:
- A) Additive (e.g. you define a custom catalog and it implicitly assumes the standard catalog is there and can reference it)
- B) Standalone (e.g. if you want to add or change the standard catalog, you must fork it and add your changes).

We've chosen B) as the model, and I'm eager to avoid confusion around this. So for this example, I'm wondering if you can instead frame this as implementing a separate catalog with its own components and functions, or else a fork of the custom catalog (with the common content ommited).

I get that making some small additions to the standard catalog is likely a common developer workflow that we need to support, but we can do that with tooling or SDK functionality that merges together two JSON schemas etc, rather than at the protocol level.

specification/v0_9/docs/a2ui_custom_functions.md
```diff
@@ -0,0 +1,120 @@
...
+If you want to incorporate functions defined in the [`standard_catalog.json`],
+those can be added too:
+```json
+{
+  "$defs": {
+    "anyFunction": {
+      "oneOf": [
+        {"$ref": "#/functions/trim"},
+        {"$ref": "#/functions/getScreenResolution"},
+        {"$ref": "catalog.json#/$defs/anyFunction" }
```
jacobsimionato: `catalog.json` is defined to refer to the *current* catalog rather than the *standard catalog*.

I think it makes sense to just show a list of functions here which are assumed to be defined in this document. As described above, we can include SDK logic to merge catalogs if necessary, in a way that special cases the `anyFunction` and `anyComponent` definitions.

specification/v0_9/test/cases/button_checks.json
```diff
@@ -16,27 +16,37 @@
...
                         {
                           "call": "required",
-                          "args": { "value": { "path": "/formData/email" } },
+                          "args": { "value": { "path": "/formData/terms" } },
                           "returnType": "boolean"
                         },
                         {
-                          "call": "required",
-                          "args": { "value": { "path": "/formData/phone" } },
+                          "call": "or",
```
jacobsimionato, line 29: This is definitely more verbose than the "LogicExpression" based syntax, but I think I still prefer it because at least there are less concepts to reason about. WDYT @gspencergoog ?

specification/v0_9/docs/a2ui_custom_functions.md
gspencergoog: It's currently the case that you can define multiple catalogs at the same time (i.e. `supportedCatalogIds` is an array).  So, a client can support the standard catalog, and also support their own custom catalog.  This makes them effectively additive.  It's true that each catalog is standalone, and they don't refer to each other, but you can add additional catalogs. My assumption was that the aliasing done in the agent (to alias `catalog.json`) would merge these catalogs into the one that it aliased.

I suppose we should make it clear when happens when an additional catalog defines its own version of a component in another catalog (the standard catalog or otherwise). It can either be not allowed, or the order in the array could describe which one wins: I'd opt for the second one, since that's more powerful.

specification/v0_9/test/cases/button_checks.json
gspencergoog: Yes, I think we should go with this for now, since it simplifies the number of special syntax mechanisms to think about.  If we want to add the logical expression stuff back in for 0.10, we can, without breaking anything.

specification/v0_9/json/standard_catalog.json
jgindin: I'd be happy to just remove this. I added it because `specification/v0_9/test/cases/contact_form_example.jsonl` refers to it. I can change that to reference another function.

jgindin: `now` has been removed.

specification/v0_9/docs/a2ui_custom_functions.md
jgindin: I've changed this to remove the reference to the "standard set" (which, I see now, is just adding more confusion by introducing a new term).

specification/v0_9/docs/a2ui_custom_functions.md
jgindin: The example above (on line 78) is, I think, what you're asking for. 

I've fixed the typo -- `catalog.json` should have been `standard_catalog.json` (which matches the description of this example on line 90).

But if you prefer that I just remove this example, that's fine too.

PR #593, opened 2026-02-03
Create a fork of the 0.9 spec as a starting point for 0.10

gspencergoog:
# Description

This forks the 0.9 spec, renaming everything to 0.10, preparing it for updates in the next version.  I'm crossing my fingers that the next version be relatively similar to the 0.9 version, so this should be a good starting point, and gives us a place to make changes without affecting the 0.9 spec, since that has been ratified.


specification/v0_8/README.md
```diff
@@ -0,0 +1,13 @@
+# Specification v0.8.1
+
+This directory contains the specification for version 0.8.1 of A2UI, a previous version of the specification, and is no longer under active development. The currently published specification is [v0.9](../v0_9).
```
jacobsimionato: Optional: Maybe remove "The currently published specification is..." so that we (read: Gemini) don't have to remember to update N specs later? Just for old specs. I like that v0.9 marks itself as the currently published spec.

jacobsimionato (approved):

specification/v0_8/README.md
gspencergoog: Sure, makes sense.  I kept the reference to v0.10 in the v0.9 README, on the theory that we'll have to update them both anyhow when v0.10 is published.

Issue #475, opened 2026-01-12
Figure out how to resolve Markdown dependency in OSS and google repository


jacobsimionato:
Currently, we use `markdown-it`, but this isn't available in the Google repo.

Some possible ideas:
- Somehow allow the google repo to use a different dependency than the Github repo
- Modularize the system so that google internal apps can share the core low-dependency code, but fork other parts of it. For example, if we clearly split the renderer framework (e.g. A2UImessageProcessor) etc from the catalog implementation, then in the Google repo we can fork the Catalog implementation and use our internal markdown renderer, while Github can use its one. This also has the advantage that Google teams can rapidly iterate on the specific visual styling for each element without modifying open source code.

ditman:
I'd claim this issue for myself but don't have permissions in the repo!

If you want to merge a fix before I do, don't feel bad (but cc me in the PR, please!)

PR #595, opened 2026-02-04
[RFC] Split markdown rendering to a separate package.

ditman:
> [!IMPORTANT]
> This PR is too big to land in one go, but luckily we can land this in stages quite easily by landing the `markdown-it-*` packages first, and then the modifications to the renderers and samples. That also gives us some time to figure out how to publish the `markdown-it-*` packages automatically without the pressure of them being "depended on", or needed within G3.

# Description

This PR separates the `markdown-it` dependency from the `lit` renderer, by allowing users to `@provide` a markdown renderer of their own.

**BREAKING CHANGE**: By default, now incoming markdown is rendered as a `pre` element, and users *must* inject a markdown renderer of their choosing for markdown to be rendered (see below).

In order to keep the "batteries" somewhat included in the sdk, a couple of new packages are introduced:

* `@a2ui/markdown-it-shared`: the pre-configured `markdown-it` instance for all web renderers. This allows us to have a single configured markdown renderer with `markdown-it` and `dompurify` (TBD) that can be reused across all packages. This is pretty much a markdown string -> html string converter, but if we want to add plugins to `markdown-it` later, all packages get the new output at once.
* `@a2ui/markdown-it-lit`: the bridge package between the `lit` renderer and `markdown-it-shared`. This is what people will use when using the `lit` renderer to `@provide` the `markdown` renderer. They can also use this package as a jumping off point to create their own markdown renderer, if they prefer anything else other than `markdown-it`.
* `@a2ui/markdown-it-angular`: the bridge package between the `angular` renderer and `markdown-it-shared`. This package provides a `MarkdownItMarkdownRenderer` class that can `provide: MarkdownRenderer`. 

The `lit` samples, and the `angular` restaurant sample are updated to inject the new markdown renderer.

## Fixes

* Fixes https://github.com/google/A2UI/issues/475

## Tests

* Added some unit tests to all new packages.

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/.github/workflows/ng_build_and_test.yml
+++ b/.github/workflows/ng_build_and_test.yml
@@ -51,13 +51,13 @@ jobs:
         working-directory: ./renderers/lit
         run: npm run build
 
-      - name: Build Angular renderer and its dependencies
-        working-directory: ./samples/client/angular
-        run: npm run build:renderer
-
       - name: Install top-level deps
         working-directory: ./samples/client/angular
         run: npm i
+
+      - name: Build Angular renderer and its dependencies
+        working-directory: ./samples/client/angular
+        run: npm run build:renderer
 
       - name: Build contact sample
         working-directory: ./samples/client/angular


renderers/lit/src/0.8/ui/directives/noop_markdown.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * "Handles" Markdown rendering by doing nothing.
+ *
+ * Configure @a2ui/lit-markdown, or your custom Markdown renderer
+ * to actually parse and render Markdown in your app.
+ */
+class NoopMarkdownRendererDirective extends Directive {
+  render(markdown: string, _tagClassMap?: Record<string, string[]>) : TemplateResult {
+    return html`<pre>${markdown}</pre>`;
```
ditman: This should probably `console.warn` once to let users know that they *need* to use `@a2ui/markdown-it-lit`, and maybe link to some documentation site.

renderers/lit/src/0.8/ui/context/context.ts
```diff
@@ -14,11 +14,11 @@
  limitations under the License.
  */
 
-import { googleAI } from "@genkit-ai/google-genai";
-import { configure } from "genkit";
+import {markdownContext} from "./markdown.js";
```
ditman, line 17:
```suggestion
import { markdownContext } from "./markdown.js";
```

? I need to setup some form of "format on save" or something I guess?

renderers/lit/src/0.8/ui/directives/directives.ts
ditman, line 1: This file is now probably redundant.

renderers/lit/src/0.8/ui/text.ts
```diff
@@ -44,6 +46,9 @@ export class Text extends Root {
   @property({ reflect: true, attribute: "usage-hint" })
   accessor usageHint: Types.ResolvedText["usageHint"] | null = null;
 
+  @consume({context: Context.markdown})
+  accessor markdownRenderer = noopMarkdown;
```
ditman: Not to toot my own horn, but this ended up looking quite nice. I like Lit!

renderers/lit/package-lock.json
ditman, line 1: Can we stop comitting the lockfiles to the repo? They add a lot of noise and can cause unnecessary merge trouble! Any advantage of checking this in??

renderers/markdown/markdown-it-lit/src/markdown.ts
ditman, line 1: (This is the `Directive` that existed before, moved to its own separate package)

renderers/markdown/markdown-it-lit/.npmrc
ditman, line 1: (Is this needed?)

renderers/markdown/markdown-it-lit/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "@types/node": "^24.10.1",
+    "typescript": "^5.8.3",
+    "wireit": "^0.15.0-pre.2"
+  },
+  "dependencies": {
+    "@lit/context": "^1.1.4",
+    "lit": "^3.3.1",
+    "@a2ui/lit": "file:../../lit",
+    "@a2ui/markdown-it-shared": "file:../markdown-it-shared",
+    "@a2ui/web_core": "file:../../web_core"
```
ditman: TODO: Review all the dependencies in this package.

renderers/markdown/markdown-it-lit/prepare-publish.mjs
```diff
@@ -0,0 +1,73 @@
+import { readFileSync, writeFileSync, copyFileSync, mkdirSync, existsSync } from 'fs';
+import { join } from 'path';
+
+// TODO: Review this script, it's copied wholesale from the lit package.
```
ditman, line 4: **This can't land as is**

Also this file is popping up in a bunch of packages, can it be converted into a reusable tool maybe?

renderers/markdown/markdown-it-shared/src/markdown.ts
```diff
@@ -0,0 +1,38 @@
...
+ * A Markdown renderer using markdown-it and dompurify.
+ */
+export const markdownRenderer = {
+  /**
+   * Renders markdown to HTML.
+   * @param value The markdown code to render.
+   * @param tagClassMap A map of tag names to classes.
+   * @returns The rendered HTML as a string.
+   */
+  render: (value: string, tagClassMap?: TagClassMap) => {
```
ditman: I think `TagClassMap` is very limiting and may introduce issues down the line if we add more features to the `markdown` renderer that need configuration... Imagine something like `LaTeX` rendering or similar.

I think the second parameter of the `render` method should be something like:

```
MarkdownOptions {
  tagClassMap?: TagClassMap,
  ... // other
}
```

So the signature of render would end up being:

```
render: (value: string, options?: MarkdownRendererOptions)...
```

That should make it much easier to add new options to the markdown rendering, in a backwards-compatible way! Also people may be able to `extend MarkdownRendererOptions` and create their own types for their non-markdown-it based renderers, if they need to.

renderers/markdown/markdown-it-shared/src/sanitizer.ts
```diff
@@ -14,5 +14,11 @@
  limitations under the License.
  */
 
-import "./generation_flow";
-import "./evaluation_flow";
+// TODO: Implement the DOMPurify sanitizer.
```
ditman, line 17: This can't land without the `dompurify` integration.

renderers/markdown/markdown-it-shared/prepare-publish.mjs
```diff
@@ -0,0 +1,73 @@
+import { readFileSync, writeFileSync, copyFileSync, mkdirSync, existsSync } from 'fs';
+import { join } from 'path';
+
+// TODO: Review this script, it's copied wholesale from the lit package.
```
ditman, line 4: (Same comment as above, this won't work as is)

renderers/markdown/README.md
```diff
@@ -0,0 +1,10 @@
+This directory contains the default markdown implementations for A2UI.
+
+* `markdown-it-shared` is a shared markdown renderer that uses markdown-it and
+  DOMPurify to render markdown content in general.
+* `markdown-it-lit` is the Lit facade of the markdown renderer for Lit.
+* `markdown-it-angular` is the Angular facade of the markdown renderer
```
ditman, line 6: > `markdown-it-angular` is the Angular...

Haha, aspirational :P

renderers/web_core/src/v0_8/schemas/a2ui_client_capabilities_schema.json
ditman, line 10: I don't understand why am I getting these .JSON updates, is my branch too old? Is this being updated off of `master`, instead of the contents of my branch?

Updating off of `master` sounds _dangerous_.

samples/client/lit/contact/theme/theme.ts
ditman, line 1: Changes to this file are 100% cosmetic, just because I thought I had broken the `contact` example with my markdown changes (turns out I hadn't, but it rubbed me the wrong way)... I can remove this from this PR since it's not too related!

samples/client/lit/contact/contact.ts
ditman, line 1: This is the gist of the "integration" for Lit users, I think it turned out fairly simple, but still it's a breaking change with the previous version of the package, unfortunately.

samples/client/lit/shell/app.ts
```diff
@@ -54,8 +55,11 @@ const configs: Record<string, AppConfig> = {
 
 @customElement("a2ui-shell")
 export class A2UILayoutEditor extends SignalWatcher(LitElement) {
-  @provide({ context: UI.Context.themeContext })
-  accessor theme: v0_8.Types.Theme = uiTheme;
+  @provide({ context: UI.Context.theme })
```
ditman, line 58: Renaming `UI.Context.themeContext` to `UI.Context.theme` is also a breaking change, but IMO it improves legibility a little bit (especially now that there's more than one inhabitant in the `UI.Context` namespace.

samples/client/lit/package.json
```diff
@@ -11,15 +11,13 @@
...
     "demo:all": "npm install && npm run build:renderer && concurrently -k -n \"SHELL,REST,CONT1\" -c \"magenta,blue,green\" \"npm run serve:shell\" \"npm run serve:agent:restaurant\" \"npm run serve:agent:contact_lookup\"",
     "demo:restaurant": "npm install && npm run build:renderer && concurrently -k -n \"SHELL,REST\" -c \"magenta,blue\" \"npm run serve:shell\" \"npm run serve:agent:restaurant\"",
-    "demo:contact": "npm install && npm run build:renderer && concurrently -k -n \"SHELL,CONT1\" -c \"magenta,green\" \"npm run serve:shell\" \"npm run serve:agent:contact_lookup\"",
-    "demo:rizzcharts": "npm install && npm run build:renderer && concurrently -k -n \"SHELL,RIZZ\" -c \"magenta,yellow\" \"npm run serve:shell\" \"npm run serve:agent:rizzcharts\"",
-    "demo:orchestrator": "npm install && npm run build:renderer && concurrently -k -n \"SHELL,ORCH\" -c \"magenta,cyan\" \"npm run serve:shell\" \"npm run serve:agent:orchestrator\""
+    "demo:contact": "npm install && npm run build:renderer && concurrently -k -n \"SHELL,REST\" -c \"magenta,blue\" \"npm run serve:contact\" \"npm run serve:agent:contact_multi_surface\""
```
ditman: I added `demo:contact` for my own convenience, but we can remove it if we don't want it.

ditman:
(Rebasing to resolve merge conflicts)

ditman:
I had to tweak the build script task a little bit to ensure the new packages I added as path dependencies for the lit renderer are being built.

samples/client/lit/package.json
ditman: This is outdated, somebody added a bunch of other commands and I don't want to step on their toes.

yuantian-agentspace:
I didn't go through all the code changes. Overall, I agree with the approach.

nit: I see the default is noop. is it possible to make the default renderer capable to handle very basic markdown? So that it can be used out of box without a markdown provider?

renderers/lit/src/0.8/ui/directives/noop_markdown.ts
ditman: This now warns users once when the directive is created.

ditman:
> nit: I see the default is noop. is it possible to make the default renderer capable to handle very basic markdown? So that it can be used out of box without a markdown provider?

@yuantian-agentspace I've made the NoopMarkdownRenderer a very very very simple MinimalMarkdownRenderer (you'll be familiar with it :P). It also warns programmers the first time it's used, pointing them to the more fully-featured one.

renderers/lit/src/0.8/ui/context/context.ts
ditman: I've setup `prettier` in the three new packages, but not `lit` yet. We may be able to hoist the configuration from the `markdown` subdirectory to `renderers`.

renderers/lit/package-lock.json
ditman: I've added a `.gitignore` to the `markdown` subdirectory so lockfiles are not committed there.

renderers/markdown/markdown-it-lit/.npmrc
ditman: This is used to publish packages through a different registry (?) it might be needed for internal compliance.

renderers/markdown/markdown-it-lit/package.json
ditman: Reviewed.

renderers/markdown/README.md
ditman: This is done now.

ditman:
I've added the Angular integration that allows for DI, and removed the `markdown-it` dependency from the Angular renderer as well.

renderers/lit/src/0.8/ui/directives/minimal_markdown.ts
jacobsimionato, line 1: I wonder if it makes sense for us to not include this, just so that it's less code to maintain. I think that having a partial implementation of Markdown could ultimately be frustrating for developers, and it shouldn't be too hard to wire up a real markdown renderer.

jacobsimionato (commented):
This looks super neat!

I think my main concern here is just that it adds a lot of boilerplate.

Another option here might be for us to focus on separating the standard catalog from the central renderer, and also breaking down catalogs into freestanding component implementations which are composed together into a catalog.

That way, Google internal customers could import the code and just fork the "Text" Component implementation, and the BasicCatalog definition, substituting it with a different implementation of Text, with a different Markdown dependency. I think that might be neater overall rather than us needing to formalize the concept of a Markdown renderer in this way.

It's a bit of a hack, because we'd still import all the code and then just have BUILD rule that excludes the sources with the unwanted dependency, but maybe simpler overall?

All that said, I'm still happy to land this, and maybe it's still beneficial. I don't understand the markdown rendering landscape across the industry and whether people have strong opinions or specific needs which make this injection approach valuable!

renderers/lit/src/0.8/ui/directives/minimal_markdown.ts
ditman: Yeah, originally I had a thing that just wrapped markdown in a `<pre>` but got some feedback from @yuantian-agentspace suggesting to minimally render something. I can go back to the true noop version, maybe make it a `<span>` instead of a `<pre>` so plain text kind of looks nice.

ditman:
(Closing in favor of the simpler: https://github.com/google/A2UI/pull/647)

PR #606, opened 2026-02-09
Implement the data model for web renderers v0.9

jacobsimionato:
Refactor: Enhance DataModel robustness and type safety in web_core v0.9

This change refactors the DataModel in `@a2ui/web_core/v0_9` to improve its robustness, type safety, and adherence to the A2UI protocol specification, particularly concerning data manipulation and path resolution.

Key changes include:

- **Improved DataModel.set path handling:**
    - Modified the `set` method to throw an error if an intermediate segment of a path points to a primitive value when attempting to set a nested property, preventing unexpected data corruption.
    - Added a check to enforce numeric indices for array segments in paths, throwing an error for non-numeric access on arrays to maintain data model integrity.
    - Updated documentation for the `set` method to explicitly describe the behavior when setting values to `undefined` for both objects (key removal) and arrays (index nullification).

- **Enhanced Type Safety:**
    - Updated the `Subscription<T>` interface to correctly type `value` and `onChange` callback parameters as `T | undefined`, reflecting potential `undefined` states in the data model and improving consumer type safety.

- **Test Coverage & Fixes:**
    - Added new test cases in `data-model.test.ts` to validate the error conditions for primitive path traversal and non-numeric array segment access.
    - Resolved TypeScript compilation errors in `web_core` and `lit` by ensuring `@types/node` is correctly installed and by updating `web_core/package.json` to explicitly export `v0_9` modules.
    - Corrected implicit `any` type errors in test callbacks.
    - Fixed a JSON syntax error in `specification/v0_9/json/standard_catalog.json`.

- **Documentation Updates:**
    - Updated `specification/v0_9/docs/a2ui_protocol.md` to reflect the refined data model behavior for `undefined` values and array index rules.

### DIFF
--- a/renderers/web_core/src/v0_8/schemas/standard_catalog_definition.json
+++ b/renderers/web_core/src/v0_8/schemas/standard_catalog_definition.json
@@ -31,9 +31,7 @@
           ]
         }
       },
-      "required": [
-        "text"
-      ]
+      "required": ["text"]
     },
     "Image": {
       "type": "object",
@@ -76,9 +74,7 @@
           ]
         }
       },
-      "required": [
-        "url"
-      ]
+      "required": ["url"]
     },
     "Icon": {
       "type": "object",
@@ -148,9 +144,7 @@
           }
         }
       },
-      "required": [
-        "name"
-      ]
+      "required": ["name"]
     },
     "Video": {
       "type": "object",
@@ -170,9 +164,7 @@
           }
         }
       },
-      "required": [
-        "url"
-      ]
+      "required": ["url"]
     },
     "AudioPlayer": {
       "type": "object",
@@ -205,9 +197,7 @@
           }
         }
       },
-      "required": [
-        "url"
-      ]
+      "required": ["url"]
     },
     "Row": {
       "type": "object",
@@ -236,10 +226,7 @@
                   "type": "string"
                 }
               },
-              "required": [
-                "componentId",
-                "dataBinding"
-              ]
+              "required": ["componentId", "dataBinding"]
             }
           }
         },
@@ -258,17 +245,10 @@
         "alignment": {
           "type": "string",
           "description": "Defines the alignment of children along the cross axis (vertically). This corresponds to the CSS 'align-items' property.",
-          "enum": [
-            "start",
-            "center",
-            "end",
-            "stretch"
-          ]
+          "enum": ["start", "center", "end", "stretch"]
         }
       },
-      "required": [
-        "children"
-      ]
+      "required": ["children"]
     },
     "Column": {
       "type": "object",
@@ -297,10 +277,7 @@
                   "type": "string"
                 }
               },
-              "required": [
-                "componentId",
-                "dataBinding"
-              ]
+              "required": ["componentId", "dataBinding"]
             }
           }
         },
@@ -319,17 +296,10 @@
         "alignment": {
           "type": "string",
           "description": "Defines the alignment of children along the cross axis (horizontally). This corresponds to the CSS 'align-items' property.",
-          "enum": [
-            "center",
-            "end",
-            "start",
-            "stretch"
-          ]
+          "enum": ["center", "end", "start", "stretch"]
         }
       },
-      "required": [
-        "children"
-      ]
+      "required": ["children"]
     },
     "List": {
       "type": "object",
@@ -358,35 +328,22 @@
                   "type": "string"
                 }
               },
-              "required": [
-                "componentId",
-                "dataBinding"
-              ]
+              "required": ["componentId", "dataBinding"]
             }
           }
         },
         "direction": {
           "type": "string",
           "description": "The direction in which the list items are laid out.",
-          "enum": [
-            "vertical",
-            "horizontal"
-          ]
+          "enum": ["vertical", "horizontal"]
         },
         "alignment": {
           "type": "string",
           "description": "Defines the alignment of children along the cross axis.",
-          "enum": [
-            "start",
-            "center",
-            "end",
-            "stretch"
-          ]
+          "enum": ["start", "center", "end", "stretch"]
         }
       },
-      "required": [
-        "children"
-      ]
+      "required": ["children"]
     },
     "Card": {
       "type": "object",
@@ -397,9 +354,7 @@
           "description": "The ID of the component to be rendered inside the card."
         }
       },
-      "required": [
-        "child"
-      ]
+      "required": ["child"]
     },
     "Tabs": {
       "type": "object",
@@ -429,16 +384,11 @@
                 "type": "string"
               }
             },
-            "required": [
-              "title",
-              "child"
-            ]
+            "required": ["title", "child"]
           }
         }
       },
-      "required": [
-        "tabItems"
-      ]
+      "required": ["tabItems"]
     },
     "Divider": {
       "type": "object",
@@ -447,10 +397,7 @@
         "axis": {
           "type": "string",
           "description": "The orientation of the divider.",
-          "enum": [
-            "horizontal",
-            "vertical"
-          ]
+          "enum": ["horizontal", "vertical"]
         }
       }
     },
@@ -467,10 +414,7 @@
           "description": "The ID of the component to be displayed inside the modal."
         }
       },
-      "required": [
-        "entryPointChild",
-        "contentChild"
-      ]
+      "required": ["entryPointChild", "contentChild"]
     },
     "Button": {
       "type": "object",
@@ -521,22 +465,14 @@
                     }
                   }
                 },
-                "required": [
-                  "key",
-                  "value"
-                ]
+                "required": ["key", "value"]
               }
             }
           },
-          "required": [
-            "name"
-          ]
+          "required": ["name"]
         }
       },
-      "required": [
-        "child",
-        "action"
-      ]
+      "required": ["child", "action"]
     },
     "CheckBox": {
       "type": "object",
@@ -569,10 +505,7 @@
           }
         }
       },
-      "required": [
-        "label",
-        "value"
-      ]
+      "required": ["label", "value"]
     },
     "TextField": {
       "type": "object",
@@ -620,9 +553,7 @@
           "description": "A regular expression used for client-side validation of the input."
         }
       },
-      "required": [
-        "label"
-      ]
+      "required": ["label"]
     },
     "DateTimeInput": {
       "type": "object",
@@ -650,9 +581,7 @@
           "description": "If true, allows the user to select a time."
         }
       },
-      "required": [
-        "value"
-      ]
+      "required": ["value"]
     },
     "MultipleChoice": {
       "type": "object",
@@ -703,10 +632,7 @@
                 "description": "The value to be associated with this option when selected."
               }
             },
-            "required": [
-              "label",
-              "value"
-            ]
+            "required": ["label", "value"]
           }
         },
         "maxAllowedSelections": {
@@ -724,7 +650,7 @@
         "filterable": {
           "type": "boolean",
           "description": "If true, displays a search input to filter the options."
-        }
+        },
       }
     },
     "Slider": {
@@ -753,9 +679,7 @@
           "description": "The maximum value of the slider."
         }
       },
-      "required": [
-        "value"
-      ]
+      "required": ["value"]
     }
   },
   "styles": {


renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,189 @@
...
+      }
+      current = current[segment];
+    }
+
+    if (value === undefined) {
+      if (Array.isArray(current)) {
+        current[parseInt(lastSegment, 10)] = undefined;
+      } else {
+        delete current[lastSegment];
+      }
```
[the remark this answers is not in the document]
jacobsimionato: Please add documentation to the 'set' method for this, and also update specification/v0_9/docs/a2ui_protocol.md to specify this behavior explicitly in the relevant section.

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * A callback function to be invoked when the value changes.
+   */
+  onChange?: (value: T) => void;
+
+  /**
+   * Unsubscribes from the data model.
+   */
+  unsubscribe(): void;
+}
```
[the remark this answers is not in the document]
jacobsimionato: Please update this.

jacobsimionato:

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    let current = this.data;
+    for (let i = 0; i < segments.length; i++) {
+      const segment = segments[i];
+      if (current[segment] === undefined || current[segment] === null) {
+        const nextSegment = (i < segments.length - 1) ? segments[i + 1] : lastSegment;
+        current[segment] = /^\d+$/.test(nextSegment) ? [] : {};
+      }
+      current = current[segment];
+    }
```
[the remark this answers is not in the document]
jacobsimionato: No, in the case that there is an existing primitive object instead of a container at an intermediate path, the update should fail as it doesn't make sense, so it should throw an exception or something here.

renderers/web_core/src/v0_9/state/data-model.test.ts
```diff
@@ -0,0 +1,188 @@
...
+  it('stops notifying after dispose', () => {
+    let count = 0;
+    const sub = model.subscribe('/');
+    sub.onChange = () => count++;
+    
+    model.dispose();
+    model.set('/foo', 'bar');
+    assert.strictEqual(count, 0);
+  });
+});
```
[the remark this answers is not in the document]
jacobsimionato: Please add a test case for this, but verify failure as explained above.

jacobsimionato:

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,199 @@
...
+      if (current[segment] !== undefined && current[segment] !== null && typeof current[segment] !== 'object') {
+        throw new Error(`Cannot set path '${path}': segment '${segment}' is a primitive value.`);
+      }
+
+      if (current[segment] === undefined || current[segment] === null) {
+        const nextSegment = (i < segments.length - 1) ? segments[i + 1] : lastSegment;
+        current[segment] = /^\d+$/.test(nextSegment) ? [] : {};
+      }
+      current = current[segment];
+    }
```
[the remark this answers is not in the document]
jacobsimionato: Yes, please update this and also update the a2ui_protocol.md doc for v0.9 to specify that this is an error.

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,199 @@
...
+      parentPath = parentPath.substring(0, parentPath.lastIndexOf('/')) || '/';
+      this.notify(parentPath);
+    }
+
+    // Notify Descendants
+    for (const subPath of this.subscriptions.keys()) {
+      if (this.isDescendant(subPath, normalizedPath)) {
+        this.notify(subPath);
+      }
+    }
```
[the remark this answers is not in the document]
jacobsimionato: Ignore this. We don't need to optimize now.

jacobsimionato:

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -408,6 +408,7 @@ When a container component (such as `Column`, `Row`, or `List`) utilizes the **T
 - **Scope instantiation:** For every item in the array, the client instantiates the template component.
 - **Relative resolution:** Inside these instantiated components, any path that **does not** start with a forward slash `/` is treated as a **Relative Path**.
   - A relative path `firstName` inside a template iterating over `/users` resolves to `/users/0/firstName` for the first item, `/users/1/firstName` for the second, etc.
+  - It is an error to use a non-numeric index on a path segment that refers to an array.
```
gspencergoog, line 411: This is a good addition.

specification/v0_9/docs/a2ui_protocol.md
```diff
@@ -519,7 +520,7 @@ The server sends `updateDataModel` messages to modify the client's data model. T
 
 - If the path exists, the value is updated.
 - If the path does not exist, the value is created.
-- If the value is `null`, the key at that path is removed.
+- If the value is omitted (or set to `undefined`), the key is removed. For arrays, the index is set to `undefined`, preserving length.
```
gspencergoog: It's the value at the index that is set to undefined, right? Not the index itself.

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,207 @@
...
+            this.subscriptions.delete(normalizedPath);
+          }
+        }
+      }
+    };
+
+    Object.defineProperty(subscription, 'value', {
+      get: () => this.get(normalizedPath),
+      enumerable: true
+    });
```
ditman: This `defineProperty` maybe hints at the code wanting `Subscription<T>` to be a proper class, rather than an interface that we coerce a plain JS object to? That way we can make sure the exposed `value` is a proper getter? Why do we prefer this creating a `new Subscription`?

specification/v0_9/docs/a2ui_protocol.md
jacobsimionato: Let's fix this.

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: We do not prefer this! Let's make Subscription a proper class!

jacobsimionato: Fixed.

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * Updates the model at the specific path and notifies all relevant subscribers.
+   * If path is '/' or empty, replaces the entire root.
+   *
+   * Note on `undefined` values:
+   * - For objects: Setting a property to `undefined` removes the key from the object.
+   * - For arrays: Setting an index to `undefined` sets that index to `undefined` but preserves the array length (sparse array).
+   */
+  set(path: string, value: any): void {
```
ditman: I think it'd be nice if this object was more like an "Observable Map", IMO this should return `this`, see `Map.prototype.set()`, for example: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/set#return_value

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,226 @@
...
+    this._unsubscribe();
+  }
+}
+
+/**
+ * A standalone, observable data store representing the client-side state.
+ * It handles JSON Pointer path resolution and subscription management.
+ */
+export class DataModel {
+  private data: any = {};
```
ditman: There's a lot of `any` in this class. can we say `data` must be a `Record` or `Map` or similar, instead of `any`?

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,226 @@
...
+  private readonly subscriptions: Map<string, Set<SubscriptionImpl<any>>> = new Map();
```
ditman, line 71:
```suggestion
  private readonly subscriptions: Map<string, Set<Subscription<any>>> = new Map();
```

We should only use the `Impl` to create new instances, but in general we should code against the interface, if we can!

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  constructor(initialData: any = {}) {
+    this.data = initialData;
+  }
+
+  /**
+   * Updates the model at the specific path and notifies all relevant subscribers.
+   * If path is '/' or empty, replaces the entire root.
```
ditman, line 79: What happens if path is `null` or `undefined`? Does that become a runtime error?

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,226 @@
...
+      }
+
+      if (current[segment] === undefined || current[segment] === null) {
+        const nextSegment = (i < segments.length - 1) ? segments[i + 1] : lastSegment;
+        current[segment] = /^\d+$/.test(nextSegment) ? [] : {};
+      }
+      current = current[segment];
+    }
+
+    if (Array.isArray(current) && !/^\d+$/.test(lastSegment)) {
```
ditman: This regexp is repeated here and in line 107 and 95, please create a helper "isOnlyNumbers" function or similar!

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,226 @@
...
+  unsubscribe(): void {
+    this._unsubscribe();
+  }
+}
+
+/**
+ * A standalone, observable data store representing the client-side state.
+ * It handles JSON Pointer path resolution and subscription management.
+ */
+export class DataModel {
```
ditman, line 69: Is there any way of retrieving the whole built `data` object?

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,226 @@
...
+  private normalizePath(path: string): string {
+    if (path.length > 1 && path.endsWith('/')) {
+      return path.slice(0, -1);
+    }
+    return path || '/';
+  }
+
+  private parsePath(path: string): string[] {
+    return path.split('/').filter(p => p.length > 0);
+  }
```
ditman, line 203: Are there any restrictions on what the syntax for a path can be, other than "separated by `/`"?

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: Yes! Call `myDataModel.get('/')`.

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: I think `Record<string, unknown>` seems to be the best fit, because the structure of the map is unknown.

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: Here, we need a reference to the Impl because it exposes the 'setValue' method which data-model needs access to e.g. on line 213. But it only exposes these objects via the more narrow Subscription interface publicly.

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: Path can't be null or undefined due to the magic of TS, right?

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: I like this suggestion, let's do this.

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: Sounds great

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: Hmmm I'm not sure. I don't think we were super clear in the spec. I think as time goes on, we'll likely be more specific about this in the spec and then can add corresponding validation here.

renderers/web_core/src/v0_9/state/data-model.ts
ditman: I don't trust anything that ends up being javascript not being null or undefined :P (also, can a key come from something at runtime?)

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,237 @@
...
+      if (current === undefined || current === null) return undefined;
+      current = current[segment];
+    }
+    return current;
+  }
+
+  /**
+   * Subscribes to changes at a specific path. Returns a Subscription object.
+   */
+  subscribe<T>(path: string): Subscription<T> {
```
ditman, line 162: I just noticed, do we need a `unsubscribe(subscription: Subscription<T>)` method? How can you stop listening to updates with this API? (If there's no way to unsubscribe, does the `Subscription` need an unsubscribe method at all?)

renderers/web_core/src/v0_9/state/data-model.ts
ditman: SGTM!

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: Well, the Subscription has an `unsubscribe` method, which does clear the reference from DataModel to Subscription (see line 167). There's no need to break the link from Subscription to the application logic after that.

The Subscription is a throwaway object, and so you can call `mySubscription.unsubscribe()` and then `mySubscription = undefined` then mySubscription should just get garbage collected, right?

ditman: Oh yeah, I didn't notice that we were already binding the unsubscribe method when we create the subscription, duh! 🤦

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  /**
+   * Retrieves data at a specific path.
+   */
+  get(path: string): any {
+    if (path === null || path === undefined) {
+      throw new Error("Path cannot be null or undefined.");
+    }
+    if (path === '/' || path === '') return this.data;
```
ditman: I totally missed this line; I think we should make brackets mandatory on statements, even if they're oneliners :P

renderers/web_core/src/v0_9/state/data-model.ts
ditman: Ahhh yes, didn't notice the use of the private setValue!

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: Very fair - this is where your web expertise is essential! I (read: a robot) have added checks for null/undefined.

renderers/web_core/src/v0_9/state/data-model.ts
jacobsimionato: I fixed this now :-D

jacobsimionato:
Thank you for the review!

PR #609, opened 2026-02-09
feat(ui): dispatch validation events from TextField

dmandar:
This PR introduces the a2ui-validation-input event to the 
TextField
 component to bubble validation state changes.

Changes
TextField (
text-field.ts
):
Added dispatch of a2ui-validation-input on every input event.
Payload: { componentId, value, valid: boolean }.
Reason: Native invalid events do not cross the Shadow DOM boundary. This custom event allows parent components (like the Shell) to observe real-time validation status.
Component Gallery (
component-gallery.ts
):
Added a listener for a2ui-validation-input at the document root.
Logs validation events to both the Console and the internal Debug Panel for visual verification.
Verification
Verified manually in the Component Gallery:
Typing in "TextField (Regex)" now triggers logs in the Debug Panel.
Valid/Invalid states strictly match the regex rules.

## Pre-launch Checklist

- [X ] I signed the [CLA].
- [ X] I read the [Contributors Guide].
- [ X] I read the [Style Guide].
- [ X] I have added updates to the [CHANGELOG].
- [ X] I updated/added relevant documentation.
- [ X] My code changes (if any) have tests.

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/lit/src/0.8/core.ts
+++ b/renderers/lit/src/0.8/core.ts
@@ -22,8 +22,8 @@ import * as Styles from "@a2ui/web_core/styles/index";
 import { A2uiMessageProcessor } from "@a2ui/web_core/data/model-processor";
 import * as Primitives from "@a2ui/web_core/types/primitives";
 import { create as createSignalA2uiMessageProcessor } from "./data/signal-model-processor.js";
-
-export { Types, Guards, Schemas, Styles, A2uiMessageProcessor, Primitives };
+import { Events as WebEvents } from "@a2ui/web_core";
+export { Types, Guards, Schemas, Styles, A2uiMessageProcessor, Primitives, WebEvents };
 
 export const Data = {
   createSignalA2uiMessageProcessor,
--- a/renderers/lit/src/0.8/ui/text-field.ts
+++ b/renderers/lit/src/0.8/ui/text-field.ts
@@ -20,6 +20,7 @@ import { Root } from "./root.js";
 import { A2uiMessageProcessor } from "@a2ui/web_core/data/model-processor";
 import * as Primitives from "@a2ui/web_core/types/primitives";
 import * as Types from "@a2ui/web_core/types/types";
+import { Events } from "@a2ui/web_core";
 import { classMap } from "lit/directives/class-map.js";
 import { styleMap } from "lit/directives/style-map.js";
 import { extractStringValue } from "./utils/utils.js";
@@ -116,15 +117,11 @@ export class TextField extends Root {
           }
 
       this.dispatchEvent(
-        new CustomEvent("a2ui-validation-input", {
-          bubbles: true,
-          composed: true,
-          detail: {
-            componentId: this.id,
-            value: evt.target.value,
-            valid: evt.target.checkValidity(),
-          },
-        })
+        new Events.A2UIValidationEvent({
+          componentId: this.id,
+          value: evt.target.value,
+          valid: evt.target.checkValidity(),
+            })
       );
 
           this.#setBoundValue(evt.target.value);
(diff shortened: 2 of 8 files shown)


renderers/lit/src/0.8/ui/text-field.ts
```diff
@@ -115,6 +115,18 @@ export class TextField extends Root {
             return;
           }
 
+      this.dispatchEvent(
```
ditman, line 119: The indentation of this block seems wrong.

renderers/lit/src/0.8/ui/text-field.ts
```diff
@@ -115,6 +115,18 @@ export class TextField extends Root {
...
+        new CustomEvent("a2ui-validation-input", {
```
ditman: Can this be an `A2UIValidationEvent extends CustomEvent`, with a specific signature for the `detail` property, so it can be accessed safely later?

There's an example of an extension of CustomEvent [here](https://github.com/google/A2UI/blob/main/renderers/lit/src/0.8/events/events.ts#L39-L47). In this case, I *think* the event should be in `web_core`, so it can be shared with other renderers; or is this a Lit-only functionality?

samples/client/lit/component_gallery/component-gallery.ts
```diff
@@ -186,16 +186,40 @@ export class A2UIComponentGallery extends SignalWatcher(LitElement) {
...
   async #initiateSession() {
     const message: v0_8.Types.A2UIClientEventMessage = {
       request: "START_GALLERY"
     };
     await this.#sendAndProcessMessage(message);
   }
 
+  // Debug Validation Events
+  #handleValidationInput = (e: Event) => {
+    const detail = (e as CustomEvent).detail;
```
ditman: (This is where having a custom event type would help to access `detail` in a type-safe way)

renderers/lit/src/0.8/core.ts
```diff
@@ -22,8 +22,8 @@ import * as Styles from "@a2ui/web_core/styles/index";
 import { A2uiMessageProcessor } from "@a2ui/web_core/data/model-processor";
 import * as Primitives from "@a2ui/web_core/types/primitives";
 import { create as createSignalA2uiMessageProcessor } from "./data/signal-model-processor.js";
-
-export { Types, Guards, Schemas, Styles, A2uiMessageProcessor, Primitives };
+import { Events as WebEvents } from "@a2ui/web_core";
+export { Types, Guards, Schemas, Styles, A2uiMessageProcessor, Primitives, WebEvents };
```
ditman: Instead of importing and renaming here, how about exporting these events from web_core directly from `./events/events.js` so they become part of the `Events` namespace offered by the lit renderer?

I think it's a little bit of an anti-pattern to re-export directly from `web_core` (unless you're coding an actual renderer?)

ditman (approved):
This LGTM, but I really think we shouldn't reexport from web_core. Instead, we should re-export from the lit renderer `Events` namespace.

(There's a similar problem to this with the styles. The Lit renderer only reexports the ones coming from `web_core`, instead of its internal version, so application users that want to use the basic styling for their own components, must use `unsafeCSS`, but that's a fix for another day :P)

renderers/lit/src/0.8/core.ts
```diff
@@ -18,12 +18,11 @@ export * as Events from "./events/events.js";
 import * as Types from "@a2ui/web_core/types/types";
 import * as Guards from "@a2ui/web_core/data/guards";
 import { Schemas } from "@a2ui/web_core";
-import * as Styles from "@a2ui/web_core/styles/index";
+import * as Styles from "./styles/index.js";
```
ditman: I think this won't work as is, because there's more stuff being exported from `styles/index` (like a `merge` function). Let's leave this part for another PR, so we land the validation events cleanly!

dmandar: sg

Issue #480, opened 2026-01-13
Autofix malformed JSON returned by the LLM


jonsharkey:
Run fixers after generating the a2ui schema to correct common issues. One I noticed was the LLM putting a trailing "," after the last element in a list, which is noncompliant with JSON.

selamw1:
PR [#619](https://github.com/google/A2UI/pull/619) raised to fix it.

PR #619, opened 2026-02-11
fix(agent): Autofix trailing commas in LLM-generated JSON

selamw1:
Fixes `json.JSONDecodeError` caused by LLMs returning JSON with trailing commas.

## Changes
- Added a regex pre-processing step in `RestaurantAgent` to strip trailing commas before calling `json.loads()`.
- Improves parsing robustness and prevents unnecessary retries.

## Fixes
Resolves [#480](https://github.com/google/A2UI/issues/480)

### DIFF
--- a/a2a_agents/python/a2ui_agent/src/a2ui/extension/send_a2ui_to_client_toolset.py
+++ b/a2a_agents/python/a2ui_agent/src/a2ui/extension/send_a2ui_to_client_toolset.py
@@ -262,7 +262,9 @@ class SendA2uiToClientToolset(base_toolset.BaseToolset):
 
           # Auto-wrap single object in list
           if not isinstance(a2ui_json_payload, list):
-            logger.info("Received a single JSON object, wrapping in a list for validation.")
+            logger.info(
+                "Received a single JSON object, wrapping in a list for validation."
+            )
             a2ui_json_payload = [a2ui_json_payload]
 
           jsonschema.validate(instance=a2ui_json_payload, schema=a2ui_schema)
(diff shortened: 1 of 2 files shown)


samples/agent/adk/restaurant_finder/agent.py
```diff
@@ -226,6 +227,9 @@ async def stream(self, query, session_id) -> AsyncIterable[dict[str, Any]]:
                     if not json_string_cleaned:
                         raise ValueError("Cleaned JSON string is empty.")
 
+                    # Autofix: Remove trailing commas from arrays and objects
+                    json_string_cleaned = re.sub(r",(?=\s*[\]}])", "", json_string_cleaned)
```
wrenj: thanks for this

can you put it in the SDK here:
https://github.com/google/A2UI/blob/main/a2a_agents/python/a2ui_agent/src/a2ui/extension/send_a2ui_to_client_toolset.py#L265

and add some tests

I think flow would be 
if not jsonschema.validate(): 
  run fixers
  if trailing comma fixer does something
    emit warning
 
jsonschema.validate

selamw1: Thanks, I've adjusted the code into the SDK and followed the suggested flow. I also added unit tests to verify the fix and the warning message.

a2a_agents/python/a2ui_agent/src/a2ui/extension/send_a2ui_to_client_toolset.py
```diff
@@ -252,17 +253,27 @@ async def run_async(
...
-        a2ui_json_payload = json.loads(a2ui_json)
+        a2ui_schema = await self.get_a2ui_schema(tool_context)
+        try:
+          # Attempt to parse and validate
+          a2ui_json_payload = json.loads(a2ui_json)
+          jsonschema.validate(instance=a2ui_json_payload, schema=a2ui_schema)
 
-        # Auto-wrap single object in list
-        if not isinstance(a2ui_json_payload, list):
-          logger.info(
```
wrenj, line 260: lets keep the list wrapping logic

zeroasterisk:
wrenj's approval is correct — just needs the list wrapping restored. The regex handles multiple trailing commas fine (each `,` before `]` or `}` is matched individually).

Once the auto-wrap logic is back, this is good to merge. 👍

selamw1:
Thanks! preserved the list wrapping logic in both paths.

nan-yu (approved):
/lgtm

PR #621, opened 2026-02-12
Refactor theme validation to use catalog schema reference

jacobsimionato:
This PR refactors the A2UI v0.9 and v0.10 protocol schemas to strictly reference theme properties from the component catalog, aligning theme validation with the existing pattern for components and functions.

### Changes:
- **Schema Refactoring:**
  - Removed the top-level `theme` object from `standard_catalog.json` (v0.9 & v0.10).
  - Inlined `primaryColor`, `iconUrl`, and `agentDisplayName` into the `$defs/theme` definition within `standard_catalog.json`.
  - Updated `server_to_client.json` (v0.9 & v0.10) to reference the theme schema via `catalog.json#/$defs/theme` instead of using a generic object type.

- **Documentation:**
  - Updated `a2ui_protocol.md` (v0.9 & v0.10) to explicitly state that the theme schema is part of the catalog and referenced via the `catalog.json` alias.

- **Testing:**
  - Added comprehensive `theme_validation.json` test suites for both versions to verify that valid themes are accepted and invalid ones (e.g., incorrect types or hex codes) are rejected.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,35 @@
+# CODEOWNERS file
+#
+# It allows us to define individuals or teams that are responsible for code in the repository.
+#
+# Syntax:
+# pattern owner1 owner2
+#
+# The pattern is a glob pattern relative to the root of the repository.
+# Owners can be:
+# - A GitHub username (e.g., @octocat)
+# - An organization team (e.g., @org/team-name)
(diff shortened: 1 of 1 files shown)


gspencergoog (approved):
Oh, I see.  Yes, this is probably an error in the spec, and was intended to work that way.

PR #622, opened 2026-02-12
docs(a2ui): Update documentation to align with v0.9 specification

gspencergoog:
# Description

This updates the docs folder to describe the 0.9 specification instead of 0.8.  I added disclaimers that the renderers are not yet up to spec, but I think it's still a little odd to have docs for 0.9 and no examples yet.

Better that the docs represent the current spec, though.

Fixes https://github.com/google/A2UI/issues/703

### DIFF
--- a/agent_sdks/python/uv.lock
+++ b/agent_sdks/python/uv.lock
@@ -124,14 +124,14 @@ wheels = [
 
 [[package]]
 name = "authlib"
-version = "1.6.6"
+version = "1.6.7"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "cryptography" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/bb/9b/b1661026ff24bc641b76b78c5222d614776b0c085bcfdac9bd15a1cb4b35/authlib-1.6.6.tar.gz", hash = "sha256:45770e8e056d0f283451d9996fbb59b70d45722b45d854d58f32878d0a40c38e", size = 164894, upload-time = "2025-12-12T08:01:41.464Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/49/dc/ed1681bf1339dd6ea1ce56136bad4baabc6f7ad466e375810702b0237047/authlib-1.6.7.tar.gz", hash = "sha256:dbf10100011d1e1b34048c9d120e83f13b35d69a826ae762b93d2fb5aafc337b", size = 164950, upload-time = "2026-02-06T14:04:14.171Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/54/51/321e821856452f7386c4e9df866f196720b1ad0c5ea1623ea7399969ae3b/authlib-1.6.6-py2.py3-none-any.whl", hash = "sha256:7d9e9bc535c13974313a87f53e8430eb6ea3d1cf6ae4f6efcd793f2e949143fd", size = 244005, upload-time = "2025-12-12T08:01:40.209Z" },
+    { url = "https://files.pythonhosted.org/packages/f8/00/3ed12264094ec91f534fae429945efbaa9f8c666f3aa7061cc3b2a26a0cd/authlib-1.6.7-py2.py3-none-any.whl", hash = "sha256:c637340d9a02789d2efa1d003a7437d10d3e565237bcb5fcbc6c134c7b95bab0", size = 244115, upload-time = "2026-02-06T14:04:12.141Z" },
 ]
 
 [[package]]
--- a/samples/agent/mcp/uv.lock
+++ b/samples/agent/mcp/uv.lock
@@ -152,14 +152,14 @@ wheels = [
 
 [[package]]
 name = "authlib"
-version = "1.6.6"
+version = "1.6.7"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "cryptography" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/bb/9b/b1661026ff24bc641b76b78c5222d614776b0c085bcfdac9bd15a1cb4b35/authlib-1.6.6.tar.gz", hash = "sha256:45770e8e056d0f283451d9996fbb59b70d45722b45d854d58f32878d0a40c38e", size = 164894, upload-time = "2025-12-12T08:01:41.464Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/49/dc/ed1681bf1339dd6ea1ce56136bad4baabc6f7ad466e375810702b0237047/authlib-1.6.7.tar.gz", hash = "sha256:dbf10100011d1e1b34048c9d120e83f13b35d69a826ae762b93d2fb5aafc337b", size = 164950, upload-time = "2026-02-06T14:04:14.171Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/54/51/321e821856452f7386c4e9df866f196720b1ad0c5ea1623ea7399969ae3b/authlib-1.6.6-py2.py3-none-any.whl", hash = "sha256:7d9e9bc535c13974313a87f53e8430eb6ea3d1cf6ae4f6efcd793f2e949143fd", size = 244005, upload-time = "2025-12-12T08:01:40.209Z" },
+    { url = "https://files.pythonhosted.org/packages/f8/00/3ed12264094ec91f534fae429945efbaa9f8c666f3aa7061cc3b2a26a0cd/authlib-1.6.7-py2.py3-none-any.whl", hash = "sha256:c637340d9a02789d2efa1d003a7437d10d3e565237bcb5fcbc6c134c7b95bab0", size = 244115, upload-time = "2026-02-06T14:04:12.141Z" },
 ]
 
 [[package]]


gspencergoog:
> The content updates look accurate, but I think we need version-specific documentation tabs before merging this.

Yeah, I wondered about this too.  As for what form this should take, the examples you gave basically just have a banner or a link to the previous versions.

React:
image

Vue:
image

And they start off on the "current" version by default.

That's *sort of* what I was going for here anyhow: the 0.8 spec docs are still there, but they're pared down.

What if we just do what React does and just keep previous versions as separate sites?  I mean, for now, we can just have them at a different URL with the version in it, so we don't need separate host names, but does that sound good?

And the bigger question is: do we start people on the 0.9 docs or the 0.8 docs when they go to [a21ui.org](https://a2ui.org)?

gspencergoog:
I assume we don't want to go as far as Python does: https://docs.python.org/3/

It has a pulldown to select the version to view.  I think only A2UI renderer implementers will be interested in that level of detail, and they can still browse the docs if they are at a separate URL like https://a2ui.org/0.8/quickstart/

So, I'm going to proceed like this: I'll convert this PR so that everything except for the Intro and FAQ tabs are under a version number, and the landing page will explain the different versions and their current state. The stuff under 0.8 will be the original content from 0.8 and the 0.9 content will be this content.  The Intro and FAQ tabs will be 0.9 specific in the places where an example is given, and I'll a FAQ about the versions.

Sound good @zeroasterisk ?

gspencergoog:
Okay, I think this reflects what you were asking for.

image

PTAL @zeroasterisk

docs/v0.10/specification/docs/a2ui_custom_functions.md
jacobsimionato, line 1: Do we need any docs for v0.10 at this stage? I think it would be nicer to not maintain them if necessary at this point.
I propose rm -r docs/v0.10

Though I guess the script just copied these, so maybe it's more work to not have them copied. If you feel the minimum maintenance option is to just check these in, I'm fine with that!

docs/scripts/prepare_docs.py
```diff
@@ -0,0 +1,170 @@
...
+    from migrate_wrappers import migrate_wrappers
+except ImportError:
+    # Fallback for when running directly or in different contexts
+    from docs.scripts.migrate_wrappers import migrate_wrappers
+
+"""
+Prepares the raw A2UI specification files for use in the MkDocs build.
+
+This script performs the following tasks:
+1.  **Copies Source Files**: Copies the raw specification files from the root `specification/` directory (e.g., `specification/v0_9`) into the versioned documentation structure (e.g., `docs/v0.9/specification`).
```
jacobsimionato: Would it make sense to .gitignore the copied files and have this be part of the build process, rather than checking in the copied files? That would keep the repo smaller and avoid confusion about what needs to be maintained etc.

Or is there some way to at least mark the copied files very clearly, e.g. by placing them in `generated/` or `script_output/` or something? I guess maybe that breaks all the links and/or makes the URLs weird?

gspencergoog: No, we should just .gitignore them and not check them in.  They're just copies, and we should only have one checked in.

docs/introduction/where-is-it-used.md
```diff
@@ -134,16 +134,16 @@ The A2UI community is building exciting projects:
...
 ## Next Steps
 
-- [Quickstart Guide](../quickstart.md) - Try the demo
-- [Agent Development](../guides/agent-development.md) - Build an agent
-- [Client Setup](../guides/client-setup.md) - Integrate a renderer
-- [Community](../community.md) - Join the community
+- [Quickstart Guide](../v0.9/quickstart.md) - Try the demo
+- [Agent Development](../v0.9/guides/agent-development.md) - Build an agent
+- [Client Setup](../v0.9/guides/client-setup.md) - Integrate a renderer
+- [Community](../v0.9/community.md) - Join the community
```
zeroasterisk: Legacy from when I was harassing you about version paths?

docs/introduction/who-is-it-for.md
```diff
@@ -46,7 +46,7 @@ Do you ship your agent into other apps you don't necessarily control?
 - Extensible: custom component catalogs
 - Open source (Apache 2.0)
 
-**Get started:** [Community](../community.md) | [Roadmap](../roadmap.md)
+**Get started:** [Community](../v0.9/community.md) | [Roadmap](../v0.9/roadmap.md)
```
zeroasterisk: version paths?

docs/v0.9/community.md
zeroasterisk, line 1: this just doesn't make sense... v0.9 community page?

mkdocs.yaml
```diff
@@ -39,37 +40,89 @@ nav:
...
-  - Agents (Server-side): agents.md
-  - Community: community.md
-  - Roadmap: roadmap.md
-  - Reference:
-      - Component Reference: reference/components.md
-      - Message Reference: reference/messages.md
+      - FAQ: introduction/faq.md
+  - v0.9 (Current):
+      - Quickstart: v0.9/quickstart.md
+      - A2UI Composer ⭐: v0.9/composer.md
```
zeroasterisk: I think these items have to move out of versioned docs ^

mkdocs.yaml
```diff
@@ -39,37 +40,89 @@ nav:
...
+          - Data Binding: v0.9/concepts/data-binding.md
+      - Specifications:
+          - A2UI Specification: v0.9/specification/docs/v0.9-a2ui.md
+          - Evolution Guide: v0.9/specification/docs/v0.9-evolution-guide.md
+          - Extension Specification: v0.9/specification/docs/v0.9-extension-spec.md
+          - Custom Functions: v0.9/specification/docs/v0.9-custom-functions.md
+      - Catalogs: v0.9/catalogs.md
+      - Renderers (Clients): v0.9/renderers.md
+      - Transports (Message Passing): v0.9/transports.md
+      - Agents (Server-side): v0.9/agents.md
```
zeroasterisk: I agree all of the above are versioned docs ^

mkdocs.yaml
```diff
@@ -39,37 +40,89 @@ nav:
...
+      - Community: v0.9/community.md
+      - Roadmap: v0.9/roadmap.md
```
zeroasterisk: not versioned

mkdocs.yaml
```diff
@@ -39,37 +40,89 @@ nav:
...
+      - Specifications:
+          - A2UI Specification: v0.9/specification/docs/v0.9-a2ui.md
+          - Evolution Guide: v0.9/specification/docs/v0.9-evolution-guide.md
+          - Extension Specification: v0.9/specification/docs/v0.9-extension-spec.md
+          - Custom Functions: v0.9/specification/docs/v0.9-custom-functions.md
+      - Catalogs: v0.9/catalogs.md
+      - Renderers (Clients): v0.9/renderers.md
+      - Transports (Message Passing): v0.9/transports.md
+      - Agents (Server-side): v0.9/agents.md
+      - Community: v0.9/community.md
```
zeroasterisk: not versioned

mkdocs.yaml
```diff
@@ -39,37 +40,89 @@ nav:
...
+      - Roadmap: v0.9/roadmap.md
+      - Reference:
+          - Component Reference: v0.9/reference/components.md
+          - Message Reference: v0.9/reference/messages.md
+          - Client-Side Functions: v0.9/reference/functions.md
```
zeroasterisk: agree versioned

mkdocs.yaml
```diff
@@ -39,37 +40,89 @@ nav:
...
-  - Renderers (Clients): renderers.md
-  - Transports (Message Passing): transports.md
-  - Agents (Server-side): agents.md
-  - Community: community.md
-  - Roadmap: roadmap.md
-  - Reference:
-      - Component Reference: reference/components.md
-      - Message Reference: reference/messages.md
+      - FAQ: introduction/faq.md
+  - v0.9 (Current):
```
zeroasterisk: not "current" until we have renderers ready

zeroasterisk (changes requested):
I now doubt everything I said before :) 

We need docs and specs in sync per version, but I think this is feeling kinda unwieldy to maintain.  
- Can we reduce scope some?
- Can we reduce what needs multiple version of concurrent docs?

I appreciate you getting it to here, just feels like a lot

docs/v0.9/community.md
gspencergoog: Yeah, good point.  We should move that up a level.

gspencergoog:
> I now doubt everything I said before :)

LOL! I do this on the daily.

> We need docs and specs in sync per version, but I think this is feeling kinda unwieldy to maintain.
> 
> * Can we reduce scope some?
> * Can we reduce what needs multiple version of concurrent docs?
> 
> I appreciate you getting it to here, just feels like a lot

It is a lot. :-)

I'm happy to keep iterating on this until we have something good, but would it be a good idea to hand this (or maybe the result of what we arrive at?) to someone with DevRel credentials?

Want me to keep iterating, or should I pause until we figure out some of the questions?

mkdocs.yaml
gspencergoog: Okay, but then what version of the spec should the examples in those docs refer to?

mkdocs.yaml
gspencergoog: Okay, I updated the language to be "Proposed" (v0.10), "Draft" (v0.9), and "Current Stable" (v0.8).

mkdocs.yaml
gspencergoog: Transports is pretty agnostic, actually.  I moved it up a level.

gspencergoog:
Okay, here's the index now:

image

I've reworded things so that 0.8 is "Stable", 0.9 is "Draft", and 0.10 is "Proposed".

I moved the things you mention above up a level, and also moved Transport up a level (out of the versioned dirs).

PTAL.  If this isn't converging, let me know and we'll try a different approach.

gspencergoog:
Okay, @zeroasterisk , this contains the docs updates you made, and a fix for the build.

I didn't make any changes from what you had in your docs.

PTAL, and approve if it looks good.

PR #624, opened 2026-02-12
Add basic validation function in python

wrenj:
To cover common errors not covered by the json schema


a2a_agents/python/a2ui_agent/src/a2ui/extension/validation.py
```diff
@@ -0,0 +1,262 @@
...
+
+def validate_a2ui_json(a2ui_json: Union[Dict[str, Any], List[Any]], a2ui_schema: Dict[str, Any]) -> None:
+    """
+    Validates the A2UI JSON payload against the provided schema and checks for integrity.
+
+    Checks performed:
+    1.  **JSON Schema Validation**: Ensures payload adheres to the A2UI schema.
+    2.  **Component Integrity**:
+        -   All component IDs are unique.
+        -   A 'root' component exists.
```
nan-yu: If the updateComponents message only include a partial list of components to update an existing surface, does it also require a `root` component?

nan-yu: Should we only enforce the root requirement if there is a CreateSurface message along with the UpdateComponents message?

a2a_agents/python/a2ui_agent/src/a2ui/extension/send_a2ui_to_client_toolset.py
```diff
@@ -262,7 +263,7 @@ async def run_async(
           a2ui_json_payload = [a2ui_json_payload]
 
         a2ui_schema = await self.get_a2ui_schema(tool_context)
-        jsonschema.validate(instance=a2ui_json_payload, schema=a2ui_schema)
+        validate_a2ui_json(a2ui_json_payload, a2ui_schema)
```
nan-yu, line 266: What is the expected format of the a2ui_schema?

v0.9 is a modularized version with multiple schemas. Do we need to bundle them into a unified one?

https://github.com/google/A2UI/pull/557 adds an A2uiValidator that builds referencing registry to work with multiple schemas: https://github.com/nan-yu/A2UI/blob/836b2858833516ec532fdaa5366e0c4f386c72ad/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/validator.py. I think we can move these additional validations into the A2uiValidator.

The send_a2ui_to_client_toolset.py will extract the A2uiCatalog instance stored in the session and invoke the validator to validate the json payload, for example, https://github.com/google/A2UI/pull/620/changes/f35082315c9638467fbbcefce590b718d382ec3a#diff-5f858e097e91850cde41d939245f10a3ef616ec3875236afcb4cfa4105f4384aR279.

nan-yu:
The build was failing because of the format check. `uv run pyink .` would reformat the files.

a2a_agents/python/a2ui_agent/src/a2ui/extension/send_a2ui_to_client_toolset.py
wrenj: its the dict produced by json.loads(), the validate function doesn't need to resolve all the refs

a2a_agents/python/a2ui_agent/src/a2ui/extension/validation.py
wrenj: this validation logic doesn't work on streaming

it would need the final state of updateComponents to validate

wrenj:
> The build was failing because of the format check. `uv run pyink .` would reformat the files.

done!

a2a_agents/python/a2ui_agent/src/a2ui/extension/send_a2ui_to_client_toolset.py
nan-yu: Got it. The A2UI full schema might be cached in the context. I can follow up to consolidate with the A2UI validator in a separate PR.

a2a_agents/python/a2ui_agent/src/a2ui/extension/validation.py
nan-yu: ACK.

PR #631, opened 2026-02-16
Add a framework-agnostic data layer to the web core library

jacobsimionato:
This PR introduces a framework-agnostic data model for A2UI within the `web_core` library, following the architecture outlined in [Unified Data Model Architecture](specification/v0_9/docs/unified_data_model.md).

### Key Changes

- **New v0.9 Data Model:** Implemented the three-layer architecture in `renderers/web_core/src/v0_9`:
    - **Processing Layer:** `A2uiMessageProcessor` for handling the A2UI message stream.
    - **Data Layer:** Persistent, "dumb" models including `SurfaceGroupModel`, `SurfaceModel`, `SurfaceComponentsModel`, and `DataModel`.
    - **Context Layer:** Transient `DataContext` and `ComponentContext` to facilitate rendering and scope resolution.
- **Dependency Updates:** Added `zod` and `zod-to-json-schema` to `web_core` to support type-safe catalog definitions and schema introspection as planned in the v0.9 evolution.
- **Package Exports:** Added a new `./v0_9` export to `@a2ui/web_core` to allow framework-specific renderers (like Lit or Angular) to opt-in to the new architecture.
- **Improved Testing:** Added a `test` script using the native Node.js test runner for the new model.

### Architectural Alignment
The implementation adheres to the principles of separation of concerns, mutable but observable "dumb" models, and granular reactivity through standard observer patterns, ensuring it remains framework-agnostic and high-performance.


renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,203 @@
...
+            if (!component) {
+                console.warn(`Cannot create component ${id} without a type.`);
+                continue;
+            }
+            surface.componentsModel.createComponent(id, component, properties);
+          }
+        }
+      } else if (msg.updateDataModel) {
+        const payload = msg.updateDataModel;
+        const path = payload.path || '/';
```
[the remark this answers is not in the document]
jacobsimionato: This is fine, because it's declared as a oneof in the JSON schema. But add a check at the start to ensure that exactly one message type is present.

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,189 @@
...
+      if (current[segment] === undefined || current[segment] === null) {
+        const nextSegment = (i < segments.length - 1) ? segments[i + 1] : lastSegment;
+        current[segment] = /^\d+$/.test(nextSegment) ? [] : {};
+      }
+      current = current[segment];
+    }
+
+    if (value === undefined) {
+      if (Array.isArray(current)) {
+        current[parseInt(lastSegment, 10)] = undefined;
```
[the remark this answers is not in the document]
jacobsimionato: This is the intention. Please leave this as-is.

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    let current = this.data;
+    for (let i = 0; i < segments.length; i++) {
+      const segment = segments[i];
+      if (current[segment] === undefined || current[segment] === null) {
+        const nextSegment = (i < segments.length - 1) ? segments[i + 1] : lastSegment;
+        current[segment] = /^\d+$/.test(nextSegment) ? [] : {};
+      }
+      current = current[segment];
+    }
```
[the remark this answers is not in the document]
jacobsimionato: Please add the check/failure. If an intermediate path exists and it's a primitive value, an error should be thrown.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,203 @@
...
+      if (msg.updateComponents) {
+        const payload = msg.updateComponents;
+        for (const comp of payload.components) {
+          const { id, component, ...properties } = comp;
+          
+          const existing = surface.componentsModel.get(id);
+          if (existing) {
+            if (component && component !== existing.type) {
+                console.warn(`Attempting to change type of component ${id} from ${existing.type} to ${component}. Ignoring new type.`);
+            }
```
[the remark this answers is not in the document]
jacobsimionato: Please recreate the component in this case.

renderers/web_core/src/v0_9/state/data-model.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  private normalizePath(path: string): string {
+    if (path.length > 1 && path.endsWith('/')) {
+      return path.slice(0, -1);
+    }
+    return path || '/';
+  }
+
+  private parsePath(path: string): string[] {
+    return path.split('/').filter(p => p.length > 0);
```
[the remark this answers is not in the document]
jacobsimionato: Don't fix this now. Filed https://github.com/google/A2UI/issues/651 to do it separately.

jacobsimionato:

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,132 @@
...
+   */
+  removeLifecycleListener(listener: SurfaceLifecycleListener<T>): void {
+    this.model.removeLifecycleListener(listener);
+  }
+
+  processMessages(messages: any[]): void {
+    for (const msg of messages) {
+      if (msg.createSurface) {
+        this.handleCreateSurface(msg.createSurface);
+      } else if (msg.blockInput) {
```
gspencergoog: What kind of message is this?  It seems made up, and it's not used.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,132 @@
...
+  private handleCreateSurface(payload: any) {
+    const { surfaceId, catalogId, theme } = payload;
+
+    // Find catalog
+    const catalog = this.catalogs.find(c => c.id === catalogId);
+    if (!catalog) {
+      console.warn(`Catalog not found: ${catalogId}`);
+      // Using first catalog as fallback or erroring? 
+      // For now, let's create a surface with no catalog or throw?
+      // Better to ignore or error.
```
gspencergoog: These comments can be deleted.

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,107 @@
...
+    // 2. Path Check: { path: "..." }
+    if ('path' in value && typeof value.path === 'string') {
+      const absolutePath = this.resolvePath(value.path);
+      return this.dataModel.get(absolutePath);
+    }
+
+    // 3. Function Call: { call: "...", args: ... }
+    if ('call' in value) {
+      // TODO: Implement function calls
+      // For now, return as is or undefined
```
gspencergoog: Is this TODO expected? If so, that's fine.  Just thought I'd check and make sure it was intended.

specification/v0_9/docs/unified_data_model.md
```diff
@@ -0,0 +1,337 @@
...
+
+interface ComponentApi {
+  name: string; // Protocol name (e.g. 'Button')
+  readonly schema: z.ZodType<any>; // Technical definition for capabilities
+}
+```
+
+## **Framework-Specific Renderers**
+
+Framework implementations (Angular, Lit, Flutter) extend the core by providing rendering logic that consumes the `ComponentContext`.
```
gspencergoog: Flutter isn't going to use this TypeScript library, though...

specification/v0_9/docs/renderer_guide.md
```diff
@@ -0,0 +1,337 @@
...
+
+This makes it less error prone, though there is no type safety around the exact value lookups in the example below.
+
+```
+{
+   "id": "tf1",
+   "component": "TextField",
+   "value": DataModelReference(absolutePath: "/formData/email"),
+}
+```
```
gspencergoog: Is this a real proposal to add types to the A2UI JSON output?  Or is this just in an interface layer?

renderers/web_core/src/v0_9/rendering/data-context.ts
jacobsimionato: Yep, this TODO is expected. I wanted to make this PR small enough to review, and implement the function stuff later.

specification/v0_9/docs/unified_data_model.md
jacobsimionato: Oh yeah, I was trying to imply that Flutter would have a Dart implementation of the core data model, and a Flutter-specific layer which would consume ComponentContext.

How about we update the wording to:
> rendering logic that consumes a `ComponentContext`.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Just an interface layer!! A2UI protocol already has as much typing as we can push into it I think. Here, I'm trying to seed the idea that in some languages, we might want to codegen typesafe interfaces based on Catalog documents one day. But not this quarter, or next probably!

renderers/web_core/src/v0_9/schema/common-types.ts
```diff
@@ -0,0 +1,102 @@
+import { z } from 'zod';
```
ditman, line 1: Should this code be part of the artifacts provided by the A2UI spec, instead of being in web_core?

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,124 @@
...
+import { SurfaceGroupModel, SurfaceLifecycleListener } from '../state/surface-group-model.js';
+import { ComponentModel } from '../state/component-model.js';
+
+export type { SurfaceLifecycleListener };
+
+/**
+ * The central processor for A2UI messages.
+ * @template T The concrete type of the Catalog, which extends CatalogApi.
+ */
+export class A2uiMessageProcessor<T extends CatalogApi> {
```
ditman: Why the "A2uiSomething" capitalization? I don't want to bikeshed too much, but should this be "A2UISomething"?

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * @param catalogs A list of available catalogs.
+   * @param actionHandler A global handler for actions from all surfaces.
+   */
+  constructor(
+    private catalogs: T[],
+    private actionHandler: ActionListener
+  ) {
+    this.model.addActionListener(this.actionHandler);
```
ditman: Can `this.model` be created in the `constructor`? I think this'd make the code more readable than having to track the initialization at the top.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /**
+   * Removes a lifecycle listener.
+   */
+  removeLifecycleListener(listener: SurfaceLifecycleListener<T>): void {
+    this.model.removeLifecycleListener(listener);
+  }
+
+  processMessages(messages: any[]): void {
```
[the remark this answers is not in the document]
ditman: Gemini is kind of right, don't we have an A2UIMessage type for here?

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,124 @@
...
+    for (const msg of messages) {
+      this.processMessage(msg);
+    }
+  }
+
+  getSurfaceModel(surfaceId: string): SurfaceModel<T> | undefined {
+    return this.model.getSurface(surfaceId);
+  }
+
+  private processMessage(msg: any): void {
```
ditman: (Same comment about `msg: any` here. Also, I think we should use `message` instead of `msg`.)

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (msg.createSurface) {
```
ditman: I think we should split the logic on this method for different message types, so have a `processCreateSurfaceMessage(message)` function that handles this bit. That way this big method can be split into smaller, single-purpose functions.

renderers/web_core/src/v0_9/rendering/component-context.ts
```diff
@@ -0,0 +1,33 @@
...
+    componentId: string,
+    dataModelBasePath: string = '/'
+  ) {
+    const model = surface.componentsModel.get(componentId);
+    if (!model) {
+      throw new Error(`Component not found: ${componentId}`);
+    }
+    this.componentModel = model;
+    this.dataContext = new DataContext(surface.dataModel, dataModelBasePath);
+    this._actionDispatcher = (action) => surface.dispatchAction(action);
```
ditman, line 28: This dynamic lambda here is a little bit odd, wouldn't `this._actionDispatcher = surface.dispatchAction` have the same effect (in case we don't want to keep a reference to the `surface` being passed in?)

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Resolves a DynamicValue to its current evaluation.
+   * Does not set up any subscriptions.
+   */
+  resolveDynamicValue<V>(value: any): V {
+    // 1. Literal Check
+    if (typeof value !== 'object' || value === null) {
+      return value as V;
```
ditman, line 35: `value as V`. Does this work well when `value === null`? Is `V` somehow nullable by default?

renderers/web_core/src/v0_9/rendering/data-context.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  it('handles root context', () => {
+    const rootContext = new DataContext(model, '/');
+    assert.strictEqual(rootContext.resolveDynamicValue({ path: 'user/name' }), 'Alice');
+  });
+
+  it('subscribes relative path', (_, done) => {
+    const sub = context.subscribeDynamicValue({ path: 'name' });
+    sub.onChange = (val) => {
+      assert.strictEqual(val, 'Charlie');
+      done();
```
ditman: Does this test fail if this callback is never called?

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    return value as V;
+  }
+
+  /**
+   * Subscribes to changes in a DynamicValue.
+   * Returns a Subscription object that provides the current value and allows listening for updates.
+   */
+  subscribeDynamicValue<V>(value: any): Subscription<V> {
```
ditman: qq here, why are we separating the subscription from the actual function? If I understand this correctly, to observe values you need:

```js
let sub = data.subscribeDynamicValue<T>({path: "path/to/value"})
sub.onChange((new_val) => {
  // do things with the onchange
});
```

Why not pass the subscription callback to the subscribe method, and let the subscription be just the unsubscribe object? Something like:


```
let sub = data.subscribeDynamicValue<T>({path: "path/to/value"}, (new_val) => {
  // do things with the onchange
});

// use sub to unsubscribe, or to retrieve the current value
```

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,107 @@
...
+    // 1. Literal: Return a static subscription
+    if (typeof value !== 'object' || value === null) {
+      return {
+        value: value as V,
+        onChange: undefined,
```
ditman: Returning `undefined` here forces users to check the Subscription type to see if they can call "onChange" or not. I think this onChange should be a noop function that warns the user that they're subscribing to a key that will never trigger this method instead.

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,107 @@
...
+    return value as V;
+  }
+
+  /**
+   * Subscribes to changes in a DynamicValue.
+   * Returns a Subscription object that provides the current value and allows listening for updates.
+   */
+  subscribeDynamicValue<V>(value: any): Subscription<V> {
+    // 1. Literal: Return a static subscription
+    if (typeof value !== 'object' || value === null) {
```
ditman: Why are being wishy-washy with the type of the `value` object, instead of defining a type or a union? For example, looking at the tests, it seems a type for "value" would be something like: "something that has a "path" string", and looking at the code it may also contain a "call"...

In which case would `value` be not an 'object' or null?

Also, a nice ergonomic improvement here would be to do "that type | string" and assume the user wants to subscribe to a "path" if they pass a raw string in, maybe?

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,107 @@
...
+      return {
+        value: value as V,
```
ditman, line 63: Isn't it weird to return the key as the value? Also, this cast is dangerous, we don't know that the incoming value is a V?

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,107 @@
...
+   * Creates a new, nested DataContext for a child component.
+   * Used by list/template components for their children.
+   */
+  nested(relativePath: string): DataContext {
+    const newPath = this.resolvePath(relativePath);
+    return new DataContext(this.dataModel, newPath);
+  }
+
+  private resolvePath(path: string): string {
+    if (path.startsWith('/')) {
```
ditman, line 93: Add a comment here `// Absolute path, whatever this.path is doesn't matter` or similar :)

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,107 @@
...
+ */
+export class DataContext {
+  /**
+   * @param dataModel The shared DataModel instance.
+   * @param path The absolute path this context is currently pointing to.
+   */
+  constructor(
+    readonly dataModel: DataModel,
+    readonly path: string
+  ) { }
```
ditman, line 16: Damn, this took me a while to parse! This TS idiom is called "Parameter Properties" and it is automatically doing `this.dataModel = dataModel` and `this.path = path`:
https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

renderers/web_core/src/v0_9/state/component-model.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export interface AccessibilityProperties {
+  label?: any;
+  description?: any;
+  [key: string]: any;
+}
+
+export class ComponentModel {
+  private _properties: Record<string, any>;
+  private listeners: Set<ComponentUpdateListener> = new Set();
```
ditman: This `ComponentModel` is very similar to a `DataModel`, but not quite, and it duplicates logic (from the DataModel class):

```
  private data: Record<string, unknown> = {};
  private readonly subscriptions: Map<string, Set<SubscriptionImpl<any>>> = new Map();
```

Would it make sense to have a base "listenable" object to handle the "listeners" part for free, and use it to implement both the ComponentModel and the DataModel?

renderers/web_core/src/v0_9/state/component-model.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export interface AccessibilityProperties {
+  label?: any;
+  description?: any;
+  [key: string]: any;
+}
+
+export class ComponentModel {
+  private _properties: Record<string, any>;
```
ditman, line 4: Why the `_` in this? It's already `private`, do we need the prefix?

renderers/web_core/src/v0_9/state/surface-components-model.ts
```diff
@@ -0,0 +1,62 @@
+import { ComponentModel } from './component-model.js';
+
+export interface ComponentsLifecycleListener {
+  onComponentCreated: (component: ComponentModel) => void;
+  onComponentDeleted?: (componentId: string) => void;
+}
+
+export class SurfaceComponentsModel {
+  private components: Map<string, ComponentModel> = new Map();
+  private listeners: Set<ComponentsLifecycleListener> = new Set();
```
ditman: `listeners` again, add it to the suggestion of having a "listenable" base class/interface!

renderers/web_core/src/v0_9/state/surface-group-model.ts
```diff
@@ -0,0 +1,101 @@
...
+ * The root state model for the A2UI system.
+ * Manages the collection of active surfaces.
+ */
+export class SurfaceGroupModel<T extends CatalogApi> {
+  private surfaces: Map<string, SurfaceModel<T>> = new Map();
+  private surfaceUnsubscribers: Map<string, () => void> = new Map();
+  private lifecycleListeners: Set<SurfaceLifecycleListener<T>> = new Set();
+  private actionListeners: Set<ActionListener> = new Set();
+
+  constructor() {}
```
ditman: I don't think you need to define an empty constructor function

renderers/web_core/src/v0_9/state/surface-group-model.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  removeActionListener(listener: ActionListener): void {
+    this.actionListeners.delete(listener);
+  }
+
+  private async dispatchAction(action: any): Promise<void> {
+    for (const listener of this.actionListeners) {
+      try {
+        await listener(action);
```
ditman: Does this need some sort of timeout so a `listener` can't make the program hang?

renderers/web_core/src/v0_9/state/surface-model.ts
```diff
@@ -0,0 +1,55 @@
...
+  readonly componentsModel: SurfaceComponentsModel;
+  private actionListeners: Set<ActionListener> = new Set();
+
+  constructor(
+    readonly id: string,
+    readonly catalog: T,
+    readonly theme: any = {}
+  ) {
+    this.dataModel = new DataModel({});
+    this.componentsModel = new SurfaceComponentsModel();
```
ditman, line 27: Why are these initialized in the body of the constructor, but the `actionListeners` is initialized inline?

renderers/web_core/src/v0_9/state/surface-model.ts
```diff
@@ -0,0 +1,55 @@
...
+   * Removes an action listener.
+   */
+  removeActionListener(listener: ActionListener): void {
+    this.actionListeners.delete(listener);
+  }
+
+  async dispatchAction(action: any): Promise<void> {
+    for (const listener of this.actionListeners) {
+      try {
+        await listener(action);
```
ditman: (Same comment as above about awaiting a promise that may never be resolved)

renderers/web_core/src/v0_9/index.ts
```diff
@@ -0,0 +1,10 @@
+
+export * from './state/data-model.js';
+export * from './rendering/data-context.js';
```
ditman, line 4: Should each of these be a different library, so you can import from `web_core/v0.9/state` or `web_core/v0.9/rendering`, instead of from a big file? That would also alleviate issues with name collisions down the line?

renderers/web_core/package-lock.json
ditman, line 1: Can we PLEASE remove the `package-lock.json` from this package? Libraries should really not commit their `package-lock.json` file IMO.

renderers/web_core/package.json
```diff
@@ -68,13 +68,23 @@
...
+      "dependencies": [
+        "build"
+      ]
     }
   },
   "author": "Google",
   "license": "Apache-2.0",
   "devDependencies": {
-    "@types/node": "^24.10.1",
+    "@types/node": "^25.2.3",
```
ditman: This is very new, wow, I'm still on 24.something

specification/v0_9/docs/renderer_guide.md
```diff
@@ -0,0 +1,337 @@
...
+  update(newProps: Record<string, any>): void; // Apply incremental updates
+  addUpdateListener(l: ComponentUpdateListener): () => void; // Watch for config changes
+}
+```
+
+#### `DataModel`
+A dedicated store for the surface's application data (the "Model" in MVVM).
+
+```typescript
+class DataModel {
```
ditman, line 276: Would it be nice to have a way of create `SomethingModel extends DataModel` and have type-safe versions of the models that "hide" the string paths and T details from users?

ditman (commented):
Some comments!

renderers/web_core/package.json
jacobsimionato: Let's revert to 24.10.1

renderers/web_core/src/v0_9/rendering/component-context.ts
jacobsimionato: Okay, this is beyond my TS expertise, but Gemini tells me (sorry to be that guy!):


> Ditman's suggestion (this._actionDispatcher = surface.dispatchAction) would actually cause a bug in
> JavaScript/TypeScript due to how this binding works.
> 
> Because dispatchAction is defined as a standard class method on SurfaceModel (i.e. async
> dispatchAction(action: any) { ... }), it relies on the implicit this context when called.
> 
> If you simply copy the reference to the function like this:
> 
> 
> 1 this._actionDispatcher = surface.dispatchAction;
> 
> 
> When you later call this._actionDispatcher(action) inside ComponentContext, the this context inside
> the executing dispatchAction function will be the ComponentContext instance (or undefined in strict
> mode), not the SurfaceModel instance (surface). This will lead to runtime errors if dispatchAction
> tries to access any of its own properties (e.g. this.listeners).
> 
> The current implementation using a lambda:
> 
> 
> 1 this._actionDispatcher = (action) => surface.dispatchAction(action);
> ...creates a closure over surface, guaranteeing that dispatchAction is always invoked with surface as
> its this context.
> 
> 
> An equally valid (but slightly older) alternative that does the exact same thing without the lambda
> would be using .bind():
> 
> 
> 1 this._actionDispatcher = surface.dispatchAction.bind(surface);
> 
> 
> So the lambda is necessary (or .bind() must be used) to preserve the correct this context for the
> method call!

renderers/web_core/src/v0_9/processing/message-processor.ts
jacobsimionato: Good point - done.

renderers/web_core/src/v0_9/processing/message-processor.ts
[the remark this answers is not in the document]
jacobsimionato: True true - I defined a type now.

renderers/web_core/src/v0_9/processing/message-processor.ts
jacobsimionato: Good idea - done.

renderers/web_core/src/v0_9/processing/message-processor.ts
jacobsimionato: Done.

renderers/web_core/src/v0_9/rendering/data-context.ts
jacobsimionato: Yeah!! I actually really like this TS feature!

renderers/web_core/src/v0_9/rendering/data-context.ts
jacobsimionato: The onChange property of a subscription is set by the application code if it wants to listen to the value being updated. So this should always be initialized to undefined by the framework.

Either way, I've removed onChange from here at your suggestion elsewhere!

renderers/web_core/src/v0_9/rendering/data-context.ts
jacobsimionato: Hmmmm I think V is not nullable by default. I think we need to decide how to handle these kinds of "inconsistent data model" errors, e.g. whether to throw or tolerate. I think tolerating as much as possible makes sense.

I filed https://github.com/google/A2UI/issues/725 and added a TODO

renderers/web_core/src/v0_9/rendering/data-context.ts
jacobsimionato: Previously, I designed it this way so that it's easy to adapt a Subscription object to some kind of Observable / Stream / Subject etc. E.g. `toObservable(listener)`. But I agree by itself it's quite awkward, so I followed your suggestion.

renderers/web_core/src/v0_9/rendering/data-context.ts
jacobsimionato: Great point. Update this to use real types.

renderers/web_core/src/v0_9/rendering/data-context.ts
jacobsimionato: In this case, 'value' is a literal value, not a key. The input here conforms to DynamicValue defined in `common_types.json`.

I filed an issue for the cast.

renderers/web_core/src/v0_9/rendering/data-context.ts
jacobsimionato: Added.

renderers/web_core/src/v0_9/processing/message-processor.ts
gspencergoog: I think I can answer that: It comes from Dart naming conventions.

https://dart.dev/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words

Since A2UI is longer than two characters, it gets capitalized like a word.

And this name (while not in Dart) came from the Dart implementation. We could make them different for TypeScript, but we don't have to, and then it seems annoying to have two different capitalizations.

ditman: Yeah, I was aware of that bit of the style guide, I'm just very used to seeing iOS capitalized without following that rule: https://github.com/search?q=repo%3Aflutter%2Fpackages%20iOS&type=code :P

renderers/web_core/src/v0_9/rendering/component-context.ts
ditman: Ah yeah, the tearoff doesn't carry the `surface` in its closure... `.bind(this)` would probably be enough, but it's still creating a new instance of the function anyways.

This point would be moot if we kept an explicit reference to the `surface` in the class, so we could use a method in the prototype, but since it seems we don't need the surface elsewhere, LGTM? :P

renderers/web_core/src/v0_9/rendering/data-context.ts
ditman: Bespoke advice about this is to "throw errors with enough info so the agent can fix them", but I don't think the framework is wired to enable this yet. Filing a TODO SGTM!

renderers/web_core/src/v0_9/rendering/data-context.ts
ditman: Ah yeah, I thought this was so multiple observers could be added to the same Subscription (but that's a bit odd). If we wanted to have a stream, we can provide a different method on `data`, like `data.stream<T>({path})` later?

renderers/web_core/src/v0_9/rendering/data-context.ts
ditman: Ah yeah, I understand now! Thanks for simplifying this API though!

renderers/web_core/src/v0_9/rendering/data-context.ts
ditman: Scaryyyy stuff :P

renderers/web_core/src/v0_9/rendering/data-context.ts
ditman: The thing that threw me off was the "emtpy constructor body" thing, what the hell is thatttt!

renderers/web_core/package-lock.json
ditman: Turns out we can't remove the package-lock.json if we end up needing to use google's build system 😢

renderers/web_core/src/v0_9/schema/common-types.ts
ditman: Are we going to have different schema_types in different parts of the code? I wonder if it'd make sense to have a "schemas" directory to put them all there:

```
schemas
  |- catalog
  -- something_else
```

Maybe? IDK, but now I don't think these should be vended from the spec, unless it vends something that we can use to codegen the ones in TS (and maybe Dart) from a single source of truth.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -0,0 +1,156 @@
...
+      return;
+    }
+  }
+
+  private processCreateSurfaceMessage(message: A2UIMessage): void {
+    const payload = message.createSurface!;
+    const { surfaceId, catalogId, theme } = payload;
+
+    // Find catalog
+    const catalog = this.catalogs.find(c => c.id === catalogId);
```
ditman, line 81: (`this.catalogs` being defined directly in the constructor keeps breaking my brain.)

renderers/web_core/src/v0_9/schema/server-to-client.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from 'zod';
+
+export const A2UIMessageSchema = z.object({
```
ditman: I like that this is called SomethingSomethingSchema. 2 comments:

* Keep the capitalization consistent, if the `A2uiMessageProcessor` uses `A2ui`, this should be `A2uiMessageSchema` (same everywhere)
* If this is a SomethingSomethingSchema, I'd argue that *everything* that is defined using Zod should be a SomethingSomethingSchema as well... It's more verbose, but also more descriptive, and frees the "good names" for the actual objects that we may want to use within the rest of the framework.

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,107 @@
...
+    if ('path' in value && typeof value.path === 'string') {
+      const absolutePath = this.resolvePath(value.path);
+      return this.dataModel.get(absolutePath);
+    }
+
+    // 3. Function Call: { call: "...", args: ... }
+    if ('call' in value) {
+      // TODO: Implement function calls
+      // For now, return as is or undefined
+    }
```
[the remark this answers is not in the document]
ditman: A `console.log('In the future we'll call a function', value.call)` might suffice for this for now.

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -0,0 +1,110 @@
...
+    // Absolute path - no resolution required.
+    if (path.startsWith('/')) {
+      return path;
+    }
+    // Handle specific cases like '.' or empty
+    if (path === '' || path === '.') {
+      return this.path;
+    }
+
+    // Normalize current path (remove trailing slash if exists, unless root)
```
ditman, line 101: Do we want/need to support relative path traversal? Something like:

`"../id"` or similar?

renderers/web_core/src/v0_9/state/data-model.test.ts
```diff
@@ -126,59 +126,61 @@ describe('DataModel', () => {
...
+  it('notifies subscribers on exact match', () => {
+    let called = false;
+    const sub = model.subscribe('/user/name', (val) => {
       assert.strictEqual(val, 'Charlie');
-      done();
-    };
+      called = true;
+    });
     model.set('/user/name', 'Charlie');
+    assert.strictEqual(called, true, 'Callback was never called');
```
ditman, line 136: I like this better!

renderers/web_core/src/v0_9/state/surface-model.ts
jacobsimionato: Fix this

renderers/web_core/src/v0_9/state/surface-group-model.ts
jacobsimionato: This is a good point, but I think it makes sense to give application developers this responsibility rather than us hardcode some timeout. If there were to be a timeout, I think it should be in one place, in MessageProcessor. But then, why not just let application developers add it, in their listening logic?

renderers/web_core/src/v0_9/index.ts
jacobsimionato: This makes sense. Maybe we can figure it out later though?

renderers/web_core/package.json
jacobsimionato: Reverted now.

renderers/web_core/src/v0_9/state/component-model.ts
jacobsimionato: This is required here because we have a public getter/setter called "properties"

renderers/web_core/src/v0_9/schema/common-types.ts
jacobsimionato: I moved things to a "schema" folder, which reflects the structure of the specification. Maybe we can generate this from json schema one day, but for now, I think translating it once per spec revision is fine.

renderers/web_core/package.json
ditman: Still not reverted?

ditman:
> `npm error Invalid: lock file's @types/node@25.2.3 does not satisfy @types/node@24.10.15`

You need to commit and push the generated lockfiles, unfortunately (just rebuild the examples)

specification/v0_9/docs/renderer_guide.md
jacobsimionato: We can't do that though, because the DataModel is an untyped thing. The exact structure is defined at runtime according to `updateDataModel` messages etc.

renderers/web_core/src/v0_9/processing/message-processor.ts
jacobsimionato: We should lobby the Dart team to add this! Then it will be the norm in our brains! :-D

renderers/web_core/src/v0_9/rendering/data-context.ts
jacobsimionato: Ummm that's interesting. It's not in the spec right now, so we don't need it. I'm sure we'll find more edge cases to handle here, but at least we will have things more centralized now so it's not so hard to tighten the implementation.

renderers/web_core/src/v0_9/schema/server-to-client.ts
jacobsimionato: Great point. I updated this to make all the zod schemas be called `SomethingSchema`. And then when we do inferType from it, we just name that as `Something` rather than `SomethingDef`.

Issue #462, opened 2026-01-09
Remove `unsafeCSS`


yuantian-agentspace:
`renderers/lit/src/0.8/ui/styles.ts` uses `unsafeCSS` which is NOT export and NOT allowed in Google3.

ditman:
@yuantian-agentspace there's an incoming PR for this: https://github.com/google/A2UI/issues/462 Do you mind taking a look as well? Thanks!

PR #633, opened 2026-02-17
fix(lit): replace unsafeCSS with CSSStyleSheet for structural styles

iamrajhans:
This PR removes `unsafeCSS` usage from the Lit renderer to fix internal build policy failures reported in issue #462.

Specifically, in `renderers/lit/src/0.8/ui/styles.ts`, `unsafeCSS(Styles.structuralStyles)` was replaced with a safer stylesheet construction flow:
- build a `CSSStyleSheet`
- load `Styles.structuralStyles` using `replaceSync`
- return `[]` as a fallback when `CSSStyleSheet` is unavailable or fails

This preserves behavior (shared structural styles still applied) while eliminating direct `unsafeCSS` usage.

Fixes: #462

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md  


renderers/lit/src/0.8/ui/styles.ts
```diff
@@ -14,7 +14,21 @@
...
 
-import { unsafeCSS } from "lit";
+import type { CSSResultGroup } from "lit";
 import * as Styles from "@a2ui/web_core/styles/index";
 
-export const structuralStyles = unsafeCSS(Styles.structuralStyles);
+const buildStructuralStyles = (): CSSResultGroup => {
+  if (typeof CSSStyleSheet === "undefined") {
+    return [];
+  }
```
ditman: The `CSSStyleSheet()` constructor seems to be very [widely supported](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/CSSStyleSheet), I don't think this check is necessary.

(In case we want to keep this, I'd suggest adding a `console.error` on the cases where we `return []` to let the user know why their stylesheets are not working!)

renderers/lit/src/0.8/ui/styles.ts
```diff
@@ -14,7 +14,21 @@
...
+
+  try {
+    const styleSheet = new CSSStyleSheet();
+    styleSheet.replaceSync(Styles.structuralStyles);
+    return styleSheet;
+  } catch {
```
ditman: In what cases would this fail? Why would we want this try-catch here? (I don't see any "Exceptions" sections in the docs?)

renderers/lit/src/0.8/ui/styles.ts
```diff
@@ -14,7 +14,21 @@
...
+    return [];
+  }
```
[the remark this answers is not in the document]
ditman: This is good advice, and if we were to keep this behavior of returning an empty array, we should also `console.error` in the cases where `CSSStyleSheet()` is not available. (But IMO, we shouldn't)

ditman (commented):
I left some comments, and I think @ava-cassiopeia should have final say, but this looks good to me! Thanks for the fix!

~(I'm also modifying the description of this PR to link it to the issue about removing `unsafeCSS`)~

iamrajhans:
hi @ava-cassiopeia @ditman could you please review, made the required changes

renderers/lit/src/0.8/ui/styles.ts
```diff
@@ -14,7 +14,18 @@
...
-export const structuralStyles = unsafeCSS(Styles.structuralStyles);
+const buildStructuralStyles = (): CSSResultGroup => {
+  try {
+    const styleSheet = new CSSStyleSheet();
+    styleSheet.replaceSync(Styles.structuralStyles);
+    return styleSheet;
+  } catch (e) {
+    console.error('Failed to construct structural styles:', e);
+    return [];
+  }
```
ava-cassiopeia, line 31: I think @ditman's comment still applies here: under what condition do we expect this to fail? And for those, do we really want the application setup to just keep going with an empty set of structural styles?

iamrajhans: The only expected failure mode is non-browser module evaluation, where `CSSStyleSheet` is unavailable (Node/test/SSR). In browser runtime, silently proceeding without structural styles is not acceptable.

I’ll update the behavior to:
- return `[]` only in non-browser contexts (`typeof window === "undefined"`), and
- throw on browser-side stylesheet construction failures (constructor/`replaceSync`) instead of falling back silently.

This preserves test/SSR import stability while ensuring browser runtime does not degrade without explicit failure.

ava-cassiopeia: SGTM, thanks!

PR #644, opened 2026-02-19
Add CallFunctionMessage and functionResponse.

jgindin:
# Description

Enables the server to request a function be invoked on the client. The server can specify whether it wants a response or not.

The new `CallFunctionMessage` can be sent at any time from the server, and is not associated with any surface. The function call leverages the existing `FunctionCall` type, but restricts the arguments and return type to scalars (or arrays of scalars). Specifically, neither data binding nor the use of client side function calls is allowed.

`CallFunctionMessage`s are not associated with any surface.

To ensure the response (including any errors) can be properly associated with the function call, a `callId` must be specified with the `CallFunctionMessage`.

The `error` response has been updated to require _either_ a `surfaceId` _or_ a `callId`.

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/specification/v0_10/json/client_to_server.json
+++ b/specification/v0_10/json/client_to_server.json
@@ -46,28 +46,24 @@
       ]
     },
     "functionResponse": {
-      "allOf": [
-        {
-          "properties": {
-            "functionCallId": {
-              "$ref": "common_types.json#/$defs/CallId",
-              "description": "Unique ID for the instance of this function. MUST be copied verbatim from the function invocation."
-            },
-            "value": {
-              "description": "The return value of the function invocation.",
-              "type": [
-                "string",
-                "number",
-                "boolean",
-                "array",
-                "object",
-                "null"
-              ]
-            }
-          },
-          "required": ["functionCallId", "value"]
+      "properties": {
+        "functionCallId": {
+          "$ref": "common_types.json#/$defs/CallId",
+          "description": "Unique ID for the instance of this function. MUST be copied verbatim from the function invocation."
+        },
+        "value": {
+          "description": "The return value of the function invocation.",
+          "type": [
+            "string",
+            "number",
+            "boolean",
+            "array",
+            "object",
+            "null"
+          ]
         }
-      ],
+      },
+      "required": ["functionCallId", "value"],
       "unevaluatedProperties": false
     },
     "error": {


specification/v0_10/json/client_to_server.json
```diff
@@ -43,6 +45,37 @@
         "context"
       ]
     },
+    "functionResponse": {
+      "allOf": [
+        {
+          "properties": {
+            "callId": {
```
wrenj: maybe functionCallId to be clearer

specification/v0_10/json/client_to_server.json
```diff
@@ -43,6 +45,37 @@
...
+              "$ref": "common_types.json#/$defs/CallId",
+              "description": "Unique ID for the instance of this function. MUST be copied verbatim from the function invocation."
+            },
+            "value": {
+              "description": "The return value of the function invocation.",
+              "oneOf": [
+                { "$ref": "common_types.json#/$defs/ScalarValue" },
```
wrenj: does dict fit in here?

specification/v0_10/json/common_types.json
```diff
@@ -201,6 +237,10 @@
       "type": "object",
       "description": "Invokes a named function on the client.",
       "properties": {
+        "allowRemote": {
```
wrenj: allowRemoteCalling ?

specification/v0_10/json/client_to_server.json
jgindin: It would fit into the 3rd option in this `oneOf`:
```
                {
                  "type": "object",
                  "description": "Zero or more named parameters comprising the response.",
                  "additionalProperties": {
                    "anyOf": [
                      { "$ref": "common_types.json#/$defs/ScalarValue" },
                      { "$ref": "common_types.json#/$defs/ScalarArrayValue" }
                    ]
                  }
                }
```

specification/v0_10/json/common_types.json
```diff
@@ -221,16 +266,12 @@
...
           "description": "The expected return type of the function call.",
-          "enum": [
-            "string",
-            "number",
-            "boolean",
-            "array",
-            "object",
-            "any",
-            "void"
+          "anyOf": [
```
gspencergoog: I realize that this is trying to encourage re-use here, but I feel like conceptually this is a lot more complex for the LLM to reason about.  An enum is easier, and it's unlikely that the list of scalar types will change.  Token-wise, it's probably a wash anyhow.

specification/v0_10/json/common_types.json
```diff
@@ -22,6 +26,38 @@
         }
       }
     },
+    "ScalarValue": {
```
gspencergoog: I think these probably should not be in `common_types.json`, since there's only one instance of their being used, and because of that it contributes to both the complexity and the token count.  Just put them inline in the `server_to_client.json`.

specification/v0_10/json/standard_catalog.json
```diff
@@ -962,6 +992,9 @@
         "call": {
           "const": "openUrl"
         },
+        "callableFrom": {
+          "const": "clientOrRemote"
```
gspencergoog: This seems pretty dangerous to me.   With this, we're saying an agent can open any URL on the client remotely.  I'm not a security expert, but that seems bad, especially in the default "basic" catalog.

specification/v0_10/json/standard_catalog.json
```diff
@@ -656,6 +656,9 @@
         "call": {
           "const": "required"
         },
+        "callableFrom": {
+          "const": "clientOnly"
```
gspencergoog: If the default is "clientOnly", then we shouldn't specify it here (or anywhere else), to save tokens.

specification/v0_10/json/common_types.json
jgindin: `ScalarValue` and `ScalarValueArray` are used in both `server_to_client` and `client_to_server`.

The `ScalarType` and `ScalarArrayType` were only used in one place in `server_to_client`, so I removed them and just inlined the usage.

specification/v0_10/json/common_types.json
jgindin: Done.

specification/v0_10/json/standard_catalog.json
jgindin: Done

specification/v0_10/json/standard_catalog.json
wrenj: clientOnly should be the default IMO

jgindin: I'll make this change.

specification/v0_10/json/common_types.json
gspencergoog: > `ScalarValue` and `ScalarValueArray` are used in both server_to_client and client_to_server.

That represents the first time that `client_to_server.json` refers to `common_types.json` at all, so I guess we'll need to start including/providing it on the client side as well now?  I can see wanting to do this for `CallId`, to have a common definition.

In any case, as the agent sees it, those types are only used once, since client and server schemas don't appear together in a prompt.  I don't see the value in consolidation here: it's not like the list of scalar types is going to change, and it makes it more complex for the agent, and uses more tokens.

specification/v0_10/json/client_to_server.json
```diff
@@ -43,6 +45,37 @@
...
+      "allOf": [
+        {
+          "properties": {
+            "functionCallId": {
+              "$ref": "common_types.json#/$defs/CallId",
+              "description": "Unique ID for the instance of this function. MUST be copied verbatim from the function invocation."
+            },
+            "value": {
+              "description": "The return value of the function invocation.",
+              "oneOf": [
```
jacobsimionato: Why do we need to put these restrictions on the function results at this level? Why not just make this an arbitrary value? Seeing as the Catalog lets you declare the functions that are available, if a client wants or server wants to only allow scalar types, then they can just only support catalogs which include only scalar typed functions, right?

What about we remove the global restriction that remote function calls can only have scalar values, and instead catalogs can choose to define functions with scalar-only inputs?

specification/v0_10/json/server_to_client.json
```diff
@@ -127,6 +128,67 @@
...
+                  "type": "object",
+                  "description": "Arguments passed to the function.",
+                  "additionalProperties": {
+                    "anyOf": [
+                      { "$ref": "common_types.json#/$defs/ScalarValue" },
+                      { "$ref": "common_types.json#/$defs/ScalarValueArray" }
+                    ]
+                  }
+                },
+                "returnType": {
```
jacobsimionato: Isn't this already specified by the function definition? Why should the server send this information to the client at runtime?

specification/v0_10/json/common_types.json
jacobsimionato: Re `client_to_server.json` - that JSON file shouldn't really need to be imported by the client or used for inference, because LLMs don't generate those events anyway. It's just a matter of the client code conforming to that schema in the JSON it generates.

That said, I also like the idea of trying to remove this from common_types.json. If we remove the scalar requirement from the core spec, individual catalogs can just inline the ScalarValue definition etc into their catalog.

jgindin: I've just inlined the types and removed the `ScalarValue` and `ScalarValueArray` altogether.

specification/v0_10/json/server_to_client.json
jgindin: This is "overriding" the `returnType` from `FunctionCall` to limit the types.

jgindin: After chat discussion, I've removed the restrictions.

specification/v0_10/json/client_to_server.json
jgindin: Removed the restrictions.

specification/v0_10/json/common_types.json
```diff
@@ -189,7 +193,7 @@
             {
               "properties": {
                 "returnType": {
-                  "const": "array"
+                  "const": "string[]"
```
gspencergoog: Instead of using TypeScript syntax, could this be a keyword, e.g. `stringArray`?

specification/v0_10/json/server_to_client.json
```diff
@@ -127,6 +127,45 @@
...
+        },
+        "callFunction": {
+          "allOf": [
+            {
+              "$ref": "common_types.json#/$defs/FunctionCall"
+            },
+            {
+              "properties": {
+                "callableFrom": {
+                  "description": "Must be remoteOnly or clientOrRemote for server initiated function calls",
```
gspencergoog: This description overrides the description in `FunctionCall`, so it should probably just be the same description as `FunctionCall` has, since the description is how the LLM will know what it is for.

specification/v0_10/json/common_types.json
jgindin: This actually shouldn't have been changed; the `returnType` in `FunctionCall` doesn't include `string[]` in the list of possible values.

specification/v0_10/json/server_to_client.json
jgindin: done

specification/v0_10/json/client_to_server.json
```diff
@@ -43,6 +45,31 @@
         "context"
       ]
     },
+    "functionResponse": {
+      "allOf": [
```
jacobsimionato: I think `allOf` is not required here and you can just move the properties and required up a level. I think `allOf` is necessary if you want to kind of merge two subschemas together.

jgindin: fixed!

PR #655, opened 2026-02-23
CPP validation library

wrenj:


a2a_agents/cpp/src/validation.cpp
```diff
@@ -0,0 +1,319 @@
...
+    }
+    return refs;
+}
+
+void _validate_component_integrity(const nlohmann::json& components, const RefFieldsMap& ref_fields_map) {
+    std::unordered_set<std::string> ids;
+    
+    for (const auto& comp : components) {
+        auto id_it = comp.find(ID);
+        if (id_it == comp.end() || !id_it->is_string()) continue;
```
gspencergoog: If the id isn't a string, isn't that a validation failure?

There not being any IDs might also be a failure, although that's less clear.  It's probably "valid" to have an empty `updateComponents` message, although I'd argue we should disallow that on efficiency grounds.

a2a_agents/cpp/src/validation.cpp
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        recursion_stack.erase(node_id);
+    };
+
+    if (all_ids.count(ROOT)) {
+        dfs(ROOT);
+    }
+
+    std::unordered_set<std::string> orphans;
```
gspencergoog, line 223: When streaming, it's allowable to send `updateComponents` messages that don't include a root yet.  You can later send one more that includes a root.

Those earlier messages won't validate because they will all be orphans until the root component arrives.

You should probably skip the orphan check (and maybe other topology checks) until you have seen a root component.

wrenj: none of these validation libraries dont work with streaming, ill add a todo

a2a_agents/cpp/src/validation.cpp
wrenj: added a test for it

all of the json schema is validated already

i think this was validated earlier but it gave some confusing error message so i updatred it to be very clear here instead of saying "unrecognized type"

a2a_agents/cpp/src/validation.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+namespace {
+
+const std::regex JSON_POINTER_PATTERN(R"(^(?:\/(?:[^~\/]|~[01])*)*$)");
+const int MAX_GLOBAL_DEPTH = 50;
+const int MAX_FUNC_CALL_DEPTH = 5;
+
+const std::string COMPONENTS = "components";
+const std::string ID = "id";
+const std::string COMPONENT_PROPERTIES = "componentProperties";
```
nan-yu, line 39: While consolidating the validation logic into A2uiValidator, I noticed that the validation in https://github.com/google/A2UI/pull/624 was actually based on an outdated spec. As seen in [this commit](https://github.com/google/A2UI/commit/b87d4aaaac1187c5aa8ca9343fbdb2fb04cd9012#diff-6ed23e33f15457ea66f0e318c2b3f01fcb16137129a0009b30c816c9f158e82c), the schema has evolved, notably replacing `componentProperties` with `component`.

I've updated Python A2uiValidator to support both v0.8 and v0.9 specs. Would you mind updating the CPP version to align with the python validator: https://github.com/google/A2UI/blob/main/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/validator.py?

ditman:
Drive by comment: Can we add a small description to this pull request? Is there an issue that this can be associated with?

nan-yu:
This is replaced by https://github.com/google/A2UI/pull/1254 and https://github.com/google/A2UI/pull/1274.

nan-yu:
> Drive by comment: Can we add a small description to this pull request? Is there an issue that this can be associated with?

Corresponding issue: #688

Issue #650, opened 2026-02-22
remove notion of "custom catalogs" from sdk


wrenj:
Hi Nan

We removed the concept of a custom catalog in our catalog docs https://github.com/google/A2UI/blob/main/docs/catalogs.md, it was causing a huge amount of confusion (eg teams asking if they need to support the basic catalog video player), and we want A2UI to be catalog neutral.

The new setup is we have catalogs, and a2ui team makes a catalog called the basic catalog which isn't special in any way and is on even footing with any other catalog.

Can we update the SDKs to reflect this. I think the the schema manager should just take in a list of catalogs and thats it. So no reference to a basic or a custom catalog in any of our SDKs, they just use catalogs.

PR #705, opened 2026-02-24
Remove notion of "custom catalogs" from agent SDK

nan-yu:
# Description

The basic catalog maintained by the A2UI team has no difference from third-party catalogs.

This PR removes the notion of custom catalogs. Each catalog provided at runtime should be independent and immutable. At build time, catalogs can refer to components from other catalogs. They need to be bundled into a
free-standing one using the `tools/build_catalog/build_catalog.py` script.

Fixes: https://github.com/google/A2UI/issues/650

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
+++ b/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
@@ -19,8 +19,8 @@ import os
 from dataclasses import dataclass, field, replace
 from typing import Any, Dict, List, Optional, TYPE_CHECKING
 
-from .catalog_provider import A2uiCatalogProvider, FileSystemCatalogProvider, BundledCatalogProvider
-from .constants import CATALOG_COMPONENTS_KEY, CATALOG_ID_KEY, BASIC_CATALOG_NAME
+from .catalog_provider import A2uiCatalogProvider, FileSystemCatalogProvider
+from .constants import CATALOG_COMPONENTS_KEY, CATALOG_ID_KEY
 
 
 @dataclass
@@ -40,17 +40,6 @@ class CatalogConfig:
   name: str
   provider: A2uiCatalogProvider
   examples_path: Optional[str] = None
-
-  @classmethod
-  def bundled(
-      cls, version: str, examples_path: Optional[str] = None
-  ) -> "CatalogConfig":
-    """Returns a CatalogConfig for the basic bundled catalog."""
-    return cls(
-        name=BASIC_CATALOG_NAME,
-        provider=BundledCatalogProvider(version),
-        examples_path=examples_path,
-    )
 
   @classmethod
   def from_path(
--- a/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog_provider.py
+++ b/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog_provider.py
@@ -15,23 +15,10 @@
 """Module for providing A2UI catalog schemas and resources."""
 
 import json
-import logging
-import os
-import importlib.resources
 from abc import ABC, abstractmethod
 from json.decoder import JSONDecodeError
-from typing import Any, Dict, Optional
-
-from .constants import (
-    A2UI_ASSET_PACKAGE,
-    BASE_SCHEMA_URL,
-    CATALOG_ID_KEY,
-    CATALOG_SCHEMA_KEY,
-    SPEC_VERSION_MAP,
-    find_repo_root,
-)
-
-ENCODING = "utf-8"
+from typing import Any, Dict
+from .constants import ENCODING
 
 
 class A2uiCatalogProvider(ABC):
@@ -39,16 +26,16 @@ class A2uiCatalogProvider(ABC):
 
   @abstractmethod
   def load(self) -> Dict[str, Any]:
-    """Loads a schema resource.
+    """Loads a catalog definition.
 
     Returns:
-      The loaded schema as a dictionary.
+      The loaded catalog as a dictionary.
     """
     pass
 
 
 class FileSystemCatalogProvider(A2uiCatalogProvider):
-  """Loads schemas from the local filesystem."""
+  """Loads catalog definition from the local filesystem."""
 
   def __init__(self, path: str):
     self.path = path
@@ -59,84 +46,3 @@ class FileSystemCatalogProvider(A2uiCatalogProvider):
         return json.load(f)
     except (FileNotFoundError, JSONDecodeError) as e:
       raise IOError(f"Could not load schema from {self.path}: {e}") from e
-
-
-def load_from_bundled_resource(version: str, resource_key: str) -> Dict[str, Any]:
-  """Loads a schema resource from bundled package resources."""
-  spec_map = SPEC_VERSION_MAP.get(version)
-  if not spec_map:
-    raise ValueError(f"Unknown A2UI version: {version}")
-
-  if resource_key not in spec_map:
-    return None
-
-  rel_path = spec_map[resource_key]
-  filename = os.path.basename(rel_path)
-
-  # 1. Try to load from the bundled package resources.
-  try:
-    traversable = importlib.resources.files(A2UI_ASSET_PACKAGE)
-    traversable = traversable.joinpath(version).joinpath(filename)
-    with traversable.open("r", encoding=ENCODING) as f:
-      return json.load(f)
-  except Exception as e:
-    logging.debug("Could not load '%s' from package resources: %s", filename, e)
-
-  # 2. Fallback to local assets
-  # This handles cases where assets might be present in src but not installed
-  try:
-    potential_path = os.path.abspath(
-        os.path.join(
-            os.path.dirname(__file__),
-            "..",
-            "assets",
-            version,
-            filename,
-        )
-    )
-    if os.path.exists(potential_path):
-      provider = FileSystemCatalogProvider(potential_path)
-      return provider.load()
-  except Exception as e:
-    logging.debug("Could not load schema '%s' from local assets: %s", filename, e)
-
-  # 3. Fallback: Source Repository (specification/...)
-  # This handles cases where we are running directly from source tree
-  # And assets are not yet copied to src/a2ui/assets
-  # manager.py is at a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py
-  # Dynamically find repo root by looking for "specification" directory
-  try:
-    repo_root = find_repo_root(os.path.dirname(__file__))
-    if repo_root:
-      source_path = os.path.join(repo_root, rel_path)
-      if os.path.exists(source_path):
-        provider = FileSystemCatalogProvider(source_path)
-        return provider.load()
-  except Exception as e:
-    logging.debug("Could not load schema from source repo: %s", e)
-
-  raise IOError(f"Could not load schema {filename} for version {version}")
-
-
-class BundledCatalogProvider(A2uiCatalogProvider):
-  """Loads schemas from bundled package resources with fallbacks."""
-
-  def __init__(self, version: str):
-    self.version = version
-
-  def load(self) -> Dict[str, Any]:
-
-    resource = load_from_bundled_resource(self.version, CATALOG_SCHEMA_KEY)
-
-    # Post-load processing for catalogs
-    if CATALOG_ID_KEY not in resource:
-      spec_map = SPEC_VERSION_MAP.get(self.version)
-      rel_path = spec_map[CATALOG_SCHEMA_KEY]
-      # Strip the `json/` part from the catalog file path for the ID.
-      catalog_file = rel_path.replace("/json/", "/")
-      resource[CATALOG_ID_KEY] = BASE_SCHEMA_URL + catalog_file
-
-    if "$schema" not in resource:
-      resource["$schema"] = "https://json-schema.org/draft/2020-12/schema"
-
-    return resource
--- a/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py
+++ b/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py
@@ -19,7 +19,7 @@ import os
 import importlib.resources
 from typing import List, Dict, Any, Optional, Callable
 from dataclasses import dataclass, field
-from .catalog_provider import BundledCatalogProvider, load_from_bundled_resource
+from .utils import load_from_bundled_resource
 from ..inference_strategy import InferenceStrategy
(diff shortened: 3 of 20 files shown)


a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py
```diff
@@ -255,37 +241,35 @@ def _determine_catalog(
...
+      return self._supported_catalogs[0]
 
-    for scid in supported_catalog_ids:
-      if scid in self._supported_catalogs:
-        # Return the first supported catalog.
-        return self._supported_catalogs[scid]
+    for cscid in client_supported_catalog_ids:
+      for catalog in self._supported_catalogs:
+        if catalog.catalog_id == cscid:
+          return catalog
```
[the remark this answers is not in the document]
nan-yu: done

yourkarma6788:
WHY IS THIS ILLEGAL API STILL OPEN THAT I DID NOT AUTHORIZE. AND NOR DID I GIVE CONSENT TO.  I REQUESTED A PULL REQUEST A LO G TIME AGO ....

tools/build_catalog/README.md
jacobsimionato, line 1: Thanks for fixing this!!

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
```diff
@@ -28,13 +28,35 @@
...
   name: str
   catalog_path: str
   examples_path: Optional[str] = None
 
+  @classmethod
+  def for_basic_catalog(cls, examples_path: Optional[str] = None) -> "CatalogConfig":
+    """Returns a CatalogConfig for the basic catalog."""
+    return cls(
+        name=BASIC_CATALOG_NAME,
+        catalog_path="",  # basic catalog doesn't need a path from the user
```
jacobsimionato: Why doesn't this need a path? Otherwise, the core code won't know where to load it from, right?

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
```diff
@@ -28,13 +28,35 @@
...
+    catalog_path: The path to the catalog JSON file.
+    examples_path: The path to the examples directory.
+  """
 
   name: str
   catalog_path: str
   examples_path: Optional[str] = None
 
+  @classmethod
+  def for_basic_catalog(cls, examples_path: Optional[str] = None) -> "CatalogConfig":
```
jacobsimionato: Can we move this to a top-level function in a different file, perhaps even a different folder, e.g. inference/basic_catalog or something?

That makes the conceptual separation clear, and avoids us accidentally sprinkling basic catalog assumptions into the core library. Ideally, it would be possible for us to publish basic catalog support as a completely separate library which depends on the core inference library.

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py
```diff
@@ -172,11 +169,7 @@ def _load_schemas(
     )
 
     # Process basic catalog
-    basic_catalog_schema = self._apply_modifiers(
-        _load_basic_component(version, CATALOG_SCHEMA_KEY)
-    )
-    if not basic_catalog_schema:
-      basic_catalog_schema = {}
+    basic_catalog_schema = _load_basic_component(version, CATALOG_SCHEMA_KEY)
```
jacobsimionato: Ideally, avoid having this special cased logic. I wonder if we can publish the basic catalog schema as a JSON file bundled with the library?

Or perhaps the parameter of CatalogConfig could actually just be a "CatalogProvider" which could either be a FileReadingCatalogProvider which reads a JSON file on disk, or just an InlineCatalogProvider which returns a catalog baked into Python code as a string constant (so it's easy for us to create for the basic catalog).

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
```diff
@@ -28,13 +28,35 @@
...
+    """Returns a CatalogConfig for the basic catalog."""
+    return cls(
+        name=BASIC_CATALOG_NAME,
+        catalog_path="",  # basic catalog doesn't need a path from the user
+        examples_path=examples_path,
```
wrenj, line 52: maybe just examples_paths=

versus doing an ls of a directory? what if there .git file in that directory?

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py
```diff
@@ -190,44 +183,36 @@ def _load_schemas(
...
+      self._supported_catalogs.append(catalog)
+      self._catalog_example_paths[catalog.catalog_id] = config.examples_path
 
   def _determine_catalog(
       self, client_ui_capabilities: Optional[dict[str, Any]] = None
   ) -> A2uiCatalog:
     """Determines the catalog to use based on supported catalog IDs.
 
-    If neither inline catalogs nor supported catalog IDs are provided, the basic catalog is used.
+    If neither inline catalogs nor client-supported catalog IDs are provided, the first agent-supported catalog is used.
```
wrenj: is it first in the client_capabitilities or first in the supported_catalogs?

samples/agent/adk/contact_lookup/agent.py
```diff
@@ -46,7 +46,12 @@ def __init__(self, base_url: str, use_ui: bool = False):
     self.base_url = base_url
     self.use_ui = use_ui
     self._schema_manager = (
-        A2uiSchemaManager("0.8", basic_examples_path="examples") if use_ui else None
+        A2uiSchemaManager(
+            "0.8",
```
wrenj: maybe make some consts here?

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
wrenj: +1 on this, then we can run the basic catalog through the sdk like any other catalog

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
nan-yu: The basic catalog and other core schemas (server-to-client and common types) are bundled directly within the SDK. [pack_specs_hook.py](https://github.com/nan-yu/A2UI/blob/268c25f0507a61d6c14b3e5b158e2ac47d1c56cf/a2a_agents/python/a2ui_agent/pack_specs_hook.py) packs the specs into the package assets. The SDK then [loads them as package resources](https://github.com/nan-yu/A2UI/blob/268c25f0507a61d6c14b3e5b158e2ac47d1c56cf/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py#L43). We only require a path when a user is loading a custom catalog from their own filesystem.

nan-yu: I refactored the CatalogConfig with 2 class methods.
- CatalogConfig.bundled(version, examples_path): load the bundled (basic) catalog with examples.
- CatalogConfig.from_path(name, catalog_path, examples_path): load the user catalog with examples.

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
nan-yu: This part is refactored with 2 class methods in the CatalogConfig:
- CatalogConfig.bundled(version, examples_path): load the bundled (basic) catalog with examples.
- CatalogConfig.from_path(name, catalog_path, examples_path): load the user catalog with examples.

I this way, the basic catalog can be loaded similar as other catalogs.

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py
nan-yu: The basic catalog schema was bundled within the library via [pack_specs_hook.py](https://github.com/nan-yu/A2UI/blob/268c25f0507a61d6c14b3e5b158e2ac47d1c56cf/a2a_agents/python/a2ui_agent/pack_specs_hook.py), and will be published with the library.

I had 2 loaders, [FileSystemLoader](https://github.com/nan-yu/A2UI/blob/268c25f0507a61d6c14b3e5b158e2ac47d1c56cf/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/loader.py#L35C7-L35C23) and [PackageLoader](https://github.com/nan-yu/A2UI/blob/268c25f0507a61d6c14b3e5b158e2ac47d1c56cf/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/loader.py#L51) to load the bundled catalog and user-provided catalog accordingly. The `_load_basic_component` was a utility function that invokes the `PackageLoader`. Maybe it was not clear enough.

I refactored to make the `CatalogConfig` to take a `CatalogProvider`, and use the provider to load the catalog.
For bundled catalog, it is `BundledCatalogProvider`, and for user-provided catalog, it is `FileSystemCatalogConfig`. Hopefully that clarifies the loading logic.

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
nan-yu: The `load_examples` method only reads files ending with `.json`: https://github.com/nan-yu/A2UI/blob/268c25f0507a61d6c14b3e5b158e2ac47d1c56cf/a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py#L173.

There might be multiple JSON example files, for example, https://github.com/nan-yu/A2UI/tree/268c25f0507a61d6c14b3e5b158e2ac47d1c56cf/samples/agent/adk/contact_lookup/examples.

Passing a directory seems simple enough.

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py
nan-yu: It is the first in the client_capabilities, which is also supported by the agent. Reworded to the following:
```
Selects the component catalog for the prompt based on client capabilities.

    Selection priority:
    1. First inline catalog if provided (and accepted by the agent).
    2. First client-supported catalog ID that is also supported by the agent.
    3. Fallback to the first agent-supported catalog (usually the bundled catalog).

    Args:
      client_ui_capabilities: A dictionary of client UI capabilities, containing
        inline catalogs and client-supported catalog IDs.

    Returns:
      The resolved A2uiCatalog.
    Raises:
      ValueError: If capabilities are ambiguous (both inline_catalogs and supported_catalog_ids are provided), if inline
        catalogs are sent but not accepted, or if no mutually supported catalog is found.
```

samples/agent/adk/contact_lookup/agent.py
nan-yu: done

nan-yu:
All comments are addressed. Ready for another round of review. Thanks @jacobsimionato @wrenj

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog_provider.py
```diff
@@ -0,0 +1,142 @@
...
+
+ENCODING = "utf-8"
+
+
+class A2uiCatalogProvider(ABC):
+  """Abstract base class for providing A2UI schemas and catalogs."""
+
+  @abstractmethod
+  def load(self) -> Dict[str, Any]:
+    """Loads a schema resource.
```
jacobsimionato: "schema" -> "catalog definition"

The catalog definition isn't exactly a schema, it's a JSON object with some fields, most of which happen to be JSON schemas. See
https://github.com/google/A2UI/blob/5db8030ece57444576ef2b62c000165dc713dd6b/specification/v0_9/json/a2ui_client_capabilities.json#L62

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
jacobsimionato: Thanks for iterating!!

I'd still much prefer moving all references to the basic catalog to a separate `basic_catalog` folder and inverting the dependency structure, so that the basic catalog support depends on the core SDK. I can see how what you have is a super ergonomic way to access the basic catalog, but that's almost part of the problem, because it privileges the basic catalog above other catalogs. We have had a similar issue in the client SDKs and it's been very difficult to unwind, so that's why I'm so sensitive to it!

E.g.
```py
BasicCatalog.getConfig(version, example_paths)
```

or

```py
CatalogConfig(BasicCatalog.provider(version), example_paths)
```

I don't feel strongly about the exact API, as long as the basic catalog stuff is in a separate folder which depends on the core SDK, rather than the other way around.

Rationale:
- Avoid giving developers the idea that they should use the basic catalog if they are using A2UI
- Avoid us accidentally baking in more basic catalog assumptions. It's easy to spot when depend on it if it's in a different folder.
- Allow us to release basic catalog as a separate package if we want
- Provide an example for other catalog developers for how to expose their catalog programmatically in the Python SDK. E.g. if there is a Material Catalog, they might want to implement a package to expose it ergonomically in Python. If you provide a separate basic catalog implementation with programmatic APIs that build on top of the core APIs, they can trivially follow your example.

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog_provider.py
nan-yu: done

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/catalog.py
nan-yu: Thanks for the explanation. That makes a lot of sense.

I've refactored the SDK to decouple basic catalog support from the core library:
- Moved the basic_catalog loading and instantiation to a separate `basic_catalog` folder.
- Inverted the dependency to make `basic_catalog` depend on the core SDK.
- Updated `pack_specs_hook.py` to generalized module loading. Basic catalog is currently bundled into the SDK the same way as other schemas (server-to-client and common-types), but it is modularized to be easily replaced with other approach if the basic catalog is moved outside of the specification directory.

a2a_agents/python/a2ui_agent/src/a2ui/inference/schema/manager.py
nan-yu: It is resolved as part of https://github.com/google/A2UI/pull/705#discussion_r2865879130.

Issue #208, opened 2025-12-05
Text being clipped with rounded corners in the Lit renderer


jacobsimionato:
We're seeing weird clipping effects. Perhaps there is a corner radius being set incorrectly somewhere?

This is affecting internal clients right now.

<img width="644" height="161" alt="Image" src="https://github.com/user-attachments/assets/1a156a79-6eab-4f30-b9d6-aa0557b0e4b8" />

selamw1:
Raised PR [#710](https://github.com/google/A2UI/pull/710) to fix this issue.

polina-c:
This issue should be considered for a P1 priority together with other P2 issues in the next planning cycle after we burn through the triage queue.

PR #710, opened 2026-02-24
Remove overflow: hidden from border-radius utility classes

selamw1:
It fixes issue [#208](https://github.com/google/A2UI/issues/208)

This PR resolves an issue where text elements using the Lit renderer were unexpectedly clipped when rendered with rounded corners.

The border-br-* utility classes in @a2ui/web_core (which handle border-radius) were indiscriminately applying overflow: hidden. This caused clipping for content like line-height ascenders/descenders in text blocks. 

This commit removes the overflow: hidden rule from the border-radius class generator, fixing the clipping while maintaining the intended rounded corners.


renderers/web_core/package.json
ava-cassiopeia, line 1: We'll be adding screenshot tests to more robustly test this in some future changes, but in the meantime, can you add this new `test` command to [.github/workflows/web_build_and_test.yml](https://github.com/google/A2UI/blob/main/.github/workflows/web_build_and_test.yml) so that these get executed automatically?

ava-cassiopeia (changes requested):

renderers/web_core/package.json
selamw1: Test command added.

ava-cassiopeia:
I think you'll need to revert your changes to the package-lock file to get the checks to pass.

selamw1:
Done, reverted the lock file and squashed all changes into one commit.

ava-cassiopeia (approved):
Great, thanks for cleaning that up!

ava-cassiopeia (changes requested):
Actually, looks like there's one merge conflict to resolve, and then I think this is good to merge.

ava-cassiopeia (changes requested):
Actually, looks like there's one merge conflict to resolve, and then I think this is good to merge.

ava-cassiopeia (changes requested):
Actually, looks like there's one merge conflict to resolve, and then I think this is good to merge.

selamw1:
Thanks, conflict resolved.

ava-cassiopeia (changes requested):
Oh weird GitHub sent my request for changes multiple times, sorry about that!

Anyway, it looks like there's a check failing, it might be a build order-of-operations issue from first glance.

selamw1:
CI failure fixed.

selamw1:
Adjusted the test script to use find and xargs to fix the globbing issues in the CI runner. Tests are discovering and passing locally now.

renderers/web_core/package.json
```diff
@@ -40,9 +40,16 @@
...
+    "copy-spec": "wireit",
     "test": "wireit"
   },
   "wireit": {
+    "test": {
+      "command": "find dist -name '*.test.js' | xargs node --test --enable-source-maps --test-reporter spec",
+      "dependencies": [
+        "build"
+      ]
+    },
```
ava-cassiopeia, line 78: Was the previous command not working here?

selamw1: Yes, the previous command failed to find the `test.js` file. I have added the missing license header to `border.test.ts`. Since the tests passed locally, it required multiple commits to successfully pass the remote tests.

renderers/web_core/src/v0_8/styles/border.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import assert from 'node:assert';
+import { border } from './border.js';
+
+test('Border Styles', async (t) => {
+  await t.test('border-radius classes should not apply overflow: hidden to avoid text clipping', () => {
+    // The issue reported clipping of text when using rounded corners because
+    // .border-br-* generated classes also blindly applied overflow: hidden.
+    
+    // Check if the generated CSS contains overflow: hidden for border-br classes
+    const match = border.match(/\.border-br-\w+ \{[^}]*overflow:\s*hidden/);
```
ditman, line 27:
```suggestion
    const match = border.match(/\.border-br-\d+ \{[^}]*overflow:\s*hidden/);
```

It's only numbers, right?

renderers/web_core/src/v0_8/styles/border.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    assert.strictEqual(match, null, 'Found overflow: hidden applied within a border-br-* class');
+    
+    // Check that we DO have the expected border-radius applied
+    const hasBorderRadius = border.match(/\.border-br-1 \{ border-radius: [^;]+; \}/);
```
ditman: This is a doozy, it's only testing that border-br-1 has *any* border radius, but there's other 24 border-br-* classes that we're not testing. Anyway, this assertion is pretty extra, I think we only need the one above where we ensure that there's no border-br-X that has an overflow: hidden!

ditman (approved):
I think @ava-cassiopeia liked this at an earlier revision.

Thanks for the fix!

renderers/web_core/src/v0_8/styles/border.test.ts
selamw1: Actually, we also generate `.border-br-50pc` in `border.ts`, so keeping `\w+` is necessary to ensure we don't skip testing the percentage-based class.

renderers/web_core/src/v0_8/styles/border.test.ts
selamw1: Good catch. I have removed the extra assertion as verifying the absence of overflow: hidden is the core requirement.

ditman:
@ava-cassiopeia PTAL when you have a minute! :)

ditman:
I'm closing this one, I think v0.9 has moved on from these. Thanks for the contribution @selamw1!

PR #747, opened 2026-03-02
Resolve some accessibility issues.

jgindin:
# Description

1. Image needs to have an `altText`. If not specified, it defaults to an empty string, which indicates to a screen reader to skip the image entirely, assuming it is decorative.
2. Icons need to ensure they are not in the tab order.

## Pre-launch Checklist

- [x ] I signed the [CLA].
- [ ]x I read the [Style Guide].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/angular/src/lib/catalog/image.ts
+++ b/renderers/angular/src/lib/catalog/image.ts
@@ -57,7 +57,7 @@ export class Image extends DynamicComponent {
   protected readonly resolvedUrl = computed(() => this.resolvePrimitive(this.url()));
   protected readonly resolvedAltText = computed(() => {
     const raw = this.altText();
-    return (raw ? this.resolvePrimitive(raw) : null) ?? '';
+    return raw ? this.resolvePrimitive(raw) : '';
   });
 
   protected classes = computed(() => {


renderers/angular/src/lib/catalog/image.ts
```diff
@@ -40,19 +40,25 @@ import { DynamicComponent } from '../rendering/dynamic-component';
   `,
   template: `
     @let resolvedUrl = this.resolvedUrl();
+    @let resolvedAltText = this.resolvedAltText();
 
     @if (resolvedUrl) {
       <section [class]="classes()" [style]="theme.additionalStyles?.Image">
-        <img [src]="resolvedUrl" />
+        <img [src]="resolvedUrl" [alt]="resolvedAltText || ''" />
```
sugoi-yuzuru: The fallback ` || ''` seems redundant at this point.

renderers/web_core/src/v0_8/types/components.ts
```diff
@@ -59,6 +59,7 @@ export interface Image {
     | "largeFeature"
     | "header";
   fit?: "contain" | "cover" | "fill" | "none" | "scale-down";
+  altText: string;
```
sugoi-yuzuru: Were you intending to make this optional? If so, would you not need a tailing `?`?

renderers/angular/src/lib/catalog/icon.ts
```diff
@@ -20,6 +20,10 @@ import * as Primitives from '@a2ui/web_core/types/primitives';
 
 @Component({
   selector: 'a2ui-icon',
+  host: {
```
sugoi-yuzuru, line 23: What a11y issue does this address?

sugoi-yuzuru (commented):
Thanks!

renderers/angular/src/lib/catalog/icon.ts
jgindin: Apparently, the icons were being included in the tab order, so this makes sure they're not.

renderers/web_core/src/v0_8/types/components.ts
jgindin: I wasn't, but it seems the suggestions are to make it optional, so I've updated.

renderers/angular/src/lib/catalog/default.ts
```diff
@@ -61,7 +61,7 @@ export const DEFAULT_CATALOG: Catalog = {
       return [
         inputBinding('url', () => properties.url),
         inputBinding('usageHint', () => properties.usageHint),
-        inputBinding('altText', () => properties.altText),
+        inputBinding('altText', () => properties.altText ?? null),
```
sugoi-yuzuru, line 64: `| | null` seems to be the approach more aligned with the rest of the file to set default value here.

renderers/angular/src/lib/catalog/image.ts
```diff
@@ -57,7 +57,7 @@ export class Image extends DynamicComponent {
   protected readonly resolvedUrl = computed(() => this.resolvePrimitive(this.url()));
   protected readonly resolvedAltText = computed(() => {
     const raw = this.altText();
-    return raw ? this.resolvePrimitive(raw) : '';
+    return (raw ? this.resolvePrimitive(raw) : null) ?? '';
```
sugoi-yuzuru: This could be as simple as 

`return `this.resolvePrimitive(this.altText()) ?? '';`

renderers/angular/src/lib/catalog/default.ts
jgindin: It seems like everywhere except 2 places, we're using `??`

PR #753, opened 2026-03-03
Move Composer app to the `tools/` folder

jacobsimionato:
https://github.com/google/A2UI/pull/365 added the `Composer` to the root of the repository.

It is a tool, so it should be in the `tools/` subfolder.


nan-yu (approved):
/lgtm

PR #768, opened 2026-03-04
Fix publishing workflow and update scripts for web packages

jacobsimionato:
These are to use temporarily while set up the automated pipeline properly.


renderers/web_core/PUBLISHING.md
ditman, line 1: Can't wait to deprecate this flow :P

renderers/markdown/markdown-it/prepare-publish.mjs
```diff
@@ -0,0 +1,85 @@
+import { readFileSync, writeFileSync, copyFileSync, mkdirSync, existsSync } from 'fs';
```
gspencergoog, line 17: Copyright?

gspencergoog (approved):
Thanks for writing this up!

renderers/angular/postprocess-build.mjs
```diff
@@ -38,6 +38,12 @@ if (!packageJson.dependencies['@a2ui/web_core']) {
 }
 
 packageJson.dependencies['@a2ui/web_core'] = '^' + coreVersion;
```
ditman, line 40: I think we should stop post-processing the `web_core` dependency, and instead, take a dependency on the version of the package that's published on npm. That way, these build script could be simplified quite a lot! WDYT?

Then, we'd need some instructions on how to develop locally to be able to modify multiple packages at once, with workspaces, or npm link (?)

renderers/markdown/markdown-it/prepare-publish.mjs
```diff
@@ -0,0 +1,85 @@
...
+import { join } from 'path';
+
+// This script prepares the markdown-it package for publishing by:
+// 1. Copying package.json to dist/
+// 2. Updating @a2ui/web_core dependency from 'file:...' to the actual version
+// 3. Adjusting paths in package.json (main, types, exports) to be relative to dist/
+
+const dirname = import.meta.dirname;
+const corePkgPath = join(dirname, '../../web_core/package.json');
```
ditman, line 26: This package is going to be fairly stable, and it most likely won't need a lot of changes to web_core. Can you please update the package.json to take a dependency on the published version of web_core? Maybe we can remove most of this file?

renderers/web_core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/web_core",
-  "version": "0.8.2",
+  "version": "0.8.0",
```
ditman: Would it make sense to call this fix 0.8.1 and publish the package, so the package.json version matches what's published (not retroactively? :P)

renderers/angular/postprocess-build.mjs
ava-cassiopeia: my 2¢ is a npm link workflow is probably the best way to approach this, whenever we document that.

yourkarma6788:
YOUR BEST INTEREST IS TO REMOVE ALL THE ILLEGAL APIS THAT I DID NOT GIVE CONSENT TO FOR THIRD PARTY ACCESS, BLOCKLISTS, NOR RESTRICTIONS. I DOD NOT AUTHORIZE NOR DID I GIVE CONSENT!!!!

renderers/angular/postprocess-build.mjs
jacobsimionato: Sorry we discussed this weeks ago and I dropped the ball. As we said in the discussion today, let's stick with the hacky scripts for now, then switch to workspaces or completely separate repos later.

renderers/markdown/markdown-it/prepare-publish.mjs
jacobsimionato: As discussed today, how about we stick with the hacky scripts for now so that *everything* developed locally on the repo is at head?

renderers/web_core/package.json
jacobsimionato: Mmmm this is probably a good idea. Though we need to think this all through and probably switch to a different versioning system anyway.

renderers/markdown/markdown-it/prepare-publish.mjs
jacobsimionato: Done.

PR #772, opened 2026-03-04
Modify components to allow application of more expressive styling.

jgindin:
# Description

A few components needed some tweaks in order to apply a new theme.

There is no change to these components when using existing themes.

## Pre-launch Checklist

- [x ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/angular/src/lib/catalog/list.ts
+++ b/renderers/angular/src/lib/catalog/list.ts
@@ -31,18 +31,19 @@ import { Renderer } from '../rendering/renderer';
       display: block;
       flex: var(--weight);
       min-height: 0;
-      overflow: auto;
     }
 
     :host([direction='vertical']) section {
       display: flex;
       flex-direction: column;
+      max-height: 100%;
+      overflow-y: auto;
     }
 
     :host([direction='horizontal']) section {
       display: flex;
       max-width: 100%;
-      overflow-x: scroll;
+      overflow-x: auto;
       overflow-y: hidden;
       scrollbar-width: none;
     }
--- a/renderers/angular/src/lib/catalog/modal.ts
+++ b/renderers/angular/src/lib/catalog/modal.ts
@@ -61,7 +61,9 @@ import { Renderer } from '../rendering';
     dialog {
       padding: 0;
       border: none;
-      background: none;
+      background: transparent;
+      box-shadow: none;
+      overflow: visible;
 
       & section {
         & .controls {


renderers/angular/src/lib/catalog/list.ts
```diff
@@ -35,7 +35,8 @@ import { Renderer } from '../rendering/renderer';
     }
 
     :host([direction='vertical']) section {
-      display: grid;
+      display: flex;
```
ditman, line 37: I think the new style makes more sense, I guess `display:grid` by default behaves like a column flex.

renderers/angular/src/lib/catalog/list.ts
```diff
@@ -44,17 +45,21 @@ import { Renderer } from '../rendering/renderer';
...
       scrollbar-width: none;
+    }
 
-      > ::slotted(*) {
-        flex: 1 0 fit-content;
-        max-width: min(80%, 400px);
-      }
+    .a2ui-list-item {
+      display: flex;
+      align-items: center;
```
ditman: Odd that we want any centering here?

renderers/angular/src/lib/catalog/slider.ts
```diff
@@ -62,13 +69,32 @@ export class Slider extends DynamicComponent {
...
   protected handleInput(event: Event) {
     const path = this.value()?.path;
 
-    if (!(event.target instanceof HTMLInputElement) || !path) {
+    if (!(event.target instanceof HTMLInputElement)) {
       return;
     }
 
-    this.processor.setData(this.component(), path, event.target.valueAsNumber, this.surfaceId());
+    const val = event.target.valueAsNumber;
```
ditman: Prefer not using abbreviations to make the code more readable, so in this case:

```suggestion
    const value = event.target.valueAsNumber;
```

renderers/angular/src/lib/catalog/slider.ts
```diff
@@ -62,13 +69,32 @@ export class Slider extends DynamicComponent {
...
+    event.target.style.setProperty('--slider-percent', percent + '%');
+
+    if (path) {
+      this.processor.setData(this.component(), path, val, this.surfaceId());
+    }
+  }
+
+  private computePercentage(val: number): number {
+    const min = this.minValue() ?? 0;
+    const max = this.maxValue() ?? 100;
```
ditman, line 85: Is this component enforcing that both min and max are being passed? It'd be a little weird for these defaults if only one of them is passed but not the other?

renderers/angular/src/lib/catalog/tabs.ts
```diff
@@ -29,12 +28,14 @@ import * as Types from '@a2ui/web_core/types/types';
     @let selectedIndex = this.selectedIndex();
 
     <section [class]="theme.components.Tabs.container" [style]="theme.additionalStyles?.Tabs">
-      <div [class]="theme.components.Tabs.element">
+      <div class="a2ui-tabs-container" [class]="theme.components.Tabs.element">
```
ditman: Would it make sense to add the `a2ui-tabs-container` class to the theme? Is there any reason it needs to be "hardcoded" in the element directly? (similar question for a2ui-tab-button below)

renderers/angular/src/lib/catalog/tabs.ts
```diff
@@ -29,12 +28,14 @@ import * as Types from '@a2ui/web_core/types/types';
...
         @for (tab of tabs; track tab) {
           <button
+            class="a2ui-tab-button"
             (click)="this.selectedIndex.set($index)"
             [disabled]="selectedIndex === $index"
-            [class]="buttonClasses()[selectedIndex]"
+            [class]="theme.components.Tabs.controls.all"
+            [class.selected]="selectedIndex === $index"
```
ditman: In the previous version of the code, the whole "theme....Tabs.controls.selected" array (?) was being added on selected... I like this simplification, but is this breaking the theming ability for the "selected" element? (Imagine that you want to add more than one class to selected items)

renderers/angular/src/lib/catalog/slider.ts
jgindin: Both the min & max are marked as `required`, but...they can also be undefined. That hasn't changed in this CL.

renderers/angular/src/lib/catalog/tabs.ts
jgindin: Yes! I was able to remove this class and push it into the new theme I'm creating.

renderers/angular/src/lib/catalog/tabs.ts
jgindin: After the other tab-related changes, this particular change has been reverted and we're again using the `buttonClasses` signal.

renderers/angular/src/lib/catalog/list.ts
jgindin: Yeah, I don't think this is necessary.

ditman (approved):
Looks good to me! Thanks for addressing my comments!

Issue #778, opened 2026-03-05
Extract a parser in the agent SDK to parse LLM response


nan-yu:
Currently, the LLM output contains both conversational text and A2UI JSON payloads. The parsing logic is tightly coupled or duplicated across the samples. The task is to move the logic to the agent SDK with a centralized and robust parser.

PR #780, opened 2026-03-05
Implement A2UI parser to extract text parts and A2UI JSON parts separately

nan-yu:
# Description

It hides the A2UI delimiter as implementation details by injecting the LLM response format (text part and JSON part separated by the delimiter) into the system prompt and extracting them with a parser.

- Tested: The orchestrator sample runs successfully from end-to-end. 
- Fixes: https://github.com/google/A2UI/issues/778
- TODO: 
  - Replace the A2UI delimiter with XML/HTML style tags
  - Support multiple pairs of text part and A2UI JSON part

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md


agent_sdks/python/src/a2ui/core/schema/constants.py
```diff
@@ -45,3 +45,13 @@
 SPECIFICATION_DIR = "specification"
 
 ENCODING = "utf-8"
+
+A2UI_DELIMITER = "---a2ui_JSON---"
```
gspencergoog, line 49: I think it would be better to try pseudo-HTML tags:

````txt
<a2ui_json>
```json
(a2ui content)
```
</a2ui_json>
````

Because 1) Claude really likes them, and Gemini understands them, and 2) they have a clear start and end, rather than just a start, and it  then allows for multiple A2UI blocks in a single stream.

gspencergoog: (and it should be case-insensitive, so `<a2ui_json>` or `<A2UI_JSON>` would work.

gspencergoog: Ahh, I didn't see the part about the followup PR to do this.  Carry on!

nan-yu: Yup. I don't want to mix everything in a single PR :)

PR #782, opened 2026-03-05
remove rizzcharts

gspencergoog:
# Description

We're removing the rizzcharts demo because we feel like we have too many demos to maintain, and there aren't clear owners for the rizzcharts example. Oh, and it's entirely broken in the Lit client, and requires an API key to run.


nan-yu:
Can we keep the rizzcharts on the agent side and the angular client? That is the only sample that uses custom catalog, and supports injecting different prompts on different requests.

renderers/web_core/src/v0_8/schema/common-types.ts
```diff
@@ -122,19 +122,23 @@ export const ActionSchema = z.object({
 
 export const TextSchema = z.object({
   text: StringValueSchema,
-  usageHint: z.enum(["h1", "h2", "h3", "h4", "h5", "caption", "body"]),
+  usageHint: z
+    .enum(["h1", "h2", "h3", "h4", "h5", "caption", "body"])
+    .optional(),
```
ditman, line 146: This is from your previous PR, so this'll need to be rebased?

gspencergoog: Yeah, rebased.

gspencergoog:
> Can we keep the rizzcharts on the agent side and the angular client? That is the only sample that uses custom catalog, and supports injecting different prompts on different requests.

Sure, but what client will connect to it?  If we're removing the Lit and Angular clients for it, then there won't be anything that can draw it because nothing will implement that catalog...

I thought the plan was to eliminate it because there wasn't a clear owner for it, and we felt like we had too many samples, so it makes sense to remove all of it and restore it if we decide we want it back.

ava-cassiopeia (approved):

PR #784, opened 2026-03-06
Add rizzcharts back again, disable Lit version

gspencergoog:
# Description

This reverts https://github.com/google/A2UI/pull/782 and instead ONLY removes the Lit demo, since that is broken.

@nan-yu pointed out that this is the only demo that can switch catalogs dynamically, so she wanted to keep just the Angular version around.


nan-yu (approved):
Thanks, Greg!

PR #788, opened 2026-03-06
Update renderer guide to comprehensively cover Catalog API design and implementation process

jacobsimionato:
This PR introduces a comprehensive refactor of the A2UI Renderer Guide to try to make it so you can vibe code based on just this design.

### Renderer Guide Enhancements (v0.9)
- **Structural Reorganization**: Clearly separated the architecture into the Agnostic Data Layer, Catalog API & Bindings, and the Framework Binding Layer.
- **Improved Developer Experience**: Added a 'Happy Path' section demonstrating how simple and type-safe component implementation becomes once the framework adapter is in place.
- **Agent Implementation Guide**: Added a strict 7-step phased guide for building new renderers, ensuring architectural consistency and providing a clear path for AI-driven implementation.
- **Static vs. Dynamic Optimizations**: Simplified core examples for statically typed languages while retaining advanced Zod/TypeScript patterns as specific optimizations.
- **Schema-First Design**: Ensured `ComponentApi` uses a formal `schema` property to reinforce the requirement for robust validation and capabilities advertising.

### DIFF
--- a/specification/v0_9/docs/renderer_guide.md
+++ b/specification/v0_9/docs/renderer_guide.md
@@ -92,18 +92,18 @@ The model is designed to support high-performance rendering through granular upd
 *   **`SurfaceComponentsModel`**: A flat collection of component configurations.
 *   **`ComponentModel`**: A specific component's raw configuration.
 *   **`DataModel`**: A dedicated store for application data.
-*   **`DataContext`**: A scoped window into the `DataModel`. Used by functions and components to resolve dependencies and mutate state. The window is specifically a view of the data model that has some defined base path, so that relative path references are resolved against this. This is typically just the root `/`, except for cases where we are using the "template" pattern to build a `ChildList` of Components. When this occurs, the parent component will generally call some function to build each child, specifying the same component ID but a different base path for each child so that the same template can be rendered with different data.
+*   **`DataContext`**: An abstraction around the data model, available functions, and the base path of a Component, which allows Component implementations to fetch and subscribe to dynamic values via a simple API. Different Component instances instantiated from the same Component ID, but with different base paths (e.g. because they are different instances of a *template*) can have a different `DataContext` instance.
 *   **`ComponentContext`**: A binding object pairing a component with its data scope.
 
 ### The Models
-These classes are designed to be "simple containers" for data. They hold the state of the UI but contain minimal logic (i.e. they do not handle side effects, layout algorithms, or framework integration). 
+These classes are designed to be "simple containers" for data. They hold a snapshot of the A2UI state and contain logic to implement observability. They may validate changes to prevent the system entering inconsistent states. Logic to decode A2UI messages and update the model layer should be within MessageProcessor. Logic to unwrap data model and function references should be within the context layer.
 
 **Key Characteristics:**
 *   **Mutable**: Their properties can be updated over time.
 *   **Observable**: They provide mechanisms to listen for those updates.
 *   **Encapsulated Composition**: Parent models hold references to children, but do not construct them.
 
-They are organized hierarchically based on the structure of the data and component tree.
+They are organized hierarchically based on the structure of the data and component tree in A2UI e.g. SurfaceGroup, Surface, Component. Within each SurfaceModel, ComponentModels are represented as a flat list, with view hierarchy construction handled in the Surface rendering logic for each UI framework.
 
 #### SurfaceGroupModel & SurfaceModel
 The root containers for active surfaces and their catalogs, data, and components.
@@ -146,7 +146,7 @@ class SurfaceModel<T> {
 }
 ```
 #### `SurfaceComponentsModel` & `ComponentModel`
-Manages the raw JSON configuration of components in a flat map.
+Manages the raw JSON configuration of components in a flat map which includes one entry per component ID. This represents the raw Component data *before* ChildList templates are resolved, which can instantiate multiple instances of a single Component with the same ID.
 
 ```typescript
 class SurfaceComponentsModel {
@@ -207,15 +207,15 @@ A change at a specific path must trigger notifications for related paths to ensu
 #### Type Coercion Standards
 To ensure the Data Layer behaves identically across all platforms (e.g., TypeScript, Swift, Kotlin), the following coercion rules MUST be followed when resolving dynamic values:
 
-| Input Type                 | Target Type | Result                               |
-| :------------------------- | :---------- | :----------------------------------- |
+| Input Type                 | Target Type | Result                                                                  |
+| :------------------------- | :---------- | :---------------------------------------------------------------------- |
 | `String` ("true", "false") | `Boolean`   | `true` or `false` (case-insensitive). Any other string maps to `false`. |
-| `Number` (non-zero)        | `Boolean`   | `true`                               |
-| `Number` (0)               | `Boolean`   | `false`                              |
-| `Any`                      | `String`    | Locale-neutral string representation |
-| `null` / `undefined`       | `String`    | `""` (empty string)                  |
-| `null` / `undefined`       | `Number`    | `0`                                  |
-| `String` (numeric)         | `Number`    | Parsed numeric value or `0`          |
+| `Number` (non-zero)        | `Boolean`   | `true`                                                                  |
+| `Number` (0)               | `Boolean`   | `false`                                                                 |
+| `Any`                      | `String`    | Locale-neutral string representation                                    |
+| `null` / `undefined`       | `String`    | `""` (empty string)                                                     |
+| `null` / `undefined`       | `Number`    | `0`                                                                     |
+| `String` (numeric)         | `Number`    | Parsed numeric value or `0`                                             |
 
 
 ### The Context Layer (Transient Windows)


specification/v0_9/docs/renderer_guide.md
```diff
@@ -1,25 +1,67 @@
...
 It consists of three sub-components: the Processing Layer, the Dumb Models, and the Context Layer.
 
+### Prerequisites
+
+To implement the Data Layer effectively, your target environment needs two foundational utilities: a Schema Library and an Observable Library.
+
+#### 1. Schema Library
+To represent and validate component and function APIs, the Data Layer requires a **Schema Library**. 
+
+*   **Ideal Choice**: A library (like **Zod** in TypeScript or **Pydantic** in Python) that allows for programmatic definition of schemas and the ability to validate raw JSON data against those definitions.
```
jgindin, line 44: Given the list of languages we know need support, is there value in adding more recommendations?

specification/v0_9/docs/renderer_guide.md
```diff
@@ -1,25 +1,67 @@
...
+*   **Capabilities Generation**: The library should ideally support exporting these programmatic definitions to standard JSON Schema for the `getClientCapabilities` payload.
+*   **Fallback**: If no suitable programmatic library exists for the target language, raw **JSON Schema strings**, `Codable` structs, or `kotlinx.serialization` classes can be used instead.
+
+#### 2. Observable Library
+A2UI relies on a standard observer pattern to reactively update the UI when data changes. The Data Layer and client-side functions must be able to return streams or reactive variables that hold an initial value and emit subsequent updates.
```
jgindin, line 49: Is there any way we can ensure the API is of the form `subscribe` / `unsubscribe` and leave the particular mechanism as an implementation detail?

This might get into over-specifying and maybe it's not so important if we don't expect users to be using more than one client framework.

There is also some tension between a completely consistent API vs. leveraging idiomatic patterns and best practices for each language.

I think I'm talking myself out of suggesting any changes, but I'm going to leave this comment in case it spurs further thought from anyone else.

(This might be getting into where we provide one canonical implementation and encourage the use of agentic coding to "port" that implementation to other languages? At this point, we're probably going beyond the scope of this document, though...)

specification/v0_9/docs/renderer_guide.md
```diff
@@ -28,196 +70,44 @@ We strictly separate **construction** from **composition**. Parent containers do
...
-
 ### Key Interfaces and Classes
 *   **`MessageProcessor`**: The entry point that ingests raw JSON streams.
 *   **`SurfaceGroupModel`**: The root container for all active surfaces.
 *   **`SurfaceModel`**: Represents the state of a single UI surface.
 *   **`SurfaceComponentsModel`**: A flat collection of component configurations.
 *   **`ComponentModel`**: A specific component's raw configuration.
 *   **`DataModel`**: A dedicated store for application data.
-*   **`DataContext`**: A scoped window into the `DataModel`.
+*   **`DataContext`**: A scoped window into the `DataModel`. Used by functions and components to resolve dependencies and mutate state.
```
jgindin: How is the appropriate "window" determined?

specification/v0_9/docs/renderer_guide.md
```diff
@@ -28,196 +70,44 @@ We strictly separate **construction** from **composition**. Parent containers do
...
-* **Description Preservation:** Additional descriptions appended to tagged types are preserved and properly formatted alongside the reference pointer.
-
 ### The "Dumb" Models
-These classes are designed to be "dumb containers" for data. They hold the state of the UI but contain minimal logic. They are organized hierarchically and use a consistent pattern for observability and composition.
-
-**Key Characteristics:**
-*   **Mutable:** State is updated in place.
-*   **Observable:** Each layer is responsible for making its direct properties observable via standard listener patterns, avoiding heavy reactive dependencies.
-*   **Encapsulated Composition:** Parent layers expose methods to add fully-formed child instances (e.g., `addSurface`, `addComponent`) rather than factory methods that take parameters.
+These classes are designed to be "dumb containers" for data. They hold the state of the UI but contain minimal logic. They are organized hierarchically.
```
jgindin: Can we be more descriptive of what "minimal logic" means?

specification/v0_9/docs/renderer_guide.md
```diff
@@ -28,196 +70,44 @@ We strictly separate **construction** from **composition**. Parent containers do
...
-* Returns a configuration object ready to be sent in the transport metadata (populating `supportedCatalogIds` and `inlineCatalogs`).
-
-**Test Cases to Include**
-When implementing or modifying the capabilities generator, you must include test cases that verify:
-* **Capabilities Generation:** The output successfully includes the `inlineCatalogs` list when requested.
-* **Component Envelope:** Generated schemas correctly wrap component properties in an `allOf` block referencing `common_types.json#/$defs/ComponentCommon` and correctly assert the `component` property `const` value.
-* **Reference Resolution:** Properties tagged with `REF:` successfully resolve to `$ref` objects instead of expanding the inline schema definition (e.g., a `title` property correctly emits `"$ref": "common_types.json#/$defs/DynamicString"`).
-* **Description Preservation:** Additional descriptions appended to tagged types are preserved and properly formatted alongside the reference pointer.
-
 ### The "Dumb" Models
```
jgindin: Can we just use "Models" without the "Dumb" adjective?

Especially since we're saying that they can have _some_ logic...

In general, maybe "simple" is a better term?

specification/v0_9/docs/renderer_guide.md
jgindin: An example of what is meant by hierarchically would be useful.

Is it hierarchical according to the components? Or, more likely, the data model, matching the JSON sent in the `UpdateDataModelMessage` ?

specification/v0_9/docs/renderer_guide.md
```diff
@@ -227,7 +117,7 @@ class SurfaceGroupModel<T extends CatalogApi> {
 
 type ActionListener = (action: any) => void | Promise<void>; // Handler for user interactions
```
jgindin: What kind of user actions? Like a button has been clicked?

Is more data required to understand, e.g., where this action is coming from (the component which is the source) and maybe the surface the component is part of?

Maybe we need an `Action` interface?
```
interface Action {
  surface: SurfaceModel, // maybe just surfaceId?
  source: ComponentModel, // maybe just componentId ?
  data: any
}
```

specification/v0_9/docs/renderer_guide.md
jgindin: Shouldn't there be `removeXxxListener` functions as well?

jgindin: "Fires"? Is this supposed to be "Invoked when any property changes"?

jgindin, line 174: This seems to be the only place we're using a subscription. In other places we're using `addXxxListener`.

Should we be consistent?

jgindin: What is the mapping for any other string, e.g., "foo"? Presumably this maps to `false`?

specification/v0_9/docs/renderer_guide.md
```diff
@@ -1,25 +1,67 @@
...
-## 1. Data Layer
+## Implementation Topologies
+Because A2UI spans multiple languages and UI paradigms, the strictness and location of these architectural boundaries will vary depending on the target ecosystem.
+
+### Dynamic Languages (e.g., TypeScript / JavaScript)
+In highly dynamic ecosystems like the web, the architecture is typically split across multiple packages to maximize code reuse across diverse UI frameworks (React, Angular, Vue, Lit).
+*   **Core Library (`web_core`)**: Implements the Core Data Layer, Component Schemas, and a Generic Binder Layer. Because TS/JS has powerful runtime reflection, the core library can provide a generic binder that automatically handles all data binding without framework-specific code. 
+*   **Framework Library (`react_renderer`, `angular_renderer`)**: Implements the Framework-Specific Adapters and the actual view implementations (the React `Button`, `Text`, etc.).
+
+### Static Languages (e.g., Kotlin, Swift)
```
gspencergoog: Would Dart also be here?

specification/v0_9/docs/renderer_guide.md
```diff
@@ -1,25 +1,67 @@
...
+### Prerequisites
+
+To implement the Data Layer effectively, your target environment needs two foundational utilities: a Schema Library and an Observable Library.
+
+#### 1. Schema Library
+To represent and validate component and function APIs, the Data Layer requires a **Schema Library**. 
+
+*   **Ideal Choice**: A library (like **Zod** in TypeScript or **Pydantic** in Python) that allows for programmatic definition of schemas and the ability to validate raw JSON data against those definitions.
+*   **Capabilities Generation**: The library should ideally support exporting these programmatic definitions to standard JSON Schema for the `getClientCapabilities` payload.
+*   **Fallback**: If no suitable programmatic library exists for the target language, raw **JSON Schema strings**, `Codable` structs, or `kotlinx.serialization` classes can be used instead.
```
gspencergoog, line 46: Maybe also code generation?

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
+
+### The Catalog API
+A catalog groups component definitions (and optionally function definitions) together so the `MessageProcessor` can validate messages and provide capabilities back to the server.
+
+```typescript
+class Catalog<T> {
+  readonly id: string; // Unique catalog URI (e.g., "https://mycompany.com/catalog.json")
+  readonly components: ReadonlyMap<string, T>;
+  readonly functions?: ReadonlyMap<string, FunctionImplementation>;
+
```
jgindin, line 306: What about `theme`?

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
+*   Implement the mechanism that binds a `ComponentBinding` stream to the native UI state (e.g., a wrapper wrapper view/widget).
+*   Implement the recursive `Surface` builder that takes a `surfaceId`, finds the "root" component, and recursively calls `buildChild`.
+*   **Crucial**: Ensure the unmount/dispose lifecycle hook calls `binding.dispose()`.
+
+### 5. Minimal Catalog Support
+Do not start with the full Basic Catalog. Target the `minimal_catalog.json` first.
+*   **Core Library**: Create definitions/binders for `Text`, `Row`, `Column`, `Button`, and `TextField`.
+*   **Core Library**: Implement the `capitalize` function.
+*   **Framework Library**: Implement the actual native UI widgets for these 5 components.
+*   Write a `createMinimalCatalog()` function that bundles these together.
```
gspencergoog: This might be too prescriptive.  It might want a different factory pattern for the catalogs.

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
+### Creating Custom Catalogs
+Extensibility is a core feature of A2UI. It should be trivial to create a new catalog by extending an existing one, combining custom components with the standard set.
+
+*Example of composing a custom catalog:*
+```python
+# Pseudocode
+myCustomCatalog = Catalog(
+  id="https://mycompany.com/catalogs/custom_catalog.json",
+  functions=basicCatalog.functions,
+  components=basicCatalog.components.append([MyCompanyLogoComponent()])
```
jgindin: This doesn't match the way the `components` member of `Catalog` is declared above. Should this be:
```
components = basicCatalog.components.update({'my-company-logo', MyCompanyLogoComponent()})
```

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
+
+### 6. Demo Application (Milestone)
+Build a self-contained application to prove the architecture works before scaling.
+*   The app should run entirely locally (no server required).
+*   It should load the JSON message arrays from `specification/v0_9/json/catalogs/minimal/examples/`.
+*   It should display a list of these examples.
+*   When an example is selected, it should pipe the messages into the `MessageProcessor` and render the surface.
+*   **Reactivity Test**: Add a mechanism to simulate delayed `updateDataModel` messages (e.g., waiting 2 seconds before sending data) to prove that the UI progressively renders and reacts to changes.
+
+**STOP HERE. Ask the human user for approval of the architecture and demo application before proceeding to step 7.**
```
gspencergoog: tiny nit: I'd remove "human", just because it might be an agent running this as a sub-agent.  Or leave it.  It probably doesn't matter.

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
+
+### 7. Basic Catalog Support
+Once the minimal architecture is proven robust:
+*   **Core Library**: Implement the full suite of basic functions (including the complex `formatString` parser).
```
gspencergoog: Somewhere, not sure where, it would be good to stress that the `formatString` function is the only place where string interpolation and parsing happens.  Twice now, I've had the agent "generalize" that all strings should have string interpolation and go wild trying to add it everywhere.

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
+*   **Core Library**: Create definitions/binders for the remaining Basic Catalog components.
+*   **Framework Library**: Implement all remaining UI widgets.
+*   **Tests**: Look at existing reference implementations (e.g., `web_core`) to formulate and run comprehensive test cases for data coercion and function logic.
```
gspencergoog: I'd also add integration tests to this, not just unit tests.

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
+}
+```
+
+#### Angular Pseudo-Adapter
+```typescript
+// Pseudo-code concept for an Angular adapter
+@Component({
+  template: `<ng-container *ngIf="props$ | async as props">
+               <!-- Render specific component here -->
+             </ng-container>`
```
jgindin: Bit of a nit, but we should use "modern" Angular. Something like:
```
@Component({
  selector: 'app-angular-wrapper',
  imports: [MatButtonModule],
  template: `
    @if (props(); as props) {
      <button mat-button>{{ props.label }}</button>
    }
  `
})
export class AngularWrapper {
  private binder = inject(BinderService);
  private context = inject(...);

  private bindingResource = resource({
    loader: async () => {
      const binding = this.binder.bind(this.context);

      return {
        instance: binding,
        props: toSignal(binding.propsStream) // Convert Observable to Signal [cite: 18]
      };
    },
  });

  props = computed(() => this.bindingResource.value()?.props() ?? null);

  constructor() {
    inject(DestroyRef).onDestroy(() => {
      this.bindingResource.value()?.instance.dispose();
    });
  }
}
```

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
-    *   **FunctionCall**: Identified by parentheses (e.g., `${now()}`).
+1.  **Recursion**: The function implementation MUST use `DataContext.resolveDynamicValue()` or `DataContext.subscribeDynamicValue()` to recursively evaluate nested expressions or function calls (e.g., `${formatDate(value:${/date})}`).
+2.  **Tokenization**: The parser must distinguish between DataPaths (e.g., `${/user/name}`) and FunctionCalls (e.g., `${now()}`).
 3.  **Escaping**: Literal `${` sequences must be handled (typically by escaping as `\${`).
-4.  **Reactive Coercion**: Results are transformed into strings using the **Type Coercion Standards** defined in the Data Layer section.
+4.  **Reactive Coercion**: Results are transformed into strings using the Type Coercion Standards.
+
+---
+
+## 5. Agent Implementation Guide
```
jgindin, line 647: Should this be a separate markdown file?

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
+*   Implement `ComponentModel`, `SurfaceComponentsModel`, `SurfaceModel`, and `SurfaceGroupModel`.
+*   Implement `DataContext` and `ComponentContext`.
+*   Implement `MessageProcessor`. Include logic for detecting schema references to generate `ClientCapabilities`.
+*   Define the `Catalog`, `ComponentApi`, and `FunctionImplementation` interfaces.
+*   Define the `ComponentBinding` interface.
+
+### 4. Framework-Specific Layer
+Implement the bridge between the agnostic models and the native UI (Section 3).
+*   Define the `ComponentAdapter` API (how the core library hands off a component to the framework).
+*   Implement the mechanism that binds a `ComponentBinding` stream to the native UI state (e.g., a wrapper wrapper view/widget).
```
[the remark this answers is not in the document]
jacobsimionato: Let's fix this

specification/v0_9/docs/renderer_guide.md
```diff
@@ -348,66 +239,437 @@ While A2UI components are designed to be self-contained, certain rendering logic
...
+Extensibility is a core feature of A2UI. It should be trivial to create a new catalog by extending an existing one, combining custom components with the standard set.
+
+*Example of composing a custom catalog:*
+```python
+# Pseudocode
+myCustomCatalog = Catalog(
+  id="https://mycompany.com/catalogs/custom_catalog.json",
+  functions=basicCatalog.functions,
+  components=basicCatalog.components.append([MyCompanyLogoComponent()])
+)
```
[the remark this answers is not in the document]
jacobsimionato: Agreed, let's fix this

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Yes, let's add it

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Well, the catalog agnostic part of the library can't refer to catalog-specific codegenned classes, so I think strings make sense in this case. Elsewhere, I think the doc already mentions codegen as an approach to universal bindings etc.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Agreed - let's make the wording more loose, like "implement a factory API that returns the minimal catalog". This could be a function like `createMinimalCatalog()`, though it doesn't have to be. Follow the idioms of the platform.

Make this change to the part about minimal catalog also.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Agreed, let's remove "human". We can just say stop here and check the app works before moving on.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Good point. Let's specify this.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Maybe! But I want to avoid doing due diligence on libraries for every language right now and start with some generic requirements. Quite possibly, framework experts can make a better choice than me, given the requirements.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: The window is specifically a view of the data model that has some defined base path, so that relative path references are resolved against this. This is typically just the root `/`, except for cases where we are using the "template" pattern to build a ChildList of Components. When this occurs, the parent component will generally call some function to build each child, specifying the same component ID but a different base path for each child so that the same template can be rendered with different data.

Let's explain this.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Let's avoid the use of the term "dumb" because it comes across as negative and judgemental.

These models classes are intended to represent specifically the information communicated in the A2UI protocol, while not implementing any parsing, decoding, validation, conversion or rendering logic. This should all be in other layers.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Agreed. Let's just say "Models".

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Let's define this based on how it's currently implemented in web_core.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Good point. The theme schema should be represented here.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: I think that the name of a Component should be a property of the Component object. That way, when people assemble Catalogs from sets of Components that are implemented in SDKs, they use the same name by default. If they need to rename some component, to avoid a collision etc, they can use a  more verbose API like  `textComponent.withName("alternativeText")`. E.g. let's say GoogleMaps release some SDKs that implement GoogleMap and PlaceCard for web, iOS and Android, and also provide some schema to use for inference (we need to figure out how to represent an individual Component in this context). It'd be great if by default, it always uses the same name.

So I think it's okay to have a constructor that accepts a list of Components, but expose it as a map for efficient lookup. Though catalogs likely aren't big enough that it matters, so I'm also fine to just use arrays everywhere and do linear search as necessary.

Let's leave this the same for now, but I'm curious what @jgindin thinks about this.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: SGTM - let's do it! Thanks for the pointer.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Maybe! There is probably a case for splitting this over several files, even separating basic catalog stuff out etc. Let's leave this as-is for now.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: I agree there is a tension here which is tricky to navigate. I think we're on the same page. I think there are two kinds of observation necessary here:

- Value updates (there is a current state e.g. data model value, which may change over time)
- Events with no current state (an action is triggered from a component)

Each codebase should choose some approach the two cases above and use it consistently, but we don't care how they do it.

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Added an actionInterface

specification/v0_9/docs/renderer_guide.md
jacobsimionato: Good point

jacobsimionato: Good point. Let's revert to addXxxListener. This is slightly different to the use case where you want to observe a creation or deletion event, because there is the concept of a current value, which frameworks would probably adapt to something like a Signal.

But I think to have a vanilla interface, and as a guide here, we can just have addXxxListener, and you can adapt it to a Signal etc by doing get(), then addXxxListener(). And implementers can still ignore the examples and decide to use some observable library if it makes sense in their language anyway. So let's keep the examples consistent and simple.

jacobsimionato: Hmmm I thought about this a little more, and I think there is value in having a different approach for the DataModel and DataContext that allow you to fetch the current value and subscribe in a single operation (similar to connecting to a Signal or BehaviorSubject), so that custom function implementations which represent a changing data source don't need to be triggered twice for this scenario.

I think I'd like to leave this for now, but if anything, I'm tempted to suggest that people use a library here, e.g. the same one they will use to represent the output of a function. We currently use rxjs observable in web_core but are thinking of migrating to signals in https://github.com/google/A2UI/pull/795 .

specification/v0_9/docs/renderer_guide.md
jgindin: Even that template case assumes that *all* the data a child needs is contained in a narrow branch of the data model. If, for example, there's some other data that the child needs to either read or write, then this "window" seems like it is always just "/".

I like the idea of being able to limit what the component can access, but I just wonder if it's actually going to work out that way...

specification/v0_9/docs/renderer_guide.md
jgindin: That description makes it sound like there should be **no** logic.

I could see some logic living here -- e.g., if a component updates a string, having the model enforce that there's no leading or trailing spaces might make sense.

It might help if we called out specific examples of what kind of logic would be acceptable in the model, and then also specifically describe where other types of logic should live.

specification/v0_9/docs/renderer_guide.md
jgindin: Ah, that makes sense, and I agree it is a **much** better approach. It just isn't clear from the documentation.

PR #793, opened 2026-03-06
Change CallId to allow for more data.

jgindin:
# Description
The agent identifier and name of the function called are now optional; the callId itself is still required.

This enables support for stateless backends, especially where there are multiple instances sharing a backing session store.

## Pre-launch Checklist

- [x ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/specification/v0_10/json/common_types.json
+++ b/specification/v0_10/json/common_types.json
@@ -16,10 +16,6 @@
           "type": "string",
           "description": "Identifies the agent initiating the function call."
         },
-        "call": {
-          "type": "string",
-          "description": "Name of the function call."
-        },
         "callId": {
           "type": "string",
           "description": "Uniquely identifies this instance of the function call."


specification/v0_10/json/common_types.json
```diff
@@ -9,8 +9,23 @@
...
-      "type": "string",
-      "description": "The unique identifier for a server initiated function call."
+      "type": "object",
+      "description": "The unique identifier for a server initiated function call.",
+      "properties": {
+        "agentId": {
+          "type": "string",
+          "description": "Identifies the agent initiating the function call."
+        },
+        "call": {
```
gspencergoog: If the `callId` uniquely identifies the call, why do we ever need `call`?

jgindin: I debated naming this `functionName` , but decided to use `call` to match the property in the `FunctionCall` type.

`call` identifies which function was called, and `callId` is the specific instance of the call.

The feedback received is that when used with ADK, they have to match the response with both the call id and the function name. (Probably need to add something in the docs that captures the experience of this integration.)

gspencergoog: I thought the callId was unique across all calls.  Why couldn't it be? Worst case you prefix the callId with the call name...

jgindin: Yes, the call name could be prefixed, but this just avoids building & parsing the string.

gspencergoog: I think I'd rather build and parse the string.  Adding the "call" parameter to `common_types.json` means it gets added to the token/context budget of every inference that does A2UI generation, which is a higher cost than building/parsing a more unique ID.

jgindin: Got it!

PR #799, opened 2026-03-09
Add GitHub workflow to enforce Google OSS license

ava-cassiopeia:
Also adds the license to any places we missed.

### DIFF
--- a/.github/workflows/check_license.yml
+++ b/.github/workflows/check_license.yml
@@ -16,6 +16,8 @@ name: Google open-source license
 
 on:
   push:
+    branches:
+      - main
   pull_request:
 
 jobs:


ava-cassiopeia:
> However, the pull request title, 'Add GitHub workflow to enforce Google OSS license', suggests that a new GitHub workflow file should be part of this change. This file seems to be missing from the current set of changes. To fully achieve the goal of this PR, please consider adding the workflow file that enforces the presence of these license headers.

Gemini my dear, you [reverse-hallucinated](https://github.com/google/A2UI/blob/b1bd275856e82c48e0212135245b12c4673dc1a9/.github/workflows/check_license.yml) lol

PR #791, opened 2026-03-06
feat: Introduce a calculator UI application, integrating it as a serv…

sugoi-yuzuru:
feat: Introduce a calculator UI application, integrating it as a server resource and tool, and updating related dependencies.

This calculator MCP App is a precursor to a future sample application that intends to integrate MCP Apps into A2UI. A2UI Custom component that will allow MCP Apps rendering will be a following up.

https://screenshot.googleplex.com/3tQRreZ5eQE9qsQ is how the calculator looks.

_List which issues are fixed by this PR. For larger changes, raising an issue first helps reduce redundant work._

## Pre-launch Checklist

- [x] I signed the [CLA].
- [x] I read the [Contributors Guide].
- [x] I read the [Style Guide].
- [x] I have added updates to the [CHANGELOG].
- [x] I updated/added relevant documentation.
- [x] My code changes (if any) have tests.

If you need help, consider asking for advice on the [discussion board].

[CHANGELOG]: ../CHANGELOG.md
[CLA]: https://cla.developers.google.com/
[Contributors Guide]: ../CONTRIBUTING.md
[discussion board]: https://github.com/google/A2UI/discussions
[Style Guide]: ../STYLE_GUIDE.md

PR #801, opened 2026-03-09
Mcp Apps 1 : Introducing McpApps A2UI Component

sugoi-yuzuru:
# Description

This PR introduces an Angular A2UI Component named `McpApps`. McpApps component leverages the McpApps API to create a double-layered iframe (iframe sandboxing) and inserts a HTML bundle (HTML/JS/CSS) passed into the component as an A2UI data input field into the inner-iframe to load arbitrary web-application into an A2UI component.

The idea is that A2UI Agent can load an MCP Application resource from an external MCP server and simply relay that application as a `content` of an McpApp A2UI Component. 

The McpApp A2UI Component will take care of the double-layered iframing and the MCP Application's message-based communications to its host application and server. Attached is an overview of what this PR achieves. 

image

In forthcoming changes, this McpApp A2UI component will integrate with the `Orchestrator` Agent sample and will be loading the `Calculator` MCP Application that was introduced in https://github.com/google/A2UI/pull/791

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
+++ b/samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
@@ -98,7 +98,7 @@ export class McpApp
 
   readonly allowedTools = input<string[]>([]);
   readonly title = input<Primitives.StringValue | null>();
-  protected readonly resolvedTitle: Signal<string | null> = computed(() =>
+  protected readonly resolvedTitle = computed<string | null>(() =>
     super.resolvePrimitive(this.title() ?? null),
   );
 


dmandar:
Can we fix the Gemini code review comments first?

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,222 @@
...
+  readonly title = input<Primitives.StringValue | null>();
+  protected readonly resolvedTitle: Signal<string | null> = computed(() =>
+    super.resolvePrimitive(this.title() ?? null),
+  );
+
+  protected readonly iframeSrc = signal<SafeResourceUrl | null>(
+    this.sanitizer.bypassSecurityTrustResourceUrl('about:blank')
+  );
+
+  @ViewChild('iframe') iframeElement!: ElementRef<HTMLIFrameElement>;
```
jgindin: Can we switch to the `viewChild` signal instead?

```
private iframe = viewChild.required<HTMLIFrameElement>('iframe');
```

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,222 @@
...
+  private appBridge: AppBridge | null = null;
+  private messageHandler: ((event: MessageEvent) => void) | null = null;
+
+  ngAfterViewInit() {
```
jgindin: Prefer using signals (including `computed`) over this lifecycle event.

go/angular-style#avoid-all-lifecycle-methods-except-oninit-and-ondestroy

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+import {
+    AppBridge,
+    PostMessageTransport,
+    SANDBOX_PROXY_READY_METHOD,
+    SANDBOX_RESOURCE_READY_METHOD
+} from '@modelcontextprotocol/ext-apps/app-bridge';
+
+const contentDiv = document.getElementById('content');
```
jgindin: Nit -- instead of embedding the _type_ into the name, I'd suggest just `content`.

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
```diff
@@ -0,0 +1,80 @@
...
+        const { html, sandbox } = data.params;
+        
+        if (html && contentDiv) {
+            // Create an inner iframe with srcdoc to enable Javascript execution if any.
+            const innerFrame = document.createElement('iframe');
+            innerFrame.srcdoc = html;
+            innerFrame.style.width = '100%';
+            innerFrame.style.height = '100%';
+            innerFrame.style.border = 'none';
+            if (sandbox) {
```
jgindin: If `sandbox` is null or undefined, what is the effect on what we're building here? Is `sandbox` truly optional?

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,222 @@
...
+import { DomSanitizer, SafeHtml, SafeResourceUrl } from '@angular/platform-browser';
+
+@Component({
+  selector: 'a2ui-mcp-app',
+  standalone: true,
+  imports: [],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  styles: `
+    :host {
+      display: flex;
```
jgindin, line 48: `flex` seems overly complex if there's just a single DOM element (the iframe) being displayed.

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
```diff
@@ -0,0 +1,80 @@
...
+        logging: {}
+    }
+);
+
+bridge.oncalltool = async (params: any) => {
+    // Forward tool calls to parent if needed, or handle locally
+    // For now, we just log
+    console.log('[Sandbox] Tool call:', params);
+    // We can also postMessage back to parent manually if bridge doesn't handle it
+    // implementation of forwarding logic would go here
```
jonsharkey: *nit* The lack of capital and periods here made it hard for for mo understand. Should it be:

// We can also postMessage back to parent manually if bridge doesn't handle it.
// Implementation of forwarding logic would go here.
// For now, throw to indicate not implemented in sandbox.

Also there's an extra leading space on the third line.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: Done.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: Ack. 
Changed to onInit.

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
sugoi-yuzuru: Good point. I have updated the code to default to block all if `sandbox` is empty. So the optional argument is a true `allowlist`.

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
sugoi-yuzuru: Done.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: I still have a TODO to dynamically update the size of the component based on the content's sizing. But as for now, this flex is critical for the host component to properly fill the empty space (especially vertically) in sync with the child iframe.

AI summary: 
This styling combination ensures that the embedded iframe dynamically expands to perfectly fill the entire width and height of the <a2ui-mcp-app> component without manually computing pixel heights, helping to prevent weird scrolling or collapsed height issues.

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
sugoi-yuzuru: Fixed.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,222 @@
...
+    iframe {
+      flex: 1;
+      width: 100%;
+      height: 100%;
+      border: none;
+      background-color: white; /* Ensure content is readable */
+    }
+  `,
+  template: `
+    @if (resolvedContent(); as content) {
```
suyangw-g: alias not used

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,222 @@
...
+{
+  private readonly sanitizer = inject(DomSanitizer);
+
+  readonly content = input.required<Primitives.StringValue | null>();
+  protected readonly resolvedContent: Signal<SafeHtml | null> = computed(() => {
+    const rawContent = super.resolvePrimitive(this.content() ?? null);
+    return rawContent
+      ? this.sanitizer.bypassSecurityTrustHtml(rawContent)
+      : null;
+  });
```
suyangw-g, line 97: Add a comment explaining what this resolvedContent is used for. It's unclear to me

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+  `,
+})
+export class McpApp
+  extends DynamicComponent<Types.CustomNode>
+  implements OnDestroy, OnInit
+{
+  private readonly sanitizer = inject(DomSanitizer);
+
+  readonly content = input.required<Primitives.StringValue | null>();
```
suyangw-g, line 79: Will this signal input `content` change dynamically?

If so, the iframe will currently ignore the update since it's sent to sandbox once during initializeBridge(), and you should use effect() to listen for updates and push updates to appBridge.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: Thx. Removed.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: I have added some logic that was left for when I integrated. 
Essentially, in the integration stage, I use url encoding to reduce the risk of new line and special characters that may be included in the HTML content compromising A2UI Json. 

I have added this logic in this PR :)

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: Good callout. It totally is possible that async dataUpdate will be supplying the HTML content. Also when server/client interaction updates the content. 

I have updated so that the signal is connected to surface updates.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,228 @@
...
+    // We must pass the iframe's contentWindow as the target
+    const transport = new PostMessageTransport(
+      // The first argument is the target window to send messages TO (the iframe).
+      // The second argument is the source window to validate messages FROM (also the iframe).
+      iframe.contentWindow!,
+      iframe.contentWindow!,
+    );
+    await bridge.connect(transport);
+
+    this.appBridge.set(bridge);
```
suyangw-g, line 236: I'm not familiar with MCP AppBridge. Based on Gemini, you may need to clean up the old AppBridge instance before you set the new one, since there could be significant performance and architectural bugs, like zombie event listeners (window.addEventListener in appBridge), memory leaks (window still holds a reference to the old appBridge's event listener), orphaned connections to the mcp server, etc.

Could you please double check?

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,228 @@
...
+  `,
+})
+export class McpApp
+  extends DynamicComponent<Types.CustomNode>
+  implements OnDestroy, OnInit
+{
+  private readonly sanitizer = inject(DomSanitizer);
+
+  readonly content = input.required<Primitives.StringValue | null>();
+  protected readonly resolvedContent: Signal<string | null> = computed(() => {
```
jgindin: Nit: Instead of the explicit `Signal<string|null>`, can we just add the type to the `computed`?

E.g., 

```
protected readonly resolvedContent = computed<string|null>(() => ...);
```

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
```diff
@@ -0,0 +1,78 @@
...
+        const { html, sandbox } = data.params;
+        
+        if (html && content) {
+            // Create an inner iframe with srcdoc to enable Javascript execution if any.
+            const innerFrame = document.createElement('iframe');
+            innerFrame.srcdoc = html;
+            innerFrame.style.width = '100%';
+            innerFrame.style.height = '100%';
+            innerFrame.style.border = 'none';
+            innerFrame.sandbox = sandbox || '';
```
jgindin: I feel like the term "sandbox" is being over used....

This file is `sandbox.ts`, which is setting up a sandbox. But here, we're setting a `sandbox` property on the iframe, and that (seemingly?) has nothing to do with this file.

Can we clarify terms to make this easier to understand?

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
```diff
@@ -0,0 +1,78 @@
...
+    // Validate the origin of incoming messages
+    if (event.origin !== window.location.origin) {
+        return;
+    }
+
+    const data = event.data;
+    if (data && data.method === SANDBOX_RESOURCE_READY_METHOD) {
+        const { html, sandbox } = data.params;
+        
+        if (html && content) {
```
jgindin, line 74: Should we move the retrieval of the content to here? If we retrieve it above, and it's added dynamically, wouldn't it be possible that it doesn't exist when retrieved on line 24, but could be present if retrieved "just in time"?

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,228 @@
...
+    let rawContent = super.resolvePrimitive(this.content() ?? null);
+    if (rawContent && rawContent.startsWith('url_encoded:')) {
+      rawContent = decodeURIComponent(rawContent.substring(12));
+    }
+    return rawContent;
+  });
+
+  private readonly contentUpdateEffect = effect(() => {
```
jgindin: Nit: Do we need to embed the type of the variable (effect) in the name? Can we just name this as `contentUpdate` or `contentUpdater` ?

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: Done.

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
sugoi-yuzuru: `sandbox` property is an iframe attribute. By default (i.e., `''`), the sandbox iframe will not allow anything. Hence, by default the iframe in this `sandbox` component will be as vacuumed environment as possible. `sandbox` argument is an allow-list instead of a block-list. 

Since this is inherited naming from the MCP Apps repo codebase, I think I will keep it as it is. 

Calling this html files as `sandbox.html/ts` might seem loaded, but IMO it is appropriate as its primary functionality is to serve a sandboxed environment for the McpApp component. Calling it otherwise could result in further confusion. 

In reality, the `sandbox.html` is nothing more than an webpage that only contains a sandboxed iframe. If you see it so, i think the naming is reasonable. 

LMK if you think there is any additional information I might be able to add to README.md to better convey this unit.

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
sugoi-yuzuru: Done.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: done.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: I think you were correct. I have added a block of code to close the bridge if there is one already.

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
jgindin: Actually, adding what you just wrote as a comment in the code here would help a lot!

samples/client/angular/projects/orchestrator/public/sandbox_iframe/README.md
```diff
@@ -0,0 +1,21 @@
+# Sandbox Iframe
```
jgindin, line 1: Please format this file.

Lines should not be more than 80 characters.

samples/client/angular/projects/orchestrator/public/sandbox_iframe/sandbox.ts
sugoi-yuzuru: Got it!
Added.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,238 @@
...
+      bridge.sendSandboxResourceReady({
+        html: rawContent,
+        sandbox: 'allow-scripts',
+      }).catch(err => console.error('Failed to update sandbox content:', err));
+    }
+  });
+
+  readonly allowedTools = input<string[]>([]);
+  readonly title = input<Primitives.StringValue | null>();
+  protected readonly resolvedTitle: Signal<string | null> = computed(() =>
```
jgindin: Sorry for missing this, but here too -- we could just have `resolvedTitle = computed<string|null>(() => ....);

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,238 @@
...
+  private appBridge = signal<AppBridge | null>(null);
+  private messageHandler: ((event: MessageEvent) => void) | null = null;
+
+  ngOnInit() {
+    this.setupSandbox();
+  }
+
+  ngOnDestroy() {
+    if (this.messageHandler) {
+      window.removeEventListener('message', this.messageHandler);
```
jgindin, line 119: Rather than repeating this code below in `setupSandbox`, should we just have a single `cleanup()` method?

Also gives us a single place to add other cleanup as necessary.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -0,0 +1,238 @@
...
+      // The second argument is the source window to validate messages FROM (also the iframe).
+      iframe.contentWindow!,
+      iframe.contentWindow!,
+    );
+    await bridge.connect(transport);
+
+    // Clean up old bridge to prevent memory leaks and zombie event listeners.
+    const oldBridge = this.appBridge();
+    if (oldBridge) {
+      oldBridge.close().catch(e => console.error('Error closing previous AppBridge:', e));
```
jgindin, line 233: Might be getting down into the weeds, but if `close` doesn't remove the references we've setup to the various `onxxxx` functions, then it might not get properly garbage collected....

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: I think I will keep as is. 

The reason is that through previous comments we set this component up to be somewhat resilient to methods being called in unexpected order. This adjustment was unnecessary for the component at the time being (and now), but added a layer of robustness for when future changes result in calls that are not in the existing order. The adjustments have added cleanup code in exact locations where it may be appropriate. 

Now, if I compile these codes into single atomic `cleanup()` method, I believe we might be losing the robustness we added there. E.g., for case when `initializeBridge` is called multiple times without the paird setupSandbox. 

As for the above reason I will lean towards NOT to compile these cleanup methods into a single call for the sake of reducing potential code duplication.

samples/client/angular/projects/orchestrator/src/a2ui-catalog/mcp-app.ts
sugoi-yuzuru: `onxxxx` are properties, and `close()` should be already designed to be cleaning this up. 
If not, this is a bug on `Mcp Apps` repo (https://github.com/modelcontextprotocol/ext-apps). 

We are in align with the specs and it is likely that these garbage is collected appropriately.

PR #803, opened 2026-03-09
Remove CODEOWNERS

gspencergoog:
# Description

Removing the CODEOWNERS file because we've decided it's just too much noise for reviewers.


gspencergoog:
> ## Code Review
> Have you considered refining the `CODEOWNERS` rules instead of removing the file? For example, the wildcard rule on line 15 (`* @dmandar ...`) is very broad and could be the primary source of noise. Perhaps removing or narrowing that rule, while keeping the more specific directory-based rules, would be a better compromise to achieve the goal of reducing noise without losing the benefits of code ownership.

The rules were already removed, but having CODEOWNERS spams all owners for every PR, and with the rules turned off already, this is no longer valuable.

jacobsimionato (approved):

PR #809, opened 2026-03-10
Revamp and expand the theming documentation based on learnings.

jgindin:
# Description

Add a lot more documentation, going into more detail on how theming actually works.

## Pre-launch Checklist

- [x ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md


docs/guides/theming.md
```diff
@@ -1,187 +1,559 @@
...
+actual code rendering the UI?
 
-TODO: Add platform-specific theming guides:
+The top-level `theme` property defined by the catalog acts as an **input
+argument** to the underlying component when it is rendered. When the agent sends
+a valid `theme` object according to the catalog schema, these semantic hints are
+passed down as **Component Overrides** via the `additionalStyles` record.
 
-**Web (Lit):**
+### Dynamic Styles via BeginRenderingMessage
```
jacobsimionato, line 125: Isn't this the same as above? The 'styles' / 'theme' property on beginRendering / createSurface is defined by the catalog and it's what the agent uses to express theme.

docs/guides/theming.md
```diff
@@ -1,187 +1,559 @@
...
+application can map them via the `additionalStyles` record.
+
+## 2. What Can Be Specified in a Client Theme?
+
+On the web clients (Angular, Lit), an A2UI `Theme` is defined as a TypeScript
+object that maps A2UI components and HTML elements to CSS classes or inline
+styles. On mobile native clients like Flutter, these semantic hints are mapped
+directly to global or widget-specific `ThemeData` configurations instead of CSS.
+
+The `Types.Theme` object consists of structurally predefined properties:
```
jacobsimionato, line 166: I actually think that we should restructure this for v0.9. Having a monolithic theme object for the basic catalog which then has sections for each component makes things less modular. Now if I want to take some components from the basic catalog and use them elsewhere, I need to define some new theme object. Instead, these theme settings should ideally be passed into the constructor of a Component instance when you instantiate it in an app's configuration. E.g. imagine something like:

```
myNewCatalog = createCatalog({
components: [
  BasicCatalog.Text(defaultSize: 10, color: 0xFFF)
]
})
```

WDYT?

jacobsimionato: Or maybe some CSS mapping should be passed in.

docs/guides/theming.md
```diff
@@ -1,187 +1,559 @@
...
+`ThemeData` objects and widget-specific theme overrides instead of CSS classes.
 
-## Layer 1: Semantic Hints
+## The A2UI Styling Philosophy
 
-Agents provide semantic hints (not visual styles) to guide client rendering:
+A2UI follows a **client-controlled styling** approach, meaning:
+
+- **Agents describe _what_ to show** (components and structure).
+- **Clients decide _how_ it looks** (colors, fonts, spacing).
```
jonsharkey, line 18: This is up the catalog in a way, no?
A catalog could expose color, font, and spacing params to its components if it wants to.

docs/guides/theming.md
```diff
@@ -1,187 +1,559 @@
...
+      "text": {
+        "literalString": "Welcome"
+      },
       "usageHint": "h1"
     }
   }
 }
 
-// ❌ Bad: Visual properties (not supported)
+// ❌ Bad: Visual properties. Breaks responsiveness and brand consistency.
```
jonsharkey, line 508: I guess this addresses my comment above: https://github.com/google/A2UI/pull/809/changes#r2918577043

Here we say you CANNOT do what I suggested there. But this isn't enforced, right? We're just strongly suggesting it?

docs/guides/theming.md
jonsharkey: See my follow up comment here: https://github.com/google/A2UI/pull/809/changes#r2918598387

docs/guides/theming.md
```diff
@@ -1,187 +1,559 @@
...
+   system. This layer guarantees that agent UI remains visually consistent and
+   secure.
+
+## 1. Theming in the Catalog Definition
+
+A Catalog in A2UI is defined using a JSON Schema (based on
+`a2ui_client_capabilities.json`). The catalog definition dictates what
+components and APIs an agent can use.
+
+**Why would an agent need to send theme properties?** While the client
```
jonsharkey, line 58: Is this using BeginRenderingMessage.styles?

docs/guides/theming.md
jgindin: The catalog can provide `theme` data, but it's still up to the client to decide how it actually gets applied.

docs/guides/theming.md
jgindin: Most likely, but it could also be the catalog's `theme` data.

docs/guides/theming.md
jgindin: The `beginRendering`'s `styles` are limited to what's in the catalog's `theme`?

Hmmm... That's not something I had realized.

Please confirm this and then I'll update the doc.

jgindin: But...I thought that the catalog's `theme` was meant to provide "hints" for rendering all the components from this catalog, and the `beginRendering` `styles` were meant as hints for this particular surface.

E.g., if _this_ surface is rendering some sort of interactive UI for Coca-cola, it would want to use, e.g., it's own color scheme.

docs/guides/theming.md
jgindin: I absolutely agree that theming should be rethought. This is confusing.

I think we should have a brainstorming session on this ASAP.

docs/guides/theming.md
jgindin: Right -- this is _bad_, but we can't stop someone from building a component that allows these sorts of properties.

docs/guides/theming.md
jgindin: @jacobsimionato Can you comment?

docs/guides/theming.md
jacobsimionato: Hey yes! I think there are two separate issues here:

## Agent-driven theming
The catalog defines  a theme schema, and createSurface must conform to it. In v0.8, it's the same, but it's called "styles" and it's set in beginRendering. See https://github.com/google/A2UI/blob/main/specification/v0_9/docs/a2ui_protocol.md#envelope-message-structure for background.

Let's stick with this mechanism for now because it's in the spec, but I agree this does hinder the ability to flexibly create new catalogs from existing catalog items, because the theme schema is defined at the catalog level, so if you merge two catalogs, there's no obvious way to combine the theming. I kind of imagine one of the following paths:
- Multiple catalogs end up at least reusing the same typical theme parameters, a sort of duck typing.
- Components reused across many catalogs just ignore agent-driven theming. E.g. you can add the GoogleMap component to your catalog which has a sophisticated custom theming system, but it's not going to adhere to your theming.

Though now this makes me think of a third option: some ability to programmatically define a binding between some catalog theming schema, and the schema accepted by some component like GoogleMap. This could be an interesting problem to solve. I think the solution could be entirely inside the GoogleMap implementation.

## Client controlled theming
The client codebase defines styles and passes them into the components programatically when configuring the renderer. This is what I was describing above. This might typically target properties which are not set by the agent, e.g. finer grained properties like specific fonts and sizes, border radiuses etc.

ditman:
I did a quick update to the theming docs to reflect the styling for the basic catalog v0.9 styling via CSS variables, here: https://github.com/google/A2UI/pull/1231. We may be able to close this? (Or we can port any improvements from this PR that are missing from the current documentation over!)

jacobsimionato:
Closing this because I believe its out of date now.

PR #817, opened 2026-03-11
New assemble_catalog.py script.

jgindin:
# Description

Much more robust and comprehensive than the previous build_catalog.py script:
  * Combines multiple catalogs into one
  * Inputs can be local files or HTTP URLs
  * Easy to extend the basic catalog with just a simple flag
  * Integrated handling for references to either basic_catalog.json or common_types.json without having to provide them as explicit inputs
  * Protect against circular references
  * Detects collisions

## Pre-launch Checklist

- [x ] I signed the [CLA].
- [x ] I read the [Contributors Guide].
- [x ] I read the [Style Guide].
- [x ] I updated/added relevant documentation.
- [x ] My code changes (if any) have tests.

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/tools/build_catalog/assemble_catalog.py
+++ b/tools/build_catalog/assemble_catalog.py
@@ -260,14 +260,15 @@ class CatalogAssembler:
             logger.warning(f"{label} collision: '{k}' already exists. Overwriting.")
           target[category][k] = v
 
-  def _init_combined_catalog(self, base_name: str, file_name: str, input_uris: list[str]) -> dict[str, Any]:
-    """Initializes the skeleton of the combined catalog."""
+  def _init_combined_catalog(self, base_name: str, file_name: str, input_uris: list[str], custom_catalog_id: Optional[str] = None) -> dict[str, Any]:
+    """Initializes the skeleton for the newly combined catalog."""
+    catalog_id = custom_catalog_id if custom_catalog_id else f"urn:a2ui:catalog:{base_name}"
     return {
         "$schema": "https://json-schema.org/draft/2020-12/schema",
         "$id": file_name,
         "title": f"{base_name} A2UI Catalog",
         "description": f"{base_name} A2UI catalog, including {', '.join(Path(uri).stem for uri in input_uris)}.",
-        "catalogId": file_name,
+        "catalogId": catalog_id,
         "components": {},
         "functions": {},
     }
@@ -310,10 +311,13 @@ class CatalogAssembler:
       if synthesized_name == "anyComponent":
         self.definitions[synthesized_name]["discriminator"] = {"propertyName": "component"}
 
-  def assemble(self, name: str, input_uris: list[str], extend_basic: bool = False) -> dict[str, Any]:
-    """Assembles multiple catalogs into one."""
+  def assemble(self, name: str, input_uris: list[str], extend_basic: bool = False, catalog_id: Optional[str] = None) -> dict[str, Any]:
+    """Assembles a list of catalog URIs into a single, unified catalog JSON."""
+    if not input_uris:
+      return {}
+
     file_path = Path(name).with_suffix(".json")
-    combined_catalog = self._init_combined_catalog(file_path.stem, file_path.name, input_uris)
+    combined_catalog = self._init_combined_catalog(file_path.stem, file_path.name, input_uris, custom_catalog_id=catalog_id)
 
     basic_uri = self._normalize_uri(self.local_basic_catalog_path or BASIC_CATALOG_URLS[self.version])
     merged_theme, theme_sources = self._load_initial_theme(basic_uri)
@@ -373,6 +377,7 @@ def main():
   parser = argparse.ArgumentParser(description="Assemble multiple A2UI Catalogs into a single file.")
   parser.add_argument("inputs", nargs="+", help="Input paths or URLs to A2UI component catalog JSONs")
   parser.add_argument("--output-name", required=True, help="Name of the combined catalog")
+  parser.add_argument("--catalog-id", type=str, help="Custom catalogId for the output. Defaults to urn:a2ui:catalog:<base_name>")
   parser.add_argument("--version", choices=["0.9", "0.10"], default="0.9", help="A2UI basic_catalog version to use if remote")
   parser.add_argument("--extend-basic-catalog", action="store_true", help="Always include the entire basic_catalog.json in the output")
   parser.add_argument("--out-dir", "-o", type=Path, default="dist", help="Output directory (default: dist)")
@@ -397,7 +402,7 @@ def main():
         local_basic_catalog_path=local_basic,
         local_common_types_path=local_common,
     )
-    final_schema = assembler.assemble(output_filename, args.inputs, extend_basic=args.extend_basic_catalog)
+    final_schema = assembler.assemble(output_filename, args.inputs, extend_basic=args.extend_basic_catalog, catalog_id=args.catalog_id)
 
     validate_catalog(final_schema)
 
--- a/tools/build_catalog/tests/test_assemble_catalog.py
+++ b/tools/build_catalog/tests/test_assemble_catalog.py
@@ -37,7 +37,7 @@ class TestAssembleCatalog(unittest.TestCase):
         result = assembler.assemble("TestCatalog", [str(self.component1_path), str(self.component2_path)])
 
         self.assertEqual(result["$id"], "TestCatalog.json")
-        self.assertEqual(result["catalogId"], "TestCatalog.json")
+        self.assertEqual(result["catalogId"], "urn:a2ui:catalog:TestCatalog")
         self.assertEqual(result["title"], "TestCatalog A2UI Catalog")
         self.assertEqual(result["description"], f"TestCatalog A2UI catalog, including {self.component1_path.stem}, {self.component2_path.stem}.")
         self.assertIn("CustomHeader", result["components"])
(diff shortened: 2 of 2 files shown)


tools/build_catalog/README.md
```diff
@@ -1,37 +1,131 @@
...
+A2UI v0.9+ requires catalogs be free standing, except for references to
+`common_types` which are automatically resolved by the A2UI SDK, to simplify LLM
+inference and dependency management.
 
-A2UI v0.9+ requires catalogs be free standing, except for references to `common_types` which are automatically resolved by the A2UI SDK, to simplify LLM inference and dependency management.
+This directory contains two tools for managing and bundling A2UI catalogs:
 
-## Use
+1. `assemble_catalog.py` (New, Recommended)
+2. `build_catalog.py` (Legacy)
```
nan-yu: Should we remove the legacy version directly?

tools/build_catalog/README.md
```diff
@@ -1,37 +1,131 @@
...
+### Usage
+
+```bash
+uv run tools/build_catalog/assemble_catalog.py [INPUTS ...] --name <OUTPUT_NAME> [--version <VERSION>] [--extend-basic-catalog] [--out-dir <DIR>] [--verbose]
+```
+
+### Arguments
+
+- `inputs`: One or more paths or URLs to A2UI component catalog JSONs.
+- `--name`: (Required) The desired name of the combined catalog (e.g.
```
nan-yu: What about `--output-name` to be more specific?

tools/build_catalog/assemble_catalog.py
```diff
@@ -0,0 +1,371 @@
...
+
+    return combined_catalog
+
+
+def validate_catalog(catalog: dict):
+  """Optional validation using jsonschema if available."""
+  try:
+    import jsonschema
+    # Basic check to see if it's a valid JSON schema itself
+    jsonschema.Draft202012Validator.check_schema(catalog)
```
nan-yu, line 357: The agent SDK has a more advanced validator that validates not only the schema, but other component integrity checks. We can leverage the existing validation.

tools/build_catalog/assemble_catalog.py
```diff
@@ -0,0 +1,371 @@
...
+    self.version = version
+    self.local_basic_catalog_path = local_basic_catalog_path
+    self.local_common_types_path = local_common_types_path
+    self.max_depth = max_depth  # Protects against circular $ref dependencies
+    self.definitions = {}
+    self.ref_mapping = {}
+    self.used_def_keys = set()
+    self.file_cache = {}
+
+  def fetch_json(self, uri: str) -> dict:
```
nan-yu: Should we consider migrating `assemble_catalog.py` into the `agent_sdks/python` directory? By leveraging the existing [packing hook](https://github.com/google/A2UI/blob/1ca2bd1eae18dcdc6e7ddd6db3cb2e2b7a99cb91/agent_sdks/python/pack_specs_hook.py) to bundle the specification into the SDK, we can then [load](https://github.com/google/A2UI/blob/1ca2bd1eae18dcdc6e7ddd6db3cb2e2b7a99cb91/agent_sdks/python/src/a2ui/basic_catalog/provider.py#L24) them as bundled resources. This would eliminate the network dependency entirely.

tools/build_catalog/assemble_catalog.py
```diff
@@ -0,0 +1,371 @@
...
+              f"{category.title()[:-1]} collision: '{k}' already exists. Overwriting.")
+          target[category][k] = v
+
+  def assemble(self, name: str, input_uris: list[str],
+      extend_basic: bool = False) -> dict:
+    """Assembles multiple catalogs into one, optionally extending basic_catalog."""
+    file_path = Path(name).with_suffix(".json")
+    base_name = file_path.stem
+
+    combined_catalog = {
```
nan-yu: [server_to_client.json](https://github.com/google/A2UI/blob/a3a96fc1d303d518fe8c706188baec1f9f5d35e3/specification/v0_10/json/server_to_client.json#L68) references components via `"catalog.json#/$defs/anyComponent"`. Hence, `combined_catalog` should have a `$defs/anyComponent` field with merged components.

tools/build_catalog/README.md
jgindin: Removed!

tools/build_catalog/README.md
jgindin: Done.

tools/build_catalog/assemble_catalog.py
jgindin: Done. And also made sure there's `anyFunction` and the `theme`s get merged properly too.

tools/build_catalog/assemble_catalog.py
jgindin: I think that might be a good idea.

Can we get this PR in first before looking into that, though?

tools/build_catalog/assemble_catalog.py
jgindin: If we move this script into the SDK, then I think it makes sense to move to that.

WDYT?

tools/build_catalog/README.md
nan-yu: Thanks for updating the python code. The doc should be updated as well.

tools/build_catalog/README.md
```diff
@@ -1,161 +1,80 @@
...
+- `--out-dir`, `-o`: The directory where the assembled catalog will be saved. Defaults to `dist`.
+- `--verbose`, `-v`: If passed, enables verbose debug logging to help diagnose issues.
+
+### Examples
+
+**Combine two local catalogs:**
 
 ```bash
-$ uv run tools/build_catalog/build_catalog.py specification/v0_9/json/sample_popup_catalog.json
+uv run tools/build_catalog/assemble_catalog.py component1.json component2.json --name merged_catalog
```
nan-yu: --out-name

tools/build_catalog/assemble_catalog.py
```diff
@@ -0,0 +1,423 @@
...
+          target[category][k] = v
+
+  def _init_combined_catalog(self, base_name: str, file_name: str, input_uris: list[str]) -> dict[str, Any]:
+    """Initializes the skeleton of the combined catalog."""
+    return {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "$id": file_name,
+        "title": f"{base_name} A2UI Catalog",
+        "description": f"{base_name} A2UI catalog, including {', '.join(Path(uri).stem for uri in input_uris)}.",
+        "catalogId": file_name,
```
nan-yu: The `catalogId` serves as a contract between the agent and the client. Should we formalize it as a URI-style identifier (e.g., `urn:a2ui:catalog:${file_name}`)? Additionally, should we allow users to provide a custom value, or should the system enforce a specific naming convention?

tools/build_catalog/README.md
jgindin: Thanks for catching! It's been fixed.

tools/build_catalog/assemble_catalog.py
jgindin: I added a `--catalog-id` command line argument. If not specified, we'll use your suggestion: `urn:a2ui:catalog:<name>`

tools/build_catalog/tests/test_assemble_catalog.py
jacobsimionato, line 1: Just confirming that these will run on CI?

tools/build_catalog/README.md
```diff
@@ -1,153 +1,99 @@
...
+
+```bash
+uv run tools/build_catalog/assemble_catalog.py [INPUTS ...] --output-name <OUTPUT_NAME> [--version <VERSION>] [--extend-basic-catalog] [--out-dir <DIR>] [--verbose]
 ```
 
-**2. Run `uv run build_catalog.py <path-to-your-catalog` to bundle all external file references into a single, independent JSON Schema file**
+### Arguments
 
-Example running build_catalog on the sample catalog
+- `inputs`: One or more paths or URLs to A2UI component catalog JSONs.
```
jacobsimionato, line 46: Where is catalog ID set? I think we should force people to pick new catalog ids when they modify a catalog!

jacobsimionato (commented):
LGTM overall - awesome to have the workflow emphasize the ability to flexibly create catalogs from existing components.

jacobsimionato (approved):

tools/build_catalog/assemble_catalog.py
jacobsimionato: +1 to this direction. I think this script is nice for some workflows, but it'd be great to unify it with the rest of the python ecosystem.

One advantage is that we can centralize our concepts of Component and Catalog. I imagine each major language will have one of these eventually, and if we all use the same model classes, then our logic will be modular and interoperable and we can mix and match different modules that we have. For example, we have Catalog and ComponentApi defined in TS here: https://github.com/google/A2UI/blob/main/renderers/web_core/src/v0_9/catalog/types.ts

Another advantage is that if we model this ability to merge together catalogs and components nicely, it will unlock people doing what this script does programmatically. Let's say I own an agent server which has the basic catalog augmented with a special GoogleMaps component . One option is that I get all the json files, run the script to make a new JSON file, then embed it in my server. Another, possibly more elegant option, is to programmatically in Python say:
```py
myNewCatalog = Catalogs.merge(catalogs=[BASIC_CATALOG], components=[GOOGLE_MAPS_COMPONENT], id='myAugmentedCatalog')

inferenceStrategy = StreamingInferenceStrategy(myNewCatalog);
```

That way, there is no build-time step etc.

tools/build_catalog/README.md
jgindin: They can set it by using the `--catalog-id` command line argument. If not specified, it gets assigned one of the form, `urn:a2ui:catalog:<base_name>`

tools/build_catalog/tests/test_assemble_catalog.py
jgindin: Not sure what you mean?

tools/build_catalog/assemble_catalog.py
nan-yu: Sounds good.

PR #820, opened 2026-03-11
Support for version 0.9 of the A2UI protocol in the Angular renderer

gspencergoog:
# Description

## Overview
This Pull Request introduces initial support for version 0.9 of the A2UI protocol in the Angular renderer and provides updates/strict validations in `web_core`. This is the **first part** of a two-part change, with the second part following in the `angular_renderer` branch.  It leaves the samples using v0.8, and a later change will convert them to use 0.9.

There some minimal import changes necessary, which is a breaking change, since v0.9 is now the default, and 0.8 needs to be explicitly imported.

### Angular Renderer (`renderers/angular`)
- **Versioned Organization**: Existing v0.8 files are moved and isolated into `src/lib/v0_8/` to make room for future proofing.
- **v0.9 Module Support**: Created a canonical `v0_9/` directory containing:
    - **Components**: Fresh implementations of v0.9 compliant components (such as `audio`, `button`, `card`, `checkbox`, `choice-picker`, `column`, `divider`, `icon`, `image`, `list`, `modal`, `slider`, `surface`, `tabs`, `text`, `text-field`, `video`).
    - **Rendering Pipeline**: Completely updated `dynamic-component`, `renderer`, and `id-generator` node traversal strategies.
    - **Styling Utility**: Introduced `styles.ts` containing ported CSS utility tokens for layout, palette-mapping, and response behaviors.
    - **Data Processing**: Enhanced context isolation patterns (`data-context.ts`) and markdown parsing hooks.
- **Default Baseline Setup**: `public-api.ts` now defaults to exporting the v0.9 pipeline set.

### Web Core Maintenance (`renderers/web_core`)
- **Strict Context Validation**: Upgraded function handlers inside `BASIC_FUNCTIONS` (arithmetic, logical operators, string lookups, validations) to explicitly throw `A2uiExpressionError` on missing or malformed arguments, preventing undefined fallback behavior passes.

### Samples & Configurations (`samples/client/angular`)
- **Client Alignments**: Patched node configurations across sample directories (`orchestrator`, `gallery`, `rizzcharts`, `contact`) to consume the v0.9 renderer exports safely.
- **Rizzcharts Fixes**: Included fixes for fallback state behaviors that prevent rendering layout snaps.


renderers/angular/src/lib/v0_9/components/surface.ts
```diff
@@ -0,0 +1,116 @@
...
+        styles['--p-50'] = value;
+        styles['--p-40'] = `color-mix(in srgb, ${value} 80%, black 20%)`;
+        styles['--p-35'] = `color-mix(in srgb, ${value} 70%, black 30%)`;
+        styles['--p-30'] = `color-mix(in srgb, ${value} 60%, black 40%)`;
+        styles['--p-25'] = `color-mix(in srgb, ${value} 50%, black 50%)`;
+        styles['--p-20'] = `color-mix(in srgb, ${value} 40%, black 60%)`;
+        styles['--p-15'] = `color-mix(in srgb, ${value} 30%, black 70%)`;
+        styles['--p-10'] = `color-mix(in srgb, ${value} 20%, black 80%)`;
+        styles['--p-5'] = `color-mix(in srgb, ${value} 10%, black 90%)`;
+        styles['--0'] = '#000000'; // Fixed typo #00000 -> #000000
```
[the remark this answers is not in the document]
gspencergoog: Fixed

renderers/angular/src/lib/v0_9/data/data-context.ts
jacobsimionato, line 1: This is defined in web_core - we should use it from there!

renderers/angular/src/lib/v0_9/data/processor.ts
```diff
@@ -0,0 +1,116 @@
...
+  };
+}
+
+export interface DispatchedEvent {
+  message: A2uiClientMessage;
+  completion: Subject<Types.A2uiMessage[]>;
+}
+
+@Injectable({ providedIn: 'root' })
+export class MessageProcessor extends A2uiMessageProcessor<any> {
```
jacobsimionato, line 55: This should also be in web_core!

renderers/angular/src/lib/v0_9/rendering/dynamic-component.ts
```diff
@@ -0,0 +1,233 @@
...
+})
+export abstract class DynamicComponent<T extends Types.AnyComponentNode = Types.AnyComponentNode> implements OnInit, OnDestroy {
+  protected readonly id = `a2ui-${idCounter++}`;
+  protected processor = inject(A2UI_PROCESSOR) as MessageProcessor;
+  protected readonly theme = inject(Theme);
+  protected cdr = inject(ChangeDetectorRef);
+
+  readonly surfaceId = input<Types.SurfaceID>();
+  readonly component = input.required<T>();
+  readonly weight = input.required<string | number>();
```
jacobsimionato, line 41: This is part of the basic catalog, not the spec, so it shouldn't be here

renderers/angular/src/lib/v0_9/rendering/renderer.ts
```diff
@@ -0,0 +1,143 @@
...
+import { MessageProcessor } from '../data';
+
+@Directive({
+  selector: 'ng-container[a2ui-renderer]',
+})
+export class Renderer implements OnDestroy {
+  private viewContainerRef = inject(ViewContainerRef);
+  private catalog = inject(CatalogToken);
+
+  private processor = inject(A2UI_PROCESSOR);
```
jacobsimionato, line 46: This should accept just SurfaceModel, not a2uiprocessor overall

renderers/angular/src/lib/v0_9/components/button.ts
```diff
@@ -0,0 +1,117 @@
...
+  `,
+  styles: `
+    :host {
+      display: block;
+      flex: var(--weight);
+      min-height: 0;
+    }
+  `,
+})
+export class Button extends DynamicComponent<Types.ButtonNode> {
```
jacobsimionato, line 46: We want to stop depending on these types and try to completely separate the component implementations from the core renderer codebase.

jacobsimionato (commented):
Hey I think the v0.9 code here still uses very v0.8 patterns.. I wonder if it's worth vibecoding a new v0.9 renderer from scratch, as an experiment. I tried it out for React here: https://github.com/google/A2UI/pull/827

renderers/angular/src/lib/v0_9/data/data-context.ts
gspencergoog: This  is a leftover file that isn't actually used.  Removed.

renderers/angular/src/lib/v0_9/components/audio.ts
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+import { ChangeDetectionStrategy, Component, computed, input } from '@angular/core';
+import { DynamicComponent } from '../rendering/dynamic-component';
+import * as Primitives from '@a2ui/web_core/types/primitives';
+
+@Component({
+  selector: 'a2ui-audio',
+  changeDetection: ChangeDetectionStrategy.Eager,
```
jgindin, line 23: Why not make the components `OnPush`?

gspencergoog:
> Hey I think the v0.9 code here still uses very v0.8 patterns.. I wonder if it's worth vibecoding a new v0.9 renderer from scratch, as an experiment. I tried it out for React here: #827

Sure, I can give that a try.  This started from the 0.8 source, which is probably why it's using those patterns.  I've been trying to cure it of some of the patterns (treating the basic_catalog as separate, making things use web_core), but not entirely successful, as you see.

renderers/angular/src/lib/v0_9/rendering/dynamic-component.ts
```diff
@@ -0,0 +1,233 @@
...
+import * as Primitives from '@a2ui/web_core/types/primitives';
+import { DataContext as WebCoreDataContext } from '@a2ui/web_core/v0_9';
+import { Types } from '../types';
+import { Directive, inject, input, signal, OnInit, OnDestroy, ChangeDetectorRef } from '@angular/core';
+import { A2UI_PROCESSOR } from '../config';
+import { Theme } from './theming';
+import { MessageProcessor, A2uiClientMessage } from '../data';
+import { ComponentModel } from '@a2ui/web_core/v0_9';
+
+let idCounter = 0;
```
jgindin, line 26: Can this just be a `static private` member of the class?

renderers/angular/src/lib/v0_9/rendering/dynamic-component.ts
```diff
@@ -0,0 +1,233 @@
...
+
+@Directive({
+  host: {
+    '[style.--weight]': 'weight()',
+  },
+})
+export abstract class DynamicComponent<T extends Types.AnyComponentNode = Types.AnyComponentNode> implements OnInit, OnDestroy {
+  protected readonly id = `a2ui-${idCounter++}`;
+  protected processor = inject(A2UI_PROCESSOR) as MessageProcessor;
```
jgindin, line 35: Can this be `readonly` ? (Same for `cdr`.)

renderers/angular/src/lib/v0_9/types.ts
```diff
@@ -0,0 +1,132 @@
...
+  // Link component wasn't in basic_catalog.json but it's used in types.ts.
+  // Re-adding it here until a proper migration is mapped.
+  export interface LinkProps {
+    text: string;
+    url: string;
+  }
+  export type LinkNode = Component<LinkProps>;
+  export type Link = LinkNode;
+  
+  export type CardNode = import('@a2ui/web_core/v0_9/basic_catalog').CardNode;
```
jgindin, line 130: Should this be in the list abvoe?

renderers/angular/src/lib/v0_9/data/data-context.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  getValue(path: string): any {
+    return this.model.get(this.resolvePath(path));
+  }
+
+  observe<T>(path: string): Observable<T> {
+    const resolvedPath = this.resolvePath(path);
+    return new Observable<T>((subscriber) => {
+      // Emit initial value
```
jgindin: Can we update the comment to explain _why_ we're eating the initial value?

renderers/angular/src/lib/v0_9/data/data-context.ts
```diff
@@ -0,0 +1,71 @@
...
+      subscriber.next(this.model.get(resolvedPath));
+
+      const subscription = this.model.subscribe<T>(resolvedPath, (value) => {
+        if (value !== undefined) {
```
jgindin: Does this mean that if there is a value at the path, and then it gets updated to be `undefined`, the subscribers won't know the value has been changed?

renderers/angular/src/lib/v0_9/data/processor.ts
```diff
@@ -0,0 +1,116 @@
...
+    for (const surfaceId of this.model.surfacesMap.keys()) {
+      this.model.deleteSurface(surfaceId);
+    }
+  }
+
+  // Override to handle the fact that we're using a different base class
+  // The base class processMessages expects v0.9 messages.
+  // We trust the server sends valid v0.9 messages.
+  override processMessages(messages: any[]): void {
+    console.log('[MessageProcessor] Received messages to process:', messages);
```
jgindin, line 90: Should these `console.log` lines be removed?

renderers/angular/src/lib/v0_9/rendering/dynamic-component.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    fixture.componentRef.setInput('surfaceId', 'site');
+    fixture.componentRef.setInput('component', { id: 'root', type: 'Box', properties: {} });
+    processorSpy = TestBed.inject(MessageProcessor) as jasmine.SpyObj<MessageProcessor>;
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  // Tests for resolve/data binding would need to be checked against how DynamicComponent uses them.
```
jgindin, line 65: These look like LLM-generated comments?

Should we implement additional tests?

renderers/angular/src/lib/v0_9/rendering/dynamic-component.ts
```diff
@@ -0,0 +1,233 @@
...
+
+let idCounter = 0;
+
+@Directive({
+  host: {
+    '[style.--weight]': 'weight()',
+  },
+})
+export abstract class DynamicComponent<T extends Types.AnyComponentNode = Types.AnyComponentNode> implements OnInit, OnDestroy {
+  protected readonly id = `a2ui-${idCounter++}`;
```
jgindin, line 34: Should we use `getUniqueId()` here?

Or should we `inject` the `IdGenerator` and use that here and replace the `getUniqueId` with that?

gspencergoog:
Sorry, this PR is clearly not ready.

I was excited that I got both v0.8 and v0.9 rendering working, but was premature in sending it out.

Sorry to waste your time with the reviews.

I'll try a clean implementation and see if I can get closer to something correct: I think this one suffers a lot from iteration creep and cruft where I started off with porting the v0.8 implementation to a pure v0.9 implementation, then moved a bunch of that into `web_core`, and then split things apart again so that it would do both 0.9 and 0.8.

PR #826, opened 2026-03-12
feat(web-core): implement client capabilities API in MessageProcessor

jacobsimionato:
### Description
Updates the `getClientCapabilities` implementation in the web core renderer to fully support functions and theme schemas, aligning with the A2UI specification. Restructures the TypeScript definitions for the capabilities payload to live natively within the `schema` layer, and adds comprehensive guidelines for managing protocol serialization boundaries across all renderers.

### Changes
- **Client Capabilities Typings**: Extracted the protocol-specific payload typings (`InlineCatalog`, `FunctionDefinition`, `JsonSchema`, `A2uiClientCapabilities`) out of `message-processor.ts` and into a dedicated `schema/client-capabilities.ts` file, establishing strong deep typings corresponding to the `client_capabilities.json` schema.
- **Catalog Types**: Added an optional `themeSchema` property to the `Catalog` class and updated its constructor.
- **Capabilities Generation**: Updated `MessageProcessor.generateInlineCatalog` to successfully serialize `functions` and `themeSchema` into the client capabilities object. Applied `processRefs` to these schemas to ensure correct JSON Schema references.
- **Schema References**: Added `REF:` tags to shared schemas in `common-types.ts` (e.g., `DataBinding`, `FunctionCall`, `DynamicValue`, `ChildList`, `Action`, `Checkable`) to enable automatic `$ref` node generation during capability exchange.
- **Testing**: Added comprehensive test cases to `message-processor.test.ts` covering edge cases for `REF:` tags, multiple catalogs, nested references, and empty schema states.
- **Documentation (`renderer_guide.md`)**: 
  - Updated the capabilities serialization logic for future A2UI renderer implementations.
  - Added a new "**Protocol Models & Serialization**" architecture section. This explicitly defines the boundary between raw JSON network streams and internal state models, instructing future renderer authors to build strict data classes (like `A2uiMessage` or `A2uiClientCapabilities`) that handle serialization/validation *before* passing objects into the `MessageProcessor`.

### DIFF
--- a/renderers/web_core/src/v0_9/processing/message-processor.ts
+++ b/renderers/web_core/src/v0_9/processing/message-processor.ts
@@ -32,6 +32,7 @@ import {
   A2uiClientCapabilities,
   InlineCatalog,
 } from "../schema/client-capabilities.js";
+import { A2uiClientDataModel } from "../schema/client-to-server.js";
 import { A2uiStateError, A2uiValidationError } from "../errors.js";
 
 /**
@@ -182,6 +183,28 @@ export class MessageProcessor<T extends ComponentApi> {
   }
 
   /**
+   * Returns the aggregated data model for all surfaces that have 'sendDataModel' enabled.
+   */
+  getClientDataModel(): A2uiClientDataModel | undefined {
+    const surfaces: Record<string, any> = {};
+
+    for (const surface of this.model.surfacesMap.values()) {
+      if (surface.sendDataModel) {
+        surfaces[surface.id] = surface.dataModel.get("/");
+      }
+    }
+
+    if (Object.keys(surfaces).length === 0) {
+      return undefined;
+    }
+
+    return {
+      version: "v0.9",
+      surfaces,
+    };
+  }
+
+  /**
    * Subscribes to surface creation events.
    */
   onSurfaceCreated(handler: (surface: SurfaceModel<T>) => void): Subscription {
@@ -243,7 +266,7 @@ export class MessageProcessor<T extends ComponentApi> {
 
   private processCreateSurfaceMessage(message: CreateSurfaceMessage): void {
     const payload = message.createSurface;
-    const { surfaceId, catalogId, theme } = payload;
+    const { surfaceId, catalogId, theme, sendDataModel } = payload;
 
     // Find catalog
     const catalog = this.catalogs.find((c) => c.id === catalogId);
@@ -255,7 +278,12 @@ export class MessageProcessor<T extends ComponentApi> {
       throw new A2uiStateError(`Surface ${surfaceId} already exists.`);
     }
 
-    const surface = new SurfaceModel<T>(surfaceId, catalog, theme);
+    const surface = new SurfaceModel<T>(
+      surfaceId,
+      catalog,
+      theme,
+      sendDataModel ?? false,
+    );
     this.model.addSurface(surface);
   }
 
--- a/agent_sdks/python/src/a2ui/core/parser/constants.py
+++ b/agent_sdks/python/src/a2ui/core/parser/constants.py
@@ -1,0 +1,32 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Constants for A2UI parsing."""
+
+DEFAULT_ROOT_ID = "root"
+
+# Message types (v0.8)
+MSG_TYPE_BEGIN_RENDERING = "beginRendering"
+MSG_TYPE_SURFACE_UPDATE = "surfaceUpdate"
+MSG_TYPE_DATA_MODEL_UPDATE = "dataModelUpdate"
+MSG_TYPE_DELETE_SURFACE = "deleteSurface"
+
+# Message types (v0.9)
+MSG_TYPE_CREATE_SURFACE = "createSurface"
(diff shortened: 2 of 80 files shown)


renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -52,6 +61,100 @@ export class MessageProcessor<T extends ComponentApi> {
     }
   }
 
+  /**
+   * Generates the a2uiClientCapabilities object for the current processor.
+   *
+   * @param options Configuration for capability generation.
+   * @returns The capabilities object.
+   */
+  getClientCapabilities(options?: CapabilitiesOptions): any {
```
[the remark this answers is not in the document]
jacobsimionato: Yes, let's do this. Define a type based on specification/v0_9/json/client_capabilities.json

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -52,6 +61,100 @@ export class MessageProcessor<T extends ComponentApi> {
...
+    if (options?.includeInlineCatalogs) {
+      capabilities["v0.9"].inlineCatalogs = this.catalogs.map((c) =>
+        this.generateInlineCatalog(c),
+      );
+    }
+
+    return capabilities;
+  }
+
+  private generateInlineCatalog(catalog: Catalog<T>): any {
```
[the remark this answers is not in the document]
jacobsimionato: Yes, lets use a Catalog type which we'll need to define as part of client capabilities anyway.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -52,6 +61,100 @@ export class MessageProcessor<T extends ComponentApi> {
...
+            node["description"] = desc;
+          } else {
+            delete node["description"];
+          }
+        }
+      } else {
+        this.processRefs(val);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
jacobsimionato: Let's do this.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -27,8 +28,20 @@ import {
...
+  A2uiClientCapabilities,
+  InlineCatalog,
+} from "../schema/client-capabilities.js";
 import { A2uiStateError, A2uiValidationError } from "../errors.js";
 
+/**
+ * Options for generating client capabilities.
+ */
+export interface CapabilitiesOptions {
+  /** If true, the full definition of all catalogs will be included. */
```
gspencergoog, line 42: I thought the idea for this option was to *allow* inline catalogs, not to include the full definition of all of them.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -52,6 +65,122 @@ export class MessageProcessor<T extends ComponentApi> {
...
+
+      // Re-add only the $ref and an optional description.
+      node["$ref"] = ref;
+      if (desc) {
+        node["description"] = desc;
+      }
+      return;
+    }
+
+    // If not a REF target, recurse into its children.
```
gspencergoog, line 173: What if an expanded `$ref` has a `REF:` in it?  Or is that not possible? Or is this just an incorrect comment?

renderers/web_core/package.json
```diff
@@ -88,11 +88,11 @@
...
     "typescript": "^5.8.3",
-    "wireit": "^0.15.0-pre.2",
-    "zod-to-json-schema": "^3.25.1"
+    "wireit": "^0.15.0-pre.2"
   },
   "dependencies": {
     "@preact/signals-core": "^1.13.0",
-    "zod": "^3.25.76"
+    "zod": "^3.25.76",
+    "zod-to-json-schema": "^3.25.1"
```
gspencergoog, line 100: This might be bloating the front end, though.  I was tempted to add this in the past too, but decided against it in order to keep the front end lean.  If you think it's worth it, we should quantify the size hit.

renderers/web_core/src/v0_9/processing/message-processor.ts
jacobsimionato: Well, for now this API just turns on sending the full capabilities for every catalog. In the future, we might want some more fine-grained API to say "just send full inline catalogs for these three schemas", but I figure this is a simple starting point? WDYT?

renderers/web_core/src/v0_9/processing/message-processor.ts
jacobsimionato: What this is doing is kind of the inverse of refs actually. It's taking fully expanded Zod schemas and then refactoring them to include references to common_types when they are detected. If this happens, e.g. it sees a DynamicString and replaces it with a reference, there is no more work to do. If not, then it needs to keep exploring, looking for internal refs further into the tree of properties.

renderers/web_core/package.json
jacobsimionato: Well, we have decided to use Zod to represent schemas, and the A2UI specification supports inline catalogs, requiring those Zod schemas to be converted to JSON. So I think our choices are:
A. Add the dep
B. Write our own more lightweight version of the same conversion
C. Not support inlineCatalogs at all
D. Support inlineCatalogs through an additional package we provide.

I vote we do A for now, then switch to D if people complain. WDYT?

gspencergoog: Yeah, let's add the dep for now.  If it's too bloated, we can evalutate B.

renderers/web_core/src/v0_9/processing/message-processor.ts
gspencergoog: Yeah, I guess start here and we can get more complicated/refined as time goes on.  I was just thinking about how it is including these with every message and thinking it was a lot.

jacobsimionato: SGTM!

renderers/web_core/package.json
jacobsimionato: SGTM. I noticed that zod 4 has this functionality built into the Zod package natively, so I don't know if there's a benefit from excluding it anyway. I guess we should update to zod 4 at at some point.

PR #831, opened 2026-03-12
Fix inline catalog validation and multi-surface support

### LINKED ISSUE #796: Inline catalog validation fails in `contact_multiple_surfaces` sample agent
## Summary

The `contact_multiple_surfaces` agent fails A2UI schema validation when the client sends `inlineCatalogs` in `a2uiClientCapabilities`. There are two issues, one in the sample agent and one in the schema manager.

## Steps to Reproduce

1. Run the `contact_multiple_surfaces` agent (`samples/agent/adk/contact_multiple_surfaces`)
2. Run the `contact` client (`samples/client/lit/contact`)
3. The client sends `inlineCatalogs` with `OrgChart` and `WebFrame` custom components
4. Validation fails with: `Additional properties are not allowed ('OrgChart' was unexpected)`

## Issue 1: `client_ui_capabilities` not passed during validation

In `samples/agent/adk/contact_multiple_surfaces/agent.py`, line 172, `get_selected_catalog()` is called **without** `client_ui_capabilities`:

```python
selected_catalog = self.schema_manager.get_selected_catalog()
```

This means validation always uses the default basic catalog, which doesn't know about the inline components (`OrgChart`, `WebFrame`). The LLM is correctly told about these components (via `agent_executor.py` which passes `client_ui_capabilities` when building the prompt), but validation checks against a schema that doesn't include them.

**Fix:** Pass `client_ui_capabilities` to `get_selected_catalog()` during validation.

## Issue 2: `_select_catalog` replaces base catalog instead of merging

Even after fixing Issue 1, validation fails with a different error: `Additional properties are not allowed ('Image' was unexpected)`. This is because the LLM uses standard components like `Image` (from the basic catalog in the system prompt) alongside inline components, but `_select_catalog` treats inline catalogs as a **full replacement** of the base catalog.

In `agent_sdks/python/src/a2ui/core/schema/manager.py`, lines 145-155:

```python
if inline_catalogs:
    inline_catalog_schema = inline_catalogs[0]
    inline_catalog_schema = self._apply_modifiers(inline_catalog_schema)
    return A2uiCatalog(
        version=self._version,
        name=INLINE_CATALOG_NAME,
        catalog_schema=inline_catalog_schema,
        s2c_schema=self._server_to_client_schema,
        common_types_schema=self._common_types_schema,
    )
```

Per the design intent (see `samples/client/lit/contact/README_CUSTOM_COMPONENTS.md`), inline catalogs are meant to be custom extensions. However, `_select_catalog` currently treats them as a full replacement of the base catalog.

### Proposed approach

Should inline catalogs be treated as **additive extensions** to a base catalog, not replacements? When `inlineCatalogs` is provided, its components would be merged on top of the agent's default catalog (`self._supported_catalogs[0]`).

This matches the design intent described in the README and how prompt construction currently works — the system prompt includes the agent's default catalog schema (`agent.py:127-135`), and inline components are appended to the query separately (`agent_executor.py:94-102`). The validation path should produce a similarly merged catalog.

Also, if the `ValueError` restriction that prevents both `inlineCatalogs` and `supportedCatalogIds` from being sent together (line 139-143) is relaxed, the base catalog could be determined by `supportedCatalogIds` instead of the agent's default, with inline components merged on top.

Is this the right approach?

sarthak96agarwal:
Fixes #796

### Summary

This PR fixes inline catalog handling in the schema manager, ensures client capabilities are extracted from all DataPart types (not just `request`), fixes an unresolvable `$ref` in the prompt builder's test JSON, and adds test coverage for the recently-landed multi-surface and incremental update validator logic.

### Problems

1. **Only the first inline catalog was merged** — The schema manager used `inline_catalogs[0]` and ignored the rest. Clients that split custom components across multiple inline catalog definitions would lose all but the first.
2. **`supportedCatalogIds` couldn't be used alongside `inlineCatalogs`** — Sending both raised a `ValueError`, but clients legitimately need both: `supportedCatalogIds` to pick the base catalog and `inlineCatalogs` to extend it with custom components.
3. **`supportedCatalogIds` was ignored when selecting the base for inline catalog merging** — The base catalog was always `_supported_catalogs[0]` even when the client specified a preferred base via `supportedCatalogIds`.
4. **`client_ui_capabilities` not extracted from `userAction` DataParts** — The executor only extracted capabilities from `request` DataParts. When a UI event fires (`userAction`), that branch is never hit, so the LLM response is validated against the wrong catalog.
5. **OrgChart inline schema had unresolvable `$ref`** — The prompt builder's test JSON used `$ref: "#/definitions/Action"` which doesn't resolve in an inline catalog context.

*Note: Multi-surface root tracking and incremental update detection were independently fixed in upstream. This PR adds test coverage for that logic.*

### Changes

**`agent_sdks/python/src/a2ui/core/schema/manager.py`**
- Merge all inline catalogs (loop), not just the first one
- Remove `ValueError` when both `inlineCatalogs` and `supportedCatalogIds` are present
- Use `supportedCatalogIds` to select which base catalog to merge onto

**`samples/agent/adk/contact_multiple_surfaces/agent_executor.py`**
- Extract `a2uiClientCapabilities` from any DataPart that has metadata, before the `userAction`/`request` branch — fixes UI events using the wrong catalog for validation

**`samples/agent/adk/contact_multiple_surfaces/prompt_builder.py`**
- Update inline catalog test JSON: inline the action schema (was unresolvable `$ref`), add `oneOf` for OrgChart chain property

**`agent_sdks/python/tests/core/schema/test_validator.py`**
- Add multi-surface validation tests (two `beginRendering` + two `surfaceUpdate` with different roots)
- Add incremental update tests: no-root passes, orphans pass, but self-refs/cycles/duplicates still fail
- All incremental update tests cover both v0.8 and v0.9

### Key Design Decisions

1. **Inline catalogs are additive extensions.** Components from each inline catalog are merged onto the base catalog via `components.update()`. This ensures standard components remain available alongside custom ones.

2. **All inline catalogs from the client are merged.** If a client sends multiple inline catalogs, all their components are merged onto the base. This supports clients that split custom components across multiple catalog definitions.

3. **Both `inlineCatalogs` and `supportedCatalogIds` are allowed simultaneously.** `supportedCatalogIds` selects the base catalog, and `inlineCatalogs` extend it.

### Validation Steps

-  All 47 validator tests pass (`pytest tests/core/schema/test_validator.py -v`)
- Run the `contact_multiple_surfaces` agent + `contact` client end-to-end
- Verify both contact card and org chart surfaces render on initial search
- Click an org chart node and verify the update validates and renders correctly
- Verify incremental data model updates (no full re-render) work without validation errors

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/agent_sdks/python/tests/core/schema/test_schema_manager.py
+++ b/agent_sdks/python/tests/core/schema/test_schema_manager.py
@@ -412,11 +412,9 @@ def test_generate_system_prompt_with_inline_catalog(mock_importlib_resources):
 
   assert "Role" in prompt
   assert "---BEGIN A2UI JSON SCHEMA---" in prompt
-  assert (
-      '### Catalog Schema:\n{\n  "$schema":'
-      ' "https://json-schema.org/draft/2020-12/schema",\n  "catalogId": "id_inline"'
-      in prompt
-  )
+  # Inline catalog is merged onto the base catalog (catalogId: "basic")
+  assert "### Catalog Schema:" in prompt
+  assert '"catalogId": "basic"' in prompt
   assert '"Button": {}' in prompt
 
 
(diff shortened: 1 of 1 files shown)


nan-yu:
The CI is currently failing. Could you please run a formatter on the python code: `uv run pyink .`?

PR #832, opened 2026-03-12
fix(lit): accept partial hinted text styles

### LINKED ISSUE #602: Lit renderer - hintedStyles assumption causes additionalStyles to be ignored
### Description

areHintedStyles() expects ["h1", "h2", "h3", "h4", "h5", "h6", "caption", "body"] and uses .every() to check if additionalStyles.Text is a hinted styles object.

This requires all 8 keys including h6 to be present. In practice, themes rarely define h6 styles (the HintedStyles interface doesn't even declare it). When any key is missing, the check returns false and the entire object is passed to styleMap() as a flat Record<string, string> — but the values are nested objects, so nothing gets applied.

### Location

lit/src/0.8/ui/text.ts [Line 125 - 132]

### Suggested fix

Change .every() to .some() and remove h6 from the expected list so any recognized key triggers hinted mode

stablegenius49:
## Summary
- stop the lit text renderer from requiring every hinted text style key before it recognizes `additionalStyles.Text` as hinted styles
- remove the impossible `h6` expectation and treat partial hinted style maps as hinted when they contain at least one recognized hinted key
- add a regression test covering partial hinted styles vs flat style maps

## Testing
- `cd renderers/web_core && npm run build`
- `cd renderers/lit && npm test`

Fixes #602

### DIFF
--- a/renderers/lit/src/0.8/hinted-styles.test.ts
+++ b/renderers/lit/src/0.8/hinted-styles.test.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


Varun-S10:
Hi @stablegenius49, Thank you for raising this PR and for your patience. We noticed that a few of the pre-checks are failing. Could you please resolve them?

jacobsimionato:
Closing because this is v0.8 which is not being extended at this point.

jacobsimionato:
Sorry we neglected this earlier @stablegenius49 !

PR #836, opened 2026-03-12
fix(shell): render text-only A2A fallbacks

### LINKED ISSUE #585: Text fallback from agent not rendered in A2UI, page refreshes silently
https://github.com/user-attachments/assets/f7d02f4d-ad2d-4d03-be00-dd17ef0650ab

When the agent fails to return data (e.g., no restaurants found, or max retries exhausted), the backend correctly sends a TextPart fallback message

Example backend log:

ERROR:agent:--- RestaurantAgent.stream: Max retries exhausted. Sending text-only error. ---

However, the frontend (A2UI) does not render this TextPart, and the UI appears to silently refresh or remain blank. Users never see the intended fallback message like:

Oops, I couldn’t find anything you requested.

Expected Behavior:
TextPart messages from the agent should be rendered in the UI, similar to how DataParts are displayed, so users see a clear error or fallback message.

Expected Behavior:
TextPart messages from the agent should be rendered in the UI, similar to how DataParts are displayed, so users see a clear error or fallback message.

Steps to Reproduce:

Query the agent with a request that triggers failure, e.g., “Chinese restaurants in Sydney” or any nonexistent location.

Observe the backend logs, which show a TextPart fallback message.

Observe the frontend: no message is displayed, and the page may refresh or remain blank.

Actual Behavior:

Backend sends the fallback TextPart.

Frontend fails to render it.

User sees no visible feedback.

Suggested Fix / Workaround:
Modify A2UI to handle TextPart messages and render them visibly. Example:

if (part.type === 'TextPart') {
    const container = document.getElementById('fallback-container');
    container.textContent = part.text;
    container.style.display = 'block';
}

stablegenius49:
## Summary

- convert text-only A2A task responses into a simple A2UI Text surface in the Lit shell client
- preserve normal A2UI data-part responses unchanged
- add regression tests for data-part passthrough and single/multi-text fallbacks

Fixes #585.

## Why

The Lit shell client currently drops `text` parts and only forwards `data` parts into the A2UI processor. When an agent returns a text-only fallback message, the UI ends up blank even though the backend sent a useful error/fallback string.

## Testing

- `npm install` in `renderers/web_core`
- `npm run build:tsc` in `renderers/web_core`
- `npm install` in `renderers/markdown/markdown-it`
- `npm run build` in `renderers/markdown/markdown-it`
- `npm install` in `samples/client/lit/shell`
- `npm run build:tsc` in `samples/client/lit/shell`
- `node --test --enable-source-maps dist/tests/client.test.js`

### DIFF
--- a/samples/client/lit/shell/client.ts
+++ b/samples/client/lit/shell/client.ts
@@ -19,6 +19,99 @@ import { A2AClient } from "@a2a-js/sdk/client";
 import { v0_8 } from "@a2ui/lit";
 
 const A2UI_MIME_TYPE = "application/json+a2ui";
+const TEXT_FALLBACK_SURFACE_ID = "@default";
+const TEXT_FALLBACK_ROOT_ID = "__a2ui_text_fallback_root";
+
+function textFallbackComponentId(index: number) {
+  return `__a2ui_text_fallback_${index}`;
+}
+
+export function convertA2APartsToMessages(
+  parts: Part[]
+): v0_8.Types.ServerToClientMessage[] {
+  const dataMessages: v0_8.Types.ServerToClientMessage[] = [];
+  const textMessages: string[] = [];
+
+  for (const part of parts) {
+    if (part.kind === "data") {
+      dataMessages.push(part.data as v0_8.Types.ServerToClientMessage);
+      continue;
+    }
+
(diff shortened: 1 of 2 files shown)


polina-c:
Thank you for your contribution. I am closing this PR as the linked issue has already been resolved.

PR #839, opened 2026-03-13
fix(lit): handle TextPart fallback messages in shell and contact sample clients

### LINKED ISSUE #585: Text fallback from agent not rendered in A2UI, page refreshes silently

Mindev27:
# Description
image

Both Lit sample clients (shell, contact) were ignoring TextPart messages
from agent responses. When the agent fails and sends back only a TextPart
fallback, the UI showed a blank screen with no feedback to the user.

Fixed by collecting TextPart in the client response and displaying it
through the existing error UI when no DataPart is available.

This is my first contribution - any feedback on the approach is welcome :)

Fixes #585


gspencergoog:
Closing because the lit contact example has been removed.

PR #856, opened 2026-03-14
feat(web_core): add `resolveAction` and improve error handling

gspencergoog:
# Changes in `renderers/web_core`

Here are some changes to web_core to prepare for the 0.9 angular renderer.

### 1. **`DataContext` (New Method)**
-   **Added**: `resolveAction(action: Action): any`
    -   **Description**: This method takes an `Action` object and resolves its top-level `DynamicValue` properties (e.g., event context values) using the current `DataContext`. This bridges the gap between static component action definitions and the actual data bound to them.

### 2. **Type-Safe Action Propagation**
-   **`SurfaceModel`**:
    -   **Change**: `onAction` stream and `dispatchAction` method are now strictly typed to use the `Action` interface from the schema instead of `any`.
-   **`SurfaceGroupModel`**:
    -   **Change**: `onAction` now emits a strictly typed `SurfaceGroupAction` (`Action & { surfaceId: string }`). Spreading an action with `surfaceId` is now statically validated for type safety.

### 3. **Error Handling (Enhanced Specificity)**
-   Functions and evaluation pipelines now throw `A2uiExpressionError` instead of generic errors or silent failures:
    -   **`DataContext`**: Both `resolveDynamicValue` and `evaluateFunctionReactive` translate `ZodError` into `A2uiExpressionError` with detail scopes (`issues` or `errors`).
    -   **`Catalog`**: `getFunction()` now catches `ZodError` during argument parsing and raises `A2uiExpressionError`.
    -   **`RegexImplementation`**: Now throws `A2uiExpressionError` on invalid regex patterns rather than returning `false` with a warning.

---

### 1. **Reactivity Upgrades (`DataContext`)**
-   **`resolveSignal`**: Substantially refactored.
    -   Switched from maintaining continuous background resolution to a structured execution model with nested `computed` argument bundles and isolated `effect` subscriptions.
    -   Ensures much safer AbortController cleanup to prevent resource leakage on unsubscription.

### 2. **`DataModel` Edge Cases**
-   **Path Forcing**: `DataModel.set` now verifies that the base reference is populated (`this.data = {}`) before traversing branches, preventing crashes on absolute root paths.

### 3. **Testing Modernization**
-   Updated continuous execution test cases in `function_execution.spec.ts` from old Node-style Node `(done)` callbacks to modern **Promise-based resolutions**.

---

## File Reorganization
-   **`FunctionInvoker`**: The type definition was moved from `src/v0_9/catalog/types.ts` into a standalone file `src/v0_9/catalog/function_invoker.ts` and re-exported via `index.ts`. This was done to avoid import circular references There are no changes to the declaration itself.


renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -209,18 +282,72 @@ export class DataContext {
...
     return signal(value as unknown as V);
   }
 
+  /**
+   * Resolves an action by evaluating its top-level dynamic values.
+   *
+   * For event actions, it resolves each value in the context map.
+   * For function call actions, it evaluates the call.
+   *
+   * This is non-recursive: it only resolves one level deep for the context record.
```
jacobsimionato: Maybe mention that this is in accordance with the spec which does not allow arbitrary nesting of context - we have `"additionalProperties": { "$ref": "#/$defs/DynamicValue" }` to force it to one layer.

The full schema is:
```json
            "event": {
              "type": "object",
              "description": "The event to dispatch to the server.",
              "properties": {
                "name": {
                  "type": "string",
                  "description": "The name of the action to be dispatched to the server."
                },
                "context": {
                  "type": "object",
                  "description": "A JSON object containing the key-value pairs for the action context. Values can be literals or paths. Use literal values unless the value must be dynamically bound to the data model. Do NOT use paths for static IDs.",
                  "additionalProperties": {
                    "$ref": "#/$defs/DynamicValue"
                  }
                }
              },
              "required": ["name"],
              "additionalProperties": false
            }
```

renderers/web_core/src/v0_9/state/surface-group-model.ts
```diff
@@ -17,6 +17,10 @@
 import { SurfaceModel } from "./surface-model.js";
 import { ComponentApi } from "../catalog/types.js";
 import { EventEmitter, EventSource, Subscription } from "../common/events.js";
+import { Action } from "../schema/common-types.js";
+
+/** An action wrapper that includes the origin surface ID. */
+export type SurfaceGroupAction = Action & { surfaceId: string };
```
jacobsimionato, line 23: I have a few changes in https://github.com/google/A2UI/pull/866 which address a similar problem in a slightly different way (adding a full A2uiClientActionSchema), though this is a strict improvement, so let's go with it and my PR can refine it if necessary.

renderers/web_core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/web_core",
-  "version": "0.8.5",
+  "version": "0.9.0",
```
jacobsimionato: Can we keep this at v0.8.5 for now, just in case @ditman and I need to push another v0.8 version of a renderer with fixes in web_core etc.

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -95,17 +102,42 @@ export class DataContext {
...
+        return (result instanceof Signal ? result.peek() : result) as V;
+      } catch (e: any) {
+        if (e?.name === "ZodError" || e instanceof z.ZodError) {
+          const err = new A2uiExpressionError(
+            `Validation failed for function '${call.call}': ${e.message}`,
+            call.call,
+            e.errors ?? e.issues,
+          );
+          this.surface.dispatchError({
+            code: "EXPRESSION_ERROR",
```
jacobsimionato, line 123: A couple of thoughts:
- Should we add this to the client_to_server.json spec in v0.10? For v0.9 could it make sense to use the VALIDATION_FAILED error type which has some structure around surfaceId and path etc? No stress for this PR, just curious about direction.

- Should we have some centralized mapping from error classes e.g. A2uiExpressionError to whatever dispatchError actually accepts, to reduce the boilerplate here?

Neither of these blocking - i'm thrilled to see error handling overall.

jacobsimionato (approved):

renderers/web_core/package.json
gspencergoog: Sure. I was thinking of bumping it with the 0.9 renderer implementation, but I don't have to.

renderers/web_core/src/v0_9/rendering/data-context.ts
gspencergoog: I thought about adding it as a type in the client-to-server schema, but expressions are part of the basic catalog, not the main spec, so I wasn't sure if it belonged there.

A centralized mapping might make sense: at least it would give us a good place to add future mappings.

renderers/web_core/src/v0_9/state/surface-group-model.ts
gspencergoog: Sure, no problem. This felt like a "minimum" solution to me: it could probably be made better with a more formalized schema.

PR #833, opened 2026-03-12
refactor: share icon ligature conversion across renderers

stablegenius49:
# Description

This consolidates the Material Symbols camelCase→snake_case conversion into a single `toMaterialSymbolLigature()` helper in `web_core`, then reuses it from the Lit and React icon renderers.

It also brings the Angular icon renderer in line with the same behavior, so icon names like `shoppingCart` and `accountCircle` resolve consistently across all three renderers.

Closes #786.

## Pre-launch Checklist

- [ ] I signed the [CLA].
- [x] I read the [Contributors Guide].
- [x] I read the [Style Guide].
- [ ] I have added updates to the [CHANGELOG].
- [ ] I updated/added relevant documentation.
- [x] My code changes (if any) have tests.

### Verification

- `node --test dist/src/v0_8/styles/icons.test.js` (web_core helper test)
- `npm test -- --run tests/unit/components/Icon.test.tsx` (React icon tests)
- `npm run typecheck` (React)
- `npm test` (Lit)
- `npm run build` (Angular)

[CHANGELOG]: ../CHANGELOG.md
[CLA]: https://cla.developers.google.com/
[Contributors Guide]: ../CONTRIBUTING.md
[Style Guide]: ../STYLE_GUIDE.md

PR #858, opened 2026-03-15
fix: centralize toSnakeCase into web_core to eliminate renderer duplication

### LINKED ISSUE #786: toSnakeCase is repeated across multiple implementations
It's a small thing, but we might want to just have one implementation that they all import.

sanjayrohith:
## Summary

Fixes #786

The `toSnakeCase` utility (used to convert camelCase icon names like
`shoppingCart` → `shopping_cart` for Material Symbols) was scattered
across renderers with no single source of truth — and worse, with
subtle behavioral differences between them.

This PR fixes that by introducing one canonical implementation in
`web_core` that all renderers import.

## What was wrong

| Renderer | Status | Implementation |
|----------|--------|----------------|
| React    | ⚠️ Duplicated | Local function using `.toLowerCase()` |
| Lit      | ⚠️ Duplicated | Inline anonymous using `.toLocaleLowerCase()` |
| Angular  | ❌ Missing    | No conversion applied at all — icons were broken for camelCase names |

The `toLocaleLowerCase()` vs `toLowerCase()` inconsistency between Lit
and React meant icon rendering could differ across renderers for the
same input, depending on locale. The Angular renderer silently passed
camelCase names straight to Material Symbols, which has no effect on
rendering (the icon just doesn't show).

## What changed

- **`web_core/src/v0_8/styles/icons.ts`** — Added `export function toSnakeCase(str: string): string` as the single canonical implementation, co-located with the icon CSS utilities where it naturally belongs. Already re-exported via `@a2ui/web_core/styles/icons`.
- **Lit renderer** — Removed inline regex+`toLocaleLowerCase()` lambda; now imports `toSnakeCase` from web_core.
- **Angular renderer** — Added missing conversion: `resolvedName` now pipes the resolved string through `toSnakeCase` before rendering.
- **React renderer** — Removed local function definition; now imports from web_core.

## Impact

- All three renderers now produce **identical output** for the same icon name input.
- The `toLowerCase()` canonical form is used consistently (locale-invariant, correct for ASCII icon names).
- Angular icons with camelCase names (e.g. `shoppingCart`) now **render correctly** for the first time.
- Zero behavior change for already-snake_case icon names.

## Testing

The existing React unit tests in `Icon.test.tsx` cover `toSnakeCase`
behavior end-to-end (camelCase → snake_case, already-lowercase passthrough,
unknown names). These tests continue to pass with the import swapped to
web_core.


polina-c:
[ava-cassiopeia](https://github.com/ava-cassiopeia), would you mind to review this?

polina-c:
Thank you. Linked to the issue and asked the issue creator to review: https://github.com/a2ui-project/a2ui/issues/786#issuecomment-5085532552

nan-yu:
@sanjayrohith Can you refresh your PR to resolve conflicts? Thanks!

sanjayrohith:
> @sanjayrohith Can you refresh your PR to resolve conflicts? Thanks!

ok i am doing that

sanjayrohith:
> @sanjayrohith Can you refresh your PR to resolve conflicts? Thanks!

Rebased onto latest main and resolved the conflicts (mostly path/formatting drift from the Angular renderer restructuring — the toSnakeCase centralization logic itself didn't need changes). Ready for another look! @nan-yu @polina-c

nan-yu:
@ditman @josemontesp Can you help review this PR? 

FYI https://github.com/a2ui-project/a2ui/pull/833 has similar changes.

ditman:
This is not a bad idea, but IMO the API is a little bit too low level to put it in web_core. I think we should close this and convert it to a github issue if the angular icons are still broken.

sanjayrohith:
Thanks for the feedback! Makes sense re: basic_catalog — I see @jacobsimionato flagged the same thing on #786 before I picked this up, I should've followed that from the start.

Happy to rework this: move the shared helper into renderers/web_core/src/v0_9/basic_catalog/index.ts and update the renderers to pull from there instead of styles/icons.ts. Two questions so I scope this right:
1. Should the Angular fix (camelCase icon names currently don't render at all) still land for v0.8, or is v0.8 considered closed to new fixes and this should target v0.9 only?
2. Since #833 is doing near-identical work — should I close this in favor of that PR once it's moved to basic_catalog, or would you rather review both and pick one?

Either way, keen to get the Angular bug fixed properly rather than have this stall — let me know which direction you'd like.

ditman:
> * Should the Angular fix (camelCase icon names currently don't render at all) still land for v0.8, or is v0.8 considered closed to new fixes and this should target v0.9 only?

IMO, the fixes should only apply to v0.9, unless somebody is explicitly affected on v0.8.

> * Since https://github.com/a2ui-project/a2ui/pull/833 is doing near-identical work — should I close this in favor of that PR once it's moved to basic_catalog, or would you rather review both and pick one?

I'd prefer reviewing a single one, maybe you and @stablegenius49 could collaborate on a fix? I haven't heard from them in the other issue though, so if you want to fix this, we can close the other PR and assign you to the issue so others know that you're actively working on it.

sanjayrohith:
> > * Should the Angular fix (camelCase icon names currently don't render at all) still land for v0.8, or is v0.8 considered closed to new fixes and this should target v0.9 only?
> 
> 
> 
> IMO, the fixes should only apply to v0.9, unless somebody is explicitly affected on v0.8.
> 
> 
> 
> > * Since https://github.com/a2ui-project/a2ui/pull/833 is doing near-identical work — should I close this in favor of that PR once it's moved to basic_catalog, or would you rather review both and pick one?
> 
> 
> 
> I'd prefer reviewing a single one, maybe you and @stablegenius49 could collaborate on a fix? I haven't heard from them in the other issue though, so if you want to fix this, we can close the other PR and assign you to the issue so others know that you're actively working on it.

Works for me — I'll scope this to v0.9 only.

I'll move the helper into `renderers/web_core/src/v0_9/basic_catalog/index.ts` per your suggestion, and update the Lit, React, and Angular renderers to import from there instead of `styles/icons.ts`.

Go ahead and close #833 and assign #786 to me. I noticed gemini-code-assist flagged the same PascalCase edge case (leading underscore, e.g. `_shopping_cart`) on both PRs, so I'll fix that in the shared helper too and add test coverage for it.

I'll push the updated version.
@ditman

ditman:
Note that I have a big PR coming to web_core to move its v0.9 imports to the root of the package that will affect this one, see #2167.

The merge should be super easy because it's just moving things around and making the index.ts a little bit more tidy, but just FYI!

sanjayrohith:
> Note that I have a big PR coming to web_core to move its v0.9 imports to the root of the package that will affect this one, see #2167.
> 
> The merge should be super easy because it's just moving things around and making the index.ts a little bit more tidy, but just FYI!

Pushed the rework — moved toMaterialSymbol (plus the ligature overrides that were also duplicated: play→play_arrow, rewind→fast_rewind, favoriteOff→favorite_border, starOff→star_border) into web_core/v0_9/basic_catalog/styles/icons.ts, and updated all three v0.9 Icon components (Angular, Lit, React) to import it instead of each keeping their own copy. Added test coverage for the PascalCase leading-underscore case gemini-code-assist flagged, plus CHANGELOG entries. Scoped to v0.9 only — v0.8 is untouched.
Ready for another look whenever you have time @ditman

PR #877, opened 2026-03-17
Switch v0.8 Angular Renderer to use `OnPush` change detection

gspencergoog:
# Description

This PR refactors all components in the v0.8 Angular renderer from `ChangeDetectionStrategy.Eager` to **`ChangeDetectionStrategy.OnPush`** for improved rendering performance and determinism.

## Changes
- **`renderers/angular/v0_8/components/`**: Updated components to explicitly consume `@Input` variables and bind with `OnPush` change detection.
- **`renderers/angular/v0_8/rendering/renderer.ts`**: Replaced signal `effect` setups with `@Input()` variables and `ngOnChanges` support to satisfy standard reactive input triggering.
- **`renderers/angular/v0_8/rendering/renderer.spec.ts`**: Updated fixture `tick(10)` triggers inside async assertions because `OnPush` requires explicit detection cycles in tests.

## Impact & Risks
- **No Breaking Changes**: Keeps public APIs stable.
- **Performance Benefits**: Avoids unnecessary full component node updates on large tree mutations.

## Testing
1. Run and verify that `renderers/angular/v0_8` unit tests successfully pass with continuous integration builds.


jacobsimionato (commented):
I saw @ditman did something similar and abandoned it in https://github.com/google/A2UI/pull/805 , and @crisbeto added `EAGER` explicitly in https://github.com/google/A2UI/pull/726 to prevent a breakage - not sure if it was a theoretical breakage or a real breakage. Maybe we should consult with one of them before submitting, in case there are unintended consequences in google's monorepo? Does this block submitting v0.9 code, or it's a nice to have for consistency?

gspencergoog:
> I saw @ditman did something similar and abandoned it in #805 , and @crisbeto added `EAGER` explicitly in #726 to prevent a breakage - not sure if it was a theoretical breakage or a real breakage. Maybe we should consult with one of them before submitting, in case there are unintended consequences in google's monorepo? Does this block submitting v0.9 code, or it's a nice to have for consistency?

It is orthogonal to 0.9, although 0.9 is set up to use OnPush.  It doesn't block anything.  I agree we should get more eyes on (once I fix the build issues).

ditman:
We abandoned my prior attempt because the testing story is not good, and we're not sure of what can break by blanket-applying this to the v0.8 codebase. The deal was to apply these as bugs were found on a per-component basis.

I think it's okay to do v0.9 fully based on OnPush, but leave 0.8 as is, and bring in fixes as we find them via bug reports (I suppose those are going to be mostly internal bugs, if any)

(/cc @jgindin)

gspencergoog:
> We abandoned my prior attempt because the testing story is not good, and we're not sure of what can break by blanket-applying this to the v0.8 codebase. The deal was to apply these as bugs were found on a per-component basis.
> 
> I think it's okay to do v0.9 fully based on OnPush, but leave 0.8 as is, and bring in fixes as we find them via bug reports (I suppose those are going to be mostly internal bugs, if any)
> 
> (/cc @jgindin)

Okay, I'll close this for now, then, and abandon the change until we see reasons to change it.

gspencergoog:
Oh, wait, there are a lot of new tests here.  I think I'll just revert the `OnPush` changes, but keep the tests.

gspencergoog:
Closing in favor of https://github.com/google/A2UI/pull/882

PR #878, opened 2026-03-17
Angular v0.9 Renderer

gspencergoog:
# Description

This PR adds the **v0.9 Angular renderer** and packages it into secondary entry points with appropriate `package.json` configurations. Additionally, it integrates the `a2ui_explorer` sample app into the `samples/` workspace and verifies application bundle generation fully compiles.

## Changes
- **`renderers/angular/v0_9/`**: Added the complete v0.9 Angular code (components and core).
- **`renderers/angular/ng-package.json` & `package.json`**: Created subpath mappings (`@a2ui/angular/v0_9`) for standard `exports` lookup.
- **`samples/client/angular/projects/a2ui_explorer/`**: Imported and decoupled `demo-app` from Lit, updating for `A2uiRendererService` bindings and dynamic rendering setup.
- **Config Adjustments**: Disabling `preserveSymlinks` and configuring `moduleResolution: "bundler"` inside `samples` workspace configurations to fix subpath resolution for linked packages.

## Impact & Risks
- **No Impact to v0.8 Core**: v0.9 is completely isolated behind standalone `@a2ui/angular/v0_9` node subpaths.
- **Fixed Resolves**: Rephrased standard builder linking so application static analysis does not fail.

## Testing
1. Run `npx ng build a2ui_explorer` successfully in `samples/client/angular/` workspace.


.github/workflows/check_license.yml
```diff
@@ -38,7 +38,12 @@ jobs:
 
       - name: Check license headers
         run: |
-          addlicense -check \
+          if ! addlicense -check \
             -l apache \
             -c "Google LLC" \
-            .
+            .; then
+            echo "License check failed. To fix this, install addlicense and run it:"
```
jacobsimionato, line 45: Thanks! This is great.

renderers/angular/src/v0_8/rendering/renderer.ts
```diff
@@ -70,27 +70,11 @@ export class Renderer {
...
-      }
+
     }
 
     if (componentType) {
-      // Evaluate custom bindings from CatalogEntry if they are present
-      const bindingsObject: Record<string, any> = {};
-      if (componentBindings) {
-        for (const binding of componentBindings) {
-           // binding is from inputBinding, which typically is { provide: ..., useFactory: ... } or similar in older angular
```
jacobsimionato, line 83: Lol! Thanks for removing. I assume this was never implemented here, and the bindings are handled some other way?

renderers/angular/src/v0_8/data/processor.ts
```diff
@@ -14,11 +14,10 @@
  * limitations under the License.
```
jacobsimionato, line 14: Ha totally unused variable! I'd have a slight preference to exclude all v0.8 changes from this PR, but I can't see how deleting this could break anyone seeing as it's private.

renderers/angular/src/v0_9/core/component-host.component.ts
```diff
@@ -0,0 +1,98 @@
...
+  private rendererService = inject(A2uiRendererService);
+  private binder = inject(ComponentBinder);
+  private destroyRef = inject(DestroyRef);
+
+  protected componentType: Type<any> | null = null;
+  protected props: any = {};
+  private context?: ComponentContext;
+
+  ngOnInit(): void {
+    const surface = this.rendererService.surfaceGroup?.getSurface(this.surfaceId());
```
jacobsimionato, line 67: Is it still possible to explicitly pass in the rendererService when you create this, e.g. so there can be multiple rendererServices in the same application, potentially in different features, with different catalogs?

renderers/angular/src/v0_9/catalog/types.ts
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+import { Type } from '@angular/core';
+import { Catalog, ComponentApi } from '@a2ui/web_core/v0_9';
+
+/**
+ * Extends the generic ComponentApi to include Angular-specific component types.
+ */
+export interface AngularComponentApi extends ComponentApi {
```
jacobsimionato: Can we call this AngularComponentImplementation instead? I know that the implementation isn't inline directly here, and it's a reference to a class, but conceptually, this does represent the implementation of a component. If I have an instance of this class for e.g. ButtonAngularComponentImplementation, this defines a specific implementation of Button which will always look a certain way.

renderers/angular/src/v0_9/catalog/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  /** The Angular component class used to render this component. */
+  readonly component: Type<any>;
+}
+
+/**
+ * Base class for Angular-specific component catalogs.
+ */
+export class AngularCatalog extends Catalog<AngularComponentApi> {}
```
jacobsimionato: Can we just do this:
```
export type AngularCatalog = Catalog<AngularComponentApi>;
```

renderers/angular/src/v0_9/catalog/minimal/minimal-catalog.ts
```diff
@@ -0,0 +1,85 @@
...
+  RowApi,
+  ColumnApi,
+  ButtonApi,
+  TextFieldApi,
+} from '@a2ui/web_core/v0_9/basic_catalog';
+import { FunctionImplementation } from '@a2ui/web_core/v0_9';
+import { FormatStringImplementation } from '@a2ui/web_core/v0_9/basic_catalog';
+
+export const MINIMAL_COMPONENTS: AngularComponentApi[] = [
+  { ...TextApi, component: TextComponent },
```
jacobsimionato, line 37: Ha, this is pretty neat. I think it would be great to find a way to create a type safe binding here, as I mentioned elsewhere, e.g. so that you do something like we have in React e.g. so you can only specify a component here that does handle the exact params defined in the Zod schema in TextApi.

jacobsimionato: (to figure out next week after merging)

renderers/angular/a2ui_explorer/src/app/demo.component.ts
```diff
@@ -0,0 +1,441 @@
...
+
+      <!-- Main Canvas Area -->
+      <div class="canvas-area">
+        <div class="canvas-header" *ngIf="selectedExample">
+          <h2>{{ selectedExample.name }}</h2>
+          <p class="subtitle">{{ selectedExample.description }}</p>
+        </div>
+        <div class="canvas-frame">
+          <div *ngIf="surfaceId" class="rendered-content">
+            <a2ui-v09-component-host [surfaceId]="surfaceId" componentId="root">
```
jacobsimionato: Can we make this default to 'root', or have some wrapper `a2ui-v09-surface` component to encapsulate this? I think the wrapper could be nice, to make the concept very clear. The developer just specifies the surface, and we handle all component-level stuff.

jacobsimionato (approved):
Hey this is great! I think it'd be beneficial to merge this ASAP and happy to defer addressing everything here until after merge if you want.

renderers/angular/src/v0_8/rendering/renderer.ts
gspencergoog: Correct.

renderers/angular/src/v0_8/data/processor.ts
gspencergoog: That was my thought process as well.  Just worried I'd forget to do it, and it's just cruft.

renderers/angular/src/v0_9/core/component-host.component.ts
gspencergoog: Yes, it is still possible to have multiple A2uiRendererService instances in the same application, but the "v0.9 way" to do this is via Angular's Dependency Injection rather than explicitly passing it as a property.   We could add an input for it later if that seems more ergonomic, I think.

Gemini example:

## How it works with multiple features

If you have different features that need different catalogs or configurations, you can provide a unique A2uiRendererService at the "feature-root" component level.

```ts
@Component({
  selector: 'feature-a',
  providers: [
    A2uiRendererService,
    { 
      provide: A2UI_RENDERER_CONFIG, 
      useValue: { catalogs: [CatalogA], ... } 
    }
  ],
  template: `
    <a2ui-v09-component-host surfaceId="surface-a" componentId="root" />
  `
})
export class FeatureAComponent {}
```

renderers/angular/src/v0_9/catalog/types.ts
gspencergoog: Sure, done.

renderers/angular/src/v0_9/catalog/types.ts
gspencergoog: If we change it to a type, then it will be erased when it's compiled, and then we can't use it as a token in a provider, and we'd have to have a separate injection token.  The class makes it both stick around (so we can use it directly in a provider), and gives us a single point where would add helper methods that are available to all catalogs.

renderers/angular/a2ui_explorer/src/app/demo.component.ts
gspencergoog: Sure, added.

PR #869, opened 2026-03-16
feat(lit): Initial Draft of v0.9 Lit Renderer and Local Gallery

jacobsimionato:
### Description of Changes

This PR introduces the initial draft implementation of the A2UI v0.9 Lit Renderer (`@a2ui/lit/v0_9`), leveraging the framework-agnostic models from `@a2ui/web_core/v0_9`.

- **A2uiController**: A Lit Reactive Controller that automatically bridges the `GenericBinder` logic from web core into the native Lit element lifecycle. It manages data subscriptions and triggers `requestUpdate()` on the host when A2UI data dependencies change.
- **Component Factories**: Added `createLitComponent` for concisely generating pure functional/stateless elements strictly typed via Zod schemas.
- **Catalogs**: Full implementation of the v0.9 `minimal_catalog` and `basic_catalog` including 18 standard components (Text, Button, Row, Column, etc.) and evaluation functions (capitalize, formatString, arithmetic).
- **Client Capabilities**: Enhanced `@a2ui/web_core`'s `MessageProcessor` to automatically generate `a2uiClientCapabilities` dynamically from the underlying Zod API schemas, correctly resolving external `$ref` URIs from the `REF:` descriptions.
- **Local Gallery**: A new agentless Vite workspace at `samples/client/lit/gallery_v0_9` featuring an interactive 3-column debugging layout:
  - Spec navigation panel (automatically syncing JSON files directly from the v0.9 specifications).
  - Progressive rendering message stepper to step through payloads incrementally.
  - A live-updating Data Model and Action Log inspector.

### Rationale

This aligns the Lit ecosystem with the 'Unified Architecture' concepts drafted in v0.9, ensuring robust separation between the A2UI state protocol (`DataModel`, `SurfaceModel`, `ComponentContext`) and the actual screen painting (DOM templates).

By adopting a Reactive Controller pattern, existing and new Lit components can 'opt-in' to A2UI bindings without requiring inheritance from an obstructive base class. Leveraging Zod schemas enables us to guarantee that the UI implementation remains tightly coupled to the capabilities we broadcast back to the AI Agent.

### Testing/Running Instructions

To run the interactive debugger and step through the newly implemented examples locally:

```bash
# 1. Build the updated Web Core models and Generic Binder
cd renderers/web_core
npm install && npm run build

# 2. Build the new Lit Renderer architecture
cd ../lit
npm install && npm run build

# 3. Start the Vite gallery application
cd ../../samples/client/lit/gallery_v0_9
[...]

PR #885, opened 2026-03-18
Lit Renderer v0.9

ditman:
# Description

Adds a Lit renderer for the v0.9 version of the protocol.

I cloned @jacobsimionato's https://github.com/google/A2UI/pull/869 and polished it a little!

This iteration of the renderer is missing two big features compared to v0.8 (that will work on as fast-follows):

* Proper Markdown rendering (integrate `@a2ui/markdown-it`)
* CSS styling/theming

(And probably some proper formatting and better tooling).

Anyway, this should suffice as a starting point. Check out the `demo:gallery09` Lit sample app to get this renderer to run. Other examples will be ported later.

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/lit/package-lock.json
+++ b/renderers/lit/package-lock.json
@@ -17,8 +17,10 @@
         "zod": "^3.25.76"
       },
       "devDependencies": {
+        "@types/jsdom": "^28.0.1",
         "@types/node": "^24.10.1",
         "google-artifactregistry-auth": "^3.5.0",
+        "jsdom": "^25.0.0",
         "typescript": "^5.8.3",
         "wireit": "^0.15.0-pre.2"
       }
@@ -47,9 +49,131 @@
       "resolved": "../web_core",
       "link": true
     },
+    "node_modules/@asamuzakjp/css-color": {
+      "version": "3.2.0",
+      "integrity": "sha512-K1A6z8tS3XsmCMM86xoWdn7Fkdn9m6RSVtocUrJYIwZnFVkng/PvkEoWtOWmP+Scc6saYWHWZYbndEEXxl24jw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@csstools/css-calc": "^2.1.3",
+        "@csstools/css-color-parser": "^3.0.9",
+        "@csstools/css-parser-algorithms": "^3.0.4",
+        "@csstools/css-tokenizer": "^3.0.3",
+        "lru-cache": "^10.4.3"
+      }
+    },
+    "node_modules/@csstools/color-helpers": {
+      "version": "5.1.0",
+      "integrity": "sha512-S11EXWJyy0Mz5SYvRmY8nJYTFFd1LCNV+7cXyAgQtOOuzb4EsgfqDufL+9esx72/eLhsRdGZwaldu/h+E4t4BA==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT-0",
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/@csstools/css-calc": {
+      "version": "2.1.4",
+      "integrity": "sha512-3N8oaj+0juUw/1H3YwmDDJXCgTB1gKU6Hc/bB502u9zR0q2vd786XJH9QfrKIEgFlZmhZiq6epXl4rHqhzsIgQ==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT",
+      "engines": {
+        "node": ">=18"
+      },
+      "peerDependencies": {
+        "@csstools/css-parser-algorithms": "^3.0.5",
+        "@csstools/css-tokenizer": "^3.0.4"
+      }
+    },
+    "node_modules/@csstools/css-color-parser": {
+      "version": "3.1.0",
+      "integrity": "sha512-nbtKwh3a6xNVIp/VRuXV64yTKnb1IjTAEEh3irzS+HkKjAOYLTGNb9pmVNntZ8iVBHcWDA2Dof0QtPgFI1BaTA==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT",
+      "dependencies": {
+        "@csstools/color-helpers": "^5.1.0",
+        "@csstools/css-calc": "^2.1.4"
+      },
+      "engines": {
+        "node": ">=18"
+      },
+      "peerDependencies": {
+        "@csstools/css-parser-algorithms": "^3.0.5",
+        "@csstools/css-tokenizer": "^3.0.4"
+      }
+    },
+    "node_modules/@csstools/css-parser-algorithms": {
+      "version": "3.0.5",
+      "integrity": "sha512-DaDeUkXZKjdGhgYaHNJTV9pV7Y9B3b644jCLs9Upc3VeNGg6LWARAT6O+Q+/COo+2gg/bM5rhpMAtf70WqfBdQ==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT",
+      "engines": {
+        "node": ">=18"
+      },
+      "peerDependencies": {
+        "@csstools/css-tokenizer": "^3.0.4"
+      }
+    },
+    "node_modules/@csstools/css-tokenizer": {
+      "version": "3.0.4",
+      "integrity": "sha512-Vd/9EVDiu6PPJt9yAh6roZP6El1xHrdvIVGjyBsHR0RYwNHgL7FJPyIIW4fANJNG6FtyZfvlRPpFI4ZM/lubvw==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT",
+      "engines": {
+        "node": ">=18"
+      }
+    },
     "node_modules/@lit-labs/signals": {
       "version": "0.1.3",
-      "resolved": "https://registry.npmjs.org/@lit-labs/signals/-/signals-0.1.3.tgz",
       "integrity": "sha512-P0yWgH5blwVyEwBg+WFspLzeu1i0ypJP1QB0l1Omr9qZLIPsUu0p4Fy2jshOg7oQyha5n163K3GJGeUhQQ682Q==",
       "license": "BSD-3-Clause",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


ditman:
OK, there's a version of the renderer that uses LitElements. Some things that I still obviously need:

* [x] Add tests
* [x] Move code around so imports from v0_9 make sense (and ensure that everything is imported consistently internally)
* [ ] (Later) Integrate with @a2ui/markdown-it package
* [ ] (Later) Add some configurable styling (CSS variables? classnames?)
* [ ] (Even later) migrate some Lit samples to v0_9
* ... And probably more stuff that I'm not aware of yet :)

/cc @jacobsimionato

renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts
```diff
@@ -0,0 +1,42 @@
...
+export class A2uiAudioPlayerElement extends A2uiLitElement<typeof AudioPlayerApi> {
+  protected createController() { return new A2uiController(this, AudioPlayerApi); }
+
+  render() {
+    const props = this.controller.props;
+    if (!props) return nothing;
+
+    return html`
+      <div class="a2ui-audioplayer">
+        ${props.description ? html`<p>${props.description}</p>` : ""}
```
jacobsimionato: 🙌 type safety! This looks neat to me!

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!props) return nothing;
+
+    const children = Array.isArray(props.children) ? props.children : [];
+    const styles = { display: "flex", flexDirection: "column", flex: props.weight !== undefined ? String(props.weight) : "initial", gap: "8px" };
+    
+    return html`
+      <div class="a2ui-column" style=${styleMap(styles as Record<string, string>)}>
+        ${map(children, (child: any) => {
+          if (typeof child === 'string') {
+             return html`${renderA2uiNode(new ComponentContext(this.context.dataContext.surface, child, this.context.dataContext.path))}`;
```
jacobsimionato: Let's try to simplify the API for rendering children. In some other frameworks (Flutter, React) there is a call back something like `renderChild(id, dataModelBasePath)`. Lit has different needs. I wonder if we can do something like `renderA2uiNode(context.surface, child)` where child is an `(id, basepath)` pair outputted by the generic binder - see https://github.com/google/A2UI/blob/6efe74ce33fb3528eaf8535a5c14daa2fbb810bc/renderers/web_core/src/v0_9/rendering/generic-binder.ts#L35 .

This would also require putting SurfaceModel as a top-level parameter of ComponentContext, which seems to make sense given the way lit works here? Unless there is some other way to propagate a reference to the current SurfaceModel throughout the hierarchy?

Another idea... What if A2uiLitElement has a `buildChild` method which hides some of this?

Also, does this path actually get hit? I thought that the generic binder should always output an array of (id, basePath) objects for children, e.g. the case below this early return.

renderers/lit/src/v0_9/surface/render-node.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param context The component context defining the data model and type to render.
+ * @returns A Lit TemplateResult representing the resolved component, or `nothing` if the component is invalid or unresolvable.
+ *
+ * For examples on how to use `renderA2uiNode` when defining custom A2UI components that contain children,
+ * refer to the implementation of `A2uiSurface`, or structural catalog components like `Row` and `Column`.
+ */
+export function renderA2uiNode(context: ComponentContext) {
+  const type = context.componentModel.type;
+  const catalog = context.dataContext.surface.catalog;
+  const implementation = catalog.components.get(type) as LitComponentImplementation | undefined;
```
jacobsimionato: Catalog, SurfaceModel etc are all templated with the type of the ComponentImplementation, so that in framework code we can refer to them as SurfaceModel<LitComponentImplementation> etc and then not have to do these kinds of casts. Given the structure of this renderer, I think we should either also templatize ComponentContext and DataContext to avoid the cast.

`DataContext.surface` was supposed to be a bit of an escape hatch that isn't normally required. I think perhaps we should expose `Catalog` on `ComponentContext` or else maybe just plumb `ComponentContext` and `Catalog` as two separate parameters here?

jacobsimionato:
> * ... And probably more stuff that I'm not aware of yet :)

What about also deferring the markdown integration to later? Having this merged would mean that we can then make changes to the web core API without breaking your pending PR.

ditman:
> > * ... And probably more stuff that I'm not aware of yet :)
> 
> What about also deferring the markdown integration to later? Having this merged would mean that we can then make changes to the web core API without breaking your pending PR.

Absolutely agree, the markdown integration should be super easy to add to the Text widget later.

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
ditman: > Another idea... What if A2uiLitElement has a buildChild method which hides some of this?

I like this idea! I added the base class very late on this PR, and forgot to encapsulate that method. It already has access to `this.context`, so there's no need to explicitly create the child context externally.

> Also, does this path actually get hit? I thought that the generic binder should always output an array of (id, basePath) objects for children, e.g. the case below this early return.

Let me double-check where we end up with this. In any case, I can also hide this if check inside this new method we're adding, so it can resolve both strings and (id, basePath) objects (or whatever we want to add later)

renderers/lit/src/v0_9/surface/render-node.ts
ditman: Good call, I'll try to see if we can get a better API for the way we're using the context here. However, I just made this method "private" (not exported); Catalog implementers are expected to use the base `A2uiLitElement` class with `buildChild`, and won't need this directly.

(The other only case where this is used is in the A2uiSurface element, which is not an A2uiLitElement)

ditman: I addressed this by passing a Catalog into the renderA2uiNode (private) function. All the users of this method *can* produce a catalog without having to deeply unroll it from the context like now, so it seemed like a good fit. Passing the rest of the context in doesn't feel so bad (but this may be a flutterism speaking, maybe we just need the componentModel or a smaller slice of the context).

Being a private API, we can change this as needed easily without breaking end users though!

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
ditman: I added this to the render method, so now end catalog elements can render a child like:

```typescript
  //...
  ${props.child ? html`${this.renderNode(props.child)}` : nothing}
  //...
```

renderers/lit/src/0.8/model.test.ts
ditman, line 1: This file needs to be reverted, or landed as a separate PR.

renderers/lit/src/v0_9/catalogs/basic/index.ts
```diff
@@ -0,0 +1,63 @@
...
+import { A2uiCard } from "./components/Card.js";
+import { A2uiDivider } from "./components/Divider.js";
+import { A2uiCheckBox } from "./components/CheckBox.js";
+import { A2uiSlider } from "./components/Slider.js";
+import { A2uiDateTimeInput } from "./components/DateTimeInput.js";
+import { A2uiChoicePicker } from "./components/ChoicePicker.js";
+import { A2uiTabs } from "./components/Tabs.js";
+import { A2uiModal } from "./components/Modal.js";
+
+export const basicCatalog = new Catalog<LitComponentApi>(
```
ditman, line 48: Needs JSDocs

renderers/lit/src/v0_9/surface/a2ui-surface.ts
```diff
@@ -0,0 +1,75 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { html, nothing, LitElement, PropertyValues } from "lit";
+import { customElement, property, state } from "lit/decorators.js";
+import { SurfaceModel, ComponentContext } from "@a2ui/web_core/v0_9";
+import { renderA2uiNode } from "./render-a2ui-node.js";
+import { LitComponentApi } from "@a2ui/lit/v0_9";
```
ditman, line 21: Imports need to be sorted

renderers/lit/src/v0_9/surface/a2ui-surface.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (this.unsub) {
+      this.unsub();
+      this.unsub = undefined;
+    }
+  }
+
+  render() {
+    if (!this.surface) return nothing;
+    if (!this._hasRoot) {
+      return html`<slot name="loading"><div>Loading surface...</div></slot>`;
```
ditman, line 102: Are we using the "loading" slot elsewhere? Why is this a slot, but the error isn't?

renderers/lit/src/v0_9/surface/render-a2ui-node.ts
```diff
@@ -0,0 +1,49 @@
...
+import { LitComponentApi } from "@a2ui/lit/v0_9";
+
+/**
+ * Pure function that acts as a generic container for A2UI components.
+ *
+ * It dynamically resolves and renders the specific Lit component implementation
+ * based on the component type provided in the context, returning a TemplateResult directly
+ * to avoid duplicate DOM node wrapping.
+ *
+ * @param context The component context defining the data model and type to render.
```
ditman, line 29: The parameter list is outdated

renderers/lit/src/v0_9/a2ui-controller.ts
```diff
@@ -0,0 +1,94 @@
...
+  private subscription?: { unsubscribe: () => void };
+
+  /**
+   * Initializes the controller, binding it to the given Lit element and API schema.
+   *
+   * @param host The LitElement acting as the component host. Must provide a component `context`.
+   * @param api The A2UI component API defining the schema for this element.
+   */
+  constructor(
+    private host: LitElement & { context: ComponentContext },
```
ditman: Extract this to a separate typedef?

renderers/lit/src/v0_9/index.ts
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export type { LitComponentApi } from "./types.js";
+export { A2uiController } from "./a2ui-controller.js";
+export { A2uiSurface } from "./surface/a2ui-surface.js";
+export { A2uiLitElement } from "./a2ui-lit-element.js";
+export { minimalCatalog } from "./catalogs/minimal/index.js";
+export { basicCatalog } from "./catalogs/basic/index.js";
```
ditman, line 22: (Ensure that at least everything here has good JSDoc!)

renderers/lit/src/index.ts
ditman, line 1: Move these changes to a new PR instead!

ditman: Done https://github.com/google/A2UI/pull/965

renderers/lit/src/v0_9/surface/a2ui-surface.ts
ditman: There should be "gsc" integrated here, to enforce and auto-fix all these.

renderers/lit/src/0.8/model.test.ts
ditman: https://github.com/google/A2UI/pull/965

renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts
```diff
@@ -25,507 +25,255 @@ import {
...
-import {ComponentApi} from '../../catalog/types.js';
+} from "../../schema/common-types.js";
+import { ComponentApi } from "../../catalog/types.js";
 
 const CommonProps = {
   accessibility: AccessibilityAttributesSchema.optional(),
-  weight: z
-    .number()
-    .describe(
-      "The relative weight of this component within a Row or Column. This is similar to the CSS 'flex-grow' property. Note: this may ONLY be set when the component is a direct descendant of a Row or Column.",
```
ditman, line 36: Moved to a separate PR: https://github.com/google/A2UI/pull/968

renderers/lit/src/v0_9/a2ui-controller.ts
ditman: In practice, this is always an A2uiLitElement, so I'm using that type instead of this custom one.

ditman:
OK, this is it! PTAL @jacobsimionato, @ava-cassiopeia (anybody else who may want to look at this?)

ditman:
Let me rebase the branch, update the locks and push then!

PR #886, opened 2026-03-18
Refactor @a2ui/react package into versioned subdirectories (v0_8)

jacobsimionato:
## Description of Changes
This pull request restructures the `@a2ui/react` renderer package into a versioned subdirectory layout, specifically moving the current implementation to a `v0_8` namespace. 

Key changes include:
* Moved `src/` files (components, core, hooks, registry, styles, theme, lib) and `tests/` files into respective `v0_8/` subdirectories.
* Introduced top-level proxy export files (`src/index.ts`, `src/types.ts`, `src/styles/index.ts`, and `src/styles/index.d.ts`) that transparently re-export the `v0_8` codebase.
* Updated `package.json` to explicitly export the `./v0_8` paths, alongside the default root exports, ensuring backward compatibility.
* Modified the build (`tsup.config.ts`, `package.json` scripts) and test (`vitest.config.ts`) configurations to account for the new file structures.
* Updated the `README.md` to document the new `Versioned Structure` and the future roadmap for the A2UI specification integrations.

## Rationale
As we prepare to introduce the `v0_9` A2UI specification, the React renderer requires a side-by-side transition phase, similar to `@a2ui/angular` and `@a2ui/web_core`. This versioned directory structure allows both versions to coexist safely in the interim. 

By utilizing top-level proxy exports, we maintain full backward compatibility for existing consumers, who can continue importing from `@a2ui/react` without making any modifications to their import paths. Once the library is unified to support multiple specifications natively, the `v0_8` implementation will eventually be deprecated and removed.

1. Navigate to the `renderers/react` directory.
2. Install dependencies: `npm install`
3. Run the TypeScript compiler: `npm run typecheck`
4. Build the package: `npm run build`
    * Confirm that `dist/` contains both root exports (`index.js`, `styles/index.js`) and versioned exports (`v0_8/index.js`, `v0_8/styles/index.js`).
5. Run the test suite: `npm test`
    * Ensure all tests pass successfully against the refactored directory.

### DIFF
--- a/renderers/react/scripts/postbuild.mjs
+++ b/renderers/react/scripts/postbuild.mjs
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import { copyFileSync, mkdirSync } from 'node:fs';
 import { join } from 'node:path';
 


renderers/react/tests/v0_8/integration/components.test.tsx
```diff
@@ -17,7 +17,7 @@
 import { describe, it, expect, vi } from 'vitest';
 import { render, screen, waitFor } from '@testing-library/react';
 import React, { useEffect } from 'react';
-import { A2UIProvider, A2UIRenderer, useA2UI } from '../../src';
+import { A2UIProvider, A2UIRenderer, useA2UI } from '../../../src';
```
ava-cassiopeia: Is this right? This says "import these from the canonical latest version of the react renderer" but from _inside_ the 0.8 implementation, which seems like it will blow up on us?

renderers/react/tests/v0_8/integration/data-binding.test.tsx
```diff
@@ -17,7 +17,7 @@
 import { describe, it, expect } from 'vitest';
 import { render, screen, fireEvent, waitFor } from '@testing-library/react';
 import React, { useEffect } from 'react';
-import { A2UIProvider, A2UIRenderer, useA2UI } from '../../src';
+import { A2UIProvider, A2UIRenderer, useA2UI } from '../../../src';
```
ava-cassiopeia: Same thing here and in a few other places in test files, I think we still want to import these from the 0.8 version.

renderers/react/package.json
```diff
@@ -36,7 +46,7 @@
     "*.css"
   ],
   "scripts": {
-    "build": "tsup && cp src/styles/index.d.ts dist/styles/index.d.ts && cp src/styles/index.d.ts dist/styles/index.d.cts",
+    "build": "tsup && cp src/styles/index.d.ts dist/styles/index.d.ts && cp src/styles/index.d.ts dist/styles/index.d.cts && mkdir -p dist/v0_8/styles && cp src/v0_8/styles/index.d.ts dist/v0_8/styles/index.d.ts && cp src/v0_8/styles/index.d.ts dist/v0_8/styles/index.d.cts",
```
ava-cassiopeia: This is getting pretty complex, do you think we should capture this in a script of some form? Bash is no good because Windows, but maybe a NodeJS script?

renderers/react/README.md
```diff
@@ -80,6 +80,32 @@ function App() {
...
+│   ├── index.ts        # Proxy export for backward compatibility -> ./v0_8/index
+│   ├── types.ts        # Proxy export -> ./v0_8/types
+│   └── styles/         # Proxy export -> ../v0_8/styles/index
+```
+
+**Backward Compatibility & Future Roadmap**
+
+The top-level exports in `package.json` and the proxy files in the `src/` root currently default to the `v0_8` implementation. This guarantees that existing applications relying on `import { A2UIProvider } from '@a2ui/react'` will continue to work without modification.
+
+Applications can also explicitly import from a specific version path if desired (e.g., `import { A2UIProvider } from '@a2ui/react/v0_8'`).
```
ava-cassiopeia, line 103: This is pretty awkward conceptually, but I guess we probably have to have this, since we're versioning NPM separately from the actual spec.

renderers/react/tests/v0_8/integration/components.test.tsx
jacobsimionato: Oh yes you're right - these should import the v0.8 sources explicitly.

renderers/react/package.json
jacobsimionato: Yes, let's make this a node.js script.

renderers/react/README.md
jacobsimionato: Yes, i think this is awkward, but a necessary stop gap.

jacobsimionato:
Hey @ava-cassiopeia I'm merging this by bypassing the rules and not waiting for your re-review, just because I did address your feedback, I think this is relatively low risk, the deadlines are tight, and I want to be able to restructure my actual renderer PR on top of this to send it for review during my day now, so that US folk can look tomorrow.

Please accept my apologies if this feels like I'm ignoring your review!

PR #915, opened 2026-03-19
add mcp guide

wrenj:


docs/guides/a2ui_over_mcp.md
```diff
@@ -0,0 +1,241 @@
+# A2UI over Model Context Protocol (MCP)
+
+This guide outlines how to use **A2UI** declarative syntax to build rich, interactive interfaces on top of **Model Context Protocol (MCP)** using Tools and Resources.
+
```
dmandar, line 4: can we add link to the MCP sample that does this

docs/guides/mcp.md
```diff
@@ -0,0 +1,241 @@
...
+  "method": "tools/call",
+  "id": "id-123",
+  "params": {
+    "name": "generate_report",
+    "arguments": { "date": "2026-03-01" },
+    "_meta": {
+      "a2ui": {
+        "clientCapabilities": {
+          "v0.10": {
+            "properties": {
```
dmandar: is this nesting correct? supportedCatalogIds inside the "properties" object?

wrenj:
updated ptal

docs/guides/a2ui_over_mcp.md
```diff
@@ -0,0 +1,241 @@
...
+```python
+a2ui_resource = types.EmbeddedResource(
+    type="resource",
+    resource=types.TextResourceContents(
+        uri="a2ui://training-plan-page",
+        mimeType="application/json+a2ui",
+        text=json.dumps(a2ui_payload)
+    ),
+    # Hide the raw JSON from the LLM, but show the UI to the user
```
sugoi-yuzuru, line 234: Amazing!!

docs/guides/a2ui_over_mcp.md
```diff
@@ -0,0 +1,241 @@
...
+```
+
+#### 2. A2UI Action MCP Payload
+
+When the button is clicked, the client resolves any absolute or relative path models (like `/dates/start` or `/dates/end`) against the surface binding state, and translates that into the MCP tool call arguments.
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "tools/call",
```
sugoi-yuzuru, line 161: Does this mean, the client (UI) needs to be aware that the A2UI data it is receiving is sourced from A2UI over MCP?

docs/guides/a2ui_over_mcp.md
```diff
@@ -0,0 +1,241 @@
...
+    "name": "generate_report",
+    "arguments": { "date": "2026-03-01" },
+    "_meta": {
+      "a2ui": {
+        "clientCapabilities": {
+          "v0.10": {
+            "supportedCatalogIds": [
+              "https://a2ui.org/specification/v0_10/basic_catalog.json"
+            ],
+            "inlineCatalogs": []
```
sugoi-yuzuru, line 64: Is inlineCatalogs also supported for Option A as a per-session catalog?

If so, can we add that in the example for option A just to be clear?

docs/guides/a2ui_over_mcp.md
```diff
@@ -0,0 +1,241 @@
...
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+## Returning A2UI Content as Embedded Resources
```
sugoi-yuzuru, line 73: How does the catalog negotiation actually take effect?

Is there any framework support that prescribes a certain behavior spec for uniformity?

The get_hello_world_ui tool call seems to simply return a payload regardless of the catalog negotiation that is expected to precede the tool call (or to be inline with the call).

docs/guides/a2ui_over_mcp.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### 3. Action Handler MCP Server Tool
+
+The MCP server receives the tool call and executes the corresponding handler. 
+
+```python
+@self.tool()
+async def action(action_payload: Dict[str, Any]) -> Dict[str, Any]:
+    if action_payload["name"] == "confirm_booking":
+        return {"response": f"Booking confirmed for {action_payload['context']['start']} to {action_payload['context']['end']}."}
```
sugoi-yuzuru, line 184: Is there also a scenario we can add here where the response will translate to a surfaceUpdate/dataModelUpdate to update the state of the A2UI Surface?

E.g., the button will be disabled and updates its label (or simply the surface updates to a completed state).

mkdocs.yaml
```diff
@@ -54,6 +54,7 @@ nav:
       - Renderer Development: guides/renderer-development.md
       - Custom Components: guides/custom-components.md
       - Theming & Styling: guides/theming.md
+      - A2UI over MCP: guides/mcp.md
```
sugoi-yuzuru: can we rename a2ui_over_mcp.md to make room for MCP Apps in A2UI?

docs/guides/a2ui_over_mcp.md
wrenj: lets do this in another PR, we need a sample working showing it in practice. the MCP samples i look at werent completely clear how this works

docs/guides/a2ui_over_mcp.md
wrenj: lets make a sample of this and check it in, and then update these docs

docs/guides/a2ui_over_mcp.md
wrenj: yeah something needs to translate what would be A2A messages of A2UI to MCP instead

docs/guides/a2ui_over_mcp.md
wrenj: we need to figure that part out, agree this is a TODO

PR #918, opened 2026-03-20
Streaming restaurant sample

nan-yu:
# Description

Add streaming support for the restaurant finder sample.

This depends on changes in PR #905.

Demo: 

[streaming demo restaurant finder.webm](https://github.com/user-attachments/assets/6be005d1-d3a3-4508-84c4-b8463f1af276)

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/samples/agent/adk/restaurant_finder/agent.py
+++ b/samples/agent/adk/restaurant_finder/agent.py
@@ -27,6 +27,7 @@ from a2a.types import (
     Part,
     TextPart,
 )
+from google.adk.agents import run_config
 from google.adk.agents.llm_agent import LlmAgent
 from google.adk.artifacts import InMemoryArtifactService
 from google.adk.memory.in_memory_memory_service import InMemoryMemoryService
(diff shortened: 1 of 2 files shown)


samples/client/angular/projects/restaurant/src/app/client.ts
```diff
@@ -36,49 +36,131 @@ export class Client {
...
+      if (done) break;
+
+      const now = performance.now();
+      buffer += decoder.decode(value, { stream: true });
+
+      // Parse SSE events. The server sends "data: <json>\n\n"
+      const lines = buffer.split('\n\n');
+      buffer = lines.pop() || '';
+
+      for (const line of lines) {
```
jacobsimionato, line 95: This seems like a reasonable amount of boilerplate. It seems like one of the issues is that this client needs to speak "A2A" to an A2A server, but it isn't using the client library for A2A, so there's quite a lot of manual request handling and stream processing. I was hoping that on the client side it should be quite simple to just take A2UI data chunks and propagate them to the message processor. All the heavy stream processing has already been accomplished on the server side.

I filed https://github.com/google/A2UI/issues/966 with some ideas on how to refactor this, but I think it makes sense to submit all your changes seeing as they work, and then we can look at refactoring to reduce boilerplate. WDYT?

PR #924, opened 2026-03-20
fix: resolve Angular build break caused by @a2ui/web_core protocol update

gspencergoog:
- Updated `A2uiRendererService` and `RendererConfiguration` to use `A2uiClientAction` instead of `SurfaceGroupAction`.
- Modified `ComponentHostComponent` to accept and pass `componentId` to rendered components.
- Updated all v0.9 catalog components to include a required `componentId` input.
- Passing `sourceComponentId` in `SurfaceModel.dispatchAction` calls (e.g., in `ButtonComponent`).
- Updated all test specs and dummy components to support the new `componentId` input.

This fix aligns the Angular renderer with the client data model synchronization changes introduced in @a2ui/web_core in https://github.com/google/A2UI/pull/866 and fixes the build.

**I'm not sure why CI didn't catch this: the angular v0_9 renderer wouldn't compile...**

### DIFF
--- a/renderers/angular/src/v0_9/catalog/basic/card.component.ts
+++ b/renderers/angular/src/v0_9/catalog/basic/card.component.ts
@@ -61,6 +61,7 @@ export class CardComponent {
    */
   props = input<Record<string, BoundProperty>>({});
   surfaceId = input.required<string>();
+  componentId = input<string>();
   dataContextPath = input<string>('/');
 
   child = computed(() => this.props()['child']?.value());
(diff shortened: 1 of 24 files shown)


renderers/angular/src/v0_9/catalog/basic/card.component.ts
```diff
@@ -61,6 +61,7 @@ export class CardComponent {
    */
   props = input<Record<string, BoundProperty>>({});
   surfaceId = input.required<string>();
+  componentId = input<string>();
```
nan-yu, line 64: Is this also required for card component, and other components?

nan-yu (approved):
LGTM

PR #926, opened 2026-03-20
Refactor A2UI extension activation

nan-yu:
The agents may support multiple versions of A2UI extensions. Based on client requested extension, the agent SDK should dynamically activate one that is supported by the agent.

## Screenshots of the custom components in contact_multiple_surfaces sample
- Org chart and contact card: 
image

- Floor plan: 
image

- Message: 
image

There is an `(empty)` block on the UI, but it sounds like a client renderer issue. 

## Screenshots of the rizzcharts components:

- image

Fixes: https://github.com/flutter/genui/issues/821

# Description

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/samples/agent/adk/contact_multiple_surfaces/agent.py
+++ b/samples/agent/adk/contact_multiple_surfaces/agent.py
@@ -16,8 +16,34 @@ import json
 import logging
 import os
 from collections.abc import AsyncIterable
+from dataclasses import dataclass
 from typing import Any, Dict, Optional
+
+import jsonschema
+
+from google.adk.agents.llm_agent import LlmAgent
+from google.adk.artifacts import InMemoryArtifactService
+from google.adk.memory.in_memory_memory_service import InMemoryMemoryService
+from google.adk.models.lite_llm import LiteLlm
+from google.adk.runners import Runner
+from google.adk.sessions import InMemorySessionService
+from a2a.types import (
+    AgentCapabilities,
+    AgentCard,
+    AgentSkill,
+    DataPart,
+    Part,
+    TextPart,
+)
+
+from google.genai import types
+from prompt_builder import get_text_prompt, ROLE_DESCRIPTION, WORKFLOW_DESCRIPTION, UI_DESCRIPTION
+from tools import get_contact_info
+from a2ui.core.schema.constants import VERSION_0_8, A2UI_OPEN_TAG, A2UI_CLOSE_TAG
 from a2ui.core.schema.common_modifiers import remove_strict_validation
+from a2ui.core.schema.manager import A2uiSchemaManager
+from a2ui.core.parser.parser import parse_response, ResponsePart
+from a2ui.basic_catalog.provider import BasicCatalog
 from a2ui.a2a import create_a2ui_part, get_a2ui_agent_extension, parse_response_to_parts
 
 logger = logging.getLogger(__name__)
@@ -48,6 +74,11 @@ class ContactAgent:
   @property
   def agent_card(self) -> AgentCard:
     return self._agent_card
+
+  def get_schema_manager(self, version: Optional[str]) -> Optional[A2uiSchemaManager]:
+    if version is None:
+      return None
+    return self._schema_managers[version]
 
   def _build_schema_manager(self, version: str) -> A2uiSchemaManager:
     return A2uiSchemaManager(
@@ -124,7 +155,7 @@ class ContactAgent:
             ui_description=UI_DESCRIPTION,
             include_examples=True,
             include_schema=True,
-            validate_examples=False,
+            validate_examples=False,  # Missing inline_catalogs for OrgChart and WebFrame validation
         )
         if schema_manager
         else get_text_prompt()
@@ -236,7 +267,9 @@ class ContactAgent:
       runner = self._ui_runners[ui_version]
       schema_manager = self._schema_managers[ui_version]
       selected_catalog = (
-          schema_manager.get_selected_catalog() if schema_manager else None
+          schema_manager.get_selected_catalog(client_ui_capabilities)
+          if schema_manager
+          else None
       )
     else:
       runner = self._text_runner
--- a/samples/agent/adk/contact_multiple_surfaces/agent_executor.py
+++ b/samples/agent/adk/contact_multiple_surfaces/agent_executor.py
@@ -56,6 +56,7 @@ class ContactAgentExecutor(AgentExecutor):
 
     logger.info(f"--- Client requested extensions: {context.requested_extensions} ---")
     active_ui_version = try_activate_a2ui_extension(context, self._agent.agent_card)
+    schema_manager = self._agent.get_schema_manager(active_ui_version)
 
     if active_ui_version:
       logger.info(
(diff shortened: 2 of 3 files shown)


agent_sdks/python/src/a2ui/a2a.py
```diff
@@ -151,20 +159,70 @@ def parse_response_to_parts(
...
-  return False
+  requested_extensions = _requested_a2ui_extensions(context)
+  if not requested_extensions:
+    return None
+
+  agent_advertised_extensions = _agent_extensions(agent_card)
+  if not agent_advertised_extensions:
+    return None
+
+  for req_uri in requested_extensions:
```
gspencergoog, line 223: So, the client is the one that determines which version gets selected if there are two, right?  If it sends [0.8, 0.9], then this will select 0.8 even if 0.9 is available, just because it is first.

Do we want to be more complicated than that, like select the "newest" version that are in both sets? Or do you think we should leave it up to the client to send them in preferred order?

nan-yu: I was under the impression that a client would typically only request a single version at a time. Is there a scenario where an Angular v0.9 renderer would need to request both v0.8 and v0.9 simultaneously?

If we do need to support multiple versions, we cannot rely on the client to determine priority or order, as `context.requested_extensions` is implemented as a set within the A2A SDK.

gspencergoog: The client can *support* more than one version at a time.  But someone has to decide.  Maybe the client is already doing this based on what the agent card has in it, and just requesting the one that it wants, and I just don't know how it works. :-)

nan-yu: Let's go with the current logic. I'll follow up with a separate PR to select the "newest" version and we can have more discussions there explicitly.

nan-yu:
I've got the contact_lookup, restaurant_finder and rizzcharts samples working individually, but when testing the orchestrator sample, the A2UI extension seems not to be propagated to the sub agents. I'll take a deeper look later and verify other samples, component_gallery and mcp_app_proxy.

agent_sdks/python/src/a2ui/a2a.py
jacobsimionato: SGTM. I think in many cases, clients will just support one version, e.g. if they only talk to one server, and they know that the server already supports the latest version. But for client which talk to multiple servers, e.g. orchestrators especially, they may want to activate multiple extension versions and then the server should pick its preferred version. So I think it makes sense for our library to automatically pick the newest version as you said in this case.

nan-yu:
Verified the following samples
- [x] contact_lookup
- [x] restaurant_finder 
- [x] rizzcharts
- [x] orchestrator
- [x] component_gallery
- [x]  mcp_app_proxy

agent_sdks/python/src/a2ui/a2a.py
```diff
@@ -151,20 +159,70 @@ def parse_response_to_parts(
...
+        ext
+        for ext in context.message.extensions
+        if isinstance(ext, str) and ext.startswith(A2UI_EXTENSION_BASE_URI)
+    ])
+
+  return requested_extensions
+
+
+def try_activate_a2ui_extension(
+    context: RequestContext, agent_card: AgentCard
```
wrenj, line 204: based on the a2a extension docs you don't technically need to advertise the extension in the agentcard to activate it
https://a2a-protocol.org/latest/topics/extensions/#extension-activation

the docs are pretty ambiguous TBH so up to you if you want to keep this, it seems reasonable to me but just wanted to flag

wrenj:
> I've got the contact_lookup, restaurant_finder and rizzcharts samples working individually, but when testing the orchestrator sample, the A2UI extension seems not to be propagated to the sub agents. I'll take a deeper look later and verify other samples, component_gallery and mcp_app_proxy.

fyi the code the orchestrator subagent extension activation is here
https://github.com/google/A2UI/blob/main/samples/agent/adk/orchestrator/agent.py#L67

it's pretty janky but thats the best we could do with the existing sdk

nan-yu:
> > I've got the contact_lookup, restaurant_finder and rizzcharts samples working individually, but when testing the orchestrator sample, the A2UI extension seems not to be propagated to the sub agents. I'll take a deeper look later and verify other samples, component_gallery and mcp_app_proxy.
> 
> fyi the code the orchestrator subagent extension activation is here https://github.com/google/A2UI/blob/main/samples/agent/adk/orchestrator/agent.py#L67
> 
> it's pretty janky but thats the best we could do with the existing sdk

Thanks for the pointer. I've got the orchestrator sample working last night by injecting the correct A2UI_EXTENSION_URI. All samples are working now.

agent_sdks/python/src/a2ui/a2a.py
nan-yu: I'd like to keep this here so the AgentCard can serve as the source of truth for server-supported versions. This allows us to implement a negotiation logic where an extension is only activated if the client's requested version matches what is advertised in the AgentCard.

The current logic is to select a random one that is both requested by the client and advertised by the AgentCard. We can enhance it to select the newest one if multiple matches.

agent_sdks/python/src/a2ui/a2a.py
nan-yu: Update the logic to select the newest version if multiple versions are compatible: https://github.com/google/A2UI/pull/936/changes/152619baababf3e2a30b33e62ed2792ac023702f.

dmandar:
Overall looks good. thx for doing this. Can we add screenshots showing the 3 custom components working for posterity? (contact org, card and location). Please also make sure clicking works in each.

samples/agent/adk/contact_lookup/agent.py
```diff
@@ -53,38 +53,47 @@ class ContactAgent:
...
-        artifact_service=InMemoryArtifactService(),
-        session_service=InMemorySessionService(),
-        memory_service=InMemoryMemoryService(),
+    self._text_runner: Optional[Runner] = self._build_runner(self._build_llm_agent())
+
+    self._schema_managers: Dict[str, A2uiSchemaManager] = {}
+    self._ui_runners: Dict[str, Runner] = {}
+
+    for version in [VERSION_0_8]:
+      schema_manager = self._build_schema_manager(version)
```
jacobsimionato, line 66: It'd be nice to try to reduce boilerplate at the call sites here. I wonder it could make sense to make the top level "inferenceStrategy" or "schemaManager" or "compositeInferenceStrategy` actually handle multiple protocols, so we only need one of them.

Then deeper inside `stream` etc, there can be logic which gets the correct PromptManager instance based on the version that is selected at runtime?

Optional though, we can always improve this later. So fantastic to have this implemented so quickly!!

nan-yu:
Screenshots added to the PR description. Merging now.

samples/agent/adk/contact_lookup/agent.py
nan-yu: Sure. Agreed, there are many boilerplate codes in both the agent and the client. We can definitely improve them later.

nan-yu: I filed https://github.com/google/A2UI/issues/969 to track the work.

Issue #898, opened 2026-03-19
SurfaceUpdate coming before BeginRendering breaks A2UI V0.8 rendering


sugoi-yuzuru:
Possible bug when surfaceUpdate is returned BEFORE beginRendering. 

I see a console error property accessing for `null` when A2UI surface is rendered. The reason is because the Surfaces returned from the MessageProcessor includes Surfaces that were created from `surfaceUpdate` message yet haven't received the corresponding `beginRendering` message. The renderer throws an error when surfaceRootId is not provided. 

Since the out-of-order A2UI DataPart response is by-design, we should apply some filtering logic to the v0.8 MessageProcessor to filter out surfaces that do not have SurfaceRootId property yet.

btw17:
This was resolved in https://github.com/google/A2UI/pull/928, so it can be closed

PR #928, opened 2026-03-20
Updated processor to return only visible surfaces

btw17:
Previously, processing `surfaceUpdate` prior to `beginRendering` would result in a surface returning via the message processor's `getSurfaces` method that could not be rendered. Now, any surface without a `rootComponentId` is filtered out.

See https://github.com/google/A2UI/issues/898 for more details.

# Description

## Pre-launch Checklist

- [x] I signed the [CLA].
  - N/A 
  - (Blake) Where is it? None exists in the repo. 
  - (Blake) Where is it? None exists in the repo. 
  - (Blake) None pertain to this specifically, as far as I've seen.

[CHANGELOG]: ../CHANGELOG.md


renderers/angular/src/v0_8/data/processor.ts
```diff
@@ -90,10 +90,26 @@ export class MessageProcessor {
...
   getSurfaces(): Map<string, WebCore.Surface> {
-    return (this.baseProcessor as any).surfaces || new Map();
+    return this.baseProcessor.getSurfaces() as Map<string, WebCore.Surface>;
   }
 
   clearSurfaces() {
     this.baseProcessor.clearSurfaces();
   }
 }
+
```
sugoi-yuzuru: Is this used at all?

btw17: Whoops - this is an unneeded artifact. All that we need is the `.getSurfaces()` update above

btw17: I originally fixed it in Angular in one commit, then when I went to fix it elsewhere, I noticed I just needed to update the core processor, so I reverted most of the Angular updates but forgot to remove that function, it seems.

PR #960, opened 2026-03-24
Add a definition for a generic signal to web core

ava-cassiopeia:
This can wrap any framework's implementation of a signal, so that we can let downstream implementations use their native signals instead of always relying on (or wrapping) Preact signals.


gspencergoog:
So, is the idea that there is no default signal type (e.g. preact), and someone implementing a new renderer would be required to implement one of these, so that all signals are "native" to the framework?

renderers/web_core/src/v0_9/reactivity/signals.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      unwrap: <T>(val: ASignal<T>) => val(),
+      set: <T>(signal: AWritableSignal<T>, value: T) => signal.set(value),
+    };
+
+    it('round trip wraps and unwraps successfully', () => {
+      const val = 'hello';
+      const wrapped = AngularSignal.wrap(val);
+      assert.strictEqual(AngularSignal.unwrap(wrapped), val);
+    });
+
```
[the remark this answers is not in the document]
ava-cassiopeia: This has nothing to do with the highlighted text.

ava-cassiopeia:
> So, is the idea that there is no default signal type (e.g. preact), and someone implementing a new renderer would be required to implement one of these, so that all signals are "native" to the framework?

Exactly. In a followup PR, I would set it up such that all of our reactive stuff that currently transacts in Preact signals would instead use a FrameworkSignal implementation to interact with the actual underlying framework signals.

ava-cassiopeia:
I could also imagine providing a "default" impl of FrameworkSignal that uses preact signals (or whatever we decide is 'generic' or 'default'), but I think in practice any downstream framework probably wants to use their own thing.

renderers/web_core/src/v0_9/reactivity/signals.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A generic representation of a Signal that could come from any framework.
+ * For any library building on top of A2UI's web core lib, this must be
+ * implemented for their associated signals implementation.
+ */
+export interface FrameworkSignal<SignalType> {
+  /**
+   * Create a computed signal for this framework.
+   */
+  computed<T>(fn: () => T): SignalType;
```
jacobsimionato, line 26: Will there also be ways to combine signals together? We need this to do stuff like subscribe to multiple data model keys and then evaluate a function which depends on all of them etc.

renderers/web_core/src/v0_9/catalog/signals.ts
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * A generic representation of a Signal that could come from any framework.
+ * For any library building on top of A2UI's web core lib, this must be
+ * implemented for their associated signals implementation.
+ */
+export interface FrameworkSignal<SignalType> {
```
jacobsimionato: I like that this represents the entire interface in a single class, and I like the wrap/unwrap structure.

renderers/web_core/src/v0_9/reactivity/signals.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  computed as aComputed,
+  Signal as ASignal,
+  WritableSignal as AWritableSignal,
+  isSignal,
+} from '@angular/core';
+
+import {FrameworkSignal} from './signals';
+
+describe('FrameworkSignal', () => {
+  // Test FrameworkSignal with two sample implemenations that wrap Angular and
```
jacobsimionato, line 31: Amazing idea!

renderers/web_core/src/v0_9/reactivity/signals.ts
jacobsimionato, line 1: How about putting this at a more generic codebase location? E.g. `renderers/web_core/signals/` or something

jacobsimionato:
> > So, is the idea that there is no default signal type (e.g. preact), and someone implementing a new renderer would be required to implement one of these, so that all signals are "native" to the framework?
> 
> Exactly. In a followup PR, I would set it up such that all of our reactive stuff that currently transacts in Preact signals would instead use a FrameworkSignal implementation to interact with the actual underlying framework signals.

Cool! I guess the open question then is how to inject the "SignalFactory" etc into all the classes which need to create signals in web_core.

I'd favor putting in the constructor of each object (e.g. MessageProcessor, SurfacesModel, SurfaceModel, DataModel, DataContext etc) and passing it through, seeing as every framework will have a different approach to DI.

One prominent place where Signals are exposed on the API for web_core is FunctionImplementation - see https://github.com/google/A2UI/blob/027307636800e4af41586a59461fb8a47882a788/renderers/web_core/src/v0_9/catalog/types.ts#L69. This is a bit unique, because here we let application developers *supply* us a signal representing the changing output of a function, rather than get a signal that we create. I think we should perhaps add `SignalFactory` or similar as a parameter of the `execute` method.

jacobsimionato (approved):
Hitting approve because my comments are minor - I mostly think it's worth putting this in a more generic location like `/signals` or `/types` or `/primitives` or something.

ava-cassiopeia:
> I'd favor putting in the constructor of each object (e.g. MessageProcessor, SurfacesModel, SurfaceModel, DataModel, DataContext etc) and passing it through, seeing as every framework will have a different approach to DI.

Yeah, the draft I'm working on takes this approach. It also means it forces upstream libraries to explicitly implement this interface, which is nice for consistency.

renderers/web_core/src/v0_9/reactivity/signals.test.ts
ava-cassiopeia: Thanks!

renderers/web_core/src/v0_9/reactivity/signals.ts
ava-cassiopeia: Seems reasonable to me, I moved these to `web_core/reactivity/signals.ts`.

renderers/web_core/src/v0_9/catalog/signals.ts
ava-cassiopeia: Thanks! Yeah I'm trying to make it cover all the common cases!

renderers/web_core/src/v0_9/reactivity/signals.ts
ava-cassiopeia: That's what `computed()` and _sometimes_ `effect()` are used for in most signals libraries. That's a little different than I think the pattern you're imagining, where there are explicit helpers for piping things around (a la rxjs).

In signals-world, you can set up things like (Angular signals syntax):

```ts
var a = signal(1);
var b = computed(() => a() + 1);
var c = computed(() => a() + b());
```

The signals framework is smart enough to produce a graph of deps like:

```mermaid
graph TD;
    a-->b;
    a-->c;
    b-->c;
```

Which implicitly allows for the combination of signals.

As such, I'm not going to make any changes here.

renderers/web_core/src/v0_9/reactivity/signals.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * Check if an arbitrary object is a framework signal.
+   */
+  isSignal(val: unknown): val is SignalType;
```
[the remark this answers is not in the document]
ava-cassiopeia: This is actually good feedback, I'll make a similar-ish change.

PR #970, opened 2026-03-25
Lit renderer example with react fixes

jacobsimionato:
Just a draft as a demo


ditman:
(Check the fixes to the react renderer here: https://github.com/google/A2UI/pull/968)

PR #972, opened 2026-03-25
fix(react-renderer): fix icon rendering in v0.9 basic catalog

### LINKED ISSUE #946: [React] Fix icon rendering
Currently, it's broken. See:

Image

jacobsimionato:
### Description of Changes
- **Automatic Font Loading**: Added a side-effect to `Icon.tsx` that programmatically injects the Google Fonts stylesheet for "Material Symbols Outlined" if it's not already detected in the document. This makes the `Icon` component self-contained.
- **Icon Name Mapping**: Implemented an `ICON_MAPPING` dictionary to bridge the gap between A2UI specification names (e.g., `play`, `rewind`, `starOff`) and the actual ligatures required by Material Symbols (e.g., `play_arrow`, `fast_rewind`, `star` with outlined setting).
- **CamelCase to Snake_case Conversion**: Added a regex transformer to convert standard A2UI camelCase icon names into the snake_case format required for font ligatures.
- **Variable Font Support**: Added support for the `FILL` variation axis. This allows icons like `star` and `favorite` to correctly render as filled when requested, while their "off" variants render as outlined.

### Rationale
The v0.9 React renderer's `Icon` component was previously just rendering the raw icon name string. This failed for two reasons:
1. The font stylesheet wasn't being loaded automatically.
2. Many icon names in the A2UI spec (like `play`) don't have exact string matches in the Material Symbols font family (which uses `play_arrow`).

These changes ensure that icons rendered via the React v0.9 catalog match the visual behavior of the v0.8 Lit/Angular renderers and follow the design requirements of the Material Symbols variable font.

### Testing/Running Instructions
1. Run the A2UI Explorer or a React application using the v0.9 renderer.
2. Load a surface that uses icons, such as the "Track List" example (`18_track-list.json`).
3. Verify that the `play` icon and other system icons (like `star`, `favorite`) render correctly as glyphs rather than raw text.
4. Verify that the "Material Symbols Outlined" font stylesheet is added to the `<head>` of the document upon the first `Icon` component mount.

Fixes https://github.com/google/A2UI/issues/946

### DIFF
--- a/renderers/react/src/v0_9/catalog/basic/components/Icon.tsx
+++ b/renderers/react/src/v0_9/catalog/basic/components/Icon.tsx
@@ -19,9 +19,44 @@ import {createReactComponent} from '../../../adapter';
 import {IconApi} from '@a2ui/web_core/v0_9/basic_catalog';
 import {getBaseLeafStyle} from '../utils';
 
+// Programmatically load the material symbols font if the application hasn't already.
+if (typeof document !== 'undefined') {
+  const isLoaded = Array.from(document.querySelectorAll('link')).some(
+    (link) => link.href.includes('Material+Symbols+Outlined')
+  );
+  if (!isLoaded) {
+    const link = document.createElement('link');
+    link.rel = 'stylesheet';
+    link.href = 'https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@24,400,0,1';
+    document.head.appendChild(link);
+  }
+}
+
+const ICON_MAPPING: Record<string, string> = {
+  play: 'play_arrow',
+  rewind: 'fast_rewind',
+  favoriteOff: 'favorite',
+  starOff: 'star',
+};
+
+const FILLED_ICONS = ['star', 'starHalf', 'favorite'];
+
 export const Icon = createReactComponent(IconApi, ({props}) => {
-  const iconName =
+  const rawIconName =
     typeof props.name === 'string' ? props.name : (props.name as {path?: string})?.path;
+  
+  // Use mapping if available, otherwise convert camelCase to snake_case
+  let iconName = '';
+  if (rawIconName) {
+    if (ICON_MAPPING[rawIconName]) {
+      iconName = ICON_MAPPING[rawIconName];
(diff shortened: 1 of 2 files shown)


renderers/react/src/v0_9/catalog/basic/components/Icon.tsx
```diff
@@ -19,9 +19,44 @@ import {createReactComponent} from '../../../adapter';
 import {IconApi} from '@a2ui/web_core/v0_9/basic_catalog';
 import {getBaseLeafStyle} from '../utils';
 
+// Programmatically load the material symbols font if the application hasn't already.
+if (typeof document !== 'undefined') {
+  const isLoaded = Array.from(document.querySelectorAll('link')).some(
+    (link) => link.href.includes('Material+Symbols+Outlined')
```
ava-cassiopeia, line 25: This will in some edge cases end up burning you. Material Symbols offers dynamic icon loading from their CDN, meaning that users can provide Material Symbols URLs that look like:

```html
<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200&icon_names=close" />
```

Note the `icon_names=close`, which will cause that to only load the close icon and nothing else, saving significantly on weight.

This code will match the above href, even though it doesn't pull in the full library, which I guess we probably need for the sake of the AI choosing any icon? More on that in my other comment.

ava-cassiopeia (commented):
Is this matching behavior seen in other, competing frameworks?

I ask because unless we're trying to match functionality, I'd actually strongly recommend against this.

This creates an opinionated link to Material Symbols, when downstream users may want to use a different icon library. We probably ought to provide a way to override the font library used (even the [MatIcon](https://material.angular.dev/components/icon/api#MatIcon), which is more Material-opinionated, provides an option for this).

I'm also wary of importing a whole icon library due to weight, as most icon libraries offer incremental loading and can be quite heavy if imported in full. Then again, I suppose the AI could produce any icon name...and speaking of, do we tell the agent what icons are available to it..? That seems like something we need if we're gonna choose the library here.

PR #982, opened 2026-03-25
fix(angular): unify preact signals instance and make tests headless

gspencergoog:
# Description

Fixes the export of signals so it comes from `web_core` instead of `preact`, because it caused duplicate definitions and tests were failing.  Not sure why that wasn't caught in CI...

Also switched the `a2ui_explorer` example tests to use Jasmine instead of vitest, since the other tests are Jasmine tests, for consistency.

Made the tests run headless, since there's no value in seeing them run (and it steals focus if you're trying to work, and doesn't work over ssh).


renderers/angular/package-lock.json
```diff
@@ -47,27 +47,17 @@
         "@angular/platform-browser": "^21.2.0"
       }
     },
-    "../web_core": {
-      "name": "@a2ui/web_core",
-      "version": "0.8.7",
+    "node_modules/@a2ui/web_core": {
+      "version": "0.8.8",
+      "resolved": "file:../web_core",
```
ava-cassiopeia: this doesn't look right, does it? Should it have a resolved that is somewhere else? I don't think this will work right after it is published.

ava-cassiopeia (changes requested):

renderers/angular/package-lock.json
gspencergoog: I don't know any other way to do it, so if you have ideas...
The only suggestion I've seen is to convert to an npm monorepo, but that seems to come with a lot of headaches.

ava-cassiopeia: Yeah, I think a NPM / yarn monorepo is the right call (using workspaces) but for now I think this is a fine approach, per discussion on the GVC.

ava-cassiopeia (approved):

PR #994, opened 2026-03-26
[ fix ] : fix memory leaks and standardize subscription lifecycle in v0.9

### LINKED ISSUE #963: [v0.9] Fix memory leaks and standardize subscription lifecycle in web_core and react renderers
This issue tracks critical memory leaks and architectural redundancies identified in the A2UI v0.9 Web, React, and Angular renderers. These issues lead to unbounded memory growth in long-lived surfaces and leaked background processes (e.g., timers) when components are unmounted.

#### Identified Issues

**Core & Shared Logic:**
1.  **Unbounded Signal Cache in `DataModel`**: The `DataModel.signals` Map caches every path ever queried. These are only cleared when the entire surface is disposed. In surfaces with many ephemeral components or large lists, this leads to significant memory growth.
2.  **`FormatString` Recursive Leaks**: The `formatString` implementation in `basic_functions.ts` resolves internal expressions into signals using `context.resolveSignal(part)` but never calls `unsubscribe` on them. Any function call (like `${now()}`) inside a string will leak its listeners/timers until the DataModel is disposed.
3.  **Broken Cleanup Chain in `DataContext`**: `resolveSignal` in `DataContext.ts` creates recursive effects for function arguments but does not properly take ownership of `unsubscribe` methods on signals returned by function implementations.

**React Renderer:**
4.  **Redundant Subscription Layers**: The React adapter wraps the `GenericBinder` in additional `useSyncExternalStore` and `useEffect` layers. While not a leak, it increases complexity and makes cleanup harder to audit.

**Angular Renderer:**
5.  **Root-Scoped Leaks in `ComponentBinder`**: `ComponentBinder.ts` injects `DestroyRef` at the root level. All subscriptions created via `bind()` are tied to the application lifetime rather than the component lifetime, causing massive leaks as users navigate surfaces.
6.  **Stale Bindings**: Angular does not listen to `ComponentModel.onUpdated`. If component properties change in the protocol (e.g., type remains same but props change), the Angular UI stays stale.

#### Goals
*   Implement reference counting or explicit signal removal in `DataModel`.
*   Ensure all recursive signals created by `formatString` and function calls are tracked and disposed of via a unified `unsubscribe` chain.
*   Fix the `DestroyRef` scoping in the Angular renderer.
*   Implement property refresh logic in Angular to match the React `GenericBinder` behavior.
*   Add exhaustive regression tests verifying that `AbortSignal` listeners and internal Maps are cleared after component unmount across all frameworks.

Dijo-404:
## Description
Fixes #963 — Addresses critical memory leaks and standardizes subscription lifecycle management across the A2UI `web_core` and Angular renderer. This PR prevents unbounded memory growth in long-lived surfaces and eliminates leaked background processes (e.g., timers) when components unmount or properties update.

## What's Fixed
- **`DataModel.signals` cache**: Added reference counting to evict unused signal entries and prevent unbounded Map growth.
- **`FormatString` leaks**: Tracked and disposed of all recursive signals created inside format strings (e.g., `${now()}`).
- **`DataContext` unsubscribe chain**: Properly propagated `unsubscribe` ownership for signals returned by function implementations.
- **Angular `ComponentBinder` scoping**: Moved `DestroyRef` injection to component scope to prevent application-lifetime subscription leaks.
- **Angular stale bindings**: Wired up `ComponentModel.onUpdated` to refresh component properties when the protocol updates them.

## Testing
- Added regression tests verifying `AbortSignal` listeners and internal Maps are cleared on component unmount.

## Verification 
1. Verify Memory Leak Fixes 
image
2. Verify Angular Build
image

## Note 
There is an lint error in ```src/v0_9/catalog/types.test.ts ``` which i haven't touched, its due unused variables, let me know should i fix that and push it in this pr itself or i need to put it as an issue and issue another pr fixing that


polina-c:
Thank you for your contribution to the A2UI ecosystem! Unfortunately, we must decline this PR at this time: we review PRs against an issue only from the contributor the issue is assigned to, and this one is not assigned to you. Issues are assigned in priority order, so we cannot say yet when this one will be picked up.

We will link this PR to the issue, so whoever takes it on can see your approach. They may reopen this PR and build on it rather than start from scratch.

PR #996, opened 2026-03-26
Pin @a2ui/web_core versions and configure React for publishing

jacobsimionato:
### Description of Changes:
- General changes:
  - Updated current version in package.json for all renderers to v0.9.0
  - Updated metadata around repository, tags, homepage etc to be correct
- **Angular:** Modified  to remove the caret () when resolving the  version.
- **Lit:** Modified  to remove the caret () when resolving the  version.
- **Markdown-it:** Modified  to remove the caret () when resolving the  version for both  and .
- **React:** Added a new  script to handle the publication transformation for the React renderer, including path resolution for , , , and .
- **React:** Updated  with  and  scripts to ensure proper publication from the  directory and prevent broken  references.
- **Web Core:** Updated  to version  and added  as a dev dependency.
- **Registry Configuration:** Updated  with the correct registry path.

### Rationale:
- To avoid issues with  API stability, each renderer should be pinned to a specific version of the core library instead of using a caret range.
- The React renderer was previously missing the necessary configuration to be published to NPM correctly. Without a rewrite of the  dependency during publication, any published package would have been broken for external users.

### Testing/Running Instructions:
- For each renderer (, , , ), verify that running the publication script (e.g.,  or ) generates a  with a pinned version of  (e.g.,  instead of ).
- Specifically for React, run  (dry run or local check) to ensure the  is correctly transformed from  and that all export paths point correctly within the  folder.

### DIFF
--- a/renderers/web_core/package.json
+++ b/renderers/web_core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/web_core",
-  "version": "0.8.1",
+  "version": "0.9.0",
   "description": "A2UI Core Library",
   "homepage": "https://a2ui.org/",
   "repository": {


renderers/react/scripts/prepare-publish.mjs
```diff
@@ -0,0 +1,112 @@
+/**
+ * Copyright 2026 Google LLC
```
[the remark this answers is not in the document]
jacobsimionato: It's 2026, so we'll keep this

renderers/react/scripts/prepare-publish.mjs
```diff
@@ -0,0 +1,112 @@
...
+if (!coreVersion) throw new Error('Cannot determine @a2ui/web_core version');
+
+// 2. Read Package
+const pkg = JSON.parse(readFileSync(pkgPath, 'utf8'));
+
+// 3. Update Dependency
+if (pkg.dependencies && pkg.dependencies['@a2ui/web_core']) {
+  pkg.dependencies['@a2ui/web_core'] = coreVersion;
+} else {
+  console.warn('Warning: @a2ui/web_core not found in dependencies.');
```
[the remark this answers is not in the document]
jacobsimionato: Lets fix this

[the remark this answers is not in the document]
jacobsimionato: Fixed.

renderers/angular/package.json
```diff
@@ -1,7 +1,24 @@
 {
   "name": "@a2ui/angular",
-  "version": "0.8.5",
+  "version": "0.8.4",
```
ditman: Rolling back the version?

renderers/lit/prepare-publish.mjs
```diff
@@ -41,9 +41,9 @@ const litPkg = JSON.parse(readFileSync(litPkgPath, 'utf8'));
 
 // 3. Update Dependency
 if (litPkg.dependencies && litPkg.dependencies['@a2ui/web_core']) {
-  litPkg.dependencies['@a2ui/web_core'] = '^' + coreVersion;
+  litPkg.dependencies['@a2ui/web_core'] = coreVersion;
 } else {
-  console.warn('Warning: @a2ui/web_core not found in dependencies.');
+  throw new Error('Error: @a2ui/web_core not found in dependencies. This is a mandatory dependency for publishing.');
```
ditman: Why is this script so different from the one on the angular renderer? Can we share this script across all packages we want to publish?

renderers/web_core/package.json
```diff
@@ -1,7 +1,24 @@
 {
   "name": "@a2ui/web_core",
-  "version": "0.8.8",
+  "version": "0.8.1",
```
ditman: Pls don't roll back the version here!

renderers/angular/postprocess-build.mjs
```diff
@@ -37,7 +37,7 @@ if (!packageJson.dependencies['@a2ui/web_core']) {
   );
 }
 
-packageJson.dependencies['@a2ui/web_core'] = '^' + coreVersion;
+packageJson.dependencies['@a2ui/web_core'] = coreVersion;
```
ditman: Should this be called `prepare-publish.mjs` like in the other renderers?

ditman (commented):
It's weird that every version of prepare-publish.mjs is slightly different than every other. Can we share that? Isn't all it's doing just fixing the built version of the package.json to point to the (about to be published) version of web_core?

renderers/angular/package.json
jacobsimionato: I meant to put v0.9.0 here sorry - updated.

renderers/lit/prepare-publish.mjs
jacobsimionato: Great point. I centralized it now.

renderers/web_core/package.json
jacobsimionato: Fixed to v0.9.0 now

renderers/angular/postprocess-build.mjs
jacobsimionato: It's centralized now.

jacobsimionato:
I believe it's the end of your workday @ditman so I'm gonna go ahead and merge this seeing as I addressed your feedback. Thanks for the feedback!

PR #1002, opened 2026-03-27
chore: unify publish scripts

jacobsimionato:
### Description of Changes

This PR updates the npm publishing configuration for all web renderer packages to default to private publishing and unifies the pre-publish preparation scripts.

- **New Public Publish Command:** Added a `publish:package` script to all packages which calls `npm run publish:package -- --access public` for releases.
- **Unified Publish Scripts:** 
  - Deleted the redundant `renderers/markdown/markdown-it/prepare-publish.mjs` script.
  - Updated the shared `renderers/scripts/prepare-publish.mjs` to resolve paths relative to `import.meta.url` (instead of `process.cwd()`), allowing it to be safely executed from nested package directories like `markdown-it`.
  - Added logic to the shared script to ensure `@a2ui/web_core` is correctly version-bumped in `peerDependencies`.
- **Documentation:** Updated `renderers/docs/web_publishing.md` to reflect the inclusion of the `react` package, explain the new default private publishing behavior, and document the `publish:public` command.

### Rationale

Unifying `prepare-publish.mjs` ensures all renderer packages share the exact same publishing metadata preparation logic (such as stripping `devDependencies` and updating `file:` dependencies), making the repository easier to maintain.

### Testing/Running Instructions

1. Check out this branch.
2. Navigate to any renderer package (e.g., `cd renderers/markdown/markdown-it`).
3. Run `npm run build` followed by `node ../../scripts/prepare-publish.mjs`.
4. Inspect the generated `dist/package.json` file. Ensure that:
   - The `@a2ui/web_core` version in `peerDependencies`/`dependencies` has been correctly updated to `^0.9.0`.
   - `devDependencies` have been removed.
   - Internal paths in `main`, `types`, and `exports` no longer contain the `./dist/` prefix.
5. Attempting to run `npm run publish:package` will prepare the dist folder and run `npm publish dist/` (which defaults to private). Running `npm run publish:public` will append `--access public`.

### DIFF
--- a/renderers/angular/package-lock.json
+++ b/renderers/angular/package-lock.json
@@ -81,7 +81,7 @@
     },
     "../web_core": {
       "name": "@a2ui/web_core",
-      "version": "0.9.0-alpha.1",
+      "version": "0.9.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@preact/signals-core": "^1.13.0",
--- a/renderers/lit/package-lock.json
+++ b/renderers/lit/package-lock.json
@@ -35,7 +35,7 @@
     },
     "../web_core": {
       "name": "@a2ui/web_core",
-      "version": "0.9.0-alpha.1",
+      "version": "0.9.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@preact/signals-core": "^1.13.0",
--- a/renderers/markdown/markdown-it/package-lock.json
+++ b/renderers/markdown/markdown-it/package-lock.json
@@ -29,7 +29,7 @@
     },
     "../../web_core": {
       "name": "@a2ui/web_core",
-      "version": "0.9.0-alpha.1",
+      "version": "0.9.0",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
--- a/renderers/react/package-lock.json
+++ b/renderers/react/package-lock.json
@@ -71,7 +71,7 @@
     },
     "../web_core": {
       "name": "@a2ui/web_core",
-      "version": "0.9.0-alpha.1",
+      "version": "0.9.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@preact/signals-core": "^1.13.0",
--- a/renderers/web_core/package-lock.json
+++ b/renderers/web_core/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@a2ui/web_core",
-  "version": "0.9.0-alpha.1",
+  "version": "0.9.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@a2ui/web_core",
-      "version": "0.9.0-alpha.1",
+      "version": "0.9.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@preact/signals-core": "^1.13.0",
--- a/renderers/web_core/package.json
+++ b/renderers/web_core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/web_core",
-  "version": "0.9.0-alpha.1",
+  "version": "0.9.0",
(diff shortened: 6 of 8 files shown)


ava-cassiopeia:
Why are we defaulting to private publishing, I don't think we can even do that, right? The NPM org is a public-only org, we'd have to upgrade it to support private publishing.

jacobsimionato:
> Why are we defaulting to private publishing, I don't think we can even do that, right? The NPM org is a public-only org, we'd have to upgrade it to support private publishing.

You're right - updated to public

renderers/docs/web_publishing.md
```diff
@@ -24,101 +25,59 @@ Ensure you have an NPM Access Token with rights to the `@a2ui` organization.
 
 ---
 
-## 📦 1. Publishing `@a2ui/web_core`
+## 📦 Publishing Packages
```
ditman: (I'd remove the box emoji 📦 :P)

renderers/docs/web_publishing.md
```diff
@@ -24,101 +25,59 @@ Ensure you have an NPM Access Token with rights to the `@a2ui` organization.
...
 **What happens during `npm run publish:package`?**
-Before publishing, the script runs the necessary `build` command which processes the code. For Lit and Markdown-it, `prepare-publish.mjs` runs, and for Angular, `postprocess-build.mjs` runs. These scripts:
-1. Copy `package.json`, `README.md`, and `LICENSE` to the `dist/` folder.
-2. Read the `version` from `@a2ui/web_core`.
-3. Update the `file:` dependency in the `dist/package.json` to the actual core version (e.g., `^0.8.0`).
-4. Adjust exports and paths to be relative to `dist/`.
-5. Remove any build scripts (`prepublishOnly`, `scripts`) so they don't interfere with the publish process.
+Before publishing, the script runs the necessary `build` command which processes the code. Then, a preparation script (usually `prepare-publish.mjs`) runs, which:
+1. Copies `package.json`, `README.md`, and `LICENSE` to the `dist/` folder.
+2. If it's a renderer, it reads the `version` from `@a2ui/web_core` and updates the `file:` dependency in the `dist/package.json` to the actual core version (e.g., `^0.9.0`).
```
ditman, line 60: Should we specify somewhere that when changing multiple packages, `web_core` **must** be published first? Otherwise we'd be publishing packages with a dependency to an unpublished version of the package?

renderers/lit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/lit",
-  "version": "0.9.0",
+  "version": "0.9.1",
```
ditman, line 3: Revert this version change?

renderers/scripts/prepare-publish.mjs
```diff
@@ -53,6 +56,9 @@ const pkg = JSON.parse(readFileSync(resolvedSourcePkg, 'utf8'));
 if (pkg.dependencies && pkg.dependencies['@a2ui/web_core']) {
   pkg.dependencies['@a2ui/web_core'] = '^' + coreVersion;
 }
+if (pkg.peerDependencies && pkg.peerDependencies['@a2ui/web_core']) {
+  pkg.peerDependencies['@a2ui/web_core'] = '^' + coreVersion;
+}
```
ditman, line 61: This is nice, I like it handling peer dependencies too!

renderers/docs/web_publishing.md
jacobsimionato: Yes, that's a good idea. I'm about to revamp this though, so I wont' worry about it now. I'll have a script that checks this.

renderers/lit/package.json
jacobsimionato: Updated this because that's the version it's publish at now I'm afraid 😳

PR #1008, opened 2026-03-27
Improve generic signal implementation

ava-cassiopeia:
Reworks the generic signals impl that I created in #960 to have better type safety.

In this implementation, downstream libraries must explicitly provide a definition for their signals, and in return we'll preserve signal types as they transit into and out of web core. An example of how this works is in the test file, which has to do the same thing to be compliant with the code.

These changes came about as I started trying to wire these through the core library and was running into annoying typing issues - this should fix all that, and I'll issue a followup PR that actually uses these in web core.

### DIFF
--- a/renderers/web_core/CHANGELOG.md
+++ b/renderers/web_core/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.9.1
+
+- Add new `FrameworkSignal` concept, which represents a generic signal from a
+  given framework like Preact or Angular.
+  - Unused in this version; future versions will introduce this throughout web
+    core and will likely be breaking changes.
+
 ## 0.8.8
 
 - Add the ability to access the `schema` of a component in a type-safe way.
--- a/renderers/web_core/package.json
+++ b/renderers/web_core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/web_core",
-  "version": "0.9.0",
+  "version": "0.9.1",
   "description": "A2UI Core Library",
   "homepage": "https://a2ui.org/",
   "repository": {


renderers/web_core/src/v0_9/reactivity/signals.test.ts
```diff
@@ -16,30 +16,53 @@
...
       wrap: <T>(val: T) => aSignal(val),
       unwrap: <T>(val: ASignal<T>) => val(),
       set: <T>(signal: AWritableSignal<T>, value: T) => signal.set(value),
+      effect: (fn: () => void, cleanupCallback: () => void) => {
+        const e = aEffect(cleanupRegisterFn => {
+          cleanupRegisterFn(cleanupCallback);
+          fn();
+        });
+        return () => e.destroy();
+      },
```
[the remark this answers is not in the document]
ava-cassiopeia: It's a test impl 🤷🏻‍♀️

renderers/web_core/src/v0_9/reactivity/signals.test.ts
```diff
@@ -79,12 +102,13 @@ describe('FrameworkSignal', () => {
...
 
   describe('Preact variation', () => {
-    const PreactSignal: FrameworkSignal<PSignal> = {
+    const PreactSignal: FrameworkSignal<'preact'> = {
       computed: <T>(fn: () => T) => pComputed(fn),
       isSignal: (val: unknown) => val instanceof PSignal,
       wrap: <T>(val: T) => new PSignal(val),
       unwrap: <T>(val: PSignal<T>) => val.value,
       set: <T>(signal: PSignal<T>, value: T) => (signal.value = value),
+      effect: (fn: () => void) => pEffect(fn),
```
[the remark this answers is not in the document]
ava-cassiopeia: This is also a test impl, so I don't mind much here

renderers/web_core/src/v0_9/reactivity/signals.ts
```diff
@@ -14,34 +14,60 @@
  * limitations under the License.
  */
 
+// SignalKinds and WritableSignalKinds are declared in such a way that
+// downstream library impls can dynamically provide their Signal implementations
+// in a type-safe way. Usage downstream might look something like:
+//
+// declare module '../reactivity/signals' {
+//   interface SignalKinds<T> {
+//     preact: Signal<T>;
```
ditman, line 23: Is this "preact" the library providing the signals? What would this look like for lit of maybe react?

ava-cassiopeia: Yes, but like you can set those words to whatever. Like, there's two phases to integrating with this. Declaring what type of signals you're dealing in:

```ts
import {Signal as PreactSignal} from 'preact blah blah';
import {Signal as AngularSignal} from '@angular/core';
import {Signal as AvaSignal} from '@ava-cassiopeia/fake-signal';

declare module '@a2ui/web_core/reactivity/signals' {
  interface SignalKinds<T> {
    preact: PreactSignal<T>;
    angular: AngularSignal<T>;
    coolestSignals: AvaSignal<T>;
  }
}
```

The key thing here is that we, as web core, **don't** define the types of signals ahead of time so that anyone can plug in any signal impl., even one we don't know about, as you can see with `coolestSignals` above.

Then, when they start constructing our core library objects, those objects will want for a `FrameworkSignal` implementation that uses one of those signals, a la:

```ts
export const AngularFrameworkSignal: FrameworkSignal<'angular'> = { /* impl */ };

export const AvaFrameworkSignal: FrameworkSignal<'coolestSignals'> = { /* impl */ };
```

So in short, our library says something like:

```ts
constructor(readonly ctx: DataContext<'react'>) {}
```

_"I need a DataContext that specifically deals in React signals"_

and then you might provide that by creating a DataContext that uses React signals:

```ts
const ctx = new DataContext(ReactFrameworkSignal);

// call the constructor from before. It won't take a DataContext that
// is using a non-matching FrameworkSignal implementation.
new FakeThing(ctx);
```

ditman: This is cool! I was iffy about using `'react'` as a string, but in reality, that accessor is only used when the framework signal type is defined!

ava-cassiopeia:
> PS: Is this a breaking change, or can this just land because we're not using the FrameworkSignals in the rest of the renderer?

Nope, we can just land this. My followup, #1009, is the breaking change.

PR #1017, opened 2026-03-27
Create signal implementations that `satisfies FrameworkSignal`. Simplify `SignalKinds`.

ditman:
# Description

Make SignalKinds just a holder for "readonly" and "writable" signal types, and use that in the FrameworkSignal type.

Users need to define their FrameworkSignal type as `satisfies FrameworkSignal` to prevent TS from erasing the internal types (similar to what we do with the component props!).

Split tests into 3 files: 2 for per-framework initialization, and a common one that runs the same tests across both implementations. The verification code remains the same as in the OG commit.

There's some shenanigans with tsc merging all our test files together and seeing the Preact and Angular types colliding with each other, and tests are not *exactly identical* to what users would define, but this is the general idea.

(I don't know if this does everything we _want_ though!)

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/web_core/src/v0_9/reactivity/signals.preact.test.ts
+++ b/renderers/web_core/src/v0_9/reactivity/signals.preact.test.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 /**
  * Copyright 2026 Google LLC
  *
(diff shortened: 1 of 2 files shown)


renderers/web_core/src/v0_9/reactivity/signals.ts
ava-cassiopeia, line 1: Thanks for putting this alternative together - do you think we should present this to the core a2ui team and see what they think?

renderers/web_core/src/v0_9/reactivity/signals.preact.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  unwrap: <T>(val: PSignal<T>) => val.value,
+  set: <T>(sig: PSignal<T>, value: T) => (sig.value = value),
+  effect: (fn: () => void) => effect(fn),
+} as unknown as FrameworkSignal; // Cast bypasses Mono-compilation interface overlap
+// The cast above is needed because tsc is merging all our test files together,
+// and the SignalKinds interface is being declared multiple times, causing a
+// type collision. Normally, the AngularSignal would `satisfies FrameworkSignal`,
+// and the declaration of SignalKinds wouldn't need to suppress anything.
+
+runFrameworkSignalTests('Preact implementation', PreactSignal);
```
ava-cassiopeia, line 45: I'm *strongly* against code reuse in tests like this, as it actively buries the lede in tests, following the philosophy in [go/tott/643](https://goto.google.com/tott/643).

ditman: Yeah, I know, I just did this to "prove" (by the test construction) that both implementations behaved the same. Since the specific implementations themselves (preact and angular) are not that important, and the testing logic is not that big, I guess we could just duplicate the test cases on each file; I don't mind!

renderers/web_core/src/v0_9/reactivity/signals.ts
ditman: If you're dialing in today, let's bring it up? I'm not sure who else is going to have opinions about this other than us? Maybe @gspencergoog?

renderers/web_core/src/v0_9/reactivity/signals.preact.test.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Test FrameworkSignal with Preact signals explicitly mapped over SignalKinds.
+const PreactSignal = {
+  computed: <T>(fn: () => T) => computed(fn),
+  isSignal: (val: unknown): val is PSignal<any> => val instanceof PSignal,
+  wrap: <T>(val: T) => new PSignal(val),
+  unwrap: <T>(val: PSignal<T>) => val.value,
+  set: <T>(sig: PSignal<T>, value: T) => (sig.value = value),
+  effect: (fn: () => void) => effect(fn),
```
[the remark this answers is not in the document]
ditman: @ava-cassiopeia what do you think about this feedback? How picky should we be with these test implementations? (I guess people may copy-paste them and then encounter bugs?)

renderers/web_core/src/v0_9/reactivity/signals.ts
```diff
@@ -14,37 +14,45 @@
...
+ * A generic representation of a read-only Signal.
+ * Resolves to the specific framework's signal type if augmented.
+ */
+export type Signal<T> = SignalKinds<T> extends { readonly: infer R } ? R : unknown;
+
+/**
+ * A generic representation of a writable Signal.
+ * Resolves to the specific framework's signal type if augmented.
+ */
+export type WritableSignal<T> = SignalKinds<T> extends { writable: infer W } ? W : unknown;
```
ditman, line 44: This is a usability problem. The way these two types are inferred is by extracting the "readonly" and "writable" properties from the `SignalKinds` that the user is defining.

HOWEVER, the `SignalKinds` type can't enforce that users define them with `readonly` and `writable`; users can do:

```
declare module './signals.js' {
  interface SignalKinds<T> {
    foo: PSignal<T>;
    bar: PSignal<T>;
  }
}
```

The type system won't flag the SignalKinds definition as invalid; instead this will fail silently and upon using the signals all the types will be `unknown` (this is bad). (Imagine this with a less obvious typo!).

ditman (commented):
Unfortunately, there's a usability snag in this solution, so we're going with @ava-cassiopeia's!

PR #1054, opened 2026-04-02
build: migrate monorepo to Yarn v4 workspaces

gspencergoog:
## Description

This pull request migrates the monorepo from independent npm packages to a centralized Yarn Modern (v4.13.0) workspace configuration. It also integrates Wireit for task caching and execution across workspaces, standardizes linting configurations, and updates CI/CD workflows to run Yarn commands.

### Key changes

* Unified workspace packages under a single root-level configuration and yarn.lock file. Removed all individual package-lock.json and pnpm-lock.yaml files.
* Configured dependency resolution to use workspace protocol (workspace:*) for internal references and enforced local workspace resolution for third-party packages and tools.
* Integrated Wireit to manage cleaning, building, and testing tasks, with defined task dependencies for topological execution.
* Enabled Corepack to manage Yarn version activation across local development and Docker containers, removing checked-in standalone Yarn binary bundles.
* Configured canonical private GCP publishing via .yarnrc.yml with authentication token injection.
* Updated repository documentation, quickstart guides, live demo runners, and Jupyter notebooks to replace legacy npm and pnpm instructions with modern Yarn workspace equivalents.
* Pinned Angular packages to version 21.2.5 across all workspaces.
* Added eslint.config.mjs configuration files to workspaces and added root-level lint commands.
* Updated GitHub Actions workflows to support Corepack and Yarn v4 commands.
* Corrected specification copying tasks to track generated v1.0 schema assets in the web_core renderer and reverted accidental changes to protocol specification files.


renderers/angular/.npmrc
```diff
@@ -1,2 +0,0 @@
-@a2ui:registry=https://us-npm.pkg.dev/oss-exit-gate-prod/a2ui--npm/
-//us-npm.pkg.dev/oss-exit-gate-prod/a2ui--npm/:always-auth=true
```
ditman, line 2: This is needed to be able to publish the package through go/oss-exit-gate-release-npm. What is the equivalent in yarn?

renderers/angular/package-lock.json
ditman, line 1: Lock files seem to be also required by go/oss-exit-gate-release-npm, are we ignoring the yarn locks?

.yarnrc.yml
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.13.0.cjs
```
ditman: This is odd, are we shrink-wrapping the whole yarn executable? Should this be a devDependency somewhere?

ditman (commented):
Removing the `.npmrc`s will be a problem to release the packages using the Exit Gate, it seems? How does publishing through yarn work?

renderers/angular/package-lock.json
ditman: There's a global yarn.lock, github didn't render the huge diff!

renderers/angular/.npmrc
ditman: Maybe related? https://github.com/yarnpkg/berry/issues/6191

gspencergoog: Okay, I *think* I updated the `.yarnrc.yml` to configure the registry this way.  I don't think we'll know if it works until we try to publish.

renderers/angular/package-lock.json
gspencergoog: Yes, I was hoping we didn't need them. I'll rethink that and see.

gspencergoog: I didn't see any mention of lock files in the docs.  I don't think we need to include them.  But no, we aren't ignoring yarn locks, but it's set up as a monorepo, so there's only one.

.yarnrc.yml
gspencergoog: Turns out this is redundant with packageManager, so I removed it and the `.yarn/releases` directory.

gspencergoog: Oh.  Actually not.  It means that you have to enable `corepack` on your machine (every dev would have to).  I went back to keeping the yarn executable.

The reason you don't put it in dev dependencies is the chicken-and-egg problem: You need a package manager to install your devDependencies. If Yarn is a devDependency, you would need Yarn to install Yarn.

gspencergoog:
I think I've resolved all the issues.  I updated the publishing doc for yarn-based instructions but crucially, I *haven't tried them*.  They look reasonable, but we would need to vet them at the next publish time to make sure they work.

renderers/angular/.npmrc
gspencergoog: Turns out we can only update the registry when we actually publish.  If I put it into the `.yarnrc.yml`, then CI will fail because it doesn't have a token.

ditman, line 2: This is quite critical for the publishing flow of the packages, where did this go?

ditman (changes requested):
This looks very cool, I like the simplifications to the CI scripts, however; have you tried following the publishing script? Where did the exit gate configuration go? I can't "find in page" in a PR so huge as this :)

renderers/angular/.npmrc
gspencergoog: Oh!  You're right.  I've restored those, and modified things so that the yarn build will use regular `npm publish` when `package:publish` is run.  I ran dry runs, and they all worked, but of course we still need to actually try it with a real publish before we'll know if it actually works.

gspencergoog: Sorry, you did mention that before.  I got distracted with the other changes.

ditman: If we use npm to publish, but yarn to manage the rest, may we end up in this state?

https://stackoverflow.com/a/55990047

I wonder if the private registry configuration can be applied at the yarn level so we can use[`yarn npm publish`](https://yarnpkg.com/cli/npm/publish)? Maybe through the [`npmPublishRegistry`](https://yarnpkg.com/configuration/yarnrc#npmPublishRegistry) with `npmAlwaysAuth`?

There's also the question of the `<auth-token>` that the `google-artifactregistry-auth` initializes ([docs](https://github.com/a2ui-project/a2ui/blob/main/renderers/docs/web_publishing.md#pre-requirement-artifact-registry-configuration)). Maybe this:
* https://github.com/GoogleCloudPlatform/artifact-registry-npm-tools/issues/16

gspencergoog: Okay, I've done some more investigating using the links you gave (thank you!), and some more searching around, and I think I've got what we need. It eliminates the `.npmrc` files again, and replaces the `<auth-token>` with a call to `gcloud auth print-access-token` right before we publish, and switches to just use `yarn npm publish`. I also updated the publishing scripts so that they will rewrite the `workspace:` protocol when publishing (along with the `file:` protocol), so that the dist directory is correct when we publish.

I was able to run the dry run, but I still won't know if it's all correct until we actually go to publish, at which point either it will work and we'll be done, or we'll fix it.

ditman: Thanks! Looking!

renderers/docs/web_publishing.md
```diff
@@ -104,24 +103,14 @@ export NPM_TOKEN="npm_YourSecretTokenHere"
...
 
-1. Ensure your local `.npmrc` in the package directory is correctly configured if you are debugging, but the automated scripts handle authentication via `google-artifactregistry-auth`.
-2. If you need to manually overwrite or create an `.npmrc` for local testing:
-   ```sh
-   echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > .npmrc
-   ```
-
 ## About the `publish:package` command
 
 Because these are scoped packages (`@a2ui/`), they require the `--access public` flag to be published to the public registry. The `publish:package` script handles this automatically, as well as replacing the path dependencies with package dependencies.
```
ditman: > as well as replacing the path dependencies with package dependencies.

With yarn and workspaces this is not true anymore, or is it?

renderers/react/package.json
```diff
@@ -75,7 +75,7 @@
   ],
   "scripts": {
     "prepublishOnly": "node -e \"if(!process.cwd().endsWith('dist')) { console.error('Error: This package must be published from the dist/ directory. Run `yarn build` then `cd dist && yarn npm publish`'); process.exit(1); }\"",
-    "publish:package": "yarn build && node ../scripts/prepare-publish.mjs && cd dist && yarn npm publish --access public",
+    "publish:package": "yarn build && node ../scripts/prepare-publish.mjs && cd dist && npm publish --ignore-scripts --access public",
```
ditman: `npm publish`? Should this be `yarn publish`?

renderers/scripts/prepare-publish.mjs
```diff
@@ -112,13 +112,16 @@ delete pkg.devDependencies;
...
 if (existsSync(join(packageDir, 'README.md'))) {
   copyFileSync(join(packageDir, 'README.md'), join(resolvedDistDir, 'README.md'));
 }
 const rootLicense = join(rootDir, 'LICENSE');
 if (existsSync(rootLicense)) {
   copyFileSync(rootLicense, join(resolvedDistDir, 'LICENSE'));
 }
+if (existsSync(join(packageDir, '.npmrc'))) {
+  copyFileSync(join(packageDir, '.npmrc'), join(resolvedDistDir, '.npmrc'));
+}
```
ditman: Do we have any `.npmrc` files anymore? Are they being generated or something?

renderers/scripts/publish_npm.mjs
```diff
@@ -310,16 +310,30 @@ export async function main(args, mocks = {}) {
...
-  maybeRunCommand('npx', ['google-artifactregistry-auth'], {}, {dryRun, runCommand: runCmd});
 
   for (const pkg of packageObjects) {
     console.log(`\n=== Publishing ${pkg.name} (${pkg.version}) ===`);
+
+    let npmToken = process.env.NPM_TOKEN || '';
+    if (!dryRun && !npmToken) {
+      console.log(`- Obtaining fresh Artifact Registry token via gcloud CLI for ${pkg.name}...`);
+      try {
+        npmToken = exec('gcloud auth print-access-token', {encoding: 'utf8'}).trim();
```
ditman, line 350: This is similar to the fallback mechanism here: 

https://github.com/GoogleCloudPlatform/artifact-registry-npm-tools/blob/9f8b4ad6b6f1911b7e103432e550e9deeb3073cc/src/auth.js#L65-L71

.yarnrc.yml
```diff
@@ -15,3 +15,10 @@
 nodeLinker: node-modules
 
 yarnPath: .yarn/releases/yarn-4.13.0.cjs
+
+npmScopes:
+  a2ui:
+    npmRegistryServer: 'https://us-npm.pkg.dev/oss-exit-gate-prod/a2ui--npm/'
+    npmPublishRegistry: 'https://us-npm.pkg.dev/oss-exit-gate-prod/a2ui--npm/'
+    npmAlwaysAuth: true
+    npmAuthToken: '${NPM_TOKEN:-}'
```
ditman: There seems to be support for the yarnrc file here:

* https://github.com/GoogleCloudPlatform/artifact-registry-npm-tools/pull/78/changes

In the [README.md](https://github.com/GoogleCloudPlatform/artifact-registry-npm-tools/blob/master/README.md) it says:

> `--repo-config-yarn` is the .yarnrc.yml file with your repository settings. If you don't specify this flag, the default location is the current directory.
> `--credential-config-yarn` is the path to the .yarnrc.yml file where you want to write the access token. The default is your user .yarnrc.yml file.

Maybe we can use this to not need our own credential management in the publish script?

specification/v0_10/catalogs/basic/catalog.json
ditman, line 1: Please, revert all the v0_10 changes from this PR, I think those are intended for a separate PR :)

ditman (changes requested):
I found some promising params on the googleauth npx script that may help with the yarn integration, instead of having to roll our own bits on the script (that way we don't own the Yarn integration of that bit)

PS: Please, remove v0_10 changes, I think those are meant for a different PR?

specification/v0_10/catalogs/basic/catalog.json
gspencergoog: Ack!  How did those get there?!?  Sure.

renderers/scripts/prepare-publish.mjs
gspencergoog: No, they don't exist.  You're right, I'll fix this part of the script.

renderers/react/package.json
gspencergoog: Aaaah!  Yes.  It's hard to find all the pieces...

.yarnrc.yml
gspencergoog: Oh, maybe the docs I found were out of date.  What I read said that it wasn't implemented yet.

specification/v0_10/catalogs/basic/catalog.json
gspencergoog: Wait, I don't see any changes in this file (or any unrelated v0_10 changes)...  Are you sure you were looking at the right commit?

ditman: I used "show me changes since my previous review"; it must have been showing me parts of the merge commit with main :/ odd

renderers/docs/web_publishing.md
gspencergoog: Actually, it still does that, for both `file:` and `workspace:` dependencies.

renderers/react/package.json
gspencergoog: I did a search over the whole codebase and fixed all the places that had incorrect `npm` usage.  I think I got them all (there were actually a lot more than I thought).

.yarnrc.yml
gspencergoog: I looked, and using `google-artifactregistry-auth` mutates your `.yarnrc.yml` file on disk to save the credential, which leaves your working tree dirty or requires cleanup logic after every publish.

Using the inline definition of NPM_TOKEN doesn't modify any files on disk, so I think I'd rather keep that method.

renderers/scripts/publish_npm.mjs
gspencergoog: Yes, very similar.

renderers/scripts/prepare-publish.mjs
gspencergoog: I also added copying the CHANGELOG.md too, so it can display it on the NPM site.

specification/v0_10/catalogs/basic/catalog.json
gspencergoog: Yeah, once I merged with main, a bunch of things showed up... That was weird.  Anyhow, reverted them.

docs/guides/a2ui-in-mcp-apps.md
```diff
@@ -166,7 +166,7 @@ MCP Apps are typically delivered as a single HTML resource from the MCP Server.
 > 1. **Install the plugin**:
 >
 >     ```bash
->     npm install -D vite-plugin-singlefile
+>     yarn add -D vite-plugin-singlefile
```
ditman: Unclear that this change was needed, this is for an end-user setup script, not our repo's I think?

gspencergoog:
Okay, I was able to run the publish scripts and push a release to the staging server using Yarn.

.yarn/releases/yarn-4.13.0.cjs
ditman, line 1: Please, remove this .yarn directory, let's fully embrace `corepack` now that we're at it, and don't redistribute the "binary" of the package manager alongside our code.

.yarnrc.yml
ditman: I say we go full-blown `corepack`, and update the setup instructions for the package. `corepack` seems to come built-in with node (at least I had it without much fuss) and installing yarn was kind of simple; I only needed to:

> 1. Ensure corepack is available (it should be if you have "node")
> ```
> $ corepack --version
> 0.24.0
> ```
> 
> 2. Enable corepack (depending on your system you may or may not need `sudo`, or you may have already done this before):
> ```
> $ corepack enable
> ```
> 
> 3. In the `a2ui` checkout, check that `yarn` is now available at the version specified in `package.json`:
> ```
> $ which yarn
> /usr/bin/yarn
> $ yarn --version
> 4.13.0
> ```

ditman (commented):
I have also managed to upload a test version of all the `renderer` packages using `yarn` from `corepack` (without the `.yarn/releases` shrink-wrap).

I think this is very cool, and that we should just go ahead and remove the shrink-wrapped version of `.yarn` that we're vending in this PR.

(In my local checkout I removed `yarnPath` from the `.yarnrc.yml` file, and deleted the whole `.yarn` directory, I left setup instructions below on what I did to have `yarn` from `corepack`)

docs/guides/a2ui-in-mcp-apps.md
gspencergoog: Yeah, good point.

.yarnrc.yml
gspencergoog: Yeah, agreed. It means some additional steps for macOS install, but it's not bad.  Done!

.yarn/releases/yarn-4.13.0.cjs
gspencergoog: Removed.

gspencergoog:
Okay, I've removed the pinned yarn package and rely entirely on `corepack` to supply yarn, which seems to work.  Updated CI and the user instructions.

.github/workflows/composer_build_and_test.yml
```diff
@@ -35,6 +35,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Enable Corepack
+        run: corepack enable
```
ditman, line 39: Should Enable Corepack happen *after* setting up Node? :P

(If this works, I wonder if the whole "Set up Node.js" step below is still needed?)

ditman (approved):
It's hard to argue with -125kloc! Small musing about being able to initialize corepack before installing node in the GH workflows (is the "install node" workflow even needed? or is that just to pin the nodejs version? Should we run corepack enable *after* the node setup?)

.github/workflows/composer_build_and_test.yml
ditman: OK, I dug a little bit deep, and in the included software of the ubuntu image we're using, Node22 is preinstalled:

* https://github.com/actions/runner-images/blob/ubuntu24/20260525.161/images/ubuntu/Ubuntu2404-Readme.md#language-and-runtime

I think it's good we're "pinning" the version of node to the specific one we want (if that even works), but we definitely should enable corepack after installing node maybe? (Or maybe we can just remove the node install step from the recipes)

gspencergoog: Hmm. I'll try enabling `corepack` after the install of node.  That seems more likely to give us what we want, while `corepack` is still part of the node install. I saw something about the node developers wanting to decouple `corepack` from node, but we can cross that bridge when they do that.

gspencergoog: Well, it turns out that we can't run `corepack enable` after the node install if we want to do caching of the yarn config.  Because the `cache: yarn` statement is there, it tries to run yarn before `corepack enable` is run, so it fails to find `yarn` because the shims aren't there.

I have reverted to running `corepack enable` before the node install, because, it will already have *some* `node` installed by the ubuntu image, and because the shims that `corepack` creates use a shebang that will resolve in the `yarn` that is installed with the updated node install, so it should always run the right one anyhow.  If they remove node from the ubuntu image (unlikely), or `corepack` starts doing something weird with the shims, we will have to reassess.

PR #1055, opened 2026-04-02
Make the client_to_server_list and server_to_client_list a dict, since the places we plan to use this DataPart.data, MCP response payload, seem to be built around dicts and don't support lists.

wrenj:
Make the client_to_server_list and server_to_client_list a dict, since the places we plan to use this DataPart.data, MCP response payload, seem to be built around dicts and don't support lists.


wrenj:
resolved comments, ptal

PR #1064, opened 2026-04-03
fix(security): remove allow-same-origin from MCP Apps sandbox iframes

zeroasterisk:
## Summary
This PR updates the iframe sandbox settings for MCP Apps to align with the **MCP Apps specification** and the **A2UI security guide**, addressing concerns about over-restriction while ensuring proper isolation. Found during review of PR #1062.

### 1. Outer Iframe (Sandbox Proxy)
*   **Change**: Added `sandbox="allow-scripts allow-same-origin"` to the outer iframe in `mcp-apps-component.ts`.
*   **Rationale**: The **MCP Apps specification** (Section: *Sandbox proxy*) explicitly mandates that if the host is a web page, the Sandbox Proxy MUST have `allow-scripts` and `allow-same-origin` permissions. The previous draft removed the sandbox attribute entirely based on the A2UI guide's advice to not sandbox the proxy, but this update restores it to strictly comply with the spec.

### 2. Inner Iframe (View)
*   **Change**: Removed `allow-same-origin` from the inner iframe's sandbox settings in both `mcp-apps-component.ts` and `sandbox.ts`.
*   **Change**: Retained (and added where missing) `allow-forms`, `allow-popups`, and `allow-modals` flags.
*   **Rationale**: Removing `allow-same-origin` ensures proper isolation of untrusted content (preventing it from accessing the parent origin). Retaining the other functional flags ensures that apps can still handle forms and popups, avoiding breaking basic functionality as flagged by reviewers. This fits the spec's requirement for "restricted permissions" and follows the A2UI guide's allowed list.

## Verification Results
*   The changes ensure that the outer iframe complies with the MCP spec.
*   The inner iframe is securely isolated without `allow-same-origin` but retains necessary UI capabilities.


samples/client/shared/mcp_apps_inner_iframe/sandbox.ts
```diff
@@ -63,7 +63,7 @@ if (!disableSelfTest) {
 // iframe on a separate origin. It creates an inner iframe for untrusted HTML content.
 const inner = document.createElement("iframe");
 inner.style.cssText = "width:100%; height:100%; border:none;";
-inner.setAttribute("sandbox", "allow-scripts allow-same-origin allow-forms");
+inner.setAttribute("sandbox", "allow-scripts allow-forms allow-popups allow-modals");
```
ditman, line 66: This line is adding "allow-popups" and "allow-modals" without it being documented in the change. Is this intended?

(This is making the allowlist the same as the custom-components-example above so I'm guessing "yes".)

ditman (approved):
If the demos continue working fine, let's do it!

(Have you tried this with a deployed version of the app? localhost is often very lax for all these sandboxing options)

samples/client/shared/mcp_apps_inner_iframe/sandbox.ts
zeroasterisk: It was intentional, but I cannot guarantee those intentions :)

PR #1074, opened 2026-04-06
support message list

wrenj:
add interfaces for schema change

https://github.com/google/A2UI/pull/1055

### DIFF
--- a/renderers/web_core/src/v0_9/schema/client-to-server.ts
+++ b/renderers/web_core/src/v0_9/schema/client-to-server.ts
@@ -137,6 +137,6 @@ export const A2uiClientMessageListWrapperSchema = z
     messages: A2uiClientMessageListSchema,
   })
   .strict()
-  .describe('An object wrapping a list of A2UI Client-to-Server messages.');
+  .describe('An object wrapping a list of client messages.');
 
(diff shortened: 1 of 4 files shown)


renderers/web_core/src/v0_9/schema/client-to-server.ts
```diff
@@ -125,3 +125,15 @@ export type A2uiClientAction = z.infer<typeof A2uiClientActionSchema>;
...
 export type A2uiClientDataModel = z.infer<typeof A2uiClientDataModelSchema>;
+
+export const A2uiClientMessageListSchema = z
+  .object({
+    version: z.literal('v0.9'),
+    messages: z.array(A2uiClientMessageSchema).describe('A list of client messages.'),
+  })
+  .strict();
+
+export declare interface A2uiClientMessageList extends z.infer<typeof A2uiClientMessageListSchema> {
```
jacobsimionato: This declaration doesn't match the rest of the file - can we just do:
```
export type A2uiClientMessageList =  z.infer<typeof A2uiClientMessageListSchema>;
```
?

I think may be we don't need the declaration for client-to-server because we don't validate?

jacobsimionato (approved):
Approval pending resolving one nit!

PR #1093, opened 2026-04-08
feat: enable streaming in the orchestrator agent

nan-yu:
# Description

Demo: 

[Streaming support in the orchestrator sample.webm](https://github.com/user-attachments/assets/fe89e9a1-4110-479b-8044-13d2cc587df0)

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/samples/client/angular/projects/orchestrator/src/services/a2a-service-impl.ts
+++ b/samples/client/angular/projects/orchestrator/src/services/a2a-service-impl.ts
@@ -16,27 +16,109 @@
 
 import { AgentCard, Part, SendMessageSuccessResponse } from '@a2a-js/sdk';
 import { A2aService } from '@a2a_chat_canvas/interfaces/a2a-service';
-import { Injectable } from '@angular/core';
+import { Injectable, inject, Injector } from '@angular/core';
+import { MessageProcessor } from '@a2ui/angular';
 
 @Injectable({
   providedIn: 'root',
 })
 export class A2aServiceImpl implements A2aService {
 
+  private readonly messageProcessor = inject(MessageProcessor);
+  private readonly injector = inject(Injector);
+
   async sendMessage(parts: Part[], signal?: AbortSignal): Promise<SendMessageSuccessResponse> {
     const response = await fetch('/a2a', {
-      body: JSON.stringify({ parts: parts }),
+      body: JSON.stringify({
+        parts: parts,
+        metadata: {
+          a2uiClientCapabilities: {
+            supportedCatalogIds: ['https://a2ui.org/specification/v0_8/standard_catalog_definition.json']
+          },
+          streaming: true
+        }
+      }),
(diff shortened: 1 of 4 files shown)


samples/client/angular/projects/orchestrator/src/services/a2a-service-impl.ts
```diff
@@ -16,27 +16,112 @@
...
-    if (response.ok) {
-      const data = await response.json();
-      return data;
+    if (!response.ok) {
+      const error = (await response.json()) as { error: string };
+      throw new Error(error.error);
+    }
+
+    const contentType = response.headers.get('content-type');
+    if (contentType?.includes('text/event-stream')) {
```
jacobsimionato, line 51: I'm surprised that we need this HTTP-level logic here to handle this. Shouldn't the A2A client libraries help us out with this low level stuff, and emit `Part` objects that we can use? Can't we do `client.sendMessageStream` the same as the orchestrator server does?

nan-yu: The current implementation in `a2a-service-impl.ts` uses manual HTTP-level logic to handle the stream because the Angular client communicates with a local server proxy rather than talking to the agent directly.

Per jetski, we do this to avoid CORS issues (the agent runs on a different port and doesn't allow the Angular origin) and to simplify the stream processing. The server uses `A2AClient.sendMessageStream` to talk to the agent, extracts the Part objects, and streams them as raw JSON to the client. Because it's a custom simplified stream of parts rather than full A2A protocol events, we can't use the A2AClient library on the client to parse it.

If we want to use `A2AClient.sendMEssageStream` directly in the client, I guess we can do via a separate refactor PR, including the following changes:
- Update the agent's CORS settings to allow "http://localhost:4000"
- Import A2AClient from @a2a-js/sdk/client and use it similarly to how the server does.
- Replace the `fetch('/a2a')` call and the manula stream reader with the `A2AClient` from the SDK.

jacobsimionato: Oh I see, this makes sense. I guess I was wondering if there is a path to instead have the Angular proxy server also act as an A2A server and the client act as an A2A client, so it's like:

agent <--- A2A transport ----> Angular proxy server <--- A2A transport ---> Angular client

Ideally, that would reduce the boilerplate etc, though I'm not 100% convince that A2A is actually a great transport mechanism to plumb all the way to the client. I think the architecture you described makes sense, but we need to significantly reduce the amount of boilerplate developers are expected to write to achieve this. I guess we need to understand the common architectures / frameworks etc and figure out an approach that works well, probably solving for next.js etc first.

Anyway, this is outside the scope of this PR, so let's leave it as-is. Thanks for explaining the situation!

PR #1129, opened 2026-04-10
Integrate Kotlin SDK with A2UI conformance tests

jiahaog:
This commit groups the changes made to Validator.kt into 4 themes, each necessitated by specific failing tests in the conformance suite:

1. Heuristic Component Reference Resolution
- Changes: Added hardcoded field names ('child', 'children', 'explicitList') to extractComponentRefFields to identify references when not explicitly declared in schema.
- Necessity: Complex schemas (like in v0.9) didn't explicitly mark these as references, but Python validator assumed them.
- Failing Test: test_custom_catalog_0_9 (failed with "Component 'c1' is not reachable from 'root'")

2. Multi-Surface Validation Support
- Changes: Refactored validate to instantiate a new A2uiTopologyValidator for each message batch based on surfaceId, and updated findRootId to look up rootId per surface.
- Necessity: Conformance tests contain batches with messages for multiple surfaces, each needing independent root validation.
- Failing Test: test_validate_multi_surface_v08 (failed with "Missing root component: No component has id='root-a'")

3. Incremental Update Support (Nullable Root)
- Changes: Allowed findRootId to return null when no root-defining message is found, and updated topology validator to skip root check and traverse all nodes for cycles.
- Necessity: Incremental updates might not contain the root component in the current batch.
- Failing Test: test_incremental_update_no_root_v08 (failed with "Missing root component: No component has id='root'")

4. v0.9 createSurface Support
- Changes: Added recognition of createSurface messages in findRootId and defaulted root ID to 'root' for v0.9.
- Necessity: In v0.9, root is assumed to be 'root' when createSurface is used.
- Failing Test: test_validate_missing_root_v09 (failed because it expected failure but passed due to fallback to null root)

Also fixed findRepoRoot in build.gradle.kts to look for 'specification' directory instead of '.git', fixing resource loading in A2uiSchemaManagerTest.

# Description

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md


jiahaog:
@jgindin @nan-yu

agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt
```diff
@@ -176,13 +187,47 @@ class A2uiValidator(private val catalog: A2uiCatalog) {
     }
 
     // Integrity validation
-    val rootId = findRootId(messages)
-    val topologyValidator = A2uiTopologyValidator(catalog, rootId)
     val recursionValidator = A2uiRecursionValidator()
 
+    // Pre-calculate root IDs for each surface to avoid O(N^2) in the loop
```
jgindin: I'd suggest pulling this out to a separate method.

agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt
```diff
@@ -176,13 +187,47 @@ class A2uiValidator(private val catalog: A2uiCatalog) {
...
+            else -> ROOT
+          }
+          surfaceRootIds[msgSurfaceId] = rootId
+        }
+      }
+      
+      if (MSG_CREATE_SURFACE in message) {
+        val createSurface = message[MSG_CREATE_SURFACE] as? JsonObject
+        val msgSurfaceId = createSurface?.get("surfaceId")?.jsonPrimitive?.content
+        if (!surfaceRootIds.containsKey(msgSurfaceId)) {
```
jgindin: I believe we could just do:
```
surfaceRootIds.putIfAbsent(msgSurfaceId, ROOT)
```

agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt
```diff
@@ -176,13 +187,47 @@ class A2uiValidator(private val catalog: A2uiCatalog) {
     }
 
     // Integrity validation
-    val rootId = findRootId(messages)
-    val topologyValidator = A2uiTopologyValidator(catalog, rootId)
     val recursionValidator = A2uiRecursionValidator()
```
jgindin: This is my fault, but this declaration could be moved down to the new line 249, which is the only place this is used.

agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt
```diff
@@ -334,7 +373,7 @@ class A2uiValidator(private val catalog: A2uiCatalog) {
         }
       }
 
-      if (rootId !in ids) {
+      if (rootId != null && rootId !in ids) {
```
jgindin, line 383: I'm wondering if this should be `||` instead of `&&`?

Given that we seem to be looking for a root component, `rootId == null` would seem to indicate that there's no root, just the same as `rootId !in ids`

agent_sdks/kotlin/src/test/kotlin/com/google/a2ui/core/schema/ConformanceTest.kt
```diff
@@ -0,0 +1,196 @@
...
+import java.io.File
+import kotlin.test.assertTrue
+import kotlin.test.assertFailsWith
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import org.junit.jupiter.api.DynamicTest
+import org.junit.jupiter.api.TestFactory
+
+class ConformanceTest {
```
jgindin, line 30: Can you run `./gradlew ktfmtFormat` to get this file formatted?

agent_sdks/kotlin/src/test/kotlin/com/google/a2ui/core/schema/ConformanceTest.kt
```diff
@@ -0,0 +1,196 @@
...
+
+    private val yamlMapper = ObjectMapper(YAMLFactory())
+    private val jsonMapper = ObjectMapper()
+
+    private fun findRepoRoot(): File {
```
jgindin: I think it would make sense to move this to a companion object:
```
private companion object {
  private val REPO_ROOT = findRepoRoot()

  private fun findRepoRoot(): File {
   . . . 
  }
}
```

agent_sdks/kotlin/src/test/kotlin/com/google/a2ui/core/schema/ConformanceTest.kt
```diff
@@ -0,0 +1,196 @@
...
+            ConformanceTestCase(name, catalog, validate, schemaMappings)
+        }
+    }
+
+    private fun buildCatalog(
+        catalogMap: Map<*, *>,
+        conformanceDir: File,
+        baseSchemaMappings: Map<String, String>
+    ): Pair<A2uiCatalog, Map<String, String>> {
+        val versionStr = catalogMap["version"] as String
```
jgindin: I suggest creating constants for the different strings used throughout here and adding them to the companion object.

This gives a good place to comment on what they all represent.

agent_sdks/kotlin/src/test/kotlin/com/google/a2ui/core/schema/ConformanceTest.kt
```diff
@@ -0,0 +1,196 @@
...
+        val version = if (versionStr == "0.8") A2uiVersion.VERSION_0_8 else A2uiVersion.VERSION_0_9
+
+        val s2cSchemaFile = catalogMap["s2c_schema"] as? String
+        val s2cSchema = if (s2cSchemaFile != null) {
```
jgindin: Rather than if/else, more idiomatic would be:
```
val s2cSchema = s2cSchemaFile?.let { 
    loadJsonFile(File(conformanceDir, it)) 
} ?: JsonObject(emptyMap())
````

File: agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt (hunk 4 above)
Comment (nan-yu) [reply agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt:383]: It should be `&&` to support incremental updates. `rootId` could be none in the case of an incremental update. In such updates, the components list might only contain a subset of components being updated, and the root component (which was sent in a previous message) may not be present in this batch.

If rootId is none, the check is skipped to allow incremental updates to pass validation.

**Review by nan-yu — APPROVED** [review]
/lgtm

File: agent_sdks/kotlin/src/test/kotlin/com/google/a2ui/core/schema/ConformanceTest.kt (hunk 7 above)
Comment (jiahaog) [reply agent_sdks/kotlin/src/test/kotlin/com/google/a2ui/core/schema/ConformanceTest.kt]: Done! The names are probably self-explanatory so I didn't leave more comments.

File: agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt (hunk 4 above)
Comment (jiahaog) [reply agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt:383]: Thanks Nan, leaving as is then :)

File: agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt [hunk 9]
```diff
@@ -176,13 +185,24 @@ class A2uiValidator(private val catalog: A2uiCatalog) {
     }
 
     // Integrity validation
-    val rootId = findRootId(messages)
-    val topologyValidator = A2uiTopologyValidator(catalog, rootId)
-    val recursionValidator = A2uiRecursionValidator()
+    // Integrity validation
```
jgindin, line 187: Nit -- this line got duplicated :-)

agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt
```diff
@@ -195,27 +215,50 @@ class A2uiValidator(private val catalog: A2uiCatalog) {
...
+        topologyValidator.validate(it)
+      }
 
+      val recursionValidator = A2uiRecursionValidator()
       recursionValidator.validate(message)
     }
   }
 
-  private fun findRootId(messages: JsonArray): String {
+  private fun calculateSurfaceRootIds(messages: JsonArray): Map<String?, String?> {
```
jgindin: Is the return type correct -- a map of a possibly null string to a possibly null string?

To have a null key for the map, I believe that would mean that `msgSurfaceId` is null.

agent_sdks/kotlin/src/test/kotlin/com/google/a2ui/core/schema/ConformanceTest.kt
```diff
@@ -0,0 +1,203 @@
...
+        loadJsonFile(File(conformanceDir, catalogSchemaObj))
+      } else if (catalogSchemaObj is Map<*, *>) {
+        val jsonStr = jsonMapper.writeValueAsString(catalogSchemaObj)
+
+        val tempFile = java.io.File.createTempFile("custom_catalog", ".json")
+        tempFile.deleteOnExit()
+        tempFile.writeText(jsonStr)
+        schemaMappings["https://a2ui.org/specification/v0_9/simplified_catalog_v09.json"] =
+          tempFile.toURI().toString()
+        schemaMappings["simplified_catalog_v09.json"] = tempFile.toURI().toString()
```
jgindin: Might be a nit, but we're repeating `simplified_catalog_v09.json` here and above on line 103. If it were to be renamed, it'd be easier to catch and fix if this was a constant (in the companion object) that was reused.

agent_sdks/kotlin/src/test/kotlin/com/google/a2ui/core/schema/ConformanceTest.kt
```diff
@@ -0,0 +1,203 @@
...
+
+        Json.parseToJsonElement(jsonStr) as JsonObject
+      } else {
+        Json.parseToJsonElement("""{"catalogId": "test_catalog", "components": {}}""") as JsonObject
```
jgindin: I'm not sure if we intend for this case to just be an empty catalog, but if so I think it'd be better to make it explicit in the conformance test file itself. That way there's less of a chance for individual conformance test implementations to accidentally diverge.

@nan-yu ?

agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt
jiahaog: Thanks for catching it!

agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Validator.kt
jiahaog: Agreed, I made this non-null. Seems like the requirement in the spec is for the keys to be required anyway. Thanks!

agent_sdks/kotlin/src/test/kotlin/com/google/a2ui/core/schema/ConformanceTest.kt
jiahaog: Agreed as well. It's simpler to just fail here. I also updated the python test to match the behavior here.

jgindin (approved):

PR #1134, opened 2026-04-10
feat(composer): production-quality A2UI Composer with AI authoring

jerelvelarde:
## Summary

Production-quality upgrade of the A2UI Composer tool (building on #987). Replaces the original prototype with a fully-featured widget editor powered by CopilotKit and the A2UI v0.9 spec.

- **Widget Editor** — Split-pane editor with Monaco code editing, live A2UI preview, and a data panel for inspecting/modifying the data model
- **AI Authoring** — Describe a widget in natural language; the AI generates valid A2UI v0.9 JSON using the full component vocabulary (Card, Row, Column, Text, Image, Button, List, Tabs, Modal, etc.)
- **Gallery** — Browse, search, and manage saved widgets with IndexedDB persistence (localforage)
- **Theater** — Replay recorded A2UI streaming scenarios chunk-by-chunk with JSONL wire inspection, lifecycle events, and variable playback speed
- **Icon Browser** — Search and preview Material Symbols icons available to the `Icon` component
- **Reactive Rendering** — `A2uiSurface` uses `useSyncExternalStore` for granular re-renders via the `SurfaceModel`
- **Catalog-Driven AI** — The built-in A2UI v0.9 basic catalog shapes both the renderer and the LLM system prompt

### Key Architecture

- **CopilotKit v2** integration for AI chat runtime (`/api/copilotkit` route)
- **Next.js 16** App Router with Turbopack, **React 19**
- **`@a2ui/web_core` v0.9** — `SurfaceModel`, `ComponentContext`, `MessageProcessor`, `Catalog`
- Client-side persistence via IndexedDB (no server-side database required)

### Commits

| Commit | Description |
|--------|-------------|
| `696b945` | Replace PR#987 app with widget-builder from cpk-deployed-composer |
| `08a12fb` | Add Apache 2.0 license headers to all source files |
| `82d2e65` | Add README, AGENTS.md, and CHANGELOG entry |
| `c25b76f` | Remove custom catalog feature, fix CopilotChat type errors |
| `3112e9b` | Fix theater JSON syntax colors for light mode, fix card clipping |
| `9ca87f2` | Restore test infrastructure (31 tests) and fix Text prop bug |
| `b9723d0` | Remove width:100% from Card to prevent right-edge clipping |
| `f45ac2f` | Add user-facing error notifications for storage failures |

## Test plan

- [ ] `pnpm install && pnpm build` — clean build with zero errors
- [ ] `pnpm test` — 31 tests pass
[...]

PR #1144, opened 2026-04-10
feat(composer): v0.9 theater, gallery fixes, and renderer improvements

jerelvelarde:
## Summary

- **Theater**: Update scenarios to v0.9 format, add JSON Pretty tab with syntax coloring, drop v0.8 support
- **Gallery**: Fix card overflow/clipping across multiple widgets (financial data grid, recipe card, task card), add full-width containment wrapper for previews
- **Renderer**: Fix `<caption>` hydration error, implement weight property for flex layouts, convert camelCase icon names to snake_case for Material Symbols, add inline markdown rendering for Text, add `overflow: hidden` to Card, make Tabs buttons flexible
- **Composer**: Default to v0.9 across gallery/components/widget creation, replace favicon, rename sidebar nav entry, set v0.9 primary color, add error notifications for storage failures

https://github.com/user-attachments/assets/7c78be8d-92b3-409c-87d3-290e0fcfbd17

image

image

Supersedes #1134.

## Test plan

- [ ] Verify `/gallery` page renders all cards without overflow or clipping
- [ ] Verify `/theater` page plays v0.9 scenarios with JSON Pretty tab
- [ ] Verify `/components` page lists v0.9 components
- [ ] Verify icons render correctly (snake_case conversion)
- [ ] Verify markdown text renders in gallery cards
- [ ] Verify no console hydration errors
- [ ] Verify new widget creation defaults to v0.9

🤖 Generated with [Claude Code](https://claude.com/claude-code)


renderers/react/src/v0_9/catalog/basic/components/Column.tsx
```diff
@@ -28,6 +28,7 @@ export const Column = createComponentImplementation(ColumnApi, ({props, buildChi
         justifyContent: mapJustify(props.justify),
         alignItems: mapAlign(props.align),
         width: '100%',
+        minWidth: 0,
```
ditman, line 32: Isn't this the default?

renderers/react/src/v0_9/catalog/basic/components/Icon.tsx
```diff
@@ -20,8 +20,9 @@ import {IconApi} from '@a2ui/web_core/v0_9/basic_catalog';
 import {getBaseLeafStyle} from '../utils';
 
 export const Icon = createComponentImplementation(IconApi, ({props}) => {
-  const iconName =
+  const rawName =
     typeof props.name === 'string' ? props.name : (props.name as {path?: string})?.path;
+  const iconName = rawName?.replace(/([a-z])([A-Z])/g, '$1_$2').toLowerCase();
```
ditman: I had something like this, but there's some icons that don't work. I ended up with:

```
const ICON_MAP: Record<string, string> = {
  favoriteOff: "favorite_border",
  play: "play_arrow",
  rewind: "fast_rewind",
  starOff: "star_border",
};

// ...

  private getIconName(rawName: string): string {
    if (ICON_MAP[rawName]) return ICON_MAP[rawName];
    return rawName.replace(/[A-Z]/g, (letter: string) => `_${letter.toLowerCase()}`);
  }
```

(Assuming Material Symbols Outlined font)

renderers/react/src/v0_9/catalog/basic/components/Image.tsx
```diff
@@ -28,7 +28,8 @@ export const Image = createComponentImplementation(ImageApi, ({props}) => {
   const style: React.CSSProperties = {
     ...getBaseLeafStyle(),
     objectFit: mapFit(props.fit),
-    width: '100%',
+    width: `calc(100% - ${2 * 8}px)`,
```
ditman, line 32: What are these 8px? Is this to offset the margin from the base leaf style?

renderers/react/src/v0_9/catalog/basic/components/Tabs.tsx
```diff
@@ -31,21 +31,26 @@ export const Tabs = createComponentImplementation(TabsApi, ({props, buildChild})
...
       <div style={{display: 'flex', borderBottom: '1px solid #ccc', marginBottom: '8px'}}>
         {tabs.map((tab: _Tab, i: number) => (
           <button
             key={i}
             onClick={() => setSelectedIndex(i)}
             style={{
-              padding: '8px 16px',
+              flex: '1 1 0',
+              minWidth: 0,
+              padding: '8px 8px',
```
ditman: This is the same as `padding: 8px`

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -19,28 +19,97 @@ import {createComponentImplementation} from '../../../adapter';
...
+    .replace(/\*\*(.+?)\*\*/g, '<strong>$1</strong>')
+    .replace(/\*(.+?)\*/g, '<em>$1</em>')
+    .replace(/\[(.+?)\]\((https?:\/\/[^\s)]+)\)/g, '<a href="$2">$1</a>');
+}
+
+/**
+ * Parses simple markdown (headings, bold, italic, lists, links) to HTML.
+ * Supports the subset described in the v0.9 spec.
+ */
+function parseSimpleMarkdown(text: string): string {
```
ditman, line 40: Prefer using the package `@a2ui/markdown-it`? it has a `renderMarkdown` function that you should be able to use. End users shoudl be able to inject it as a dependency into the renderer, and then the renderer use it (we can discuss this on chat later!)

specification/v0_9/json/catalogs/basic/examples/33_financial-data-grid.json
```diff
@@ -80,71 +72,29 @@
...
-              "returnType": "string"
-            },
+            "text": { "path": "change" },
             "weight": 1
-          },
-          {
-            "id": "asset-market-cap",
-            "component": "Text",
-            "text": {
-              "call": "formatCurrency",
```
ditman, line 143: Why did this example change? It'd be interesting to call the `formatCurrency` function, right?

tools/composer/scripts/sync-gallery.mjs
```diff
@@ -15,10 +15,10 @@
  */
 
 /**
- * Syncs gallery widget data from the canonical spec examples.
+ * Syncs gallery widget data from the canonical v0.9 spec examples.
  *
- * Reads JSON examples from specification/v0_8 and v0_9 catalogs and
```
ditman, line 20: Nice! 🟥🟥🟥🟥!

ditman (commented):
Please, add an entry to the CHANGELOG of the react renderer describing the changes that happened in this PR! That way we know if we need to release it as a major/minor version.

specification/v0_9/json/catalogs/basic/examples/33_financial-data-grid.json
jerelvelarde: Oh, the table was too wide that it was clipping on the gallery view so I thought of ways to make it fit the card. Are all columns required or is this one in particular the highlight of this example?

renderers/react/src/v0_9/catalog/basic/components/Icon.tsx
jerelvelarde: this mustve been affected while i was debugging the issue with some icons not working...

jerelvelarde:
Thanks for the thorough review @ditman! Addressed everything in `9fcb685`:

**CHANGELOG** — Added entries for all renderer fixes under `Unreleased`.

**Column.tsx `minWidth: 0`** — Not the default actually — CSS `min-width` for flex items defaults to `auto`, which prevents children from shrinking below their content size. Added an inline comment to clarify.

**Icon.tsx** — Great call on the `ICON_MAP` approach. Adopted your map (`favoriteOff` → `favorite_border`, `play` → `play_arrow`, etc.) with the regex as fallback for standard camelCase names.

**Image.tsx `calc(100% - 16px)`** — Correct, it offsets the 8px leaf margin on each side. Added a comment explaining. Happy to explore a cleaner pattern (e.g. moving to padding) as a follow-up.

**Tabs.tsx** — Fixed, `padding: 8` now.

**Text.tsx / `@a2ui/markdown-it`** — Noted! Happy to discuss migrating to the shared package in a follow-up. Left the inline parser as-is for now.

**Financial data grid** — Restored the original example with `formatCurrency`, icons, symbols, and market cap column. Good point — the function calls are a better demo.

specification/v0_9/json/catalogs/basic/examples/33_financial-data-grid.json
ditman: I think the coolest part of this example is the `formatCurrency` call!

renderers/react/src/v0_9/adapter.tsx
```diff
@@ -89,9 +89,16 @@ export function createComponentImplementation<Api extends ComponentApi>(
       return () => binding.dispose();
     }, [binding]);
 
-    return (
+    const rendered = (
       <MemoizedRender props={props || ({} as Props)} buildChild={buildChild} context={context} />
     );
+
+    // Apply the common `weight` property as flex-grow when present.
```
jacobsimionato: The 'weight' property is part of the basic catalog and not part of the overall A2UI specification.

Please implement this *within* each Component of the basic catalog, rather than in the common React framework bindings. It should be possible for someone to implement a different catalog without weight and not be affected by this logic!

jerelvelarde: Good call, moved the weight handling out of the adapter entirely in 3bd5c98.

Added a withWeight() helper in the basic catalog's utils and each component now handles its own weight wrapping. The adapter is fully catalog-agnostic now so a custom catalog won't inherit any of the basic catalog's layout logic

let me know if theres a better way of approaching this

ditman:
Oh no, this has dropped sharply. @jerelvelarde are you still in landing these, or should we close the PR?

jacobsimionato:
Closing this, because we haven't heard back from @jerelvelarde :-D

PR #1149, opened 2026-04-12
Use component ID for input and label to avoid ID collisions

zeroasterisk:
This PR addresses a bug where the input IDs were hardcoded as `data` in the Lit renderer, which made browser automation difficult.

I made what I believe are reasonable changes, but since Paul wrote the initial version of this, I seriously doubt my changes and worry that his initial version is better. I defer to @paullewis but present this as a possible PR

### DIFF
--- a/renderers/lit/src/0.8/ui/datetime-input.ts
+++ b/renderers/lit/src/0.8/ui/datetime-input.ts
@@ -87,7 +87,7 @@ export class DateTimeInput extends Root {
       class=${classMap(this.theme.components.DateTimeInput.container)}
     >
       <label
-        for="data"
+        for="${this.id}-input"
         class=${classMap(this.theme.components.DateTimeInput.label)}
         >${this.#getPlaceholderText()}</label
       >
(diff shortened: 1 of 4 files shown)


renderers/lit/src/0.8/ui/datetime-input.ts
```diff
@@ -87,7 +87,7 @@ export class DateTimeInput extends Root {
       class=${classMap(this.theme.components.DateTimeInput.container)}
     >
       <label
-        for="data"
+        for="${this.id}-input"
```
ditman, line 90: how are you setting this `id` property? Is it directly in the DOM, or does it come from the agent?

ditman (approved):
Please update the `## Unreleased` section of the CHANGELOG with a description of this change.

Also, please consider adding this functionality to the other renderers so we don't have diverging capabilities in the basic catalog components!

(Also, also: only land this if it's a super critical need! I'd rather have this fix in the v0.9 stack :))

zeroasterisk:
> also: only land this if it's a super critical need! I'd rather have this fix in the v0.9 stack :)

That's enough warning for me to mark this as closed.  I was nervous about it. We can always come back to it later.

PR #1150, opened 2026-04-12
Populate error state in UI when agent communication fails

zeroasterisk:
This PR improves error handling in the shell client by populating the existing #error state when agent communication fails, making errors visible in the main UI.


ditman (changes requested):
I actually removed the whole snackbar shebang when migrating the shell app to lit v0.9, so this won't land cleanly. If we want this, we need to restore the snackbar into the (new) shell app. Sorry!

zeroasterisk:
can abandon for now

PR #1170, opened 2026-04-14
fix(web_core): Deep recursive type resolution for GenericBinder

jacobsimionato:
### Description of Changes

This PR modifies the `ResolveA2uiProps` and `ResolveA2uiProp` utility types in `generic-binder.ts` to perform a **deep recursive resolution** of component schemas rather than just a shallow top-level pass. 

Additionally, it adds a suite of strict compile-time type tests in `generic-binder-types.test.ts` to enforce that these generic types never inadvertently degrade to `any` during future development.

### Rationale

Previously, the `GenericBinder` types in `web_core` (which drive the strongly-typed Developer Experience in the React and Angular renderers) only resolved the outermost properties of a component's Zod schema. Because of this, deeply nested arrays of objects (like the `options` array in `ChoicePicker`) remained fully unresolved (defaulting to their raw `DynamicString` or `DynamicValue` union payloads like `string | DataBinding | FunctionCall`).

Similarly, the `ChildList` routing structure originally fell back to `any` rather than strictly defining the output as an array of `{id: string, basePath?: string}` nodes.

By migrating to recursive mappings, components requiring complex structural loops or nested properties now receive perfect type inference, improving IDE autocomplete and strictness downstream.

### Testing / Running Instructions

To test these changes locally:

1. Clone or checkout the branch.
2. Navigate to `renderers/web_core`.
3. Run `npm run build`. The build invokes `tsc`, which will process `generic-binder-types.test.ts`. This file uses native `@ts-expect-error` directives and `Assert<IsEqual<...>>` constraints. If the compiler succeeds, the types are mathematically sound and have not fallen back to `any`.
4. Navigate to `renderers/react`.
5. Run `npm run typecheck` to verify that the generated `ResolveA2uiProps` mappings accurately propagate through the React implementation without throwing mismatched assignment errors.


renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -135,26 +135,31 @@ function getFieldBehavior(
...
  */
-export type ResolveA2uiProp<T> = [NonNullable<T>] extends [Action]
-  ? (() => void) | Extract<T, undefined>
-  : [NonNullable<T>] extends [ChildList]
-    ? any | Extract<T, undefined>
-    : Exclude<T, DynamicTypes> extends never
-      ? any
-      : Exclude<T, DynamicTypes>;
+export type ResolveA2uiProp<T> = IsAction<T> extends true
+  ? (() => void) | Extract<T, undefined | null>
```
ditman, line 148: Please, test this in antigravity/vscode. I think I tried something similar to this and at some point the typescript inference was giving up attempting to resolve zod stuff (look at `props.something` in a component implementation, for example)

ditman:
@jacobsimionato did you close these on purpose, or were you moving your a2ui fork?

PR #1189, opened 2026-04-15
Port Angular Restaurant Sample to 0.9 and Stabilize Tests

josemontespg:
This PR ports the Angular restaurant sample app to version 0.9.

Credits to @gspencergoog for the first port attempt in https://github.com/google/A2UI/pull/1101, I reused most of his work to create this PR.

### Key Changes
- **Angular Renderer**: Updated `ComponentHostComponent` to handle property updates reactively using Angular Signals.
- **Tests**: Modernized tests in `renderers/angular` to use the real `ComponentBinder` service and formal `ComponentModel` instances instead of object-literal mocks.
- **Web Core**: Minor updates to `data-context`.
- **Agent**: Updated the restaurant finder agent and examples for 0.9.
- Added client support for streaming responses.

Fixes https://github.com/google/A2UI/issues/701

### Screenshots

| | Light | Dark |
|---|---|---|
| **Single Column** | light_mode_5_tall | dark_mode_5_tall |
| **Double Column** | light_mode_10_tall | dark_mode_10_tall |

PR #1217, opened 2026-04-16
Revert Python changes that broke Lit renderer in restaurant sample

josemontespg:
This PR reverts the changes to agent.py in PR #1189 that enforced a flat array schema and altered the streaming logic. These changes broke the Lit renderer which expected a different structure or handling. Reverting these fixes the Lit renderer and allows running both samples against the same agent for comparison. Verification: Confirmed both Lit and Angular samples run and render correctly after this revert.


ditman (approved):
Thanks for the fix! /cc @yuantian-agentspace @holtskinner this is touching stuff that you may want to rely on!

PR #1218, opened 2026-04-17
(Angular) Fix Tabs component.

ditman:
# Description

* Use the correct props in tabs rendering.
* Update the component host so it looks at its componentKey every time it changes, not just on init.

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/angular/src/v0_9/core/component-host.component.ts
+++ b/renderers/angular/src/v0_9/core/component-host.component.ts
@@ -19,10 +19,10 @@ import {
   ChangeDetectorRef,
   Component,
   DestroyRef,
-  OnInit,
   Type,
   inject,
   input,
+  effect,
 } from '@angular/core';
 import {NgComponentOutlet} from '@angular/common';
 import {ComponentContext, ComponentModel, SurfaceModel} from '@a2ui/web_core/v0_9';
@@ -64,7 +64,7 @@ import {BoundProperty} from './types';
   `,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class ComponentHostComponent implements OnInit {
+export class ComponentHostComponent {
   /** The key of the component to render, either an ID string or an object with ID and basePath. Defaults to 'root'. */
   componentKey = input<string | {id: string; basePath: string}>('root');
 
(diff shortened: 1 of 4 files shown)


renderers/angular/src/v0_9/core/component-host.component.ts
```diff
@@ -108,17 +126,15 @@ export class ComponentHostComponent implements OnInit {
...
 
-      const sub = surface.componentsModel.onCreated.subscribe((comp) => {
+      this.createSub = surface.componentsModel.onCreated.subscribe((comp) => {
         if (comp.id === id) {
           this.initializeComponent(surface, comp, id, basePath);
           this.cdr.markForCheck();
-          sub.unsubscribe();
+          this.createSub?.unsubscribe();
         }
       });
```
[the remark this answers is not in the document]
ditman: I refactored this to use signals

renderers/angular/src/v0_9/core/component-host.component.ts
```diff
@@ -139,26 +153,31 @@ export class ComponentHostComponent implements OnInit {
...
-    const onPropsUpdateSub = componentModel.onUpdated.subscribe(() => {
-      this.props = this.binder.bind(this.context!);
-      this.cdr.markForCheck();
-    });
-
-    this.destroyRef.onDestroy(() => {
-      // ComponentContext itself doesn't have a dispose, but its inner components might.
-      // However, SurfaceModel takes care of component disposal.
-      onPropsUpdateSub.unsubscribe();
+    this.propsSub = componentModel.onUpdated.subscribe(() => {
```
josemontespg, line 166: With the changes in this PR, I think now initializeComponet may be called multiple times during the lifecycle of the component. This line can create multiple subscriptions, but only the last one is tracked for unsubscribing.

We should keep track of all subscriptions to avoid memory leaks and side effects after destroy.

You can create a Subscription object, and use that to track multiple subs, like this:

```
private readonly propsSub = new Subscription();

...
this.propsSub.add(componentModel.onUpdated.subscribe( ... ));

```

https://rxjs.dev/guide/subscription

josemontespg: Oops I see you are handling subs manually below. Resolving

josemontespg:
Changes look good, but I'm curious: How can the componentType change after component init?

josemontespg (approved):

ditman:
> Changes look good, but I'm curious: How can the componentType change after component init?

@josemontespg realistically, it only changes when the item is first initialized, but I made it a signal so I don't need to call `markForCheck` or to rely in other signals mutating at the same time for the contents to re-render.

I guess this could be simplified and just leave the props as the signal?

josemontespg:
> > Changes look good, but I'm curious: How can the componentType change after component init?
> 
> @josemontespg realistically, it only changes when the item is first initialized, but I made it a signal so I don't need to call `markForCheck` or to rely in other signals mutating at the same time for the contents to re-render.
> 
> I guess this could be simplified and just leave the props as the signal?

I see, that makese sense. I think it would be more direct to have a boolean signal isReadyToRender that gates the *ngIf, and we set it to true when we are sure the component has initialized its internal state. I'd like to avoid manually coordinating the subscriptions of multiple observables and side effects if we can. What do you think?

josemontespg (approved):
Looks good, thanks for the discussion!

PR #1220, opened 2026-04-17
(Angular) Match restaurant sample styling.

### LINKED ISSUE #1200: Align angular theme to the new lit theme for the restaurant sample apps
@josemontespg I landed [the styling changes](https://github.com/google/A2UI/pull/1166) for angular. You could *potentially* rebase this PR and apply [this stylesheet](https://github.com/google/A2UI/blob/main/samples/client/lit/shell/theme/restaurant-theme.ts) to make the sample look more like the Lit one? And see what I ended up messing up in the Angular theming :P

_Originally posted by @ditman in https://github.com/google/A2UI/issues/1189#issuecomment-4255011276_

ditman:
# Description

This PR matches Angular's restaurant finder app style with the one in Lit by applying some CSS overrides.

It also tweaks the style of a couple of components on the renderer to actually use the overridden styles, and clears the screens in between interactions (to preserve the v0.8 behavior, even though this should be driven by the server!)

## Issues

* Fixes https://github.com/google/A2UI/issues/1200#event-24545378025

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/samples/client/angular/projects/restaurant/src/app/client.ts
+++ b/samples/client/angular/projects/restaurant/src/app/client.ts
@@ -40,6 +40,12 @@ export class Client {
     try {
       this.isLoading.set(true);
 
+      // Clear existing surfaces on interaction
+      const surfaceGroup = this.renderer.surfaceGroup;
+      for (const surfaceId of Array.from(surfaceGroup.surfacesMap.keys())) {
+        surfaceGroup.deleteSurface(surfaceId);
+      }
+
       const isString = typeof request === 'string';
       const bodyData = isString
         ? { query: request, contextId: this.contextId }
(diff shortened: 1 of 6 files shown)


samples/client/angular/projects/restaurant/src/app/client.ts
```diff
@@ -40,6 +40,12 @@ export class Client {
     try {
       this.isLoading.set(true);
 
+      // Clear existing surfaces on interaction
```
josemontespg, line 43: This is fine to align with the other sample app, but I feel like the right UX here would be to do a navigation to a fresh page on action. This would allow the user to use the browser back button to go back to the previous surface state. Otherwise the previous surface state is lost

josemontespg (approved):
Very well organized styles across the samples, thank you!

samples/client/angular/projects/restaurant/src/app/client.ts
ditman: Hmmm, I'm not sure if the demo app is setup like this (for example it doesn't have any routing); HOWEVER, this sounds like a cool feature for the framework itself to build multi-page apps, I wonder how hard would it be to have some sort of routing integration with the protocol (like: "this response causes a navigation to a different URL")

PR #1221, opened 2026-04-17
Don't use z.infer to fix closure compiler issues

### LINKED ISSUE #896: Figure out ergonomic and scalable way to ensure Zod schemas and types are portable
Also see b/495161036

We have had many issues with Zod schemas. Ideally, we would use schemas in the simplest possible way.

We should investigate:
- Do we need to document best practices around how to declare types, or can we enforce them programmatically?
- Can we avoid workarounds in open source code to suit Google's monorepo? Can we just change the compiler settings in Google's mono repo instead?

## Google monorepo issue
When imported into google's mono repo, the compiler there erases the types in some circumstances, breaking stuff at runtime

See https://github.com/google/A2UI/pull/872 which addresses this, but adds extra boilerplate.

## Lit type inference issue
@ditman has hit issues getting type inference to work properly for Lit. This may affect React too.

He recommends
```
export const ButtonApi = {
  name: "Button",
  schema: z.object({
    ...
  }).strict(),
} satisfies ComponentApi;
```

Instead of:

```
export const ButtonApi: ComponentApi = {
  ...
}
```

jiahaog:
# Description

- The Closure Compiler renames properties and variables to minimize code size.
- `z.infer` Relies on Type Names and works by inspecting the shape of the schema definition at compile time to create a TypeScript type. However, the runtime validation logic within Zod likely relies on the property names defined in the schema.

The two are mutually exclusive. To fix this, introduce some duplication of the types. The duplication should be mostly okay as if they are changed to not match, typescript will fail the typecheck except if there are optional fields introduced in the schema.

Coding agents can easily address this duplication. Also added a github actions to check for `z.infer`.

### API changes

•  renderers/web_core/src/v0_9/schema/common-types.ts : Added  ChecksSchema  as a new top-level
export (to avoid using  .shape.checks  on strictly typed schemas).
•  renderers/web_core/src/v0_9/schema/client-to-server.ts : The interface  A2uiValidationError  and
its schema  A2uiValidationErrorSchema  were renamed to  A2uiValidationErrorData  and
A2uiValidationErrorDataSchema  respectively (to resolve the duplicate export clash with  errors.ts
in  index.ts ).

Related: b/502243995
Fixes b/495161036
Fixes https://github.com/google/A2UI/issues/896
## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/.github/workflows/web_build_and_test.yml
+++ b/.github/workflows/web_build_and_test.yml
@@ -32,6 +32,17 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Check for z.infer usage
+        run: |
+          RESULTS=$(grep -r --include="*.ts" --exclude-dir="node_modules" "z\.infer" renderers | grep -v "v0_9/catalog/types" || true)
+          if [ -n "$RESULTS" ]; then
+            echo "Error: z.infer found in codebase! Use manual interfaces instead to ensure Closure Compiler compatibility."
+            echo "Instead of: type Foo = z.infer<typeof Schema>;"
+            echo "Do: interface Foo { ... } AND const Schema: z.ZodType<Foo> = z.object({ ... });"
+            echo "$RESULTS"
+            exit 1
+          fi
+
       - name: Set up Node.js
         uses: actions/setup-node@v6
         with:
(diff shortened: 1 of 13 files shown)


jacobsimionato (commented):
I think I like this. It seems easier to understand and for an agent to follow than doing a pattern like `z.infer extends` etc.

Let's chat with @sugoi-yuzuru though, who has been pursuing an alternative approach within the google repo - see CL/899744784

jiahaog:
@sugoi-yuzuru any thoughts about this?

jiahaog:
An alternative approach was taken internally that uses codegen instead.

PR #1186, opened 2026-04-14
Update React Shell to A2UI v0.9 and Add A2A Middleware

jacobsimionato:
## Description of Changes
This pull request updates the React Shell sample to support A2UI version 0.9. Key changes include:
- **Protocol Upgrade:** Migrated the React Shell from v0.8 to v0.9, utilizing the `A2uiSurface` component and `MessageProcessor` from `@a2ui/web_core`.
- **A2A Middleware:** Added a new Vite middleware (`samples/client/react/shell/middleware/a2a.ts`) to handle communication with A2A agents. This middleware supports both standard POST requests and Server-Sent Events (SSE) for streaming responses.
- **Streaming Support:** Refactored `A2UIClient` in the React Shell to support streaming A2UI messages from the agent via the new middleware.
- **Agent Updates:** Updated the Restaurant Finder agent (ADK) with v0.9 examples and improved prompt instructions for relative data binding paths in templates.
- **Python SDK Fix:** Updated the Python A2uiStreamParser to correctly handle path normalization based on the protocol version (ensuring leading slashes only for v0.8).
- **Mock Data Update:** Updated the React Shell's mock response system to produce valid v0.9 messages.

## Rationale
The A2UI protocol is evolving to v0.9, which introduces a more streamlined message format and better support for dynamic UI updates. Updating the React Shell ensures it remains a reference implementation for the latest version of the protocol. The addition of the A2A middleware simplifies client-side development by providing a unified way to talk to A2A agents with built-in streaming support.

### Prerequisites
- Ensure you have `uv` installed for running the Python agent.
- Ensure you have `npm` installed.

### Steps
1. **Start the Restaurant Finder Agent:**
   ```bash
   cd samples/agent/adk/restaurant_finder
   # Make sure you have a .env file with GEMINI_API_KEY
   uv run .
   ```
2. **Build Dependencies:**
   ```bash
   # From the root of the repo
   cd renderers/web_core
   npm install && npm run build
   cd ../react
   npm install && npm run build
   ```
3. **Run the React Shell:**
   ```bash
   cd samples/client/react/shell
   npm install
   npm run dev
   ```
4. **Verify:**
   - Open `http://localhost:5003` in your browser.
   - Enter a query like "Find Chinese restaurants in New York".
   - Verify that the UI renders correctly using v0.9 components and that the responses are streamed.

PR #1225, opened 2026-04-17
fix(v0.9): make createSurface idempotent + fix React README quick-start

### LINKED ISSUE #1223: [web_core,v0.9] processCreateSurfaceMessage crashes on duplicate surfaceId (breaks React StrictMode)
## Description

`MessageProcessor.processCreateSurfaceMessage` throws `A2uiStateError: Surface <id> already exists` when `createSurface` is called with a surfaceId that has already been created. This crashes the published README quick-start in any React app using StrictMode (default in dev environments nowadays AFAICT).

## Minimal reproduction

```bash
npm create vite@latest a2ui-repro -- --template react-ts
cd a2ui-repro
npm install
npm install @a2ui/react @a2ui/web_core
```

Replace `src/App.tsx` with the quick-start example (https://github.com/google/A2UI/blob/main/renderers/react/README.md?plain=1#L30).

```tsx
import { useState, useEffect } from 'react';
import { MessageProcessor } from '@a2ui/web_core/v0_9';
import { A2uiSurface, basicCatalog } from '@a2ui/react/v0_9';

function App() {
  const [processor] = useState(() => new MessageProcessor([basicCatalog]));
  const [surface, setSurface] = useState(null);

  useEffect(() => {
    const sub = processor.onSurfaceCreated(s => {
      if (s.id === 'main-chat') setSurface(s);
    });

    processor.processMessages([{
      version: 'v0.9',
      createSurface: {
        surfaceId: 'main-chat',
        catalogId: basicCatalog.id
      }
    }]);

    return () => sub.unsubscribe();
  }, [processor]);

  return (
    <div className="a2ui-container">
      {surface ? <A2uiSurface surface={surface} /> : <div>Initializing Agent UI...</div>}
    </div>
  );
}

export default App;
```

Run:

```bash
npm run dev
```

## Expected behavior

The app loads without errors.

## Actual behavior

Unhandled error on page load:

```
A2uiStateError: Surface main-chat already exists.
  at MessageProcessor.processCreateSurfaceMessage (message-processor.js:216:18)
  at MessageProcessor.processMessage (message-processor.js:191:17)
  at MessageProcessor.processMessages (message-processor.js:177:17)
```

## Root cause

React StrictMode (enabled by default in Vite/CRA/Next.js templates) double-fires `useEffect` in development to surface side-effect bugs. The first fire creates the surface; the second fire attempts to create it again and throws.

This also affects:
- **A2A streaming**: `status-update` events carry cumulative parts, redelivering `createSurface` on every chunk.
- **ADK conversation history replay**: the full history is replayed with each new request, including prior `createSurface` messages.

## Proposed fix

Make `processCreateSurfaceMessage` idempotent: if a surface with the given ID already exists, silently return instead of throwing. This aligns with the spec, which says "once a surface is created, its surfaceId and catalogId are fixed; to reconfigure them, the surface must be deleted and recreated," implying a duplicate `createSurface` is a no-op.

## Environment

- `@a2ui/react@0.9.0`
- `@a2ui/web_core@0.9.1`
- `react@19.2.5` / `react-dom@19.2.5
- Vite 8.0.8, `react-ts` template

### LINKED ISSUE #1224: React renderer quick-start code doesn't run: `The requested module '/src/App.tsx' does not provide an export named 'default'`
## Repro:

```
npm create vite@latest a2ui-repro -- --template react-ts
cd a2ui-repro
npm install
npm install @a2ui/react @a2ui/web_core
```

Use code from quick-start instructions: https://github.com/google/A2UI/blob/main/renderers/react/README.md?plain=1#L30

```
npm run dev
```

Open your browser at the listed url and then observe a blank page with the following console output:

```
main.tsx:4 Uncaught SyntaxError: The requested module '/src/App.tsx' does not provide an export named 'default' (at main.tsx:4:8)
(index):1 Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist.
client:969 [vite] server connection lost. Polling for restart...
````

### PR DESCRIPTION
**Pull request by andrewkolos**
Fixes #1223. Fixes #1224. Bug descriptions and repro steps can be found on the issues.

Also augments the quick-start example with some messages so that the user actually seeing something instead of [Loading root...] forever.

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/react/README.md
+++ b/renderers/react/README.md
@@ -36,7 +36,7 @@ import { useState, useEffect } from 'react';
 import { MessageProcessor } from '@a2ui/web_core/v0_9';
 import { A2uiSurface, basicCatalog } from '@a2ui/react/v0_9';
 
-function App() {
+export default function App() {
   // 1. Initialize the MessageProcessor with your catalogs
   const [processor] = useState(() => new MessageProcessor([basicCatalog]));
   const [surface, setSurface] = useState(null);
@@ -49,12 +49,32 @@ function App() {
     });
 
     // 3. Process messages from your agent (e.g., via WebSocket or SSE)
-    // For demo purposes, we manually process a creation message:
+    // For demo purposes, we manually process some static messages:
     processor.processMessages([{
       version: 'v0.9',
       createSurface: { 
         surfaceId: 'main-chat', 
         catalogId: basicCatalog.id // 'https://a2ui.org/specification/v0_9/basic_catalog.json'
+      }
+    }, {
+      version: 'v0.9',
+      updateComponents: {
+        surfaceId: 'main-chat',
+        components: [
+          { id: 'root', component: 'Column', children: ['greeting', 'description'] },
+          { id: 'greeting', component: 'Text', variant: 'h1', text: { path: '/title' } },
+          { id: 'description', component: 'Text', text: { path: '/body' } },
+        ],
+      }
+    }, {
+      version: 'v0.9',
+      updateDataModel: {
+        surfaceId: 'main-chat',
+        path: '/',
+        value: {
+          title: 'Hello from A2UI!',
+          body: 'This surface was rendered from static messages. Replace these with real agent responses to build interactive UIs.',
+        },
       }
     }]);
 
--- a/renderers/react/README.md
+++ b/renderers/react/README.md
@@ -90,6 +90,12 @@ export default function App() {
 }
 ```
 
+Running this example should display a heading "Hello from A2UI!" followed by a description paragraph. The example also demonstrates three message types:
+
+- [`createSurface`](../../specification/v0_9/docs/a2ui_protocol.md#createsurface) initializes the rendering surface.
+- [`updateComponents`](../../specification/v0_9/docs/a2ui_protocol.md#updatecomponents) defines the UI tree. Here, a `Column` containing two `Text` components.
+- [`updateDataModel`](../../specification/v0_9/docs/a2ui_protocol.md#updatedatamodel) provides the data that the components reference via [`path` bindings](../../specification/v0_9/docs/a2ui_protocol.md#path-resolution--scope) (e.g. `{ path: '/title' }` resolves to the `title` field in the data model).
+
 ## Defining Custom Components
 
 A2UI v0.9 strictly separates a component's API (Schema) from its implementation.
--- a/renderers/web_core/src/v0_9/processing/message-processor.ts
+++ b/renderers/web_core/src/v0_9/processing/message-processor.ts
@@ -280,6 +280,7 @@ export class MessageProcessor<T extends ComponentApi> {
     }
 
     if (this.model.getSurface(surfaceId)) {
+      console.warn(`[A2UI] createSurface: surface "${surfaceId}" already exists, ignoring.`);
       return;
     }
 
--- a/renderers/web_core/src/v0_9/processing/message-processor.ts
+++ b/renderers/web_core/src/v0_9/processing/message-processor.ts
@@ -280,7 +280,9 @@ export class MessageProcessor<T extends ComponentApi> {
     }
 
     if (this.model.getSurface(surfaceId)) {
-      console.warn(`[A2UI] createSurface: surface "${surfaceId}" already exists, ignoring.`);
+      console.warn(
+        `[A2UI] createSurface: surface "${surfaceId}" already exists, ignoring.`,
+      );
       return;
     }
 
--- a/renderers/web_core/src/v0_9/processing/message-processor.ts
+++ b/renderers/web_core/src/v0_9/processing/message-processor.ts
@@ -280,7 +280,7 @@ export class MessageProcessor<T extends ComponentApi> {
     }
 
     if (this.model.getSurface(surfaceId)) {
-      throw new A2uiStateError(`Surface ${surfaceId} already exists.`);
+      return;
     }
 
     const surface = new SurfaceModel<T>(
--- a/renderers/web_core/CHANGELOG.md
+++ b/renderers/web_core/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+- (v0_9) Duplicate `createSurface` messages are now silently ignored instead
+  of throwing. [#1225](https://github.com/google/A2UI/pull/1225)
+
 ## 0.9.1
 
 - Add new `FrameworkSignal` concept, which represents a generic signal from a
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: renderers/react/README.md [hunk 1]
```diff
@@ -36,27 +36,63 @@ import { useState, useEffect } from 'react';
...
   // 1. Initialize the MessageProcessor with your catalogs
   const [processor] = useState(() => new MessageProcessor([basicCatalog]));
   const [surface, setSurface] = useState(null);
 
   useEffect(() => {
     // 2. Listen for surface creation
     const sub = processor.onSurfaceCreated(s => {
-      // You can filter by surfaceId or manage multiple surfaces
       if (s.id === 'main-chat') setSurface(s);
     });
```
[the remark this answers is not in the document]
andrewkolos: Edge-case handling does not seem appropriate for a hello world/quick-start

andrewkolos:
I'm going to address each issue in separate PRs since I'm conflating problems here.

andrewkolos:
Reopening. I had to think about apps using `useEffect` should just guard themselves against calling non-idempotent APIs twice, but I recalled there are other reasons we want idempotency anyway as documented in #1223.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -280,7 +280,7 @@ export class MessageProcessor<T extends ComponentApi> {
     }
 
     if (this.model.getSurface(surfaceId)) {
-      throw new A2uiStateError(`Surface ${surfaceId} already exists.`);
+      return;
     }
```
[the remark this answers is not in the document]
andrewkolos: Should we be hiding malformed retries?

andrewkolos:
Side note from testing: `A2uiSurface` requires the root component to have exactly `id: 'root'`. If you use a different id, you get `[Loading root...]` forever with no console diagnostic. Might be worth calling this out in the quick-start comments or the component docs.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -280,7 +280,7 @@ export class MessageProcessor<T extends ComponentApi> {
...
 
     if (this.model.getSurface(surfaceId)) {
-      throw new A2uiStateError(`Surface ${surfaceId} already exists.`);
+      return;
```
ditman, line 286: I'm not sure about this, I think this is a feature, not a bug, I'd follow up with @jacobsimionato here. However, in case we want to hide the surface duplication, we should at least `console.warn` so users are aware that maybe their agent is misbehaving, instead of hiding this silently.

renderers/react/README.md
```diff
@@ -36,25 +36,41 @@ import { useState, useEffect } from 'react';
...
+      }
+    }, {
+      version: 'v0.9',
+      updateDataModel: {
+        surfaceId: 'main-chat',
+        path: '/',
+        value: {
+          title: 'Hello from A2UI!',
+          body: 'This surface was rendered from static messages. Replace these with real agent responses to build interactive UIs.',
+        },
```
ditman, line 77: Now that this example is more complicated, I think you should add a paragraph under the code block explaining what should the user *expect* to see, so they can verify whether the code works as intended.

This may be a good opportunity to explain something along the lines of: see? We defined the components first, and then the data that they're supposed to render. Also talk about the "path"s a little bit.

ditman (commented):
I think the throwiness of the message-processor is intended! /cc @jacobsimionato for a 2nd opinion :)

renderers/web_core/src/v0_9/processing/message-processor.ts
andrewkolos: AFAICT Jacob already blessed this approach in #1186: https://github.com/google/A2UI/pull/1186#issuecomment-4257122586

Re: `console.warn`: in the React StrictMode case (which is what the quick-start hits), a warning would fire on every dev-mode render, adding noise. The React strict mode It could be avoided but it would need additional "already-initialized?" guard logic that would bloat the example code more than it already is. Even with that adjustment, the error would appear even more often when #1186 lands barring some major overhaul.

In the interest of getting to a state where the quick start example it runs, I will change this to console.warn anyway.

ditman: > Re: console.warn: in the React StrictMode case (which is what the quick-start hits), a warning would fire on every dev-mode render, adding noise. The React strict mode It could be avoided but it would need additional "already-initialized?" guard logic that would bloat the example code more than it already is.

I think this is all right, the browser console should collapse two console entries that are identical:

Screenshot 2026-04-17 at 2 20 06 PM

> Even with that adjustment, the error would appear even more often when https://github.com/google/A2UI/pull/1186 lands barring some major overhaul.

Yeah, the sample code needs some love; in my case it's duplicating every protocol message, and I'm not sure if that's *expected*

> AFAICT Jacob already blessed this approach in https://github.com/google/A2UI/pull/1186: https://github.com/google/A2UI/pull/1186#issuecomment-4257122586

Hmmm, I'm not sure he approves of this approach in that text. The protocol is not very clear, but it seems to indicate that duplicate createSurfaces are an error. Also, what'd be the behavior? IMO we should not drop the newer messages, but instead overwrite the previously existing ones maybe?

jacobsimionato: Hey! Sorry in my previous PR I was trying all kinds of hacks to resolve this, but I never fully understood the underlying problem here of the `useEffect` behavior. Thank you for drilling to the root issue!

I agree with @ditman that the renderer library should fail here. We discussed an intention for duplicate `createSurface` messages with the same `surfaceId` to fail, partly to cover a collision scenario where two independent subagents try to create a surface with the same ID - one of them should fail. But I see that this never made it into the specification. I'll send a PR to clarify this.

Re how to deal with this issue practically, I'm not a React expert, but I think that we should rearchitect the system as necessary to avoid message duplication. I believe that `useEffect` has this behavior in dev because it's intended to be used to drive idempotent actions, e.g. rendering. But A2UI messages are not all idempotent, so I guess `useEffect` is not the right way to add placeholder content or convey messages from the transport layer to the `MessageProcessor`. Could we resolve this by avoiding ever calling `messageProcessor.processMessage()` or `messageProcessor.processMessages()` etc from `useEffect`?

Apologies that I likely caused this issue to emerge by relying on `useEffect` in the wrong place! I really appreciate having your real React expertise here to fix this kind of stuff.

jacobsimionato: Mailed you https://github.com/google/A2UI/pull/1238 to update the spec :-D

andrewkolos:
I've decided to supersede this with a new PR, as it drops the "make createSurface idempotent" idea: #1250

PR #1242, opened 2026-04-20
Propagate primaryColor to components in the basic catalog in Angular

### LINKED ISSUE #978: [v0.9] Wire up agent-provided primary color in Angular renderer basic catalog
Implement plumbing to apply the surface theme (specifically `primaryColor`) to components in the Angular basic catalog. This involves setting CSS custom properties on the surface host and updating component styles.

This is a child issue of https://github.com/google/A2UI/issues/686.

josemontespg:
### Summary

This PR propagates the primaryColor theme property to components in the basic catalog, and applies it as an override.

Fixes #978

#### Changes
- Added `BasicCatalogComponent` base directive to bind `--a2ui-color-primary` to host elements.
- Fixed Button component background override by using `background` shorthand instead of `background-color`.
- Fixed Checkbox component by using `accent-color` to apply primary color.

#### Screenshots

| Component | Default (no theme in createSurface message) | theme: {primaryColor: red} |
|---|---|---|
| **Button** (`button.component.ts`) <br> Primary variant styles | image | image |
| **Tabs** (`tabs.component.ts`) <br> Active tab indicators | image | image |
| **ChoicePicker** (`choice-picker.component.ts`) <br> Selected states | image |  image |
| **Slider** (`slider.component.ts`) <br> Range input highlights | image | image |
| **Checkbox** (`check-box.component.ts`) <br> Checked state | image | image |

### DIFF
--- a/renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.spec.ts
+++ b/renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.spec.ts
@@ -1,0 +1,83 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Component } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { BasicCatalogComponent } from './basic-catalog-component';
+import { A2uiRendererService, A2UI_RENDERER_CONFIG } from '../../core/a2ui-renderer.service';
+import { BasicCatalog } from './basic-catalog';
+
+@Component({
+  selector: 'test-basic-comp',
+  template: '<div>Test</div>',
+  standalone: true,
+})
+class TestBasicComp extends BasicCatalogComponent {
+
+}
+
+describe('BasicCatalogComponent', () => {
+  let fixture: ComponentFixture<TestBasicComp>;
+  let rendererService: A2uiRendererService;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [TestBasicComp],
+      providers: [
+        A2uiRendererService,
+        BasicCatalog,
+        {
+          provide: A2UI_RENDERER_CONFIG,
+          useFactory: (basicCatalog: BasicCatalog) => ({
+            catalogs: [basicCatalog],
+          }),
+          deps: [BasicCatalog],
+        },
+      ],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(TestBasicComp);
+    rendererService = TestBed.inject(A2uiRendererService);
+    rendererService.processMessages([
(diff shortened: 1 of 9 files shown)


renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
```diff
@@ -14,14 +14,16 @@
...
+import { Directive, computed, HostBinding, input, inject } from '@angular/core';
 import { injectBasicCatalogStyles } from '@a2ui/web_core/v0_9/basic_catalog';
 import { BoundProperty } from '../../core/types';
+import { A2uiRendererService } from '../../core/a2ui-renderer.service';
 
 /**
  * Base class for A2UI basic catalog components in Angular.
  *
  * Automatically injects the basic catalog styles when the component is instantiated.
+ * Also binds the primary brand color to the host element.
```
jacobsimionato, line 26: Ha, this seems like a neat way to share the implementation across the basic catalog components

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.spec.ts
```diff
@@ -0,0 +1,83 @@
+/**
```
ditman, line 1: Thanks for adding this!

renderers/angular/src/v0_9/catalog/basic/button.component.ts
ditman:
```suggestion
if (this.surface) {
```

You added a computed getter for `surface`, can it be used here, instead of the rendererService directly.

~In fact, is the `rendererService` needed as a public property after your refactor, could we expose other properties of the `rendererService` from computed properties in the base class?~ Never mind, it's `protected`!

renderers/angular/src/v0_9/catalog/basic/text.component.spec.ts
```diff
@@ -31,11 +32,16 @@ describe('TextComponent', () => {
 
     await TestBed.configureTestingModule({
       imports: [TextComponent],
-      providers: [{ provide: MarkdownRenderer, useValue: mockMarkdownRenderer }],
+      providers: [
+        { provide: MarkdownRenderer, useValue: mockMarkdownRenderer },
+        A2uiRendererService,
+        { provide: A2UI_RENDERER_CONFIG, useValue: { catalogs: [] } },
```
ditman, line 38: Why do we need to inject the A2UI_RENDERER_CONFIG in these (seemingly unrelated) tests? Is it a dependency of the A2uiRendererService?

Can both the A2uiRendererService and its config default to something so we don't have to inject them everywhere just for default values?

ditman (approved):
Small comment about the refactor (opportunity to simplify some more, I think).

Also a comment about providing defaults for the injected components, so test setup doesn't get too bloated (but this is more a nit than an actual comment)

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
josemontespg: Exactly, thanks!

renderers/angular/src/v0_9/catalog/basic/button.component.ts
josemontespg: Good point, thanks!

renderers/angular/src/v0_9/catalog/basic/text.component.spec.ts
josemontespg: Yes, the A2uiRendererService requires a value for A2UI_RENDERER_CONFIG.

We could reduce the testing boilerplate by creating a TestingModule which provides some sensible defaults for testing. Then most test would have to import that module.

Not sure if it is worth it to refactor at the moment, because it is not that much boilerplate, and adding a layer of abstraction also adds indirection.

Issue #1245, opened 2026-04-21
Select examples/platforms to support and setup continuos validation.


polina-c:
Problem: some examples/platforms in GenUI SDK and A2ui repo do not work.
See: https://github.com/flutter/genui/issues/878

Internal notes: go/ci-for-examples.

Subtasks:

- [ ] agree on examples and platforms to support
- [ ] cleanup not supported examples/platforms
- [ ] setup evals
- [ ] setup manual validation to backup evals

gspencergoog:
A2UI Triage: Marked as P2. Adding `CI/CD` and `component: samples` labels.

PR #1251, opened 2026-04-21
Define workflow for tests with ai and create initial tests for Dart SDK.

polina-c:
Contributes to:
- https://github.com/flutter/genui/issues/870
- https://github.com/google/A2UI/issues/1245

The eval verifies that these entities work well together:
1. Python agent, plus whatever libraries it uses
2. Dart SDK
3. AI model

This PR:

1. Adds eval workflow that:
  - reads api key from secret
  - runs eval/run.sh
4. Creates project eval/dart_and_flutter with test, that verify that app can:
  - ask a gemini model for a joke directly
  - start restaurant finder agent 
  - request the agent to request model
5. Adds configurations to start processes via vscode UI


jacobsimionato (commented):
Hey! Thanks for digging into restaurant finder! I agree with you that we need better testing and evaluations, but my opinion is that it would be better not to add this specific test implemented in Dart.

So I think there are a couple of different types of infra we need, which could be implemented using some of the same frameworks, but have different properties overall:

- End-to-end / integration tests for agents and clients: Tests that spin up actual servers or clients and send queries or data to them, verifying they behave correctly. These are primarily to improve dev velocity and user experience by making maintenance easier and quality higher. These can be relatively slow to run each test case because they spin up separate processes. They don't need to have complex analytics - just pass or fail. Ideally, they are written in the same language as the code under test, or a super common language. I think we should strive to have an end-to-end test for *every* important sample (prioritising as you suggested).

- Inference quality evaluation harness: Some dev tools which can trigger inferences based on some relatively large set of test data points and do analysis so that we can see how a change impacts overall output quality or latency. I think we should have a sophisticated eval system implemented in just *one* language, because if we make our SDKs consistent in every language, there is no need to do extensive evaluations using multiple parallel rigs. I think the eval stuff should be implemented in Python seeing as it's so commonly used for LLM-related stuff, or another super common language. I love Dart, but I think that implementing it in Dart is a barrier for broad usage, because many people don't have Dart installed on their dev machines.

I think what you have here is an end-to-end test for the restaurant finder agent, because it spins up a real binary and verifies that it works correctly for some simple cases. I think the functionality you have here would be super valuable to the project. But, I personally think Dart is not the best language choice, because the agent under test is already written in Python, so now to run the agent and the test, you need two separate runtimes installed. Also, I believe this is the first use of Dart in this entire repository - I think we already have too many languages in one repository, so it'd be great to avoid adding a new one.

So...... can you reimplement this in Python 😬. Or just use it for your local development for now and we can figure out end-to-end testing later?

I hope I'm not being too picky! I'm curious what @dmandar who wrote this sample originally thinks!

polina-c:
I agree that we need other types of testing.

And, I agree we should choose one language for infrastracture and Python is a good candidate. Thank you for pushing for consistency here.

But, i think your interpretation of this exact test is NOT correct (I updated PR description above to make it clearer).

This exect test does not test agent.

Instead, it tests that combination of these three things work well together:

1. python restaurant agent and whatever libraries it uses
2. dart genui sdk
3. AI model

I believe this test:
1. Have to be in Dart, because sdk is in dart, and we want to be able to do white-box checks, not just black-box observations.
2. Is critical for Dart SDK, because it will guarantee the functionality will not erode as it happens normally with features, that are not test covered.
3. Should be as simple as possible for maintainability, so i use just unit testing.

I have a [task](https://github.com/flutter/genui/issues/870) to enable example that connects dart renderer with python agent. I want my work to stay and serve users. I do not want it to erode with minor changes in sdk or dependencies change. So, this kind of Dart testing is a prerequisite for the task.

Again, existence of this test does not mean we should not do other types of testing.

About unit vs integration testing technologies. 

First, the words "unit" and "integration" mean different things for technologies and concepts, right?
Both unit and integration testing of some level can be done with both unit and integration technologies.

Tests that use Unit testing technology are much cheaper to maintain, then ones that use integration testing technology. Thus, I believe we should cover whatever is possible to cover with unit testing technology. And then cover the remaining gaps with integration testing. 

Thoughts?

polina-c:
About "just use it for your local development; we can figure out end-to-end testing later":

Yes, I am practicing TDD. What's the point to remove valuable test? We can figure out needed end-to-end testing later without deleting this test, right?

jacobsimionato:
Ahhhhhh you're right that I misunderstood this test and missed some things when reading it. I'm sorry about that! I'm embarrassed to admit that I didn't notice it was exercising the Gen UI SDK.

So I understand this is an end-to-end test for Gen UI SDK + Restaurant finder, right? How about putting it at `samples/agent/adk/restaurant_finder/end_to_end_tests/dart`?

Rationale:
- Put the test close to the restaurant finder agent that it's coupled to
- I think this is closer to an end-to-end test than an eval system, because it is focused more on testing the infrastructure (e.g. agent frameworks, communication etc) rather than inference quality.
- We will eventually have an eval suite written in Python, and that could use the top level `/eval` location

polina-c:
Thank you. It seems we are almost on the same page here.

### Some comments:

> this is an end-to-end test for Gen UI SDK + Restaurant finder, right? 

Almost. It is for three things: Gen UI SDK + Restaurant finder agent + AI Model

AI model is important component here, not stubbed. We want to see Gen UI SDK is working reliably with real responses, produced by Restaurant finder + AI Model, where AI Model is not stubbed. 
If we will want to test SDK for some concrete responses, we will create tests for a stubbed finder agent.

>  I think this is closer to an end-to-end test than an eval system

I am observing that engineering community is not consistent defining what "eval" is. So far I saw these definitions:

1. A test that tests prompt
2. A test that tests anything that involves AI
3. A test that is ok to produce X% of failures.

It seems you use definition #1. I updated README.md in the folder 'eval' to contain this definition. 

> How about putting it at samples/agent/adk/restaurant_finder/end_to_end_tests/dart?

Yes, it make sense to have this test closer to the code it tests. The code it tests is the code of the finder's Dart client.
Initially I wanted to put it to GenUI SDK repo. But now I see it will be easier to have in in a2ui repo, to test it together with the agent.

I am thinking to place it to:

samples/client/flutter/restaurant_finder/app

So, it will make sense to put the test to:

samples/client/flutter/restaurant_finder/e2e_test

This is PR that removes the client from GenUI SDK: https://github.com/flutter/genui/pull/885

How does all this sound for you?

### Changes I did:

1. Renamed .github/workflows/eval.yaml to .github/workflows/test_with_ai.yaml
2. Moved eval/run.sh to scripts/test_with_ai.sh
3. Updated .github/PULL_REQUEST_TEMPLATE.md to include instruction to run the tests on forked branched manually
4. Moved the test to samples/client/flutter/restaurant_finder/e2e_test

Does this structure look right for you?

scripts/e2e_test.sh
```diff
@@ -0,0 +1,31 @@
...
+# 
+# If your PR is from forked branch, please, run this script locally before merging.
+# If the script fails both locally and
+# [on the main branch](https://github.com/flutter/genui/actions?query=event%3Aschedule+workflow%3Atest_with_ai_workflow),
+# file an issue and link in your PR. 
+#
+# To run script locally, you need to set API key as an environment variable.
+# Example: export GEMINI_API_KEY=your_api_key
+
+cd "$(dirname "$0")/../samples/client/flutter/restaurant_finder/e2e_test"
```
jacobsimionato, line 29: This script is specific to the Flutter e2e tests for the restaurant finder agent. How about moving this script into `samples/client/flutter/restaurant_finder/e2e_test` and reference it there?

BTW part of my motivation here is that there is a plan to break out the A2UI repository into many repositories - https://docs.google.com/document/d/1914FKcF5LOXrj8y7-xf-45bJAwaPtcix_nBhOCuMECc/edit?resourcekey=0-g9pTHlodm5jGpkbknk-TJw&tab=t.0#heading=h.sd9tbgmewn0l

When we do this, it will be easier if the different parts (e.g. samples, renderers etc) are clearly separated to begin with.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,6 +12,7 @@ _List which issues are fixed by this PR. For larger changes, raising an issue fi
 - [ ] I have added updates to the [CHANGELOG].
 - [ ] I updated/added relevant documentation.
 - [ ] My code changes (if any) have tests.
+- [ ] I have verified that [scripts/test_with_ai.sh](../scripts/test_with_ai.sh) passes.
```
jacobsimionato: This is covered by CI, so I don't think we need it specifically here - we'll see if this is necessary / has worked based on the results of the CI.

If we did add an item to this checklist, I think it should be a way to run all the CI workflows locally, rather than just this one. In most cases, other tests will be more relevant, because this only covers the Python SDK, restaurant finder demo, and Flutter SDK. Maybe we can experiment with something like https://github.com/nektos/act ?

.github/workflows/test_with_ai.yaml
```diff
@@ -0,0 +1,62 @@
+# Copyright 2025 The Flutter Authors.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+name: test_with_ai_workflow
```
jacobsimionato: How about calling this something like `Restaurant Finder E2E Test` and the file `restaurant_finder_e2e_test`, to be consistent with the other workflows, e.g. https://github.com/google/A2UI/blob/main/.github/workflows/lit_build_and_test.yml etc. E.g. I think that it should be "x test" rather than "test x", and the name here should be in regular English rather than snake case etc.

.github/workflows/e2e_test.yaml
```diff
@@ -0,0 +1,62 @@
...
+name: test_with_ai_workflow
+
+on:
+  push: 
+    # Workflow runs on push to any branch in the repo (not forks).
+  schedule:
+    - cron: "0 * * * *" # hourly
+
+jobs:
+  test_with_ai_job:
```
jacobsimionato: Also update this name

jacobsimionato (approved):
Thank you for iterating! This looks great to me. I think it will be nice for developers to have a Flutter example in this repo, alongside the other client samples.

.github/PULL_REQUEST_TEMPLATE.md
polina-c: Nope, CI will run only for local branches, not for forked. Forked branches do not have access to key.

Updated requirement. 

People will need to configure key to run this test. I do not see how act will help.

.github/workflows/test_with_ai.yaml
polina-c: I was supposing this workflow, as well as scripts/test_with_ai.sh to run all tests that require key. 

But, yes, it may become test for all e2e. I want to put all them together, to make it easier to run them with one command locally.

Made renamings you requested.

PR #1252, opened 2026-04-21
Exclude SVG elements from CSS reset

### LINKED ISSUE #1175: reset-styles breaks SVG rendering
[reset-styles](https://github.com/google/A2UI/blob/ba5ff2b13f2b78dbdc7d3b49420a97777a225cbc/renderers/react/src/v0_8/styles/reset.ts#L33) break rendering of the SVG elements:

Image

selamw1:
Fixes #1175

The CSS reset rule `all: revert` in `v0_8` broke SVG rendering by overriding presentational attributes (e.g., `fill`, `stroke`).

### Solution
Excluded SVG elements and their descendants from the reset:
`:where(.a2ui-surface) :where(*:not(svg):not(svg *))`

| Before Fix | After Fix |
| :---: | :---: |
| [Screenshot 2026-04-21 at 1 20 36 PM] | [Screenshot 2026-04-21 at 1 22 19 PM] |

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/renderers/react/CHANGELOG.md
+++ b/renderers/react/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- (v0_8) Exclude SVG elements and descendants from CSS reset to restore SVG rendering. [#1252](https://github.com/google/A2UI/pull/1252)
+
 - **BREAKING CHANGE**: Renamed `createReactComponent` to `createComponentImplementation`.
 - **BREAKING CHANGE**: Renamed `createBinderlessComponent` to `createBinderlessComponentImplementation`.
 - **BREAKING CHANGE**: Removed `minimalCatalog`.
--- a/renderers/react/src/v0_8/styles/reset.ts
+++ b/renderers/react/src/v0_8/styles/reset.ts
@@ -32,7 +32,7 @@
  */
 export const resetStyles: string = `
 @layer a2ui-reset {
-  :where(.a2ui-surface) :where(*) {
+  :where(.a2ui-surface) :where(*:not(svg, svg *:not(foreignObject *))) {
     all: revert;
   }
 }
--- a/renderers/react/tests/v0_8/unit/components/SvgReset.test.tsx
+++ b/renderers/react/tests/v0_8/unit/components/SvgReset.test.tsx
@@ -1,0 +1,43 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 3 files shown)


jacobsimionato:
@ditman does this look safe? Do you think it has much chance of breaking things once imported into the google monorepo especially?

ditman:
Yeah, this looks good to me. The current implementation of the reset is very hardcore, and is probably breaking in the same way internally (it's just that nobody has hit this corner case yet).

I'll bring it up to the internal users' chat to see if anybody has any concerns.

ditman:
@selamw1 can we apply this to other versions of the renderers, is this needed on lit v0.9 or any other renderer? We could land them all at the same time!

selamw1:
> @selamw1 can we apply this to other versions of the renderers, is this needed on lit v0.9 or any other renderer? We could land them all at the same time!

@ditman  yes, I checked and both versions(_React v0.8 & v0.9_) share the exact same `reset.ts` file under the hood. This PR fixed both versions at the same time.
Lit Renderers are not affected, Lit components use Shadow DOM to encapsulate styles. It never used the `all: revert` rule and was never broken.

ditman:
Thanks for checking! One last thing, do you mind updating the CHANGELOG.md of the React renderer in the "Unreleased" section listing with a small sentence about this fix (it can be the commit message, or the title of the PR) and a link to the PR (similar to [this](https://github.com/google/A2UI/blob/main/renderers/angular/CHANGELOG.md)), so we can compute the next version number when we release this package next? Thanks!!

selamw1:
> Thanks for checking! One last thing, do you mind updating the CHANGELOG.md of the React renderer in the "Unreleased" section listing with a small sentence about this fix (it can be the commit message, or the title of the PR) and a link to the PR (similar to [this](https://github.com/google/A2UI/blob/main/renderers/angular/CHANGELOG.md)), so we can compute the next version number when we release this package next? Thanks!!


sure, `CHANGELOG.md` updated.

ditman:
I've rebased this branch with `main`. There has been a big [formatting change](https://github.com/google/A2UI/pull/1338) today that might flag the changes here as being wrong, but the fix should be to run the [`scripts/fix_format.sh` script](https://github.com/google/A2UI/blob/main/scripts/fix_format.sh).

selamw1:
I made several commits to pass the formatting changes, and all checks are now passing successfully.

ditman (approved):
Thanks for fixing this and adding a test!

PR #1265, opened 2026-04-22
Improve Angular catalog components props typing

### LINKED ISSUE #1003: [Angular] Application developer can access type-safe props when implementing Components
Consider the choice picker implementation for each of the three renderers:
- https://github.com/google/A2UI/blob/main/renderers/angular/src/v0_9/catalog/basic/choice-picker.component.ts
- https://github.com/google/A2UI/blob/main/renderers/lit/src/v0_9/catalogs/basic/components/ChoicePicker.ts
- https://github.com/google/A2UI/blob/main/renderers/react/src/v0_9/catalog/basic/components/ChoicePicker.tsx

In angular, we have:

```
  displayStyle = computed(() => this.props()['displayStyle']?.value());
```

While in react, it's like:

```
          if (props.displayStyle === 'chips') {
```

It'd be nice of Angular provided a type safe way to access the props, even including deeply nested props etc. This makes it *way* easier to develop Components. It affects the API, so let's do it as soon as we can.

@ava-cassiopeia @gspencergoog

### LINKED ISSUE #1216: Create a CatalogComponentBase class that all catalog components extend, with basic common functionality
> This seems helpful. I wonder if it could make sense for this to be in a base class that is not specific to the basic catalog, i.e. at the framework level, seeing as these concepts are generic to A2UI?
> 
> I don't feel strongly though, and out of scope for this PR - just an idea. 

 _Originally posted by @jacobsimionato in [#1211](https://github.com/google/A2UI/pull/1211/changes#r3096515973)_

josemontespg:
This PR improves the type safety in Angular catalog components with a strongly-typed props architecture using the generic `BaseCatalogComponent<typeof Api>` pattern.

Fixes #1003
Fixes #1216

Key changes:
- Updated all 18 basic catalog components to extend `BaseCatalogComponent<typeof SomethingApi>`.
- Using strict types for the props(), not using `any` in the basic catalog.
- Fixed type bugs and discrepancies between component implementation and schema.


renderers/angular/src/v0_9/catalog/basic/choice-picker.component.ts
```diff
@@ -125,13 +126,13 @@ import { BasicCatalogComponent } from './basic-catalog-component';
...
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class ChoicePickerComponent extends BasicCatalogComponent {
-  readonly displayStyle = computed(() => this.props()['displayStyle']?.value());
-  readonly choices = computed(
-    () => this.props()['choices']?.value() || this.props()['options']?.value() || [],
+export class ChoicePickerComponent extends BasicCatalogComponent<typeof ChoicePickerApi> {
+  readonly displayStyle = computed(() => this.props().displayStyle?.value());
+  readonly options = computed(
+    () => this.props().options?.value() || this.props().options?.value() || [],
```
josemontespg: This fixes a bug. The schema declares 'options', but this component expected 'choices'.

renderers/angular/src/v0_9/catalog/basic/choice-picker.component.ts
```diff
@@ -167,7 +168,7 @@ export class ChoicePickerComponent extends BasicCatalogComponent {
       this.props()['value']?.onUpdate(next);
     } else {
       if (active) {
-        this.props()['value']?.onUpdate(value);
+        this.props()['value']?.onUpdate([value]);
```
josemontespg, line 171: This fixes a bug

renderers/angular/src/v0_9/catalog/basic/list.component.ts
```diff
@@ -93,9 +94,9 @@ import { Child } from '../../core/component-binder.service';
   ],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class ListComponent extends BasicCatalogComponent {
+export class ListComponent extends BasicCatalogComponent<typeof ListApi> {
   readonly listStyle = computed(() => this.props()['listStyle']?.value());
-  readonly orientation = computed(() => this.props()['orientation']?.value() || 'vertical');
+  readonly direction = computed(() => this.props()['direction']?.value() || 'vertical');
```
josemontespg, line 99: This fixes another discrepancy found by strict types.

renderers/angular/src/v0_9/catalog/basic/slider.component.ts
```diff
@@ -44,7 +45,7 @@ import { BasicCatalogComponent } from './basic-catalog-component';
         [min]="min()"
         [max]="max()"
         [step]="step()"
-        [value]="value() ?? min()"
+        [value]="value()"
```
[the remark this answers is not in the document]
josemontespg: Value is a required field in the schema. If it is absent, it will fail zod validation before this component gets rendered.

renderers/angular/src/v0_9/catalog/basic/text-field.component.ts
```diff
@@ -48,9 +49,9 @@ import { BasicCatalogComponent } from './basic-catalog-component';
         [value]="value()"
         (input)="handleInput($event)"
         [placeholder]="placeholder()"
-        [class.invalid]="props()['isValid']?.value() === false"
+        [class.invalid]="props().isValid.value() === false"
       />
-      @for (message of props()['validationErrors']?.value(); track message) {
+      @for (message of props().validationErrors.value(); track message) {
```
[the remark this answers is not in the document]
josemontespg: Good point, done

renderers/angular/src/v0_9/catalog/basic/text-field.component.ts
```diff
@@ -93,11 +94,11 @@ import { BasicCatalogComponent } from './basic-catalog-component';
...
 })
-export class TextFieldComponent extends BasicCatalogComponent {
-  readonly label = computed(() => this.props()['label']?.value());
-  readonly value = computed(() => this.props()['value']?.value() || '');
-  readonly placeholder = computed(() => this.props()['placeholder']?.value() || '');
-  readonly variant = computed(() => this.props()['variant']?.value());
+export class TextFieldComponent extends BasicCatalogComponent<typeof TextFieldApi> {
+  readonly label = computed(() => this.props().label?.value());
+  readonly value = computed(() => this.props().value?.value() || '');
+  readonly placeholder = computed(() => (this.props() as any).placeholder?.value() || '');
```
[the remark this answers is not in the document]
josemontespg: Added placeholder to the schema

renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts
josemontespg, line 1: These added props were missing from the schema, but were being read in the component implementation. Now fixed.

renderers/angular/src/v0_9/catalog/basic/slider.component.ts
```diff
@@ -79,15 +80,15 @@ import { BasicCatalogComponent } from './basic-catalog-component';
...
 })
-export class SliderComponent extends BasicCatalogComponent {
-  readonly label = computed(() => this.props()['label']?.value());
-  readonly value = computed(() => this.props()['value']?.value());
-  readonly min = computed(() => this.props()['min']?.value() ?? 0);
-  readonly max = computed(() => this.props()['max']?.value() ?? 100);
-  readonly step = computed(() => this.props()['step']?.value() ?? 1);
+export class SliderComponent extends BasicCatalogComponent<typeof SliderApi> {
+  readonly label = computed(() => this.props().label?.value());
+  readonly value = computed(() => this.props().value.value());
```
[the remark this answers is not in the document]
josemontespg: Outdated.

josemontespg:
@jacobsimionato @ditman Initially I used dot notation to access props, like `props().foo`, but then realized that this might cause a bug with build configurations that rename object properties. Reverted back to `props()['foo']`.

Has there been property renaming issues before? Would it be safe to use dot notation?

renderers/angular/src/v0_9/catalog/basic/audio-player.component.ts
```diff
@@ -65,7 +66,7 @@ import { BasicCatalogComponent } from './basic-catalog-component';
   ],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class AudioPlayerComponent extends BasicCatalogComponent {
+export class AudioPlayerComponent extends BasicCatalogComponent<typeof AudioPlayerApi> {
```
ditman, line 69: (I think I just left a comment about this in a separate issue :P)

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
```diff
@@ -50,12 +42,11 @@ export abstract class BasicCatalogComponent {
     return this.theme()?.primaryColor;
   });
 
-  /**
-   * Computes the weight of the component from the properties.
-   */
-  protected readonly weight = computed(() => this.props()['weight']?.value() ?? null);
+  /** Weight is applied as flex css property on the component host HTML element. */
+  protected readonly weight = computed(() => (this.props() as any)['weight']?.value() ?? null);
```
ditman: Why do we need `props() as any` here? IIRC all the Zod schemas for components of the basic catalog should contain the `weight` property; see [the `CommonProps` schema](https://github.com/google/A2UI/blob/main/renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts#L31-L39).

Isn't `this.props()` typed as `Api`?

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.spec.ts
```diff
@@ -19,14 +19,24 @@ import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { BasicCatalogComponent } from './basic-catalog-component';
 import { A2uiRendererService, A2UI_RENDERER_CONFIG } from '../../core/a2ui-renderer.service';
 import { BasicCatalog } from './basic-catalog';
+import { ComponentApi } from '@a2ui/web_core/v0_9';
+import z from 'zod/v3';
```
ditman: Hmmm, I don't remember seeing `from 'zod/v3'` elsewhere, isn't this just `from 'zod'`?

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
```diff
@@ -26,16 +27,7 @@ import { A2uiRendererService } from '../../core/a2ui-renderer.service';
...
-  /**
-   * Reactive properties resolved from the A2UI ComponentModel.
-   */
-  props = input<Record<string, BoundProperty>>({});
-
-  surfaceId = input.required<string>();
-  componentId = input.required<string>();
-  dataContextPath = input<string>('/');
-
+export abstract class BasicCatalogComponent<Api extends ComponentApi> extends BaseCatalogComponent<Api> {
```
ditman: The names of this class and the base class are too similar. Any chance of renaming:

* `BaseCatalogComponent` -> `A2uiAngularElement`
* `BasicCatalogComponent` -> `BasicCatalogA2uiAngularElement`

Or [similar](https://github.com/google/A2UI/blob/main/renderers/lit/src/v0_9/catalogs/basic/basic-catalog-a2ui-lit-element.ts#L28-L30)? My note would be to make the name of the superclass something that contains the name of the base class :P

renderers/angular/src/v0_9/catalog/basic/choice-picker.component.ts
ditman: Damn!

renderers/angular/src/v0_9/catalog/basic/icon.component.ts
```diff
@@ -82,8 +83,8 @@ const ICON_NAME_OVERRIDES: Record<string, string> = {
   ],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class IconComponent extends BasicCatalogComponent {
-  readonly color = computed(() => this.props()['color']?.value());
+export class IconComponent extends BasicCatalogComponent<typeof IconApi> {
+  readonly color = computed(() => this.props().color?.value());
```
ditman: This is so much nicer!

renderers/angular/src/v0_9/catalog/basic/list.component.ts
ditman: Why do we need to stringly access values here, instead of `this.props().direction?`, like in the Icon component?

renderers/angular/src/v0_9/core/component-binder.service.ts
```diff
@@ -48,7 +48,7 @@ export class ComponentBinder {
    */
   bind(context: ComponentContext): Record<string, BoundProperty> {
     const props = context.componentModel.properties;
-    const bound: Record<string, any> = {};
+    const bound: Record<string, BoundProperty<any>> = {};
```
ditman, line 51: If this is not bound to the type of the Api, are we losing type safety later whe we do `.value` on a bound property or similar? How does this relate to the `props` of the object?

renderers/angular/src/v0_9/core/types.ts
```diff
@@ -50,3 +53,41 @@ export interface BoundProperty<T = any> {
...
+*   schema: z.object({
+*     text: z.string(),
+*   })
+*   .strict(),
+* } satisfies ComponentApi;
+* export type TextComponentProps = ComponentApiToProps<typeof TextComponentApi>; // outputs { text: BoundProperty<string>; }
+*/
+export type ComponentApiToProps<Api extends ComponentApi> = InferredInterfaceToProps<ExtendedProps<z.infer<
+  Api['schema']
+>>>;
```
ditman, line 93: `>>>;` -> Looks like a merge conflict :P

renderers/angular/.gitignore
```diff
@@ -1 +1,2 @@
 a2ui_explorer/src/app/generated
+.angular/
```
ditman: (I added something similar to this in my latest PR too! Why are we getting this `.angular` dir now all of a sudden?)

renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts
ditman: It seems the `Icon` specification on the basic catalog does NOT contain a color property:

* https://github.com/google/A2UI/blob/main/specification/v0_9/json/basic_catalog.json#L87-L186

For now, I'd drop the `color` from the implementation of the widget (or default it to the foreground color).

Can you check if the other properties you needed to add here to implement the renderer are part of the specification or not?

renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts
```diff
@@ -395,6 +403,9 @@ export const TextFieldApi = {
           'A regular expression used for client-side validation of the input.',
         )
         .optional(),
+      placeholder: DynamicStringSchema.describe(
```
ditman: This one is also not in the `TextField` [specification](https://github.com/google/A2UI/blob/main/specification/v0_9/json/basic_catalog.json#L529).

These might be legit bugs on the specification, we need to check with Greg to see if they need to be added to 0.9 or dropped from the renderer :/

renderers/web_core/src/v0_9/schema/common-types.ts
```diff
@@ -37,6 +38,7 @@ export const FunctionCallSchema = z
   .describe(
     'REF:common_types.json#/$defs/FunctionCall|Invokes a named function on the client.',
   );
+export type FunctionCallType = z.infer<typeof FunctionCallSchema>;
```
ditman, line 41: An experiment that would be nice would be to have a definition of the FunctionCallType in typescript, and have the schema being generated by zod; but I'm not sure how to populate the "describe" parts :/

ditman (commented):
Thanks for adopting this pattern in the Angular renderer! I left some small comments; the most "urgent" one is that you're adding some properties to web_core that don't seem to be defined in the specification of the basic catalog, and we need to decide what to do with those.

renderers/angular/src/v0_9/catalog/basic/icon.component.ts
josemontespg: I agree. But we need to also consider the implications of property renaming on optimized builds. I'm reverting the dot access to string access to preserve the prop renaming behavior.

Not in scope for this PR

renderers/angular/src/v0_9/catalog/basic/list.component.ts
josemontespg: I'm keeping the string access because I'm pretty sure that dot access here would break optimized builds for 1P.

Until we have a clearer picture on how to ensure that this code is safe with the closure compile property renaming, I think it is safer to keep this as string for now.

renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts
josemontespg: Mmm good point, these properties are not part of the spec. I'll keep their implementations untouched, and follow up on what to do in https://github.com/google/A2UI/issues/1303

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
josemontespg: Yes, good catch 👍. Removed the `any`

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.spec.ts
josemontespg: Nice catch! Switched to 'zod'

renderers/angular/src/v0_9/core/types.ts
josemontespg: Lol yes, this way the formatter is happy

renderers/angular/.gitignore
josemontespg: I don't know why this wasn't a problem before. I merged #1291 so this shouldn't be a problem anymore. Reverting this change.

renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts
josemontespg: Good point. I'm going to leave the component implementation as is, and revert these changes to basic_components.ts. We can track the fixing of the discrepancy in #1303

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
josemontespg: I agree that Base and Basic are pretty close in name. `Element` is not ideal, because it collides with the HTML Element, and could be confusing. Adding `A2ui` and `Angular` is a bit redundant, since this will be imported from `@a2ui/angular/` and will be part of the public API of that package.

I could rename `BaseCatalogComponent` to `CatalogComponent`. WDYT?

renderers/web_core/src/v0_9/schema/common-types.ts
josemontespg: We need to figure what is the best way to represent the source of truth for the schema and types. The current approach works pretty well with z.infer for Angular. Not sure yet about Lit and React.

ditman:
> Initially I used dot notation to access props, like props().foo, but then realized that this might cause a bug with build configurations that rename object properties. Reverted back to props()['foo'].
>
> Has there been property renaming issues before? Would it be safe to use dot notation?

Yes! IIRC the dot notation is safer than the bracket access. Bracket property names normally don't get renamed, so if your "props()" object ends up being minified (imagine that `props.foo` becomes `p.f`), `props()["foo"]` would break, but `props().foo` would work, unless minifiers have become much smarter :P

There's problems with property renaming and zod internally, I can point you to some chat rooms :P

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
ditman: Yep, sounds good. It'd be nice if all the classes across our renderers were named similarly, but we may end up with a generic catalog down the line so this is not a problem anymore. `CatalogComponent` and `BasicCatalogComponent` represents the class hierarchy well in this case!

josemontespg:
> Yes! IIRC the dot notation is safer than the bracket access. Bracket property names normally don't get renamed, so if your "props()" object ends up being minified (imagine that `props.foo` becomes `p.f`), `props()["foo"]` would break, but `props().foo` would work, unless minifiers have become much smarter :P

The props object is built at runtime in the binder, using the keys present in the JSON of the A2UI response, so it will be using keys that are not renamed. This is why I think using strings here is actually safer, because it will match the keys from the JSON.

For ergonomics we should aim to use dot notation. There are ways to disable property renaming for an object using `declare interface`, but I'm not sure if that works for arbitrarily deep objects. Until we figure this out, I think we will need to stick with strings.

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
josemontespg: Renamed BaseCatalogComponent to CatalogComponent 👍

ditman (approved):
Ship it! Shame we had to use stringly typed props :/

PR #1266, opened 2026-04-23
Migrate the rest of python unit tests to conformance tests

nan-yu:
# Description

_Replace this paragraph with a description of what this PR is changing or adding, and why. Consider including before/after screenshots._

_List which issues are fixed by this PR. For larger changes, raising an issue first helps reduce redundant work._

## Pre-launch Checklist

- [ ] I signed the [CLA].
- [ ] I read the [Contributors Guide].
- [ ] I read the [Style Guide].
- [ ] I have added updates to the [CHANGELOG].
- [ ] I updated/added relevant documentation.
- [ ] My code changes (if any) have tests.

If you need help, consider asking for advice on the [discussion board].

[CHANGELOG]: ../CHANGELOG.md
[CLA]: https://cla.developers.google.com/
[Contributors Guide]: ../CONTRIBUTING.md
[discussion board]: https://github.com/google/A2UI/discussions
[Style Guide]: ../CONTRIBUTING.md#coding-style

PR #1273, opened 2026-04-24
Implement glob pattern support for file path matching in A2uiCatalog and enable corresponding conformance tests

nan-yu:
# Description

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/agent_sdks/cpp/src/schema/catalog.cc
+++ b/agent_sdks/cpp/src/schema/catalog.cc
@@ -19,6 +19,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <algorithm>
+#include <regex>
 #include <queue>
 #include <set>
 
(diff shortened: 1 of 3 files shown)


nan-yu:
This PR is based on #1266 .  The actual change is in https://github.com/google/A2UI/pull/1273/changes/34ec768b20da238335f00a10b6e926898619a46c.

agent_sdks/cpp/src/schema/catalog.cc
```diff
@@ -146,7 +204,8 @@ std::string A2uiCatalog::load_examples(const std::string& path, bool validate) c
 
 void A2uiCatalog::validate_example(const std::string& full_path, const std::string& content) const {
     try {
-        nlohmann::json::parse(content);
+        auto parsed = nlohmann::json::parse(content);
+        (void)parsed;
     } catch (const std::exception& e) {
```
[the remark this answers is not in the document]
nan-yu: this will be changed later

agent_sdks/cpp/src/schema/catalog.cc
```diff
@@ -103,18 +104,75 @@ std::string A2uiCatalog::render_as_llm_instructions() const {
     return ss.str();
 }
 
+static std::string glob_to_regex(const std::string& glob) {
```
jacobsimionato, line 107: I worry a little about the security of these kinds of transformations, e.g. what if it ends up loading unintended files due to some error, but I guess there are tests, and this is only used by trusted server code to load files controlled by the same developer.

I guess if we were developing C++ in isolation we might just expose a regexp API to the application developer rather than glob, but I can see how having the uniformity with Python allows for the shared conformance test suites which make the maintenance possible for us.

jacobsimionato (approved):

agent_sdks/cpp/src/schema/catalog.cc
nan-yu: > I guess if we were developing C++ in isolation we might just expose a regexp API to the application developer rather than glob,

Totally agreed. That's why the conformance test was skipped initially. The glob pattern in C++ is purely for feature parity between the C++ SDK and the Python one.

I feel like it might be okay with the conformance tests.

PR #1280, opened 2026-04-26
Add Dart restaurant finder.

### LINKED ISSUE #870: Dart GenUI SDK should work with python restaurant_finder via a2a.
Create a Flutter client using [our A2A package](https://github.com/flutter/genui/tree/main/packages/genui_a2a) connected to the [Restaurant finder agent example](https://github.com/google/A2UI/tree/main/samples/agent/adk/restaurant_finder) from the A2UI repository.

Lit app that is tested with restaurant finder example: https://github.com/google/A2UI/tree/main/samples/client/lit/shell.

polina-c:
Closes https://github.com/flutter/genui/issues/870.

https://github.com/user-attachments/assets/a2ac7024-e56e-4502-9db4-29ab37d8eef3


samples/client/flutter/restaurant_finder/app/lib/src/session.dart
```diff
@@ -0,0 +1,173 @@
...
+  Future<void> _sendMessageToAgent(ChatMessage message) async {
+    if (_isRequesting) return;
+    _isRequesting = true;
+    _hasSentMessage = true;
+    _error = null;
+    _startLoadingAnimation();
+
+    // Clear existing surfaces on each new request.
+    for (final String id in [..._surfaceController.activeSurfaceIds]) {
+      _surfaceController.handleMessage(DeleteSurface(surfaceId: id));
```
jacobsimionato, line 135: I like this workaround for the current agent surface management behavior. I think this is as neat as we can make it!

samples/client/flutter/restaurant_finder/app/test/widget_test.dart
```diff
@@ -0,0 +1,31 @@
...
+// utility in the flutter_test package. For example, you can send tap and scroll
+// gestures. You can also use WidgetTester to find child widgets in the widget
+// tree, read text, and verify that the values of widget properties are correct.
+
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:restaurant_finder_client/main.dart';
+
+void main() {
+  testWidgets('Counter increments smoke test', (WidgetTester tester) async {
```
jacobsimionato: Maybe rename this test? "Main screen smoke test" or "Main screen loads successfully"?

samples/client/flutter/restaurant_finder/e2e_test/test/session_test.dart
```diff
@@ -0,0 +1,75 @@
...
+      // 1. Ask the agent to find restaurants.
+      await session.sendMessage(
+        'Find me 3 italian restaurants in New York.',
+      );
+
+      expect(session.hasSentMessage, isTrue);
+      expect(session.isRequesting, isFalse);
+      expect(session.error, isNull);
+      expect(
+        session.activeSurfaceIds,
```
jacobsimionato, line 48: Wooooo! This is the most complete end-to-end test we've had in all of Gen UI and A2UI so far!

Flutter application -> Gen UI SDK -> Agent server -> LLM -> Agent Server -> Gen UI SDK -> Flutter application !

samples/client/flutter/restaurant_finder/e2e_test/test/session_test.dart
```diff
@@ -0,0 +1,75 @@
...
+        isNotEmpty,
+        reason:
+            'Agent should respond with at least one surface listing the '
+            'restaurants.',
+      );
+
+      // 2. Ask the agent to book a table at one of them. This simulates the
+      // user tapping "Book Now" — both paths funnel into _sendMessageToAgent.
```
jacobsimionato, line 56: This is great for now, but it'd be awesome to eventually implement custom test infra in our library, so developers can do something like:
`surface.tester.findButton("Book now").triggerAction()`

Or else I guess they could achieve something similar with a widget test and the regular `.find` APIs.

pubspec.yaml
jacobsimionato, line 1: I know this was already here, but can we move this file to `samples/client/flutter` ?

Just trying to avoid clutter at the root of this monorepo, and make it easier to branch to a set of separate repos later!

This is non-blocking seeing as this wasn't introduced in this PR.

jacobsimionato (approved):
Looks great! I really like that this works around the eccentricities of the restaurant_finder server in a simple way. The Dart code is smaller and more elegant than I was expecting.

samples/client/flutter/restaurant_finder/e2e_test/test/session_test.dart
polina-c: Yes, there is a lot of room for improovement. We will work it out gradually as we get more and more tests.

pubspec.yaml
polina-c: Monorepo is very convenient thing for getting dependencies and running tests.
 Let's reorganize things when we know exactly what we are doing and how.

Issue #1296, opened 2026-04-28
TypeScript code contains hundreds of references to `any` type, which disables type analysis


andrewkolos:
If we enabled `@typescript-eslint/no-explicit-any: "error"` and ran `tsc --noImplicitAny`, **381 errors** would surface in production code that the currently are not flagged.

Each of these place where a type was either explicitly declared `any` (effectively turing off the type checker) or never specified at all (compiler couldn't infer). Each one is a hole through which a wrong type can flow into otherwise type-checked code.

Regardless of how many references of `any` turn out to actually be problematic, I suggest we turn on `no-explicit-any`, at least in non-test code. The whole purpose of the `any` super type, as far I am aware, is to allow gradual migration of JavaScript to TypeScript. This doesn't apply to this repo since the relevant code was written in TS to begin with. Even in places were the shape of an object cannot be known ahead of time, types like `Record` or `unknown` can be used.

I suggest we turn on the rule sooner rather than later. The longer we put it off, the more expensive the issue becomes. I can personally say I've seen agents spending several inferences troubleshooting a bug that would have been immediately flagged with typechecking in my very brief time with this repo.

## Numbers

The following was gathered via an agent, but the full methodology is included afterwards to facilitate verification.

| Category | Errors |
|---|---:|
| **Published library code** *(`renderers/web_core`, `renderers/{angular,lit,react}`, `renderers/markdown/markdown-it`)* | **146** |
| **All other non-test code** *(samples, tools, spec eval)* | **235** |
| **Total** | **381** |

Test/spec code was excluded, though of course this could mask bugs as well.

Of the 381: **2 are implicit-any** (compiler couldn't infer) and **379 are explicit `any`** (written deliberately in source). Implicit-any is more or less fine here since we have `strict: true` in all our tsconfigs.

### Library breakdown

| Package | Errors |
|---|---:|
| `renderers/web_core` | 99 |
| `renderers/angular` | 29 |
| `renderers/lit` | 12 |
| `renderers/react` | 4 |
| `renderers/markdown/markdown-it` | 2 |

## Suggested fix

Enable `@typescript-eslint/no-explicit-any: "error"` in every workspace and gate CI on it. Fix or add `// eslint-disable-next-line` each existing site as a tracked decision rather than silent permission. Focus on library code first; samples/demos/tests can follow.

[...]

PR #1312, opened 2026-04-30
[Angular] Enable strict types on props().children.raw

josemontespg:
The `props().children.raw` property was typed as `any`, bypassing the compiler type checks. With this PR I'm marking it as `unknown`, forcing call sites to appropriately handle the type at compile time, or narrow the type at runtime.

This `raw` property was being used to support the children template pattern. I modified the binder to instead provide a type safe `template` property.

## Changes
- Updated the catalog components to use this new template property. 
- Adopted `setComponentProps` for setting component inputs in tests, to make sure that the inputs used in tests match the actual prop type. Fixed tests that were using incorrect types.
- Created a test-utils.ts file to provide helper function to interact with catalog components in unit tests in a type safe manner.
- Deleted the renderers/angular/a2ui_explorer/src/app/card.component.ts because it depended on `raw`, but it was unused.

#1296


renderers/angular/a2ui_explorer/src/app/card.component.ts
ditman, line 1: Is this change unrelated to the fix?

renderers/angular/src/v0_9/public-api.ts
```diff
@@ -27,6 +27,7 @@
 export * from './core/a2ui-renderer.service';
 export * from './core/component-host.component';
 export * from './core/surface.component';
+export * from './core/catalog_component';
```
ditman, line 30: Why do we need to export the catalog_component through the public-api here?

ditman (commented):
This change looks breaking to people who have already created their own container widget with A2UI children, right? I think this needs to be marked as a breaking change in the changelog so we can rev the version of the package appropriately?

renderers/angular/a2ui_explorer/src/app/card.component.ts
josemontespg: This component broke with the changes introduced in the PR, and instead of fixing it, I'm deleting it because it is unused.

renderers/angular/src/v0_9/public-api.ts
josemontespg: It's a base class that all catalog components should extend. It provides type safe props and other fields that are part of the spec.

Extending this is optional, but encouraged.

josemontespg:
Thanks for the review! I updated the changelog and merged with main. Please take a look when you can.

renderers/angular/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - (v0_9) Re-style the v0_9 catalog components using the default theme from
   `web_core`. [#1166](https://github.com/google/A2UI/pull/1166)
+- [Breaking change] the `children` property value changed from `props().children().value() -> Child[]` to 
+`props().children().value() -> Children`. This provides a type safe wrapper to access the child list, path and component template ID.
```
ditman:
```suggestion
- [Breaking change] the `children` property value changed from `props().children().value() -> Child[]` to 
`props().children().value() -> Children`. This provides a type safe wrapper to access the child list, path and component template ID. [#1312](https://github.com/google/A2UI/pull/1312)
```

ditman (commented):
I'd link the PR to the CHANGELOG entry so people can dive deeper into the change if they care about it.

Also funny that the one above it says "Unreleased", afaik we've already released the angular restyling; I guess we didn't update the changelog file properly!

renderers/angular/CHANGELOG.md
josemontespg: Done

josemontespg:
Added the PR link to the changelog, and updated the heading to "0.9"

renderers/angular/src/v0_9/catalog/basic/column.component.ts
```diff
@@ -73,17 +73,14 @@ export class ColumnComponent extends BasicCatalogComponent<typeof ColumnApi> {
...
 
-  protected readonly children = computed(() => {
-    const raw = this.props()['children']?.value() || [];
-    return Array.isArray(raw) ? raw : [];
-  });
+  protected readonly children = computed(() => this.props()['children']?.value()?.children || []);
 
   protected readonly isRepeating = computed(() => {
-    return !!this.props()['children']?.raw?.componentId;
+    return !!this.props()['children']?.value()?.templateId;
```
andrewkolos: I noticed the row component code doesn't have the second `?.` operator that the column does here. I think it's redundant here since `value()` has a type of `Children`.
```suggestion
    return !!this.props()['children']?.value().templateId;
```

I feel like this kind of thing should be flagged by a linter rule; but, to my surprise, it appears we don't have a linter set up for the angular renderer 👀

renderers/angular/src/v0_9/catalog/basic/column.component.ts
```diff
@@ -73,17 +73,14 @@ export class ColumnComponent extends BasicCatalogComponent<typeof ColumnApi> {
...
   });
 
   protected readonly templateId = computed(() => {
-    return this.props()['children']?.raw?.componentId;
+    return this.props()['children']?.value()?.templateId;
```
andrewkolos:
```suggestion
    return this.props()['children']?.value().templateId;
```

renderers/angular/src/v0_9/catalog/basic/column.component.ts
```diff
@@ -97,6 +94,7 @@ export class ColumnComponent extends BasicCatalogComponent<typeof ColumnApi> {
   });
 
   protected getNormalizedPath(index: number) {
-    return getNormalizedPath(this.props()['children']?.raw?.path, this.dataContextPath(), index);
+    return getNormalizedPath(this.props()['children']?.value()?.path, this.dataContextPath(), index);
```
andrewkolos:
```suggestion
    return getNormalizedPath(this.props()['children']?.value().path, this.dataContextPath(), index);
```

renderers/angular/src/v0_9/core/component-binder.service.ts
```diff
@@ -25,6 +25,15 @@ export interface Child {
...
 }
 
+/** A collection of child components. */
+export interface Children {
+  children: Child[];
+  /** Optional component ID to be used as a template when instantiating the children. */
+  templateId?: string;
+  /** Optional path to the list of children. */
+  path?: string;
+}
```
andrewkolos: I have thoughts (and confusion) around how data binding in the Angular renderer works. That being said, since we are already planning on overhauling it to use `GenericBinder` from web_core, I don't think a breaking change to `BoundProperty` is needed here as it's going to be broken again by the GenericBinder` migration anyway:

Could the template metadata go onto a sibling field of `BoundProperty` instead of in the `value()` wrapper? Something like:

```ts
  interface BoundProperty<T = unknown> {
    value: Signal<T>;
    raw: unknown; // Aside: I feel like this should never have been visible this far down the rendering stack.
    onUpdate: (v: T) => void;
    template?: { id: string; path: string }; 
  }
```

And then for the type mapping in `InferredInterfaceToProps`:
```ts
[K in keyof InferredSchema]: K extends 'children' ? BoundProperty<Child[]> 
```

This should satisfy type checking inside the Angular renderer (though of course customers will probably still break if they access `raw`).

I do think that `BoundProperty` would have been better off written as a discriminating union[^1] with a `kind` field that can narrow down the shape according to whether we are dealing with a literal/binding/template, but—again—we are going to discard this with the `GenericBinder` migration anyway, so I'm okay this more conservative approach.

[^1]:https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions

andrewkolos (commented):
~~I'm still diving into the Angular renderer's code base to figure out whether a breaking change is actually necessary here (~~

edit: oops, forgot to delete the above from my review draft, ignore :)

renderers/angular/CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
-## Unreleased
+## 0.9
```
ditman: You don't need to update the "Unreleased" tag, but when we release this I'll go to what actually is published now and restructure this a little bit (when we prepare the package for release).

The thing is that the "re-style the v0_9 catalog" is actually released and should have a version number attached to it, but the props type safety is not released, and since it's a breaking change, the version for it needs to be 0.10.

josemontespg: Ah, got it. Updated the changelog. PTAL

renderers/angular/CHANGELOG.md
sarashed731-prog: Feedback

sarashed731-prog: Feedback

renderers/angular/src/v0_9/catalog/basic/column.component.ts
josemontespg: With the refactor you proposed, the `?` is no longer needed here. 

To apply a linter rule for this, we would need to configure eslint @typescript-eslint/no-unnecessary-condition. Created #1346 to track this

renderers/angular/src/v0_9/catalog/basic/column.component.ts
josemontespg: Good suggestion. No longer needed after the refactor.

renderers/angular/src/v0_9/core/component-binder.service.ts
josemontespg: Adding a field to BoundProperty is a great idea to keep this PR backwards compatible! Thanks for the suggestion. I applied this concept, please take a look when you can.

renderers/angular/src/v0_9/catalog/basic/column.component.ts
josemontespg: Done

josemontespg:
Applied changes, ready for another review when you can @ditman and @andrewkolos

renderers/angular/src/v0_9/catalog/basic/column.component.ts
andrewkolos: Thanks for filing! I think just adding a basic linter config (ideally similar to whatever we do for our other TS code bases) to the Angular renderer alone would be a great change.

renderers/angular/CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 ## Unreleased
 
-- (v0_9) Re-style the v0_9 catalog components using the default theme from
-  `web_core`. [#1166](https://github.com/google/A2UI/pull/1166)
-- (v0_9) Improve type safety of `props()` in Catalog components. Custom catalog
+- Improve type safety of `props()` in Catalog components. Custom catalog
   components should extend the base class `CatalogComponent` from
   `import {CatalogComponent} from '@a2ui/web_core/v0_9/'` or implement the
   interface `CatalogComponentInstance`. [#1320](https://github.com/google/A2UI/pull/1320)
+- `children` props now expose a typed `value` signal, and have a new `template` field. This provides a type safe structure to access the child list, path and component template ID. [#1312](https://github.com/google/A2UI/pull/1312/changes#top)
```
andrewkolos: Technically, this change is still breaking (necessarily so to avoid `any`). Consider adding something like

```
- [Breaking] `BoundProperty.raw` is now `unknown` instead of `any`. Recommended migration: replace `raw` access with typed sibling fields where available (e.g. use the new `template` field instead of `raw.componentId`/`raw.path`).
- `BoundProperty<T = unknown>`: the default generic is now `unknown` instead of `any`. Recommended migration: provide explicit type arguments (e.g. `BoundProperty<string>`) at usage sites that previously relied on the default. Code typed via `ComponentApiToProps<Api>` is unaffected.
```

andrewkolos: This makes `value` sound like a new thing in the case of `"children"` being used as the key. Should we make it more clear that we are just correcting the type? Consider

```
- `props()['children']?.value()` is now typed `Child[]` (was `Child`, despite runtime always returning an array). [#1312]```

andrewkolos (approved):
LGTM with nits. Thank you for the fix and cleanup!

josemontespg:
Marked the children prop change as breaking in the changelog.

renderers/angular/CHANGELOG.md
josemontespg: Nice improvement, thanks!

josemontespg: I like it, thank you!

PR #1313, opened 2026-04-30
Fix: Pin dompurify to 3.3.1 in markdown-it package

josemontespg:
Pin the dompurify dependency in the markdown-it package to a stable 3.3.1 to avoid installation errors caused by caret matching versions not yet mirrored in internal registries.

Literal error encountered:
```
npm error code ETARGET
npm error notarget No matching version found for dompurify@3.4.0.
```


ditman (approved):
Uuuugghhhh

PR #1314, opened 2026-04-30
Refactor: Remove any from surface model generic fallback

josemontespg:
Allows the SurfaceModel generic fallback to default to ComponentApi, removing targeted any usage in the Angular component host.

Part of #1296


renderers/angular/src/v0_9/core/component-host.component.ts
```diff
@@ -141,7 +141,7 @@ export class ComponentHostComponent {
   }
 
   private initializeComponent(
-    surface: SurfaceModel<any>,
+    surface: SurfaceModel,
```
ditman, line 144: Was this `<any>` erasing the type of the `surface` computed in the setupComponent method?

ditman (commented):
Small question about removing an `<any>` from the code.

renderers/angular/src/v0_9/core/component-host.component.ts
josemontespg: Yes, although the type of `surface` in the setupComponent method is also not typed very well. The `any` here was just not needed. I'll improve the generic types of the Surface and Catalog in an upcoming PR

renderers/angular/src/v0_9/core/component-host.component.ts
```diff
@@ -24,12 +24,12 @@ import {
   input,
   effect,
 } from '@angular/core';
-import {NgComponentOutlet} from '@angular/common';
-import {ComponentContext, ComponentModel, SurfaceModel, Subscription} from '@a2ui/web_core/v0_9';
-import {A2uiRendererService} from './a2ui-renderer.service';
-import {AngularCatalog} from '../catalog/types';
-import {ComponentBinder} from './component-binder.service';
-import {BoundProperty} from './types';
+import { NgComponentOutlet } from '@angular/common';
```
ditman: (Is this formatting change caused by some new config? Are other files consistent with this?)

renderers/angular/src/v0_9/core/component-host.component.ts
ditman: Same question :P

ditman (commented):
This LGTM, I have two very small questions before this gets auto-submitted :P

renderers/angular/src/v0_9/core/component-host.component.ts
josemontespg: I don't know, I've seen both styles in the codebase. My IDE prefers a space between { }. I created https://github.com/google/A2UI/issues/1337 to track having a consistent formatter and linter in the repo, with CI enforcement, so that we can reduce noise during code review.

For now, I think we should ignore these white space changes, they are not worth human time IMO. Eventually we will format and lint the entire repo and get this to a consistent state.

renderers/angular/src/v0_9/core/component-host.component.ts
josemontespg: Sorry I forgot to post the response above 😅

ditman (approved):
Hmmm, this change probably needs a changelog update. Before `SurfaceModel` without a type would be of type `SurfaceModel<any>` but now it's `SurfaceModel<ComponentApi>`, but that shouldn't break users of the SurfaceModel unless they were using a very wishy-washy type to define their surfaces.

I'm not sure about the breakingness of the angular renderer though, since you didn't have to change any test code, it's probably not that breaking! (or completely untested :P)

PR #1317, opened 2026-05-01
[ci] Improve renderer publishing scripts.

ditman:
# Description

This PR modernizes the renderer publishing scripts:

* General
  * Add a `maybeRunCommand` function to handle dry-run cases.
  * Add a shared `ansi` colors object to colorize terminal output.
  * Make the `scripts` directory a private package so we can run `npm test` on it. (Also maybe in the future this can be published separately if we want to reuse the logic across repos)
* `upload_manifest.mjs`
  * Fixes the malformed format of the manifest that was being generated that didn't match the internal requirements.
  * Allows the user to pass specific package names to upload. Default to "publish_all: true" if none are set.
  * Adds a `dry-run` flag that is true by default. Use `maybeRunCommand` to run the upload manifest command.
  * Adds some tests to the script to assert the format of the manifest, and the dry-run functionality.
* `publish_npm.mjs`
  * Adopt `parseArgs` and rename `--packages` to repeatable `--package`.
  * Enable `dry-run` by default; remove `--test-only` and `--yes` flags.
  * Extract core logic into separate functions (`checkNpmVersions`, `buildAndTestPackages`, etc.).
  * Use `maybeRunCommand` and `ansi` colors from `workspace.mjs` to reduce duplication.
  * Resolve package names to objects early to simplify function signatures.
  * Adds some more test cases to test the dry-run behavior and others.

## Testing

* Tests run, but also I run the script to publish markdown-it 0.0.4, and now it's [here](https://www.npmjs.com/package/@a2ui/markdown-it/v/0.0.4) (we should bump that script to a "closer to 1.0.0 version" and settle on that, we've experimented enough with it)

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/scripts/upload_manifest.mjs
+++ b/renderers/scripts/upload_manifest.mjs
@@ -15,18 +15,18 @@
  * limitations under the License.
  */
 
-import { writeFileSync } from 'node:fs';
-import { join } from 'node:path';
-import { getPackageGraph, ROOT_DIR, ansi, maybeRunCommand } from './lib/workspace.mjs';
-import { parseArgs } from 'node:util';
-import { fileURLToPath } from 'node:url';
+import {writeFileSync} from 'node:fs';
+import {join} from 'node:path';
+import {getPackageGraph, ROOT_DIR, ansi, maybeRunCommand} from './lib/workspace.mjs';
+import {parseArgs} from 'node:util';
+import {fileURLToPath} from 'node:url';
 
 // Configuration - adjust these as needed for your environment
 const GCS_URI =
   process.env.A2UI_NPM_MANIFEST_GCS_URI ||
   'gs://oss-exit-gate-prod-projects-bucket/a2ui/npm/manifests';
 
-const { yellow, red, green, reset } = ansi;
+const {yellow, red, green, reset} = ansi;
 
 /**
  * Generates and uploads the npm manifest to GCS.
@@ -52,7 +52,7 @@ export async function main(args, mocks = {}) {
     },
   };
 
-  const { values } = parseArgs({ args, options, allowNegative: true });
+  const {values} = parseArgs({args, options, allowNegative: true});
   const packagesToPublish = values.package;
   const isDryRun = values['dry-run'];
 
@@ -77,9 +77,9 @@ export async function main(args, mocks = {}) {
       publishing_groups: [
         {
           namespace: '@a2ui',
-          packages: resolvedPackages.map(name => ({ name }))
-        }
-      ]
+          packages: resolvedPackages.map(name => ({name})),
+        },
+      ],
     };
   } else {
     // A manifest to publish all packages.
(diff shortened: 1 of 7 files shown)


renderers/scripts/upload_manifest.mjs
```diff
@@ -18,42 +18,110 @@
...
+        throw new Error(`Package "${name}" not found in renderers directory.`);
+      }
+      return pkg.name.replace('@a2ui/', '');
+    });
+    // A manifest to only publish the specified packages.
+    manifest = {
+      publish_all: false,
+      publishing_groups: [
+        {
+          namespace: '@a2ui',
```
ditman, line 79: This could be made smarter by looking at the prefix of the package name and setting the appropriate namespace, instead of hardcoding it, but I don't think we should be publishing non-namespaced packages from this repo.

renderers/scripts/upload_manifest.test.mjs
```diff
@@ -0,0 +1,98 @@
...
+    assert.strictEqual(manifest.publish_all, false, 'Should set publish_all: false');
+    assert.ok(manifest.publishing_groups, 'Should have publishing_groups');
+    assert.strictEqual(manifest.publishing_groups.length, 1);
+    assert.strictEqual(manifest.publishing_groups[0].namespace, '@a2ui');
+    
+    const packages = manifest.publishing_groups[0].packages;
+    assert.strictEqual(packages.length, 3);
+    assert.strictEqual(packages[0].name, 'angular');
+    assert.strictEqual(packages[1].name, 'lit');
+    assert.strictEqual(packages[2].name, 'react');
```
ditman, line 79: This test is dependent of the getPackageGraph output, which we maybe should mock.

renderers/scripts/upload_manifest.mjs
```diff
@@ -18,42 +18,110 @@
...
 import { getPackageGraph, runCommand, ROOT_DIR } from './lib/workspace.mjs';
+import { parseArgs } from 'node:util';
+import { fileURLToPath } from 'node:url';
 
 // Configuration - adjust these as needed for your environment
 const GCS_URI = process.env.A2UI_NPM_MANIFEST_GCS_URI || 'gs://oss-exit-gate-prod-projects-bucket/a2ui/npm/manifests';
 
-const graph = getPackageGraph();
-const renderers = Object.values(graph).filter(p => p.dir.includes('/renderers/'));
+// Some ANSI escape codes to colorize the output.
```
josemontespg: Nice

renderers/scripts/upload_manifest.mjs
josemontespg: Good enough for now. We can tweak if needed in the future

josemontespg (approved):
Thank you for doing this! Could you also update the publish docs with the full commands? in renderers/docs/web_publishing.md

ditman:
> Could you also update the publish docs with the full commands? in renderers/docs/web_publishing.md

Yes! I thought of that last night after posting this, let me update the commands a little bit. Also, I'm going to update the other script to handle separate packages and dry-run in the same way as this one, so usage is consistent between them (right now, dry-run defaults to false in the other script, and when passing multiple packages it's a comma-separated list of values, rather than passing "--package" multiple times)

renderers/scripts/upload_manifest.mjs
ditman: Thanks!!

ditman:
OK, I've tweaked both scripts, and the deployment docs.

The only big difference remaining is that the publish_npm.mjs does NOT default to "all packages" when none are specified (making the `--package` parameter pretty much mandatory). I also added a few extra tests.

ditman:
Please PTAL @josemontespg!

josemontespg (approved):
LGTM , thanks for updating the docs!

ditman:
One last time, @josemontespg, I've applied the new formatting here:

* https://github.com/google/A2UI/pull/1317/commits/7717a6631796a0cc09a855082c6368a314a508ad

josemontespg (approved):
LGTM 🎉

ditman:
I'm merging this as soon as checks pass!

Issue #931, opened 2026-03-22
Centralize Basic Catalog implementation using Web Components via Entrypoint Isolation


jacobsimionato:
# Centralize Basic Catalog implementation using Web Components via Entrypoint Isolation

Currently, the A2UI Basic Catalog (Text, Button, Row, Column, TextField, etc.) is implemented separately in every framework-specific renderer (React, Angular). This duplication leads to visual inconsistencies, repeated bug fixes, and significant maintenance overhead.

This proposal centralizes the visual implementation of the 18 Basic Catalog components into standard **Web Components (Custom Elements)** authored with **Lit**. The framework-specific renderers will then act as thin "adapters" that wrap these shared components, allowing the styling and DOM structure to be written exactly once.

## Architecture: Centralized Visuals, Thin Adapters

The architecture shifts the responsibility of painting pixels down to the `web_core` layer, while keeping the structural reconciliation in the framework layer.

1. **Shared Visuals in `web_core`:** We will add `lit` as a dependency to `@a2ui/web_core` and implement the Basic Catalog components as standard Custom Elements (e.g., `<a2ui-button>`, `<a2ui-text-field>`).
2. **Property-First Data Passing:** Frameworks will pass the fully resolved A2UI `props` object directly to the Web Component as a JavaScript property, avoiding expensive HTML attribute serialization.
3. **Recursive Slotting:** Structural components (like `Row` or `Column`) will use standard HTML `<slot>` elements. The host framework (React/Angular) will continue to manage the lifecycle of children, rendering them *into* the slots of the shared components.

### 1. The Shared Lit Component (`web_core`)
The Lit component handles the Shadow DOM encapsulation and translates CSS Custom Properties for theming. It emits standard `CustomEvent`s for interactions.

```typescript
import { LitElement, html, css } from 'lit';
import { customElement, property } from 'lit/decorators.js';

@customElement('a2ui-button')
export class A2uiButton extends LitElement {
  // Accepts the resolved A2UI props object directly
  @property({ type: Object }) props: any = {};

  static styles = css`
    :host { display: inline-block; }
    button { 
      padding: 8px 16px; 
      /* Theme injection via CSS Variables */
      background-color: var(--a2ui-primary-color, #007bff);
      color: var(--a2ui-on-primary-color, white);
      border: 1px solid var(--a2ui-border-color, #ccc);
      border-radius: var(--a2ui-border-radius, 4px);
      cursor: pointer;
    }
  `;

  render() {
    return html`
      <button 
[...]

PR #1320, opened 2026-05-01
Add strict type to enforce that catalog components implement the required properties

josemontespg:
This PR replaces `Type<any>` with `Type<CatalogComponentInstance>` in the `AngularComponentImplementation` interface to enforce that catalog components implement the required properties: `props`, `surfaceId`, `componentId`, and `dataContextPath`. 

By doing this, the compiler found that  the `CustomSliderComponent` was not correctly implementing props. It used a `Record<string, BoundProperty>`, but should be `Signal<Record<string, BoundProperty>>`.

By having strict types, we can make sure that devs creating catalog components correctly implement these required properties. As an added bonus, they get typed access to props().


renderers/angular/a2ui_explorer/src/app/demo-catalog.ts
```diff
@@ -18,8 +18,17 @@ import { Injectable } from '@angular/core';
 import { z } from 'zod';
 import { BasicCatalogBase, BASIC_FUNCTIONS } from '@a2ui/angular/v0_9';
 import { CustomSliderComponent } from './custom-slider.component';
-import { AngularComponentImplementation } from '@a2ui/angular/v0_9';
-import { createFunctionImplementation, FunctionImplementation } from '@a2ui/web_core/v0_9';
+import { ComponentApi, createFunctionImplementation, FunctionImplementation } from '@a2ui/web_core/v0_9';
+
+export const customSliderApi = {
```
ditman: Should this API be exported from `custom-slider.component.ts`, and then imported here, so the API and the component implementation stay close?

renderers/angular/a2ui_explorer/src/app/demo-catalog.ts
```diff
@@ -58,7 +57,7 @@ export class DemoCatalog extends BasicCatalogBase {
     super({
       id: 'https://a2ui.org/specification/v0_9/basic_catalog.json',
       components: {},
-      extraComponents: [customSliderApi],
+      extraComponents: [{ ...customSliderApi, component: CustomSliderComponent }],
```
ditman: Actually, the `customer-slider.component.ts` could define exactly this component object:

```ts
export const customSliderComponent = {
  ...customSliderApi,
  component: CustomSliderComponent,
};
```

And then this demo-catalog gets very much simplified, because the only thing we're exporting from the custom slider component file is this exact definition.

(We wouldn't need to export neither the API nor the Component class for this to work, I think?)

ditman (commented):
This looks good to me. Can you please update the CHANGELOG to describe this change? This needs to be released formally IMO.

renderers/angular/a2ui_explorer/src/app/demo-catalog.ts
josemontespg: It's a good idea to keep the API and implementation close in code. For custom components I think it is the way to go. I'll do this for the custom slider.

For our internal catalogs, there is an idea to have a central place for the component API declarations consumed by each of the web frameworks, then each framework implements the API. This design pushes the API away from the component implementations unfortunately.

Another idea is to implement the basic catalog in a portable way, which can then be adapted for each of the frameworks with a utility. This way, we could have a single implementation for the basic catalog, and would allow us to co-locate the API with the implementation.

ditman: > Another idea is to implement the basic catalog in a portable way, which can then be adapted for each of the frameworks with a utility. This way, we could have a single implementation for the basic catalog, and would allow us to co-locate the API with the implementation.

Yes, the "portable" catalog idea has been floated a few times, for example: https://github.com/google/A2UI/issues/931

renderers/angular/a2ui_explorer/src/app/custom-slider.component.ts
```diff
@@ -59,3 +70,5 @@ export class CustomSliderComponent extends CatalogComponent<typeof customSliderA
     this.props()['value']?.onUpdate(val);
   }
 }
+
+export const CustomSliderComponentDeclaration = { ...customSliderApi, component: CustomSliderComponent };
```
ditman: Please, add a new line at the end of the file.

Also: do we need to export CustomSliderComponent and the customSliderApi from this same file, or is the CustomSliderComponentDeclaration enough?
(Also also, being an object instance... its name should start with a lowercase `c`?)

renderers/angular/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 - (v0_9) Re-style the v0_9 catalog components using the default theme from
   `web_core`. [#1166](https://github.com/google/A2UI/pull/1166)
+- (v0_9) Improve type safety of `props()` in Catalog components. Custom catalog
+  components should extend the base class `CatalogComponent` from
+  `import {CatalogComponent} from '@a2ui/web_core/v0_9/'` or implement the
+  interface `CatalogComponentInstance`.
```
ditman: Thanks for updating the changelog, please add the PR link here!

ditman (commented):
Two minor nits!

renderers/angular/a2ui_explorer/src/app/custom-slider.component.ts
josemontespg: Added line, and renamed the const. About the export, we don't need to export the API object anymore, but I'm not so sure about the angular component. I think it might be needed by the compiler in some way. Keeping it just in case it is needed.

renderers/angular/a2ui_explorer/src/app/demo-catalog.ts
josemontespg: Done

renderers/angular/CHANGELOG.md
josemontespg: Done

ditman (approved):

PR #1322, opened 2026-05-01
Fix duplicate surfaces error in restaurant sample app streaming

josemontespg:
This PR fixes an issue where the client receives duplicate `createSurface` messages for the same `surfaceId` in streaming mode, triggering the error 'Surface default already exists'.

## Changes

1. Pass a `useStreaming` flag from the client to the server via `DataPart` to explicitly control streaming behavior.
2. Update the restaurant agent backend to return an empty list of parts `[]` at the end of the stream when `useStreaming` is enabled and parts were already streamed. This avoids sending duplicate messages.
3. Disabled streaming in the Lit shell app by default to match its capabilities and verify the non-streaming path.

This approach avoids protocol violations and provides a cleaner solution.

Manually verified that the restaurant sample app works correctly for Angular, Lit and React.


renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -279,9 +279,8 @@ export class MessageProcessor<T extends ComponentApi> {
       throw new A2uiStateError(`Catalog not found: ${catalogId}`);
     }
 
-    if (this.model.getSurface(surfaceId)) {
-      throw new A2uiStateError(`Surface ${surfaceId} already exists.`);
-    }
+    // We don't throw here if surface already exists to support cumulative payloads in streaming mode.
+    // SurfaceGroupModel.addSurface will handle duplicates by ignoring them and logging a warning.
```
ditman: Hadn't we discussed already that duplicate surfaces should throw when received?

Should the fix be in the example app that is attempting to process duplicate messages, instead of here?

josemontespg: Changed the approach, please take a look when you can. Thanks!

samples/client/lit/shell/client.ts
```diff
@@ -81,6 +81,12 @@ export class A2UIClient {
       ];
     }
 
+    parts.push({
+      kind: 'data',
+      data: {useStreaming: false},
+      mimeType: 'application/json',
+    } as Part);
+
     const response = await client.sendMessage({
```
ditman, line 90: The client-side API for this is a little bit weird; I would expect `client.sendMessage` to have an `useStreaming` option (and internally we'd handle the protocol specifics to enable/disable streaming on the responses).

I think this is enough for now, but maybe we can create an issue and revisit this later?

Since this PR is mostly python changes, I'll defer to @nan-yu for the final approval. Other than that, LGTM!

nan-yu: The python changes look good.

Regarding the streaming flag, I'm wondering whether https://github.com/google/A2UI/blob/main/specification/v0_9/json/client_capabilities.json would be a better place. Besides `supportedCatalogIds` and `inlineCatalogs`, we can add a `supportStreaming` boolean. It sounds like a more explicit protocol between client and server.

WDYT, @gspencergoog ?

nan-yu (approved):
The changes in this PR looks good. We can follow up in a separate PR if we decided to move the streaming flag to a different place.

josemontespg:
Thanks for the reviews! 🙏

Issue #932, opened 2026-03-22
Create test to reproduce zod type inference property renaming issues in google monorepo


jacobsimionato:
See b/495155654 for details

ditman:
@josemontespg's findings attempting this: https://github.com/google/A2UI/pull/1331.

polina-c:
This issue should be considered for a P1 priority together with other P2 issues in the next planning cycle after we burn through the triage queue.

josemontespg:
This is already implemented in 1P code, see http://b/495155654 for details.

ditman:
We added some tests in github too! https://github.com/a2ui-project/a2ui/pull/1902

PR #1326, opened 2026-05-02
Use strings for the catalog component names.

josemontespg:
When testing A2UI in 1P, I found that the catalog component names were being renamed by the compiler, causing the renderer to not find the components by name. 

This makes the A2UI core code more resilient to property renaming issues. I manually tested that this works in 1P code with `jsmode=o`.

Long term solution is to not rely on `Object.entries()` in the codebase, because this doesn't play well with property renaming.  Created #1327 to track


ditman:
Related: https://github.com/google/A2UI/issues/932

Is there any way that we can add a test in the open source repo to ensure that we don't break this by accident later, or that we don't need to be specially vigilant about it?

Also cc/ @sugoi-yuzuru

josemontespg:
> Is there any way that we can add a test in the open source repo to ensure that we don't break this by accident later, or that we don't need to be specially vigilant about it?

That would be ideal, specially if we can include a test like this in CI.

Potentially, we could use https://www.npmjs.com/package/google-closure-compiler to replicate how property renaming works in 1P, but I'm not sure if this would produce exactly the same results. Maybe it would be enough to catch most prop renaming bugs.

We have an integration test for 1P prop renaming in http://b/495155654#comment2

josemontespg:
I tried creating a test with the the closure compiler, but I kept running into compilation issues when using property renaming. Added my findings in https://github.com/google/A2UI/pull/1331.

ditman:
Didn't this land as part of another PR? I though I had seen this code in a bigger PR but I may be mistaken! (x_x)

josemontespg:
> Didn't this land as part of another PR? I though I had seen this code in a bigger PR but I may be mistaken! (x_x)

Nope, only this PR so far. Might need more fixes related to this in the future.

renderers/angular/src/v0_9/catalog/basic/basic-catalog.ts
```diff
@@ -60,26 +60,29 @@ import {FunctionImplementation} from '@a2ui/web_core/v0_9';
 
 /**
  * The set of default Angular implementations for each component in the basic catalog.
+ * Using string literals as keys, to survive property renaming, as these names need to match the JSON payload.
  */
+// Ignore Prettier to preserve quoted keys, needed to survive property renaming.
+// prettier-ignore
```
ditman, line 66: Can we configure prettier to do "preserve" [quote props](https://prettier.io/docs/options#quote-props) so we don't need to completely disable prettier for the file?

josemontespg: It is possible. This // prettier-ignore statement only applies to the AST node of the next line, so it only disables lint checks for the const. It's not ideal, because it not only disables quote props, but all other lints.

I'd like to keep it as is for now. This is not a long term solution, because we will need to refactor this approach to fix #1327

ditman (approved):
Hit it then!

PR #1330, opened 2026-05-04
fix: pin @google/genai to 1.22.0 in lit shell

josemontespg:
This PR pins the version of @google/genai to 1.22.0 in samples/client/lit/shell/package.json to avoid installation failures caused by corporate proxy resolution issues with newer versions.

Literal error encountered during npm install:
```
npm error code ETARGET
npm error notarget No matching version found for @google/genai@1.50.1.
npm error notarget In most cases you or one of your dependencies are requesting
npm error notarget a package version that doesn't exist.
```


ditman:
Are these pins something that's going to get fixed by itself tomorrow? (when some cache refreshes?)

josemontespg:
> Are these pins something that's going to get fixed by itself tomorrow? (when some cache refreshes?)

I'm not sure, I've been getting these errors and I would guess others in the team as well. Pinning it sounds like a good workaround until we figure out a way to reliably use ^ notation for these internal packages.

ditman:
@google/genai 1.51.0 is already available internally, I don't think we need this one? :P

josemontespg:
> @google/genai 1.51.0 is already available internally, I don't think we need this one? :P

Mmm yes, closing for now. Is this going to continue to happen whenever there is a patch version released?

ditman:
> Mmm yes, closing for now. Is this going to continue to happen whenever there is a patch version released?

@josemontespg Probably yes, the thing is that I'm not sure why this is resolving a version from the future that is not already imported. Maybe that's an "airlock" bug (?) (or maybe resolution happens in public NPM, but then the download comes from the internal cache?)

PR #1331, opened 2026-05-05
Draft: Angular Closure Compiler Investigation

josemontespg:
# Findings: Angular Explorer App with Google Closure Compiler

This document summarizes the findings and challenges encountered while attempting to compile the Angular Explorer application using the Google Closure Compiler with advanced optimizations and property renaming.

## Objective
Compile the Angular explorer app using `google-closure-compiler` with property renaming on (`ADVANCED` optimizations) and verify it in an integration test.

### 1. Environment Setup and Dependencies
- **Java Requirement**: Closure Compiler requires Java. Resolved by installing `openjdk` via Homebrew and adding it to the `PATH`.
- **Input**: Initially targeted the production build output of Angular CLI (`dist/a2ui_explorer/browser/*.js`), which was already minified, making debugging difficult.

### 2. Resolving Initial Compiler Errors
- **Undeclared Variables**: The compiler failed on `localStorage` being undeclared. Resolved by creating a custom `externs.js` file.
- **Charset Issues**: The compiler mangled the `ɵ` character (used in Angular Ivy properties like `ɵcmp`) into `?` when defaulting to `US_ASCII` output. Resolved by forcing `--charset UTF-8`.

### 3. Advanced Optimizations (`ADVANCED`) Challenges
- **Error**: `NG0906: The xu is not an Angular component, make sure it has the @Component decorator.`
- **Root Cause**: In `ADVANCED` mode, Closure Compiler aggressively renames or removes static properties like `ɵcmp` on component classes because it cannot trace their dynamic usage by Angular's runtime (which often uses string lookups).
- **Attempted Fixes**:
    - Added externs to protect `ɵcmp` on `Function.prototype` and `Object.prototype`. This did not resolve the error, suggesting properties were still being stripped or renamed in a way the externs didn't cover.
    - Attempted to use `tsickle` to generate Closure-compatible annotations. Installation failed due to a peer dependency conflict with TypeScript 5.9 (tsickle expected ~4.7.2). Forced installation revealed no CLI tool was available in that version, requiring a complex Node API integration.

### 4. Working Solution (`SIMPLE`)
- **Result**: Compiling with `--compilation_level SIMPLE` succeeded, and the application loaded successfully without runtime errors.
- **Trade-off**: `SIMPLE` mode does not perform property renaming or aggressive dead code elimination, resulting in a larger file size (~1.84 MB vs ~462 KB in `ADVANCED` mode).

## Recommendations for Future Work
- To achieve successful `ADVANCED` compilation, a proper integration with a tool like `tsickle` or a compatible Angular-Closure builder is needed to ensure metadata properties are preserved or accessed via property access rather than strings.
- For now, `SIMPLE` mode serves as the only working configuration for minifying the app with Closure Compiler in this environment.

### DIFF
--- a/renderers/angular/compile_explorer.js
+++ b/renderers/angular/compile_explorer.js
@@ -1,0 +1,32 @@
+const fs = require('fs');
+const path = require('path');
+const closureCompiler = require('google-closure-compiler-js');
+
+const inputPath = path.join(__dirname, 'dist/a2ui_explorer/browser/main-N7JESC74.js');
+const outputPath = path.join(__dirname, 'dist/a2ui_explorer/browser/main-compiled.js');
+
+console.log('Reading input file...');
+const src = fs.readFileSync(inputPath, 'utf8');
+
+console.log('Compiling...');
+const flags = {
+  jsCode: [{src: src}],
+  compilationLevel: 'SIMPLE',
+};
+
+const out = closureCompiler(flags);
+
+if (out.errors && out.errors.length > 0) {
+  console.error('Compilation errors:');
+  out.errors.forEach(err => console.error(err));
+  process.exit(1);
+}
+
+if (out.warnings && out.warnings.length > 0) {
+  console.warn('Compilation warnings:');
+  out.warnings.forEach(warn => console.warn(warn));
+}
+
+console.log('Writing output file...');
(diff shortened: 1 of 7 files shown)


josemontespg:
Was not able to get it to work. Closing for now

ditman:
Related: https://github.com/google/A2UI/issues/932

PR #1333, opened 2026-05-05
feat(eval): introduce A2UI evaluation framework based on Inspect AI

gspencergoog:
## Summary

This PR introduces a complete evaluation framework for the A2UI project, built on top of the Inspect AI ecosystem. The framework is designed to verify that LLMs can correctly generate valid, contextually aware, and renderable A2UI JSON payloads based on high-level user instructions.

## Changes
- **Core Framework**:
  - Introduced the `eval` directory with Inspect AI configuration.
  - Added datasets, solvers, and scorers tailored for A2UI protocol validation.
- **SDK Integration**:
  - Leveraged the A2UI Python SDK for robust system prompt generation (`A2uiSchemaManager`) and response validation (`A2uiValidator`).
  - Used the SDK's parser to handle JSON extraction and healing.
- **Dataset & Security**:
  - Added evaluation scenarios in YAML format.
  - Configured **Transcrypt** for transparent encryption of datasets at rest to prevent base model contamination in open-source environments.
- **Ergonomics**:
  - Set up the project using `uv` with the SDK installed in editable mode, eliminating the need for manual `PYTHONPATH` environment variables.
- **Documentation**:
  - Added a comprehensive technical design document `DESIGN.md`.
  - Added a user-friendly `README.md` with setup and execution instructions.

## Impact & Risks
This is a new, isolated testing framework and does not impact any runtime application code. It provides a foundation for continuous evaluation of agent performance.

## Testing
- Unit tests are included in the `tests/` directory and pass successfully.
- Verified execution with `uv run inspect eval tasks.py` yielding successful results on sample sets.


jiahaog (approved):
lgtm if lgt Nan :)

nan-yu:
That's a great start. I really like all the details of the eval result provided by `inspect-ai` out of the box. 

Based on our offline conversation, here is a summary of the initial feedback
- The current `duration` column tracks the overall runtime (inference + evaluation). To properly benchmark the inference SDK's performance, we need isolated metric for the inference step only.
- Detailed token counts are visible under the `MODELS` tab, but the main dashboard view seems to show the overall usage. I don't think we care that much on the tokens consumed in the evaluation step. It would be good to show inference tokens (input and output) only.
- The text-based model grader is currently too rigid. For example, it marked an answer as `Incorrect` purely due to a capitalization discrepancy ("SCREEN" vs "Screen").
  - In the short term, we can define a global and reusable set of prompt constraints (e.g. "Variations in capitalization and punctuation are OK") and append them to all target descriptions.
  - In the long term, we could explore using expected reference images as ground truth, using an LLM-as-a-judge to evaluate which visual/structural discrepancies are critical vs. negligible.
- The `ANSWER` column is empty. Would it be possible to add the generated A2UI messages there?

nan-yu (approved):

gspencergoog:
> The ANSWER column is empty. Would it be possible to add the generated A2UI messages there?

No, it's not empty: it just isn't formatted great in the table.  I think it's a bug in the CSS for Inspect.  It looks empty because the "Explanation" column is reduced to one character wide, and then it makes that row super tall.  If you inspect the page with web tools and change the max width of the ANSWER column to 500px or something, then it formats much nicer.  I was thinking of filing a PR for Inspect AI that fixed this.

gspencergoog:
> In the short term, we can define a global and reusable set of prompt constraints (e.g. "Variations in capitalization and punctuation are OK") and append them to all target descriptions.

I added these blanket instructions for the LLM Judge (the part before "Notes for grading" is the default prompt for the judge):

```
After assessing the submitted answer, reply with 'GRADE: $LETTER' (without quotes) where LETTER is one of C, P or I.  Please choose ONE option for the grade: either C for correct answers, P for partial credit, or I for incorrect answers.

For example, after reviewing a correct answer you might write 'GRADE: C' or after reviewing an incorrect answer you might write 'GRADE: I'.

First, write out in a step by step manner your reasoning about the criterion to be sure that your conclusion is correct. Avoid simply stating the correct answers at the outset. Then, end with your answer formatted as 'GRADE: $LETTER' (without quotes) where LETTER is one of C, P or I.

Notes for grading:
1. Variations in capitalization, punctuation, and minor spacing differences should be considered acceptable as long as the semantic intent and required components are present.
2. Unless a specific vertical or horizontal order is explicitly requested in the task, variations in the order of components within a container should be considered acceptable.
3. Generated component IDs do not need to match any specific pattern or example in the target, as long as they are unique and correctly establish the requested parent-child relationships.
4. Minor variations in label text that preserve the core semantic meaning (e.g., 'Submit' vs 'Send', or 'First Name' vs 'Given Name') are acceptable unless exact literal text was requested.
5. The inclusion of valid optional properties defined in the schema (such as accessibility hints or default values) that were not explicitly requested should not be penalized as long as they make sense in context.
6. If data binding paths are not explicitly specified in the prompt, accept any logically sound path structure (e.g., accepting `/user/email` or simply `/email` when the prompt asks to "bind to the user's email" without specifying a full path).
7. Partial credit "P" can be awarded when the submitted answer is a correct answer with only minor cosmetic variations or additional valid optional properties that do not substantially change the meaning of the component.  When an answer is missing components or contains substantive errors, it should be considered incorrect and awarded an "I" grade.
```

PR #1334, opened 2026-05-05
fix: add null check for node properties before iterating to prevent r…

sugoi-yuzuru:
…untime errors

# Description

The absence of null/undefined check will result in run time error if node.property is null/undefined.

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/angular/src/v0_8/rendering/renderer.ts
+++ b/renderers/angular/src/v0_8/rendering/renderer.ts
@@ -165,14 +165,16 @@ export class Renderer {
     componentRef.setInput('component', node);
     componentRef.setInput('weight', node.weight ?? 0);
 
-    const props = node.properties as Record<string, unknown>;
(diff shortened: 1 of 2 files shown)


renderers/angular/src/v0_8/rendering/renderer.ts
gspencergoog, line 1: It would also be good to apply a similar fix to the v0_9 renderer, otherwise it will get lost when the upgrade happens.

sugoi-yuzuru: I checked v0_9 and it seems to be a bit more immune from this problem as it is not using Type casting which Jose has been working hard to add better Type-safety for v0_9

wrenj (approved):

PR #1339, opened 2026-05-05
Fix non-determinisitc test failure and add format check for Kotlin

nan-yu:
# Description

- The `test_load_examples` test was failing randomly because `dir.listFiles()` returns files in a non-deterministic order depending on the filesystem. This caused the concatenated output string to differ from the expected output in `catalog.yaml`. This PR updates `Catalog.kt` to sort the loaded JSON files by name before processing them, ensuring a deterministic output order.
- Applied `ktfmt` formatting to the Kotlin code and added a `Check formatting` step in CI.

### DIFF
--- a/agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Catalog.kt
+++ b/agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Catalog.kt
@@ -147,7 +147,9 @@ data class A2uiCatalog(
       return ""
     }
 
-    val files = dir.listFiles { _, name -> name.endsWith(".json") } ?: emptyArray()
+    // Sort files by name to ensure deterministic output order for tests.
+    val files =
+      dir.listFiles { _, name -> name.endsWith(".json") }?.sortedBy { it.name } ?: emptyList<File>()
 
     return files
       .mapNotNull { file ->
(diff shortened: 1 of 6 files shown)


agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Catalog.kt
```diff
@@ -147,7 +147,8 @@ data class A2uiCatalog(
       return ""
     }
 
-    val files = dir.listFiles { _, name -> name.endsWith(".json") } ?: emptyArray()
+    val files =
+      dir.listFiles { _, name -> name.endsWith(".json") }?.sortedBy { it.name } ?: emptyList<File>()
```
jgindin, line 152: What happens when a new JSON file is added that causes a different order?

nan-yu: That would require an update to the expected_output in `catalog.yaml`.

jgindin: I suggest adding a comment explaining why we're sorting here.

jgindin (approved):

agent_sdks/kotlin/src/main/kotlin/com/google/a2ui/core/schema/Catalog.kt
nan-yu: done

PR #1340, opened 2026-05-05
Infrastructure: Expand Python formatting check in script

josemontespg:
## Purpose
This PR updates the `scripts/fix_format.sh` script to be more comprehensive and future-proof.

## Changes
1.  **Script Update**: The script now scans all Python files in the repository (including samples) instead of being limited to `src` and `tests` in the Python SDK directory. It also separates the checks for SDK and samples to better align with how CI runs them.
2.  **Minor Formatting**: A clean run with the updated script revealed that the repository was already mostly compliant, with only one documentation file (`docs/ecosystem/renderers.md`) needing minor adjustments.

## Rationale
While files were found to be compliant, this update ensures that any new Python files added anywhere in the repo in the future will be covered by the formatting checks.

### DIFF
--- a/scripts/fix_format.sh
+++ b/scripts/fix_format.sh
@@ -31,12 +31,20 @@ else
   npx -y prettier --config .prettierrc --write .
 fi
 
-echo "Running Pyink for Python..."
+echo "Running Pyink for Python SDK..."
 cd "$REPO_ROOT/agent_sdks/python"
 if [ "$CHECK_ONLY" = true ]; then
-  uv run pyink --check src tests
+  uv run pyink --check .
 else
-  uv run pyink src tests
+  uv run pyink .
+fi
+
+echo "Running Pyink for Python Samples..."
+cd "$REPO_ROOT/samples/agent/adk"
+if [ "$CHECK_ONLY" = true ]; then
+  uv run pyink --check .
+else
+  uv run pyink .
 fi
 
 echo "Running Dart format..."


ditman:
(There's some conflicts that need to be resolved)

ditman (commented):
I don't `python` much, but LGTM!

QQ: in the old change, we didn't fail even with python files misconfigured. Have you verified that the repo checks fail after this change if a file is not correctly formatted?

josemontespg:
[Jose's AI agent] Hi @ditman, yes, I have verified that the repo checks will fail if a file is not correctly formatted.

1.  **CI Failures**: In previous runs of this PR (and PR #1338), both `enforce-formatting` (via Prettier) and `Build samples` (via Pyink) failed in CI when they encountered unformatted files.
2.  **Local Verification**: I also verified this locally by introducing a formatting error in a Python file and running `./scripts/fix_format.sh --check`. It failed with exit code 1 and reported that the file would be reformatted.

So the enforcement is working as intended!

nan-yu:
> QQ: in the old change, we didn't fail even with python files misconfigured. Have you verified that the repo checks fail after this change if a file is not correctly formatted?

I also had the same question. We have the format check for python SDK in CI: https://github.com/google/A2UI/blob/main/.github/workflows/python_agent_sdk_build_and_test.yml#L51.

It was passing before. I guess the format changes in Python might be introduced by IDE settings.

josemontespg:
Hi @nan-yu, you are correct that the check in python_agent_sdk_build_and_test.yml was passing before, as the files in agent_sdks/python were already mostly compliant.

The reason we saw many formatting changes in this PR is that we extended the enforcement to cover the rest of the Python files in the repository, primarily in the samples/ directory. Those files were not being checked by the SDK workflow, and running pyink on them revealed many files that needed reformatting to match the Google style.

nan-yu:
> Hi @nan-yu, you are correct that the check in python_agent_sdk_build_and_test.yml was passing before, as the files in agent_sdks/python were already mostly compliant.
> 
> The reason we saw many formatting changes in this PR is that we extended the enforcement to cover the rest of the Python files in the repository, primarily in the samples/ directory. Those files were not being checked by the SDK workflow, and running pyink on them revealed many files that needed reformatting to match the Google style.

Actually, we also have python format checks for the `samples/` directory: https://github.com/google/A2UI/blob/main/.github/workflows/python_samples_build.yml#L52. They were also compliant, but not compliant with the Google style in your IDE settings :)

nan-yu (approved):

ditman:
It's awesome that this landed, thanks!

PR #1342, opened 2026-05-06
Stop formatting from asking to install prettier

gspencergoog:
# Description

`fix_format.sh` was asking every time it ran to approve installing prettier when running npx.  This just bypasses the prompt to install.


josemontespg (approved):
Nice, thank you!

PR #1348, opened 2026-05-07
Use strict types for props Angular tests

josemontespg:
This PR fixes inconsistent props in Angular tests to improve type safety and consistency.

This is important, because by having strict types in tests, we can make sure that the test are actually testing the components in a way that is realistic based on the schemas. Also, we make sure that this remains consistent over time.


ditman (approved):
I like this refactor, for example being able to bind the cmoponentProps in the setup method instead of on each test. Very nice! Ship it!

PR #1357, opened 2026-05-07
Optimize Angular CI workflow

josemontespg:
# Optimize Angular CI Workflow

This PR aims to speed up the Angular CI workflow (`ng_build_and_test.yml`).

## Approach
1.  **Aggressive Caching**: Added caching for `node_modules` and `.wireit` directories across the repository to speed up the install and build phases.
2.  **Matrix Strategy**: Replaced the single job parallelization with a matrix strategy to run tests and sample builds in parallel across 4 separate jobs. This avoids CPU contention on a single runner.
3.  **Removed Redundant Step**: Removed the manual build of the renderer since it was already handled by the `postinstall` script during `npm i`.

## Results
- **Original Time**: ~3 minutes
- **Final Time**: ~1 minute 15 seconds (longest matrix job)
- **Speedup**: ~2.5x

image


ditman:
(I left a comment in the React one that applies to this one as well!)

josemontespg:
> (I left a comment in the React one that applies to this one as well!)

The tasks in the matrix are ran in parallel. Having a sequential lint check would add to the total execution time. Also added complexity to the workflow. I would prefer to keep this as is. There is a separate formatter action that should do most of the linting, although currently that does not include eslint.

Please approve if you agree with the changes. We can always tweak after for incremental improvements

ditman (approved):
Let's go!

PR #1358, opened 2026-05-07
Optimize React renderer CI workflow

josemontespg:
# Optimize React Renderer CI Workflow

This PR aims to speed up the React renderer CI workflow (`react_renderer.yml`).

## Final Approach
1.  **Matrix Strategy**: Unified separate jobs into a matrix job to run tests, demo tests, and lint in parallel.
2.  **Aggressive Caching**: Added caching for `node_modules` and `.wireit` directories to speed up dependency builds.
3.  **Switched back to `npm i`**: Reverted from `npm ci` to `npm i` because `npm ci` forced clean installs that negated most of the speedup benefits from caching.

## Results & Findings
- **Original Time**: ~1m 30s
- **With `npm ci` and `setup-node` cache**: ~1m 28s (Marginal improvement)
- **With `npm i` and direct `node_modules` cache**: ~1m 10s (Longest matrix job)
- **Speedup**: ~1.3x with `npm i` approach.

We decided to stick with `npm i` to achieve a meaningful speedup, despite the theoretical safety benefits of `npm ci` pointed out by @andrewkolos.

All checks passed in the `npm i` iteration previously. We are running them again.


.github/workflows/react_renderer.yml
```diff
@@ -32,6 +32,10 @@ on:
 jobs:
   build-and-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        task: [test, demo, lint]
```
ditman, line 37: What do you think about running "lint" first, and if that fails not run anything else (fail fast)?

andrewkolos:
I'm not familiar with the CI setup, but doesn't `npm ci` exist for the sake of ensuring clean installs on CI to keep dependency-related issues reproducible and help prevent package-lock drift from masking issues that can be caused by new dependency versions (e.g. new patch version of dependency introduces a bug or unexpected breaking change). If the concern is avoiding pointlessly reinstalling tarballs for verions of packages we've already downloaded, is there a way to configure that specifically in the setup action instead?

.github/workflows/react_renderer.yml
josemontespg: These 3 run in parallel. If we lint first that will add to the overall execution time

andrewkolos (changes requested):
See prior comment about dependency version caching concerns.

While we are optimizing this, is it possible to skip building markdown-it and web_core for the lint task? I'm not an expert on job matrices, but I think this could be possible. Not sure what the practical speedup would be if any.

edit: nvm, I just saw the most recent commit

andrewkolos:
> While we are optimizing this, is it possible to skip building markdown-it and web_core for the lint task? I'm not an expert on job matrices, but I think this could be possible. Not sure what the practical speedup would be if any.

On second thought, even if I'm onto something here, this is probably not worth the added complexity since the `lint` task is probably the fastest running one anyway. Any speedup would probably be from edge cases like runner contention or whatever (if that's even a thing here). The only other benefit is maybe reduced wallclock time/cost but I'm assuming that's marginal compared to time spent on this level of optimization.

josemontespg:
> this is probably not worth the added complexity since the `lint` task is probably the fastest running one anyway.

I agree

I reverted back to `npm ci` and now the improvements are marginal. I agree that is more correct to use `ci`, but it is a real tradeoff in performance. I would optimize for performance, because the kinds of bugs that `ci` would catch should not block submit of a PR. Also, we would anyway catch these bugs when creating new branches and running `npm i` locally during normal development cycles. @andrewkolos WDYT?

This is an easily reversible decision, so I would not spend too much time on this. Whatever we decide can tweak later if needed.

andrewkolos:
> I reverted back to `npm ci` and now the improvements are marginal. I agree that is more correct to use `ci`, but it is a real tradeoff in performance. I would optimize for performance, because the kinds of bugs that `ci` would catch should not block submit of a PR. Also, we would anyway catch these bugs when creating new branches and running `npm i` locally during normal development cycles. @andrewkolos WDYT?
> 
> This is an easily reversible decision, so I would not spend too much time on this. Whatever we decide can tweak later if needed.

Fair enough and I'm probably over-indexing on my experience of working on the Flutter CLI tool. There is probably some hybrid setup where we only rebuild node_modules when the lock file actually changes, but it's probably not worth it.

andrewkolos (approved):
LGTM. Thanks for the optimization and simplification :)

PR #1409, opened 2026-05-12
fix: escape ${expression} in formatString description to prevent ADK KeyError

### LINKED ISSUE #1388: adk session: KeyError: 'Context variable not found: expression' when using adk run with A2UI agents
## Description

When running an A2UI-enabled agent using `adk run`, the CLI crashes with a `KeyError` as soon as the first user message is sent.

```
KeyError: 'Context variable not found: `expression`.'
```

Full traceback excerpt:

```
File ".../google/adk/utils/instructions_utils.py", line 124, in inject_session_state
    return await _async_sub(r'{+[^{}]*}+', _replace_match, template)
  File ".../google/adk/utils/instructions_utils.py", line 122, in _replace_match
    raise KeyError(f'Context variable not found: `{var_name}`.')
KeyError: 'Context variable not found: `expression`.'
```

---

## Steps to Reproduce

1. Create an ADK agent that uses `A2uiSchemaManager.generate_system_prompt()` with `include_schema=True` to build its instruction.
2. Expose `root_agent` at module level so it is compatible with `adk run`.

   ```python
   # agent.py
   root_agent = LlmAgent(
       model=Gemini(model="gemini-2.0-flash"),
       name="my_agent",
       instruction=schema_manager.generate_system_prompt(
           role_description="...",
           include_schema=True,
           include_examples=True,
       ),
       tools=[...],
   )
   ```

3. Run the agent:

   ```bash
   adk run path/to/my_agent
   ```

4. Type any message and press Enter → crash.

---

## Root Cause

The `BasicCatalog` JSON schema bundled at `agent_sdks/python/src/a2ui/assets/0.9/basic_catalog.json` contains the following text in the description of the `formatString` function:

```
The value string can contain interpolated expressions in the `${expression}` format.
```

When `generate_system_prompt()` is called with `include_schema=True`, this description is embedded verbatim into the agent's `instruction` string.

When `adk run` processes the first user message, the ADK framework calls `inject_session_state()` in `instructions_utils.py`. This function scans the full instruction string with the regex `r'{+[^{}]*}+'`. It matches `{expression}`, strips the braces to extract `"expression"`, validates it as a Python identifier (which passes), looks it up in the active session state dict (which is empty because `adk run` manages sessions internally), and raises a `KeyError`.

The key line in the ADK source:

```python
# google/adk/utils/instructions_utils.py
return await _async_sub(r'{+[^{}]*}+', _replace_match, template)
```

---

## Why Existing A2A Server Samples Are Unaffected

The existing A2A-based samples (e.g. `samples/agent/adk/restaurant_finder/`) manually create the ADK session and inject a dummy variable before running:

```python
# restaurant_finder/agent.py
session_state = {"base_url": self.base_url, "expression": "{expression}"}
session = await runner.session_service.create_session(
    ...,
    state=session_state,
)
```

This satisfies the ADK template engine — it finds `"expression"` in the state and silently replaces the tag. `adk run` manages sessions internally and provides no hook to pre-seed the session state before the first message, so this workaround is not available when using the CLI.

---

brucearctor:
## Summary

Fixes #1388

The `formatString` function description in `basic_catalog.json` contained the literal text `${expression}`, which the ADK's `inject_session_state()` matched as a template variable. Since `expression` passes Python's `str.isidentifier()` check, the ADK raised a `KeyError` when it couldn't find it in the session state.

## Changes

Changed `${expression}` → `${<expression>}` in the `formatString` description for both `v0_9` and `v0_10` catalog schemas.

The angle brackets cause `<expression>` to fail the `isidentifier()` check, so the ADK's `_is_valid_state_name()` returns `False` and the match is returned as-is (line 108 of `instructions_utils.py`).

## Why only `${expression}` was affected

The other interpolation examples in the same description were already safe:
- `${/absolute/path}` → `/absolute/path` is not a valid identifier (contains `/`)
- `${now()}` → `now()` is not a valid identifier (contains `()`)
- `${relative/path}` → `relative/path` is not a valid identifier (contains `/`)

Only `expression` was a bare, valid Python identifier.

## Files changed

- `specification/v0_9/json/basic_catalog.json`
- `specification/v0_10/json/basic_catalog.json`

### DIFF
--- a/specification/v0_10/json/basic_catalog.json
+++ b/specification/v0_10/json/basic_catalog.json
@@ -811,7 +811,7 @@
     },
     "formatString": {
       "type": "object",
-      "description": "Performs string interpolation of data model values and other functions in the catalog functions list and returns the resulting string. The value string can contain interpolated expressions in the `${expression}` format. Supported expression types include: JSON Pointer paths to the data model (e.g., `${/absolute/path}` or `${relative/path}`), and client-side function calls (e.g., `${now()}`). Function arguments must be named (e.g., `${formatDate(value:${/currentDate}, format:'MM-dd')}`). To include a literal `${` sequence, escape it as `\\${`.",
+      "description": "Performs string interpolation of data model values and other functions in the catalog functions list and returns the resulting string. The value string can contain interpolated expressions in the `${<expression>}` format. Supported expression types include: JSON Pointer paths to the data model (e.g., `${/absolute/path}` or `${relative/path}`), and client-side function calls (e.g., `${now()}`). Function arguments must be named (e.g., `${formatDate(value:${/currentDate}, format:'MM-dd')}`). To include a literal `${` sequence, escape it as `\\${`.",
       "properties": {
         "call": {
           "const": "formatString"
--- a/specification/v0_9/json/basic_catalog.json
+++ b/specification/v0_9/json/basic_catalog.json
@@ -962,7 +962,7 @@
     },
     "formatString": {
       "type": "object",
-      "description": "Performs string interpolation of data model values and other functions in the catalog functions list and returns the resulting string. The value string can contain interpolated expressions in the `${expression}` format. Supported expression types include: JSON Pointer paths to the data model (e.g., `${/absolute/path}` or `${relative/path}`), and client-side function calls (e.g., `${now()}`). Function arguments must be named (e.g., `${formatDate(value:${/currentDate}, format:'MM-dd')}`). To include a literal `${` sequence, escape it as `\\${`.",
+      "description": "Performs string interpolation of data model values and other functions in the catalog functions list and returns the resulting string. The value string can contain interpolated expressions in the `${<expression>}` format. Supported expression types include: JSON Pointer paths to the data model (e.g., `${/absolute/path}` or `${relative/path}`), and client-side function calls (e.g., `${now()}`). Function arguments must be named (e.g., `${formatDate(value:${/currentDate}, format:'MM-dd')}`). To include a literal `${` sequence, escape it as `\\${`.",
       "properties": {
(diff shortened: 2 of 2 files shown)


ditman:
Would it be possible to maybe escape the `$` symbol as `\$` or similar, so python doesn't even attempt to resolve the identifier at all? /cc @nan-yu

brucearctor:
@ditman  `\$` won't work due to JSON escaping semantics, and `${<expression>}` is the least-invasive fix at the A2UI layer. The root cause (ADK's overly greedy template scanning) should be addressed upstream in ADK, not here.

there might be other options, but I filed:  https://github.com/google/adk-python/issues/5706

I can work on that issue with the adk team to try to get that fix in place.  This should be OK in the meantime?

ditman:
Thanks for fixing this! Yes, the fix LGTM, but let's have @nan-yu or @gspencergoog take a second look, I normally don't work on the python code (or the spec :P)

nan-yu:
It feels like this should be fixed in the ADK library. @gspencergoog has filed an issue against the ADK project: https://github.com/google/adk-python/issues/5179.

brucearctor:
> It feels like this should be fixed in the ADK library. @gspencergoog has filed an issue against the ADK project: [google/adk-python#5179](https://github.com/google/adk-python/issues/5179).

I think this is here too ?   https://github.com/google/adk-python/issues/5706

Or, should I then not address in favor of whoever doing 5179 ?

ditman:
Here's a proposed fix for google/adk-python#5179:

* https://github.com/google/adk-python/pull/5711

nan-yu:
Waiting for the fix from the ADK library as this seems to be an ADK specific issue.

PR #1420, opened 2026-05-12
Use working package-lock.json for web_core to fix CI

josemontespg:
This PR copies the working package-lock.json from fix-angular-modal worktree to renderers/web_core to resolve dependency issues that were breaking the Angular CI.


josemontespg:
Not needed

PR #1425, opened 2026-05-13
Explicitly cast the type of props().weight in the internal implementation of the basic-catalog-component.ts

josemontespg:
This PR improves the type of props()["weight"] to be statically defined. Before this change it was ambiguous to the compiler. This fixes a type error that appeared when importing the repo to 1P.


renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
```diff
@@ -45,7 +49,9 @@ export abstract class BasicCatalogComponent<
   });
 
   /** Weight is applied as flex css property on the component host HTML element. */
-  protected readonly weight = computed(() => this.props()['weight']?.value() ?? null);
+  protected readonly weight = computed(
+    () => (this.props() as CommonPropsType)['weight']?.value() ?? null,
```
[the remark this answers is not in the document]
josemontespg: Outdated

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
```diff
@@ -19,6 +19,19 @@ import {injectBasicCatalogStyles} from '@a2ui/web_core/v0_9/basic_catalog';
...
+
+type CommonPropsOutput = z.infer<typeof CommonPropsSchema>;
+type CommonPropsInput = z.input<typeof CommonPropsSchema>;
+
+type ComponentApiWithCommonProps = ComponentApi & {
+  schema: z.ZodType<
+    CommonPropsOutput & Record<string, unknown>,
+    z.ZodTypeDef,
+    CommonPropsInput & Record<string, unknown>
+  >;
```
ditman: Why do we need this? The `ComponentApi` coming from web_core already has "weight" and "accessibility" baked in the definition?

See how `CommonProps` already belongs in `TextApi`, for example:

https://github.com/google/A2UI/blob/69c16a2a4ba85a0459004b3fd2ca77c64963d142/renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts#L43-L48

andrewkolos: Yes, this doesn't add any type safety to component implementations, it just corrects the types _within_ `BasicCatalogComponent`. The default type for the `Schema` of `ComponentApi` is effectively `any` (`z.ZodTypeAny`). 

https://github.com/google/A2UI/blob/9526ab2ec05549443aa592cab056940aa360b37d/renderers/web_core/src/v0_9/catalog/types.ts#L93

So prior to this change, `BasicCatalogComponent.props()['weight']` has a type of `any`, for example.

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
```diff
@@ -19,6 +19,19 @@ import {injectBasicCatalogStyles} from '@a2ui/web_core/v0_9/basic_catalog';
...
+import {z} from 'zod';
+
+type CommonPropsOutput = z.infer<typeof CommonPropsSchema>;
+type CommonPropsInput = z.input<typeof CommonPropsSchema>;
+
+type ComponentApiWithCommonProps = ComponentApi & {
+  schema: z.ZodType<
+    CommonPropsOutput & Record<string, unknown>,
+    z.ZodTypeDef,
+    CommonPropsInput & Record<string, unknown>
```
andrewkolos: Are the instances of `& Record<string, unknown>` necessary here?

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
andrewkolos: If the primary goal is to satisfy internal builds rather than increase type safety, we can instead do scoped casts within `BasicCatalogComponent`. Example:

```ts
const props = this.props() as {weight?: BoundProperty<number>};
return props.weight?.value() ?? null;
```

andrewkolos: Alternatively, if the primary goal is cleaner type semantics, one route I would consider is adding a second type parameter, `Props`, to `CatalogComponent`:

```ts
export abstract class CatalogComponent<
  Api extends ComponentApi,
  Props extends Record<string, unknown> = ComponentApiToProps<Api>,
> {
  readonly props = input<Props>({} as Props);
}
```

```ts
type ComponentApiWithCommonProps = Omit<ComponentApi, 'schema'> & {
  schema: z.ZodType<
    CommonPropsOutput & object, // use object instead of record to preserve the finite set of keys we want to be valid.
    z.ZodTypeDef,
    CommonPropsInput & object
  >;
};

type CommonBoundProps = ComponentApiToProps<ComponentApiWithCommonProps>;

export abstract class BasicCatalogComponent<
  Api extends ComponentApiWithCommonProps,
> extends CatalogComponent<Api, ComponentApiToProps<Api> & CommonBoundProps> {
```

This should remove `any` while also preventing access to unexpected keys (e.g. `this.props()['typo']?.value()` won't compile).

andrewkolos (commented):
I proposed some alternatives that might be preferable depending on what needs to prioritized here. I'm also happy to stamp this as-is if something is blocked on this fix

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
ditman: I think in Lit we have solved this without an extra property:

https://github.com/google/A2UI/blob/83a395c939e97bd356535a43188626611650642b/renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts#L24

Is this needed in angular because the way templates are defined or something else?

andrewkolos: Assuming by "extra property" you meant "extra type parameter", Lit avoids this by using a blanket `any` cast:

https://github.com/google/A2UI/blob/8c907e954929f79d89a751c76592418a8db2ceaf/renderers/lit/src/v0_9/catalogs/basic/basic-catalog-a2ui-lit-element.ts#L39

My first alternative proposal (https://github.com/google/A2UI/pull/1425#discussion_r3241076519) would be a compromise between the simplicity of cast(s) while preventing any internal tooling from panicking about `any` or unsafe object property accesses.

ditman: Uuuughghghgh, I didn't notice that. I thought the prop should be part of the ComponentApi anyway, I wonder where that goes :/

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
josemontespg: Refactored to a simpler approach.

renderers/angular/src/v0_9/catalog/basic/basic-catalog-component.ts
josemontespg: @andrewkolos exactly right! I like your suggestion to simplify the types. Although the original code in the PR produces the correct type, it is a bit convoluted and hard to understand. In this case a simple casting is more maintainable, and since it only affects the internal implementation of the basic-catalog-component.ts file it is fine IMO.

josemontespg:
Thanks for the reviews! I applied the suggestion, please take another look when you can!

andrewkolos (approved):
LGTM!

PR #1426, opened 2026-05-13
Move CatalogComponentInstance to its own file

josemontespg:
This PR moves CatalogComponentInstance to its own file in the Angular renderer (src/v0_9/core/catalog_component_instance.ts) to avoid cyclic dependencies when importing the repo to 1P.


ditman (approved):
Files are free, ship it!

renderers/angular/src/v0_9/core/catalog_component.ts
```diff
@@ -15,16 +15,9 @@
...
 import {ComponentApiToProps} from './types';
-
-/** Describes the properties that a Catalog component needs to implement. For ease of use, please extend CatalogComponent. */
-export interface CatalogComponentInstance {
-  readonly props: Signal<Record<string, unknown>>;
-  readonly surfaceId: Signal<string>;
-  readonly componentId: Signal<string>;
-  readonly dataContextPath: Signal<string>;
-}
+import {CatalogComponentInstance} from './catalog_component_instance';
```
ditman, line 20: Does this not need to be reexported?

josemontespg: Mmm good point, I think from /renderers/angular/src/v0_9/public-api.ts

PR #1427, opened 2026-05-13
feat(web-core): implement locale support and fix pluralize

### LINKED ISSUE #1385: `pluralize` hard-codes `en-US` and renderers are inconsistent
## Problems

**Spec vs reference renderer** — The specification does not define which locale drives plural categories, while the reference implementation hard-codes `en-US`. Renderer SDK authors have no normative target and cannot tell integrators what behavior is “correct.”

**1. WebCore hard-codes `en-US`**
`basic_functions.ts` line 326: `new Intl.PluralRules('en-US').select(args.value)` — wrong category for any non-English content (e.g. Polish `select(2)` → `few`, not `other`).

**2. Flutter ignores CLDR entirely**
`basic_functions.dart` uses `if (count == 0) / if (count == 1) / other` — no `Intl.PluralRules` equivalent, always wrong for languages like French (`0` → `one`) or Arabic.

**3. No `locale` in the spec**
`pluralize.args` has no `locale` field and uses `unevaluatedProperties: false`, so agents cannot pass a locale even if they want to.

## Fix

- Add optional `locale` (BCP 47) to `pluralize.args` in the spec.
- Update WebCore to use `args.locale` when present, fall back to client locale, then `en-US`.
- Fix Flutter to read `value` (not `count`) and use a proper CLDR plural lookup.

gspencergoog:
# Description

This PR adds locale support to `DataContext` and `SurfaceModel` in `web-core`, and updates `pluralize`, `formatNumber`, and `formatCurrency` to use the resolved locale instead of hardcoding 'en-US'. It also includes removing `.passthrough()` from `PluralizeApi` and adding tests for Welsh locale.

## Changes
- **`surface-model.ts`**: Added optional `locale` parameter to `SurfaceModel` constructor.
- **`data-context.ts`**: Added `locale` getter to `DataContext` that delegates to `surface.locale`.
- **`basic_functions.ts`**: Updated `PluralizeImplementation`, `FormatNumberImplementation`, and `FormatCurrencyImplementation` to use `context.locale`.
- **`basic_functions_api.ts`**: Removed `.passthrough()` from `PluralizeApi` schema for stricter validation.
- **`basic_functions.test.ts`**: Updated `createTestDataContext` to support passing a locale, and added tests for `pluralize` with the Welsh locale (`cy`) covering all 6 categories.
- **`CHANGELOG.md`**: Added entries under "Unreleased".

## Impact & Risks
- **Low Risk**: The `locale` parameter is optional and defaults to `undefined`. If not provided, the `Intl` APIs will fall back to the runtime's default locale, maintaining current behavior for existing usage.

## Testing
- Added new unit tests for Welsh pluralization in `basic_functions.test.ts`.
- Verified all changes by running `npm run test` in `renderers/web_core`. All 258 tests passed.

Fixes https://github.com/google/A2UI/issues/1385

### DIFF
--- a/renderers/web_core/CHANGELOG.md
+++ b/renderers/web_core/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+- Add locale support to `SurfaceModel` and `DataContext` in v0.9.
+- Update `pluralize`, `formatNumber`, and `formatCurrency` to use the context locale instead of hardcoding 'en-US'.
+- Remove `.passthrough()` from `PluralizeApi` schema for stricter validation.
+
 ## 0.10.0
 
 - **BREAKING CHANGE**: Rename Icon `path` property to `svgPath` to fix type collision with `DataBindingType`.
--- a/renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.test.ts
+++ b/renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.test.ts
@@ -382,6 +382,8 @@ describe('BASIC_FUNCTIONS', () => {
         invoke('pluralize', {value: 5, one: 'apple', other: 'apples'}, context),
         'apples',
       );
+      assert.strictEqual(invoke('pluralize', {value: 1, other: 'apples'}, context), 'apples');
+      assert.strictEqual(invoke('pluralize', {value: 0, other: 'apples'}, context), 'apples');
     });
   });
 
--- a/renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.test.ts
+++ b/renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.test.ts
@@ -34,11 +34,13 @@ const createTestDataContext = (
   model: DataModel,
   path: string,
   functionInvoker: any = testCatalog.invoker,
+  locale?: string,
 ) => {
   const mockSurface = {
     dataModel: model,
     catalog: {invoker: functionInvoker},
     dispatchError: () => {},
+    locale: locale,
   } as any;
   return new DataContext(mockSurface, path);
 };
@@ -351,6 +353,33 @@ describe('BASIC_FUNCTIONS', () => {
       );
       assert.strictEqual(
         invoke('pluralize', {value: 2, one: 'apple', other: 'apples'}, context),
+        'apples',
+      );
+    });
+
+    it('pluralize with Welsh locale', () => {
+      const cyContext = createTestDataContext(dataModel, '/', testCatalog.invoker, 'cy');
+      // Welsh for various numbers of "cat".  Welsh because all six cases have different rules.
+      const args = {
+        zero: 'cathod',
+        one: 'gath',
+        two: 'gath',
+        few: 'cath',
+        many: 'chath',
+        other: 'cath',
+      };
+
+      assert.strictEqual(invoke('pluralize', {...args, value: 0}, cyContext), 'cathod');
+      assert.strictEqual(invoke('pluralize', {...args, value: 1}, cyContext), 'gath');
+      assert.strictEqual(invoke('pluralize', {...args, value: 2}, cyContext), 'gath');
+      assert.strictEqual(invoke('pluralize', {...args, value: 3}, cyContext), 'cath');
+      assert.strictEqual(invoke('pluralize', {...args, value: 6}, cyContext), 'chath');
+      assert.strictEqual(invoke('pluralize', {...args, value: 4}, cyContext), 'cath');
+    });
+
+    it('pluralize fallback to other', () => {
+      assert.strictEqual(
+        invoke('pluralize', {value: 5, one: 'apple', other: 'apples'}, context),
         'apples',
       );
     });
--- a/renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.ts
+++ b/renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.ts
@@ -271,14 +271,17 @@ export const FormatStringImplementation = createFunctionImplementation(
  * Implementation of the number formatting function.
  * Formats a number using Intl.NumberFormat with specified decimals and grouping.
  */
-export const FormatNumberImplementation = createFunctionImplementation(FormatNumberApi, args => {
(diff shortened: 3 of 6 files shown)


renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.ts
```diff
@@ -267,36 +267,80 @@ export const FormatStringImplementation = createFunctionImplementation(
...
-  if (isNaN(args.value)) return '';
-  return new Intl.NumberFormat('en-US', {
-    minimumFractionDigits: args.decimals,
-    maximumFractionDigits: args.decimals,
-    useGrouping: args.grouping,
-  }).format(args.value);
-});
+export const FormatNumberImplementation = createFunctionImplementation(
+  FormatNumberApi,
+  (args, context) => {
```
ditman, line 296: Do we have a good type for `args` and `context`? Or does this need to be "any"?

renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.ts
```diff
@@ -318,14 +362,34 @@ export const FormatDateImplementation = createFunctionImplementation(FormatDateA
     return date.toISOString();
   }
 });
+const pluralRulesCache = new Map<string, Intl.PluralRules>();
+
+function getPluralRules(locale: string | undefined): Intl.PluralRules {
+  const key = locale ?? 'default';
+  let rules = pluralRulesCache.get(key);
+  if (!rules) {
```
ditman, line 370: `putIfAbsent` is-a-comin'!

https://caniuse.com/mdn-javascript_builtins_map_getorinsertcomputed

renderers/web_core/src/v0_9/state/surface-model.ts
```diff
@@ -53,12 +53,14 @@ export class SurfaceModel<T extends ComponentApi = ComponentApi> {
...
    * @param theme The theme to apply to this surface.
    * @param sendDataModel If true, the client will send the full data model.
+   * @param locale The locale to use in locale-sensitive functions.
    */
   constructor(
     readonly id: string,
     readonly catalog: Catalog<T>,
     readonly theme: any = {},
     readonly sendDataModel: boolean = false,
+    readonly locale?: string,
```
ditman, line 63: (I always need to look up why this "primary constructor"-ish thing works, ugh!)

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -53,6 +53,13 @@ export class DataContext {
     this.functionInvoker = surface.catalog.invoker;
   }
 
+  /**
+   * Gets the locale for this context, inherited from the surface.
+   */
+  get locale(): string | undefined {
```
ditman, line 59: If it's good [for TypeScript](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined), it's good for me :P

renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.test.ts
```diff
@@ -354,6 +356,33 @@ describe('BASIC_FUNCTIONS', () => {
...
+      assert.strictEqual(invoke('pluralize', {...args, value: 3}, cyContext), 'cath');
+      assert.strictEqual(invoke('pluralize', {...args, value: 6}, cyContext), 'chath');
+      assert.strictEqual(invoke('pluralize', {...args, value: 4}, cyContext), 'cath');
+    });
+
+    it('pluralize fallback to other', () => {
+      assert.strictEqual(
+        invoke('pluralize', {value: 5, one: 'apple', other: 'apples'}, context),
+        'apples',
+      );
```
ditman, line 384: Add tests for values `0` and `1` as well?

renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.ts
ditman: (Context can probably be typed easily, but I guess for args we have a similar situation to the shenanigans we need to do for Component APIs, ugh!)

renderers/web_core/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+- Add locale support to `SurfaceModel` and `DataContext` in v0.9.
+- Update `pluralize`, `formatNumber`, and `formatCurrency` to use the context locale instead of hardcoding 'en-US'.
+- Remove `.passthrough()` from `PluralizeApi` schema for stricter validation.
```
ditman, line 5: Removing `passthrough` is technically a breaking change, but I wouldn't expect anybody to be using pluralize with ad-hoc settings, so we can probably release as a minor version.

ditman (approved):
Minor nitpicks. We should have an issue (if it doesn't exist already) to track improving the types of the arguments to function calls?

renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.test.ts
gspencergoog: Sure.  Done.

renderers/web_core/CHANGELOG.md
gspencergoog: Yeah, but it definitely shouldn't have it.  We only want specific values.

renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.ts
gspencergoog: Cool!

renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions.ts
gspencergoog: They are actually already strongly typed.

Thanks to the generic inference in `createFunctionImplementation` defined in `types.ts`, you don't need to use `any` or add manual type annotations.

ditman: Ah cool, it is unclear whether this is strongly typed through inference rather than `any` just by looking at this PR!

PR #1431, opened 2026-05-14
refactor(web_core): allow overriding recursion depth in DataValueSchema for v0_8

### LINKED ISSUE #1424: Allow a way to bypass DataModel recursion depth check
Currently, if the depth of a data model path is more than 5, an error is thrown:

Image

I assume this is to make sure the LLM isn't producing deeply nested JSONs. However, there are cases where nested data model is necessary. In my case, my data model depth is 6 because my component is referencing data located in paths like "/ai_insights_1/items/0/sources/0/title". It would be great if we allow a way to bypass this check or define our own max depth.

gspencergoog:
# Description

Fixes https://github.com/google/A2UI/issues/1424

This PR allows developers to override the hard-coded recursion depth limit in `DataValueSchema` while maintaining backward compatibility.

## Changes
- Refactored `DataValueSchema` in `renderers/web_core/src/v0_8/schema/common-types.ts` from a static constant to a factory function `createDataValueSchema(options?: { maxDepth?: number })`.
- Exported the original `DataValueSchema` as a constant using the default limit of 5.
- Added a new test file `renderers/web_core/src/v0_8/schema/common-types.test.ts` to verify the recursion depth limits.

## Impact & Risks
- **Backward Compatibility**: The default `DataValueSchema` behavior is unchanged, so there should be no breaking changes for existing usages.
- **Risk**: Low, as it only affects schema validation if someone explicitly opts into a custom depth.

## Testing
- Added new unit tests in `common-types.test.ts` to verify default and custom depth limits.
- Ran tests manually using `node --test dist/src/v0_8/schema/common-types.test.js`.


josemontespg (approved):
LGTM. Should we add this change to the changelog?

gspencergoog:
> LGTM. Should we add this change to the changelog?

Good idea.  Done.

PR #1433, opened 2026-05-14
Fix v0.8 Icon component snake_case handling

josemontespg:
This PR fixes the v0.8 Icon component to handle camelCase and TitleCase names by converting them to snake_case, as expected by g-icon. It also adds unit tests covering this behavior.


renderers/angular/src/v0_8/components/icon.ts
```diff
@@ -45,5 +45,17 @@ import type {IconNode, StringValue} from '../types';
...
+    const rawName = this.resolvePrimitive(this.name());
+    if (!rawName) return '';
+    return this.toSnakeCase(rawName);
+  });
+
+  private toSnakeCase(str: string): string {
+    return str
+      .replace(/^[A-Z]/, letter => letter.toLowerCase())
+      .replace(/[A-Z]/g, letter => `_${letter.toLowerCase()}`);
+  }
```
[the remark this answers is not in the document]
josemontespg: That's a bit overkill. It would break TitleCase handling also.

renderers/angular/src/v0_8/components/icon.ts
```diff
@@ -45,5 +45,17 @@ import type {IconNode, StringValue} from '../types';
...
-  protected readonly resolvedName = computed(() => this.resolvePrimitive(this.name()));
+
+  // g-icon uses snake_case for the icon names. We convert camelCase and TitleCase here.
+  protected readonly resolvedName = computed(() => {
+    const rawName = this.resolvePrimitive(this.name());
+    if (!rawName) return '';
+    return this.toSnakeCase(rawName);
+  });
+
+  private toSnakeCase(str: string): string {
```
ditman, line 56: The Lit implementation is much less considerate than this:

https://github.com/google/A2UI/blob/8c907e954929f79d89a751c76592418a8db2ceaf/renderers/lit/src/0.8/ui/icon.ts#L70

ditman (approved):
In V0.9 we have some icon names that don't directly match the g-icon name when snake-cased, I guess that's not the case for v0.8!

renderers/angular/src/v0_8/components/icon.ts
josemontespg: Interesting. I think this util should be shared in core. Not worth optimizing at this point in my opinion

PR #1435, opened 2026-05-14
Move v0.9 integration tests to tests/v0_9 directory

josemontespg:
This PR moves the v0.9 integration tests of the Angular explorer app from tests/ to a new tests/v0_9/ directory to prepare for adding v0.8 tests in a separate folder. It also updates the relative imports to test_utils in the moved files.


ditman (approved):
Thanks for the move!

Issue #1432, opened 2026-05-14
Delete minimal catalog and examples


josemontespg:
From the review of https://github.com/google/A2UI/pull/1405

@ditman 
> I think we should remove the "minimal" catalog and examples. They don't add much value once the renderer is fully built (instead, they add maintenance burden like here)

ditman:
I actually removed it from Lit a while ago:

https://github.com/google/A2UI/pull/1079/changes/6c0296d09cc2bedfa9f3d57a4d8524ba5d3351fe

ditman:
PS: I actually removed the "minimal" catalog implementation from everything lit, even the a2ui_explorer demo. There were some hacks in the code needed to load the appropriate examples and catalog, and I removed everything. 🔥🔥🔥

josemontespg:
From #1497

> BTW Can you also make an equivalent change to specificaiton/v0_9_1 (sorry about that). Ideally, the generate-examples.mjs will also pull from v0_9_1. I will send a proposal later this week arguing that we should never do patch versions of the specification, to avoid this kind of churn.

gspencergoog:
A2UI Triage: Triaged as P2.

We want to keep the minimal catalog in the specification, as it makes for a good stepping stone when asking an agent to implement a renderer. We don't need support for it anywhere else, though, IMO.

ditman:
As a concept, it's probably good, but given the minimal catalog looks like a subset of the basic catalog, once the basic catalog is implemented, the minimal catalog implementation is redundant. I think we've removed it already from most of our renderers though.

gspencergoog:
Yeah, it's supposed to be removed after you implement the basic catalog, and it's also supposed to be a minimal subset of the basic catalog.  The idea is that you ask an agent to implement that one first so it doesn't get bogged down in the nitty gritty of implementing the more complex components, but is required to implement the reactive data model updates, at least one function, etc.

I think that is still useful.  We could, however, just tell it in the "build a renderer" instructions to ignore all but specific components until those components are implemented and tested.  So we don't need an actual separate catalog.

ditman:
Yep, yep, this issue was created because at some point the minimal/basic catalogs that we had were diverging (we had two ever so slightly different Text widgets to maintain, for example). Anyway, they've been removed from most of the renderers that we're vending. It's probably fine to leave them in the build a renderer instructions.

[...]

PR #1436, opened 2026-05-14
Add v0.8 support to Angular explorer app

josemontespg:
This PR adds support for the v0.8 specification to the Angular explorer app. It includes a version toggle in the UI, updates the code generator to bundle v0.8 examples. Also refactors AgentStubService to be abstract, with different implementations based on the version.


renderers/angular/scripts/generate-examples.mjs
```diff
@@ -148,31 +120,64 @@ async function main() {
...
-import { Example } from '../types';
+import { Example, Example_08 } from '../types';
+
+export const EXAMPLES_V08: Example_08[] = ${JSON.stringify(examplesV08, null, 2)};
+
+export const EXAMPLES_V09: Example[] = ${JSON.stringify(examplesV09, null, 2)};
 
-export const EXAMPLES: Example[] = ${JSON.stringify(examples, null, 2)};
+// Default to v0.9 for backward compatibility
+export const EXAMPLES: Example[] = EXAMPLES_V09;
```
ditman: I don't think we need backwards-compatibility here?

renderers/angular/scripts/generate-examples.mjs
```diff
@@ -39,7 +34,6 @@ const DEFAULT_CATALOGS = ['minimal', 'basic'];
 const options = {
   help: {type: 'boolean', short: 'h'},
   'out-file': {type: 'string', short: 'o', default: DEFAULT_OUT_FILE},
-  'spec-path': {type: 'string', short: 's', default: DEFAULT_SPEC_PATH},
   catalog: {type: 'string', short: 'c', multiple: true, default: DEFAULT_CATALOGS},
   'override-minimal-catalog-id': {type: 'boolean', default: true},
```
ditman, line 38: (We could remove this if we got rid of the minimal catalog :P)

renderers/angular/a2ui_explorer/tsconfig.app.json
```diff
@@ -8,5 +8,5 @@
     "moduleResolution": "bundler"
   },
   "include": ["src/**/*.ts"],
-  "exclude": ["src/**/*.spec.ts"]
+  "exclude": ["src/**/*.spec.ts", "src/app/tests/test_utils.ts"]
```
ditman: Instead of having a big test_utils.ts file, I'd put them in a `tests/utils` directory or similar, so you can exclude `src/app/tests/utils/**`. That way, there's no friction to creating new files and keeping them small.

renderers/angular/a2ui_explorer/src/app/agent-stub.service.ts
```diff
@@ -137,3 +121,98 @@ export class AgentStubService {
...
+
+/**
+ * A stub service that simulates an A2UI agent.
+ * It listens for actions and responds with data model updates or new surfaces.
+ * Supports the v0.8 A2UI spec.
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class AgentStubV08Service {
```
ditman: can this be in an `agent-stub-v08.service.ts` file?

renderers/angular/a2ui_explorer/src/app/agent-stub.service.ts
```diff
@@ -137,3 +121,98 @@ export class AgentStubService {
...
+  private surfaceVersions = new Map<string, '0.8' | '0.9'>();
+
+  constructor(
+    private v09Service: AgentStubV09Service,
+    private v08Service: AgentStubV08Service,
+  ) {}
+
+  get actionsLog() {
+    return [...this.v09Service.actionsLog, ...this.v08Service.actionsLog].sort(
+      (a, b) => a.timestamp.getTime() - b.timestamp.getTime(),
```
ditman: Are we trying to get v0.8 and v0.9 working simultaneously? Wouldn't it be simpler to just reload the application when switching protocol versions? That way everything can be less tied together?

ditman (commented):
In the agent-stub, are we trying to get logs from both 0.9 and 0.8 to be merged and work together nicely? I think we should reset the state of the app when switching protocol versions (and simplify all the code made to get things to work seamlessly :) )

renderers/angular/scripts/generate-examples.mjs
josemontespg: Updated the comment

renderers/angular/scripts/generate-examples.mjs
josemontespg: Yep, I'll do this after in https://github.com/google/A2UI/issues/1432

renderers/angular/a2ui_explorer/tsconfig.app.json
josemontespg: Done

renderers/angular/a2ui_explorer/src/app/demo.component.ts
```diff
@@ -581,28 +677,64 @@ export class DemoComponent implements OnInit, OnDestroy {
...
         this.eventsLog.unshift({timestamp: new Date(), action});
         this.cdr.detectChanges();
       });
+    } else if (this.version === '0.8') {
+      this.actionSub = this.messageProcessorV08.events.subscribe(event => {
+        const message = event.message;
+        if (message.userAction) {
+          const action = message.userAction as unknown as A2uiClientAction;
+          this.agentStub.handleAction(action);
+          this.eventsLog.unshift({timestamp: new Date(), action: {userAction: action} as any});
```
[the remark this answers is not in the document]
josemontespg: Done

renderers/angular/src/v0_9/catalog/basic/column.component.ts
```diff
@@ -85,7 +85,7 @@ export class ColumnComponent extends BasicCatalogComponent<typeof ColumnApi> {
 
   protected readonly normalizedChildren = computed(() => {
     if (this.isRepeating()) return [];
-    return this.children().map(child => {
+    return this.children().map((child: any) => {
```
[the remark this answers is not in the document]
josemontespg: Done

renderers/angular/src/v0_9/catalog/basic/row.component.ts
```diff
@@ -83,7 +83,7 @@ export class RowComponent extends BasicCatalogComponent<typeof RowApi> {
 
   protected readonly normalizedChildren = computed(() => {
     if (this.isRepeating()) return [];
-    return this.children().map(child => {
+    return this.children().map((child: any) => {
```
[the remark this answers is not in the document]
josemontespg: Done

renderers/angular/a2ui_explorer/src/app/agent-stub.service.ts
josemontespg: Done

renderers/angular/a2ui_explorer/src/app/agent-stub.service.ts
josemontespg: Good point. I refactored the code so that a version switch requires a full page reload. This simplified the code. Thanks!

josemontespg:
Refactored the code so that the version is resolved at page load, and a version change requires a full page reload. This way the code coupled to the specific version can be separated into independent files, making it easy to add/remove support for versions in the future, without changes to the main app.

renderers/angular/a2ui_explorer/src/app/agent-stub-v09.service.ts
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * A stub service that simulates an A2UI agent.
+ * It listens for actions and responds with data model updates or new surfaces.
+ * Supports the v0.9 A2UI spec.
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class AgentStubV09Service implements AgentStubService {
```
ditman: `AgentStubService` is an `abstract class`; I find it a little bit weird that we're treating it as an interface here. This should be `extends` and mark the methods as `override`. (Or the `AgentStubService` made an `interface`, pick your poison :P)

renderers/angular/a2ui_explorer/src/app/agent-stub.service.ts
```diff
@@ -14,126 +14,23 @@
...
 
 /**
- * Context for the 'update_property' event.
+ * Abstract base class for agent stub services.
  */
-interface UpdatePropertyContext {
-  path: string;
-  value: any;
-  surfaceId?: string;
+export abstract class AgentStubService {
```
ditman, line 24: (Should be an `interface`?)

renderers/angular/a2ui_explorer/src/app/agent-stub.service.ts
```diff
@@ -14,126 +14,23 @@
...
-            ]);
-          }
-        }
-      }
-    }
-    this.rendererService.processMessages(initialMessages);
-  }
-}
+export {AgentStubV09Service} from './agent-stub-v09.service';
+export {AgentStubV08Service} from './agent-stub-v08.service';
```
ditman: I wouldn't re-export the specific implementations from the interface file, I'd just import from here for the interface, and from the specific implementation file for the implementation I need.

renderers/angular/a2ui_explorer/src/app/demo.component.ts
```diff
@@ -312,6 +365,11 @@ import {ActionDispatcher} from './action-dispatcher.service';
         box-shadow: 0 4px 24px rgba(0, 0, 0, 0.4);
         padding: 24px;
       }
+      .rendered-content[data-version='v0.8'] {
```
ditman: I wouldn't use data-attributes to target styling. Maybe this can be a top-level `.protocol-version-08` class or similar when rendering the app?

renderers/angular/a2ui_explorer/src/app/demo.component.ts
```diff
@@ -496,8 +555,18 @@ import {ActionDispatcher} from './action-dispatcher.service';
...
+    provideMarkdownRenderer(),
     ActionDispatcher,
-    AgentStubService,
+    {
+      provide: AgentStubService,
+      useFactory: (v09: AgentStubV09Service, v08: AgentStubV08Service, version: Version) => {
+        return version === Version.V0_8 ? v08 : v09;
+      },
+      deps: [AgentStubV09Service, AgentStubV08Service, A2UI_VERSION],
+    },
```
ditman, line 573: This is a little bit weird, why not inject the right thing based off of A2UI_VERSION?

Wouldn't something like this work?

```
provide: AgentStubService,
useClass: A2UI_VERSION === Version.V0_8 ? AgentStubV08Service : AgentStubV09Service,
```

ditman: (Also something that I like is to provide the provider alongside the class, so from each of the files defining the agentstubv0Xservice files you could export a:

```
export const v08AgentStubService = {
  provide: AgentStubService,
  useClass: AgentStubV08Service,
}

// Similar for v09
```

And then here you just inject the correct provider object depending on the A2UI_VERSION variable value?

renderers/angular/a2ui_explorer/src/app/demo.component.ts
```diff
@@ -509,21 +578,48 @@ import {ActionDispatcher} from './action-dispatcher.service';
...
   private rendererService = inject(A2uiRendererService);
   private agentStub = inject(AgentStubService);
   private cdr = inject(ChangeDetectorRef);
 
-  examples = EXAMPLES;
-  selectedExample: Example | undefined = undefined;
-  surfaceId: string | null = null;
+  readonly version: Version = inject(A2UI_VERSION);
+  readonly examples: Array<Example | Example_08> = inject(A2UI_EXAMPLES);
+  selectedExample: Example | Example_08 | undefined = undefined;
```
ditman: I think it'd make sense to have a separate type alias for `Example | Example_08` so you don't have to do `Example | Example_08` wherever you want some example?

renderers/angular/a2ui_explorer/src/app/demo.component.ts
```diff
@@ -683,15 +729,19 @@ export class DemoComponent implements OnInit, OnDestroy {
...
       this.dataModelError = null;
-      const surface = this.rendererService.surfaceGroup?.getSurface(this.surfaceId!);
+      const surface = this.rendererService.surfaceGroup?.getSurface(this.surfaceId());
       surface?.dataModel.set('/', parsed);
     } catch (e) {
       this.dataModelError = e instanceof Error ? e.message : 'Invalid JSON';
       console.error(e);
     }
   }
+  onDataModelFocus() {
```
ditman, line 745: Is this file prettier'd? It's weird that there's no whitespace between L738 and L739

renderers/angular/a2ui_explorer/src/app/demo.component.ts
```diff
@@ -509,21 +578,48 @@ import {ActionDispatcher} from './action-dispatcher.service';
   ],
 })
 export class DemoComponent implements OnInit, OnDestroy {
+  readonly Version = Version;
```
ditman, line 587: I don't understand this `readonly Version = Version`; why do we need to shadow the type?

renderers/angular/a2ui_explorer/src/app/examples_injector.ts
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {InjectionToken, inject} from '@angular/core';
+import {EXAMPLES_V08, EXAMPLES_V09} from './generated/examples-bundle';
+import {Version, Example, Example_08} from './types';
+import {A2UI_VERSION} from './version_injector';
+
```
ditman, line 21: This seems like a very good file for the `export type A2uiExample = Example | Example_08` type alias :P

renderers/angular/a2ui_explorer/src/app/types.ts
```diff
@@ -15,15 +15,41 @@
...
+ */
+export interface Example_08 {
+  version: '0.8';
+  /** The name of the example, displayed in the sidebar. */
+  name: string;
+  /** A short description of what the example demonstrates. */
+  description: string;
+  /** The sequence of A2UI messages to send to the renderer. */
+  messages: ServerToClientMessage[];
+}
```
ditman, line 60: Hmm, I wonder if instead of having two different Example interfaces, we could have one that looks like:

```
export interface Example {
  version: Version;
  name: string;
  description: string;
  messages: A2uiMessage[] | ServerToClientMessage[];
}
```

We lose the bit of type safety where the version is exactly tied to the type of messages, but it should simplify the code a good amount. Also, since we have two different ways of processing messages, we don't really need to know the exact type of message? (because in the logic of the program the "version" would decide where to send the message for processing?

renderers/angular/a2ui_explorer/tsconfig.app.json
```diff
@@ -8,5 +8,5 @@
     "moduleResolution": "bundler"
   },
   "include": ["src/**/*.ts"],
-  "exclude": ["src/**/*.spec.ts"]
+  "exclude": ["src/**/*.spec.ts", "src/app/tests/utils/**"]
```
ditman, line 11: 💙

renderers/angular/scripts/generate-examples.mjs
```diff
@@ -66,40 +59,16 @@ function overrideMessagesCatalogId(messages) {
...
   for (const msg of messages) {
-    // For v0.9 (and up?)
     if (msg.createSurface && msg.createSurface.catalogId) {
       msg.createSurface.catalogId = overrideCatalogId(msg.createSurface.catalogId);
     }
-    // The minimal catalog examples in 0.8 contain a catalogId (but not the basic
-    // catalog ones). That's probably copy-pasta from when catalogIds were
-    // introduced later, as the v0.8 renderers didn't use catalogIds. We don't
-    // need to handle the overrides of the catalogId for the beginRendering
-    // messages from the v0.8 spec.
```
ditman, line 77: Is this not true anymore? Why remove the literature? :P

renderers/angular/a2ui_explorer/src/app/demo.component.ts
ditman: Wait a seeecond... Does this need to be modeled like this because we're reading the URL to figure out the version, but the dependency injection needs to be static? 😦

ditman (commented):
Some comments about minor OOP things and the types. Also I'm a little saddened by the way we must read the version query string parameter from the URL :/

renderers/angular/a2ui_explorer/src/app/agent-stub-v09.service.ts
josemontespg: Needs to be a class, because we use the symbol to provide the class. Changed the code to use `extends`

renderers/angular/a2ui_explorer/src/app/agent-stub.service.ts
josemontespg: Keeping as class, because we need the symbol when providing the class like 

```typescript
    {
      provide: AgentStubService,
      useFactory: (v09: AgentStubV09Service, v08: AgentStubV08Service, version: Version) => {
        return version === Version.V0_8 ? v08 : v09;
      },
      deps: [AgentStubV09Service, AgentStubV08Service, A2UI_VERSION],
    },
```

renderers/angular/a2ui_explorer/src/app/agent-stub.service.ts
josemontespg: Good point, done.

renderers/angular/a2ui_explorer/src/app/demo.component.ts
josemontespg: Yep, good point. Done.

renderers/angular/a2ui_explorer/src/app/demo.component.ts
josemontespg: Exactly, and I think it is better, because it allows us to provide a different version in unit tests in a way that doesn't use global state (`window`) to prevent leaking state between tests

renderers/angular/a2ui_explorer/src/app/demo.component.ts
josemontespg: To make the `Version` symbol available in the HTML template

renderers/angular/a2ui_explorer/src/app/examples_injector.ts
josemontespg: Sure, done.

renderers/angular/a2ui_explorer/src/app/types.ts
josemontespg: I crafted it this way for better type safety, and ease of removing old versions in the future. Downstream code has to anyway know how to handle each case. I don't think the structure here matters too much regarding code complexity.

renderers/angular/scripts/generate-examples.mjs
josemontespg: Still true. Reverted this. Good catch!

renderers/angular/a2ui_explorer/src/app/demo.component.ts
josemontespg: Exported a union type alias from a central location to improve this.

renderers/angular/a2ui_explorer/src/app/demo.component.ts
josemontespg: Having a line between class methods is not enforced currently. I'll check if this can be enforced repo-wide

renderers/angular/a2ui_explorer/src/app/agent-stub-v09.service.ts
ditman: Ah, because interfaces get erased at runtime? sounds good! Thanks for fixing!

renderers/angular/a2ui_explorer/src/app/demo.component.ts
ditman: I noticed this elsewhere too (when formatting markdown I think!)

renderers/angular/a2ui_explorer/src/app/agent-stub-v08.service.ts
```diff
@@ -0,0 +1,208 @@
...
+  providedIn: 'root',
+})
+export class AgentStubV08Service extends AgentStubService {
+  override eventsLog = signal<Array<{timestamp: Date; action: A2uiClientAction}>>([]);
+  override surfaceId = signal<string>('demo-surface', {equal: () => false});
+  override currentCreateSurfaceMessage = signal<ServerToClientMessage[] | null>(null);
+  private actionSub?: {unsubscribe: () => void};
+
+  override dataModel = computed(() => {
+    this.messageProcessorV08.version();
```
ditman, line 48:
```suggestion
```

Do we need this line? What a weird call!

ditman (approved):
Let me end my endless nitpicking, and let's go with this. Anything that breaks can be addressed later. PS: what is that call to `.version()` doing? :P

renderers/angular/a2ui_explorer/src/app/agent-stub-v08.service.ts
josemontespg: Mmm I don't know, I did not notice this line. Since it is a signal, calling it inside a `computed` has side effects. It is possible that this is somehow needed? I have not checked

Issue #1440, opened 2026-05-14
[ci] e2e_tests seems to be flaky


ditman:
It seems that the e2e_tests.sh script is flaky, for example:

* https://github.com/google/A2UI/actions/runs/25881317555/job/76061759428#step:9:87
* https://github.com/google/A2UI/actions/runs/25834367642/attempts/1
* https://github.com/google/A2UI/actions/runs/25758114131/job/75651870151#step:9:32
* Maybe more in [search](https://github.com/google/A2UI/actions/workflows/e2e_test.yaml?query=is%3Afailure+branch%3Amain+event%3Apush) (?)

Anyway, this is turning the tree red at seemingly random moments. This looks like an LLM capacity issue.

We need to fix/alleviate this in order to start enforcing "do not land on red" policies.

polina-c:
Statuses for test run:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#check-statuses-and-conclusions

ditman:
> Statuses for test run:
> 
> https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#check-statuses-and-conclusions

This seems unusable directly from github actions: https://github.com/orgs/community/discussions/9875

ditman:
Polina has a fix to make the affected test in an hourly schedule, rather than on post-submit:

* https://github.com/google/A2UI/pull/1441

PR #1441, opened 2026-05-14
Update e2e_test.yaml to run on schedule, because it is flacky.

polina-c:
Contribiutes to https://github.com/google/A2UI/issues/1440


ditman (approved):
This LGTM, how do we check these runs, now that they're not attached to "on push"?

Can it happen that the top of the tree becomes red/green on an hourly schedule?

polina-c:
Let's see how it goes.

PR #1443, opened 2026-05-15
Improve e2e and eval workflows.

### LINKED ISSUE #1440: [ci] e2e_tests seems to be flaky

polina-c:
Fixes https://github.com/google/A2UI/issues/1440

1. Update eval to run on schedule too.
2. Factor out issue creation to script and invoke from both workflows.
3. Update the script to check if an open issue with this label and title already exists.


scripts/create_issue.sh
```diff
@@ -0,0 +1,73 @@
...
+#
+# Arguments:
+#   $1 - WORKFLOW_NAME: The name of the workflow that failed (e.g., "Evals", "E2E tests").
+#   $2 - LABEL_NAME: The label to apply to the newly created issue (e.g., "eval_failure", "e2e_failure").
+#
+# Expected Environment Variables (provided by GitHub Actions):
+#   GITHUB_REPOSITORY, GITHUB_SHA, GITHUB_SERVER_URL, GITHUB_RUN_ID
+#   GH_TOKEN or GITHUB_TOKEN (required for the gh cli to authenticate)
+
+set -e
```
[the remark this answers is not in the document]
polina-c: let all tests to execute

.github/workflows/run_evals.yml
```diff
@@ -28,7 +28,11 @@ on:
       - "specification/v0_9/**"
       - ".github/workflows/run_evals.yml"
       - "agent_sdks/python/**"
-
+  # Run evals hourly on main, in order to:
+  # - catch regressions introduced by changes in environment
+  # - have more data for observing dynamics of performance degradation
+  schedule:
+    - cron: "0 * * * *" # hourly
```
gspencergoog: I don't think this is a good use of our AI budget.  The next time we submit something that triggers evals, it will catch any updated environment changes anyhow, so this just feels like a waste.

It also seems way too fast.  If we did this, I'd run it at most once a day, since otherwise a single change in environment could result in 48 issues being filed over a weekend.

.github/workflows/e2e_test.yaml
```diff
@@ -12,8 +12,11 @@ jobs:
   e2e_test:
     # Do not run on forked branches,
     # because the test does not have access to secrets in forks.
-    if: github.repository == 'google/a2ui'
+    if: github.repository == 'flutter/genui'
```
gspencergoog: Whoa, so I guess it wasn't running?

.github/workflows/docs.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       contents: write
       actions: read
 
-    if: github.repository == 'google/A2UI'
+    if: github.repository == 'flutter/genui'
```
gspencergoog: Where do our docs end up?  Is it published anywhere?

scripts/create_issue.sh
```diff
@@ -0,0 +1,73 @@
...
+
+BODY="The ${LOWER_WORKFLOW_NAME} workflow failed on main for commit ${GITHUB_SHA}.
+${PR_LINK}
+See logs: ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+
+echo "Creating issue..."
+gh issue create \
+  --title "$TITLE" \
+  --body "$BODY" \
+  --label "$LABEL_NAME"
```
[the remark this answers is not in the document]
gspencergoog: This is a reasonable enhancement. What do you think?  I'd check for more than the label, though: I'd check for one with the label and an identical title.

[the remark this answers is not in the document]
polina-c: The script is already checking for label. Added check for title. Thank you!

.github/workflows/run_evals.yml
polina-c: Switched to daily. Thanks.

.github/workflows/e2e_test.yaml
polina-c: nope

we are in A2UI

reverted

.github/workflows/docs.yml
polina-c: oh, I missed it

reverted

thank you

.github/workflows/run_evals.yml
```diff
@@ -28,7 +28,11 @@ on:
       - "specification/v0_9/**"
       - ".github/workflows/run_evals.yml"
       - "agent_sdks/python/**"
-
+  # Run evals hourly on main, in order to:
```
gspencergoog: I think this is daily now?

gspencergoog (approved):

PR #1445, opened 2026-05-15
ci: declare workflow-level `contents: read` on the 17 build/test workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on every workflow in `.github/workflows/`. None of them call a GitHub API beyond the initial checkout. They are all per-language build/test/lint workflows (composer, flutter, kotlin, lit, ng, python, react, web) plus license, private-index, and spec validation checks.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.


ditman:
Thanks @arpitjain099! The token should be read-only by default, but this change makes it more explicit!

ditman (approved):

PR #1462, opened 2026-05-20
feat(specification): add steps property to Slider in v0.10 basic catalog

### LINKED ISSUE #1156: [Slider] Slider needs a `steps` property
# Specification Version
v0_10

# According doc
https://github.com/google/A2UI/blob/main/specification/v0_10/json/basic_catalog.json#L562

# Issue
While implementing the Slider recently, our team noticed that the `steps` parameter was missing. This parameter is extremely useful in many practical scenarios (such as the temperature range for air conditioners or the speed range for fans), so we propose adding it as an optional parameter to the standard.

Similarly, we can also see this parameter in the Compose implementation on the Android platform.
https://developer.android.com/develop/ui/compose/components/slider#advanced-implementation

Image

# Break-down
To make it easier for large language models to understand, We propose that this value be described as “the difference between every two options,” to distinguish it from Compose’s native definition: “The number of notches on the slider to which the thumb snaps.”

Taking the Android platform using Jetpack Compose as an example:

If we want to draw [0,2,4,6,8,10], according to Compose’s native definition, the `steps` value should be 4, corresponding to the 4 points “2,4,6,8.”
However, to make the model easier to understand and less prone to errors, we propose that in this scenario, `steps` should be 2, representing “a separation every 2 numbers.” The rendering layer can calculate the required `steps` value for the control using `(maxValue - minValue) / steps - 1`.

This helps reduce the model's complexity and prevent errors.

# Follow-up
We may submit a PR regarding this issue later; for now, please feel free to discuss it here.

gspencergoog:
# Description

This change adds an optional `steps` property to the `Slider` component definition in A2UI protocol version 0.10. It is modeled after the number of discrete divisions in a slider range (similar to the divisions property in Flutter's Slider).

### Schema and test cases
* `specification/v0_10/catalogs/basic/catalog.json`: Add `steps` as an optional integer property (minimum 1) inside the `Slider` component definition.
* `specification/v0_10/test/cases/checkable_components.json`: Add three test cases to validate the `steps` property behavior (valid integer value, invalid string value, invalid zero/negative integer value).

Fixes https://github.com/google/A2UI/issues/1156


josemontespg (approved):
Nice!

PR #1474, opened 2026-05-20
feat: configure eslint lines-between-class-members rule repository-wide

### LINKED ISSUE #1473: Configure eslint on all typescript projects in the repo
Some projects are missing an eslint config

josemontespg:
Configure eslint lines-between-class-members rule repository-wide

This PR configures ESLint with the lines-between-class-members rule (requiring empty lines between class members) repository-wide.

Key Changes:
1. Configured existing ESLint projects (web_core, react, composer) to include the rule.
2. Set up new ESLint Flat Configurations with Google's GTS style guide for all other TypeScript packages that lacked linting (angular, lit, markdown-it, editor, inspector).
3. Configured the quotes rule in all configs to allow backticks (template literals) for standard string literals.
4. Ran formatting/linting repository-wide to align all files with the new rules.
5. Added ESLint verification steps to all corresponding GitHub Actions CI workflows (ng_build_and_test.yml, lit_build_and_test.yml, editor_build.yml, inspector_build.yml) to ensure the rules are fully enforced on push/PR.

Fixed #1473


josemontespg:
This approach works, but makes the formatter CI script much slower, because it has to run `npm install` on every project, then run the eslint.

This is a limitation of eslint. Until we figure out a way to run this in a quick way O(10s), I don't think we should include it in CI

PR #1475, opened 2026-05-20
refactor(angular): remove redundant version() call in dataModel computed property

josemontespg:
Removes the redundant `this.messageProcessorV08.version()` call in the `dataModel` computed property of `AgentStubV08Service`.


samples/client/flutter/restaurant_finder/e2e_test/test/infra_test.dart
josemontespg, line 1: This change was automatic by the formatter :(

ditman: Please, revert this file, you have this change in a separate PR, and left a question there about it :P

ditman (approved):
LGTM! Thanks for correcting this so quickly!

PR #1481, opened 2026-05-21
Fix test imports, add !, and restore OnPush change detection

josemontespg:
Fixes test imports to use namespace imports, adds non-null assertions (!) to restore successful compilation and passing tests under 1P / google3 environment, and removes ChangeDetectionStrategy overrides in tests while restoring components/samples to OnPush.


renderers/angular/src/v0_8/components/card.spec.ts
```diff
@@ -65,7 +65,6 @@ describe('Card Component', () => {
     })
       .overrideComponent(Card, {
         set: {
-          changeDetection: ChangeDetectionStrategy.Default,
```
ditman, line 68: Wasn't the problem of this change that the google3 defaults are different than the open source ones?

josemontespg: The problem is that ChangeDetectionStrategy.Default does not exist in google3. Only Eager and OnPush. 
These tests don't really need to override the change detection, so we can just remove these.

ditman (approved):
LGTM! Good luck with the OnPush!

PR #1494, opened 2026-05-26
ci: use dart-lang/setup-dart for formatting checks

josemontespg:
Addresses feedback from PR #1479. Replaces subosito/flutter-action with the lighter dart-lang/setup-dart action since we only need the Dart SDK for formatting.

This makes the formatting faster


renderers/angular/src/v0_9/catalog/basic/text.component.ts
```diff
@@ -108,7 +108,10 @@ export class TextComponent extends BasicCatalogComponent<typeof TextApi> {
   private markdownRenderer = inject(MarkdownRenderer);
 
   readonly variant = computed(() => this.props()['variant']?.value() || 'body');
-  readonly text = computed(() => this.props()['text']?.value() || '');
+  readonly text = computed(() => {
+    const val = this.props()['text']?.value();
+    return val == null ? '' : String(val);
```
ditman: Two things:

* Unrelated change, please move to a separate PR
* How can the 'text' property of the Text component be any other than a String already? ([Spec](https://github.com/google/A2UI/blob/0e0375749bf8a940d57969d612e74f7221ad246e/specification/v0_9/catalogs/basic/catalog.json#L23-L26).)

ditman (approved):
Let's go!

renderers/angular/src/v0_9/catalog/basic/text.component.ts
josemontespg: Not needed. Reverted this file.

PR #1495, opened 2026-05-26
test(angular): resolve flaky explorer location test using waitForCondition

josemontespg:
## Description
This PR resolves the flakiness of the Angular Explorer location selection tests.

### Root Cause
In zoneless Angular applications, when chips/buttons are clicked, the state changes are dispatched asynchronously to Preact signals, which batch and trigger effects in the next microtask queue. This means the DOM updates asynchronously. The previous test used a fixed `wait(100)` which flaked on slow CI runners if the microtask queue hadn't fully settled.
Additionally, the `waitFor` helper that was initially tried relied on catching expectation exceptions. In Jasmine, `expect()` does not throw an error on failure, so try-catch based wait helpers resolve immediately with silent failures.

### Fixes Applied
1. Introduced a robust `waitForCondition` helper in `test_utils.ts` that checks boolean assertions in a loop and only asserts via `expect` once resolved, making it fully compatible with Jasmine.
2. Re-wrote the flaky location selection test inside `30_live-invitation-builder.spec.ts` to use this `waitForCondition` helper dynamically.


renderers/angular/a2ui_explorer/src/app/tests/utils/test_utils.ts
```diff
@@ -68,3 +68,25 @@ export function wait(ms: number) {
...
+  while (true) {
+    if (condition()) {
+      return true;
+    }
+    if (Date.now() - start > timeout) {
+      return false;
+    }
+    await wait(interval);
+  }
+}
```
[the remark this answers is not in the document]
josemontespg: Done.

renderers/angular/a2ui_explorer/src/app/tests/utils/test_utils.ts
```diff
@@ -85,3 +85,25 @@ export function wait(ms: number) {
...
+ * Compatible with Jasmine because it uses standard boolean checks instead of throwing expectations inside the loop.
+ */
+export async function waitForCondition(
+  condition: () => boolean | Promise<boolean>,
+  timeout = 1000,
+  interval = 50,
+): Promise<boolean> {
+  const start = performance.now();
+  while (true) {
+    if (await condition()) {
```
ditman: What happens if `condition` never resolves? Does the `timeout` actually trigger, or does the test hang forever in this `await`?

ditman (commented):
QQ about the `waitForCondition` helper function.

renderers/angular/a2ui_explorer/src/app/tests/utils/test_utils.ts
josemontespg: Mm good point, it would wait forever. It is possible to race with the timeout, but it would be more complicated. Supporting async conditions is not actually needed for the current tests. Refactored to only support sync conditions.

ditman: You could have something like (note: completely untested and human hallucinated):

```javascript
function timeout(millis) {
  return new Promise((_, reject) => {
    window.setTimeout(() => {
      reject(); // timeout exceeded
    }, millis);
  });
}
```

And then:

```javascript
await Promise.race([condition(), timeout(1000)])
```

Anyway, sync sounds good to me too!

renderers/angular/a2ui_explorer/src/app/tests/utils/test_utils.ts
```diff
@@ -85,3 +85,26 @@ export function wait(ms: number) {
...
+  interval = 50,
+): Promise<boolean> {
+  const start = performance.now();
+  while (true) {
+    if (condition()) {
+      return true;
+    }
+    if (performance.now() - start > timeout) {
+      return false;
+    }
```
ditman, line 107: You could also do:

```javascript
// Periodically poll `condition()` to see if it becomes true.
while (performance.now() - start <= timeout) {
  if (condition()) {
    return true;
  }
  await wait(interval);
}
return false;
```

And then you don't need the explicit return inside the loop?

ditman (approved):
Proposed a small logic change to the while loop, but other than that, I hope this removes the flake!

renderers/angular/a2ui_explorer/src/app/tests/utils/test_utils.ts
josemontespg: That is a clean and readable suggestion!

However, there is a subtle but important trade-off here regarding "last-chance" condition checks on the timeout boundary.

josemontespg:
Undid the unrelated formatting, thanks for the review!

PR #1496, opened 2026-05-26
fix(infra): run dart pub get before formatting to avoid resolution warnings

josemontespg:
### Description

When running `./scripts/fix_format.sh` (or during CI format enforcement checks) in a fresh clone or worktree, `dart format` can print package resolution warnings:
`Failed to resolve package URI "package:lints/recommended.yaml" in include at "Instance of 'IOFileSystemPath'".`

This occurs because Dart dependencies must be resolved (fetching packages like `lints`) so the Dart analyzer can locate referenced linters/rule files listed in the `analysis_options.yaml`.

This PR updates `scripts/fix_format.sh` to run `dart pub get` prior to formatting. If the project is in an offline environment or `dart pub get` fails, it logs a warning and continues formatting gracefully.


scripts/fix_format.sh
```diff
@@ -51,6 +51,8 @@ echo "Running Dart format..."
 cd "$REPO_ROOT"
 # Check if dart is available before running
 if command -v dart >/dev/null 2>&1; then
+  echo "Resolving Dart workspace dependencies..."
+  dart pub get || echo "Warning: 'dart pub get' failed. Formatting might have package resolution warnings."
```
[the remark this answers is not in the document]
ditman: 1. I'd let `dart` handle that for me
2. I'm not sure that'd be needed just to run `dart format` later

scripts/fix_format.sh
ditman, line 43: Unrelated, but instead of having to always cd into `$REPO_ROOT`, the script could `pushd agent_sdks/python` then `popd` at the end of the block, so that way the shell kind of guarantees that we're entering (and exiting) the directory where it needs to work.

ditman (approved):
This LGTM, I'm not sure we need to `flutter pub get` just to run `dart format`

ditman:
Would it make sense to merge this PR with the PR where you're replacing the subosito action?

scripts/fix_format.sh
josemontespg: Interesting, will keep this in mind for the next time I update the script

josemontespg:
The other PR was merged, and this one is updated and using dart format. All looks good, sending it!

PR #1502, opened 2026-05-28
refactor(angular): use declarative HTML for known Text variants

josemontespg:
## What

Refactors Angular 0.9 Text component to render headings (h1-h5) and caption variants using declarative HTML elements in the template with {{ text }} notation instead of using the markdown renderer.

Markdown rendering is still used for unknown/default variants (like 'body').

## Why?

In 1P code, there are some builds that don't allow the g3mark implementation of the markdown renderer, due to a strict CSP (blocking 'unsafe-eval'). The only reason to provide a markdown renderer is because the Text component requires some kind of markdown renderer, even to display simple variants like headings. With the changes in this PR, apps can use `provideMarkdownRenderer(async text => text)` and still get all simple variants working correctly.

Tests have been updated to assert this new behavior.

TAG=agy
CONV=3f5af79d-d201-4229-8c8a-40e91c4eb343

### DIFF
--- a/renderers/angular/src/v0_9/catalog/basic/text.component.spec.ts
+++ b/renderers/angular/src/v0_9/catalog/basic/text.component.spec.ts
@@ -60,8 +60,8 @@ describe('TextComponent', () => {
     await fixture.whenStable();
     fixture.detectChanges();
 
-    const span = fixture.debugElement.query(By.css('span'));
-    expect(span.nativeElement.innerHTML.trim()).toBe('<p>Hello World</p>');
+    const element = fixture.debugElement.query(By.css('.a2ui-text'));
+    expect(element.nativeElement.innerHTML.trim()).toBe('<p>Hello World</p>');
     expect(mockMarkdownRenderer.render).toHaveBeenCalledWith('Hello World');
   });
 
@@ -75,9 +75,9 @@ describe('TextComponent', () => {
     fixture.detectChanges();
 
     expect(mockMarkdownRenderer.render).not.toHaveBeenCalled();
-    const span = fixture.debugElement.query(By.css('span.a2ui-text.h1'));
-    expect(span).toBeTruthy();
-    const h1 = span.query(By.css('h1'));
+    const element = fixture.debugElement.query(By.css('.a2ui-text.h1'));
+    expect(element).toBeTruthy();
+    const h1 = element.query(By.css('h1'));
     expect(h1).toBeTruthy();
     expect(h1.nativeElement.textContent.trim()).toBe('Heading');
   });
@@ -92,9 +92,9 @@ describe('TextComponent', () => {
     fixture.detectChanges();
 
     expect(mockMarkdownRenderer.render).not.toHaveBeenCalled();
-    const span = fixture.debugElement.query(By.css('span.a2ui-text.caption'));
-    expect(span).toBeTruthy();
-    const em = span.query(By.css('em'));
+    const element = fixture.debugElement.query(By.css('.a2ui-text.caption'));
+    expect(element).toBeTruthy();
+    const em = element.query(By.css('em'));
     expect(em).toBeTruthy();
     expect(em.nativeElement.textContent.trim()).toBe('Caption');
   });
@@ -109,9 +109,9 @@ describe('TextComponent', () => {
     fixture.detectChanges();
 
     expect(mockMarkdownRenderer.render).not.toHaveBeenCalled();
-    const span = fixture.debugElement.query(By.css('span.a2ui-text.h2'));
-    expect(span).toBeTruthy();
-    const h2 = span.query(By.css('h2'));
+    const element = fixture.debugElement.query(By.css('.a2ui-text.h2'));
+    expect(element).toBeTruthy();
+    const h2 = element.query(By.css('h2'));
     expect(h2).toBeTruthy();
     expect(h2.nativeElement.textContent.trim()).toBe('Heading');
   });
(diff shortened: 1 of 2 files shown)


renderers/angular/src/v0_9/catalog/basic/text.component.ts
```diff
@@ -110,39 +125,23 @@ export class TextComponent extends BasicCatalogComponent<typeof TextApi> {
   readonly variant = computed(() => this.props()['variant']?.value() || 'body');
   readonly text = computed(() => this.props()['text']?.value() || '');
 
+  readonly isKnownVariant = computed(() => {
+    const v = this.variant();
+    return ['h1', 'h2', 'h3', 'h4', 'h5', 'caption'].includes(v);
+  });
```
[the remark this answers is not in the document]
josemontespg: Done.

renderers/angular/src/v0_9/catalog/basic/text.component.ts
```diff
@@ -38,7 +38,22 @@ import {TextApi} from '@a2ui/web_core/v0_9/basic_catalog';
...
+          @case ('h3') { <h3>{{ text() }}</h3> }
+          @case ('h4') { <h4>{{ text() }}</h4> }
+          @case ('h5') { <h5>{{ text() }}</h5> }
+          @case ('caption') { <em>{{ text() }}</em> }
+        }
+      </span>
+    } @else {
+      <span [class]="'a2ui-text ' + variant()" [innerHTML]="resolvedText()"></span>
+    }
+  `,
```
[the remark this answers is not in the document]
josemontespg: Sure, but this PR is a pure refactor. Should not result in a change in DOM output. We can look at a11y holistically in the future.

renderers/angular/src/v0_9/catalog/basic/text.component.spec.ts
```diff
@@ -74,7 +74,12 @@ describe('TextComponent', () => {
     await fixture.whenStable();
     fixture.detectChanges();
 
-    expect(mockMarkdownRenderer.render).toHaveBeenCalledWith('# Heading');
+    expect(mockMarkdownRenderer.render).not.toHaveBeenCalled();
+    const span = fixture.debugElement.query(By.css('span.a2ui-text.h1'));
+    expect(span).toBeTruthy();
+    const h1 = span.query(By.css('h1'));
+    expect(h1).toBeTruthy();
+    expect(h1.nativeElement.textContent.trim()).toBe('Heading');
```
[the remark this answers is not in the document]
josemontespg: Done.

renderers/angular/src/v0_9/catalog/basic/text.component.spec.ts
```diff
@@ -60,8 +60,8 @@ describe('TextComponent', () => {
     await fixture.whenStable();
     fixture.detectChanges();
 
-    const span = fixture.debugElement.query(By.css('span'));
-    expect(span.nativeElement.innerHTML.trim()).toBe('<p>Hello World</p>');
+    const element = fixture.debugElement.query(By.css('.a2ui-text'));
+    expect(element.nativeElement.innerHTML.trim()).toBe('<p>Hello World</p>');
```
ditman: The way we're asserting this here is inconsistent with how we're asserting it in the h1 case below (this should try to do `const p = element.query(By.css('p'))`?)

renderers/angular/src/v0_9/catalog/basic/text.component.ts
```diff
@@ -38,7 +38,34 @@ import {TextApi} from '@a2ui/web_core/v0_9/basic_catalog';
 @Component({
   selector: 'a2ui-v09-text',
   standalone: true,
-  template: ` <span [class]="'a2ui-text ' + variant()" [innerHTML]="resolvedText()"> </span> `,
+  template: `
+    @if (isKnownVariant()) {
+      <span [class]="'a2ui-text ' + variant()">
```
ditman, line 43: In general, wrapping a block-level element (like h1 or p) inside an inline element (like span) is not correct HTML5.

Consider removing the wrapping span here, and maybe adding the a2ui-text + variant class to the actual tag we want to render?

renderers/angular/src/v0_9/catalog/basic/text.component.ts
```diff
@@ -107,42 +134,34 @@ import {TextApi} from '@a2ui/web_core/v0_9/basic_catalog';
...
-          value = `#### ${text}`;
-          break;
-        case 'h5':
-          value = `##### ${text}`;
-          break;
-        case 'caption':
-          value = `*${text}*`;
-          break;
+      if (this.isKnownVariant()) {
+        return;
```
ditman, line 161: Add a comment here explaining why this effect is aborted when isKnownVariant

ditman (approved):
LGTM, but can use some love!

renderers/angular/src/v0_9/catalog/basic/text.component.spec.ts
josemontespg: Done.

renderers/angular/src/v0_9/catalog/basic/text.component.ts
josemontespg: I agree, but it is what we are currently doing implicitly with [innerHtml], and I'd like to reduce the amount of change

josemontespg: Left a comment in code to refactor later

renderers/angular/src/v0_9/catalog/basic/text.component.ts
josemontespg: Good point. I improved the name of the variables. I think now the comment is not needed.

josemontespg:
Thanks for the quick review!

PR #1503, opened 2026-05-28
test(angular): resolve flaky explorer tests and fix text renderer stringification

josemontespg:
Addresses reviewer feedback from PR #1494 by moving these unrelated fixes to a separate PR.


josemontespg:
The text value should already be a string based on the spec. This PR is not needed.

PR #1518, opened 2026-06-02
refactor: Refactor FunctionCall wire schema and align protocol specification

### LINKED ISSUE #718: Consider simplifying function "call sites" at the transport level or in inference formats
I noticed that when defining function calls, we use JSON schema with constants etc to enforce that certain aspects of the function signature and configuration match the context in which it is being called. For example, to ensure that a `checks` function is actually a boolean returning function, or that a remotely triggered function has `callableFrom` set to `clientOrRemote` or `remoteOnly`.

This is awesome in that it achieves much of the validation we need using only JSON schema, so there is no need to describe additional rules to the LLM, and implement additional validation logic.

However, it does have the downside that function call definitions include redundant data which the client could already infer.

I think transporting the redundant data is no big deal, and it's nice that it makes validation very easy. But for Gen UI inference tasks, it seems like a big waste of output tokens. I wonder if there is a way we can tweak the schema, at least for inference, to avoid this redundancy.

For example in a CallFunctionMessage:
```json
"callFunction": {
   "call": "required",
    "args": { "value": "bar" },
    "returnType": "boolean",
    "callableFrom": "clientOrRemote"
}
```

Here, the return type and `callableFrom` don't really need to be transported, because those are just properties of the function declaration which both the client and server already know after the catalog has been negotiated.

See the definition:
```json
    "required": {
      "type": "object",
      "description": "Checks that the value is not null, undefined, or empty.",
      "properties": {
        "call": {
          "const": "required"
        },
        "args": {
          "type": "object",
          "properties": {
            "value": {
              "description": "The value to check."
            }
          },
          "required": ["value"],
          "additionalProperties": false
        },
        "returnType": {
          "const": "boolean"
        }
      },
      "required": ["call", "args"],
    },
```

Another example for updateComponents:
```json
"updateComponents": {
  "surfaceId": "test_surface",
  "components": [
    {
      "id": "tf1",
      "component": "TextField",
      "label": "Email",
      "value": { "path": "/formData/email" },
      "checks": [
        {
          "condition": {
            "call": "required",
            "args": { "value": { "path": "/formData/email" } },
            "returnType": "boolean"
          },
          "message": "Email is required"
        },
        {
          "condition": {
            "call": "email",
            "args": { "value": { "path": "/formData/email" } },
            "returnType": "boolean"
          },
          "message": "Must be valid email"
        }
      ]
    }
  ]
}
```

gspencergoog:
# Description

This pull request refactors the `FunctionCall` schema definition in the A2UI v0.10 specification to completely remove the redundant `callableFrom` and `returnType` properties from the wire-level payload. Enforcements and validations (such as execution boundaries and return type correctness) are moved entirely to client-side runtime verification against catalog registries rather than being validated on the wire.

Additionally, this pull request adds support for static `callableFrom` and `returnType` metadata inside the client capabilities catalog definition schema, aligns the official protocol documentation (`a2ui_protocol.md`) with all updated schemas, and comprehensively updates all test cases, JSONL streams, and basic catalog examples to remove these obsolete wire-level fields.

### Schemas

- **`common_types.json`**:
  - Removed the `callableFrom` and `returnType` properties from the core `FunctionCall` schema definition.
  - Removed static `returnType` constraints from `DynamicString`, `DynamicNumber`, `DynamicBoolean`, and `DynamicStringList` schema unions so they accept any `FunctionCall` on the wire and rely on client-side runtime type checking.
- **`server_to_client.json`**: Simplified the `CallFunctionMessage` payload's `callFunction` property, removing both `callableFrom` and `returnType` overrides and required constraints.
- **`testing_catalog.json` & `catalogs/minimal/catalog.json` & `catalogs/basic/catalog.json`**: Removed the `callableFrom` and `returnType` property definitions from catalog function schemas, keeping them strictly as validators for wire payloads.
- **`client_capabilities.json`**: Declared static `callableFrom` and `returnType` properties inside the `$defs/FunctionDefinition` catalog schema. This allows clients to statically advertise their custom functions' boundaries and return types to the server as metadata, separate from the wire payload.

### Test cases and Examples

- **Test Suites (`cases/*.json`, `cases/*.jsonl`)**: Performed an automated smart cleanup of positive and unrelated test payloads across all test cases (e.g., `checkable_components.json`, `button_checks.json`, `call_function_message.json`, `function_catalog_validation.json`) to strip `returnType` and `callableFrom` fields, while preserving their values in dedicated negative tests checking for schema rejection of unknown properties.
- **Catalog Examples (`catalogs/basic/examples/*.json`)**: Cleaned up all 36 basic catalog examples to strip the obsolete wire-level `returnType` field from their function call declarations.

### Documentation (`a2ui_protocol.md`)

- **`callFunction` specifications**: Aligned the `callFunction` property documentation and examples to remove wire-level `returnType` properties and updated the JSONL stream example.
- **Server-to-client envelopes**: Aligned the introductory list under `## Envelope message structure` to include the `callFunction` key. Added a new section detailing the `callFunction` message, its properties, and payload examples. Clarified how the client runtime extracts the `callableFrom` configuration from catalog definitions/annotations (defaulting to `"clientOnly"`).
- **Client-to-server event messages**: Restructured and expanded the documentation for client events. Fully detailed the property lists, mutual exclusion rules, and JSON examples for `action` events (including `wantResponse` and `actionId`), `functionResponse` execution results, and client `error` messages.
- **Capabilities and metadata**: Promoted transport-level capability and data model exchanges to a dedicated section. Aligned the documented properties of `server_capabilities.json`, `client_capabilities.json`, and `client_data_model.json` to correctly describe version-wrapped objects (`"v0.10"`) and the required `"version"` property. Documented that custom inline catalog function definitions support static `callableFrom` and `returnType` metadata to advertise function execution details.

## Verification results

All 107 test cases in the A2UI v0.10 schema validation suite pass successfully. The automated tests validated the modified JSON schemas against the updated test files under `/specification/v0_10/test/cases/`.

Fixes https://github.com/google/A2UI/issues/718


specification/v0_10/docs/a2ui_protocol.md
```diff
@@ -324,6 +324,70 @@ Server responds with:
...
+- `callFunction` (object, required): The description of the function call.
+  - `call` (string, required): The registered name of the function to execute.
+  - `args` (object, optional): Arguments passed to the function, as defined by its schema in the catalog.
+  - `returnType` (string, required): The expected return type of the function call. Snaps to one of `array`, `boolean`, `number`, `object`, `string`, or `void`.
+
+**Security Boundaries and Verification:**
+
+Execution boundary verification (`remoteOnly` vs `clientOnly`) is enforced strictly at runtime by the client application:
+- When a client receives a `callFunction` message, it MUST look up the requested function name in its active catalog registry.
+- If the requested function is configured in the catalog as `clientOnly`, or if the function is not registered at all, the client MUST immediately reject the call and return a client-to-server `error` message with `code: "INVALID_FUNCTION_CALL"`.
```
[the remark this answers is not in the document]
gspencergoog: No, this is incorrect.  There is still a property for declaring `callableFrom` in `FunctionDefinition` (client_capabilities.json) which is can be used for defining functions in catalogs.

specification/v0_10/docs/a2ui_protocol.md
```diff
@@ -324,6 +324,70 @@ Server responds with:
...
+```
+
+If the server attempts to call a `clientOnly` function (e.g., a local-only component validator), the client responds with an error:
+
+```json
+{
+  "version": "v0.10",
+  "error": {
+    "code": "INVALID_FUNCTION_CALL",
+    "message": "Function 'validateLocalInput' is clientOnly and cannot be invoked remotely.",
```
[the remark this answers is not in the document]
gspencergoog: Added a note about how the client determines status. (from the catalog definition)

specification/v0_10/docs/a2ui_protocol.md
```diff
@@ -324,6 +324,70 @@ Server responds with:
...
+This message is sent by the server to execute a function registered on the client. Functions are catalog-defined abstractions that avoid sending raw executable code across the wire.
+
+**Properties:**
+
+- `functionCallId` (string, required): A unique identifier for this invocation instance. The client MUST copy this ID verbatim into the subsequent `functionResponse` or `error` message.
+- `wantResponse` (boolean, optional, default `false`): Specifies whether the server expects a response payload back from the client. If set to `true`, the client MUST reply with either a `functionResponse` or an `error` message.
+- `callFunction` (object, required): The description of the function call.
+  - `call` (string, required): The registered name of the function to execute.
+  - `args` (object, optional): Arguments passed to the function, as defined by its schema in the catalog.
+  - `returnType` (string, required): The expected return type of the function call. Snaps to one of `array`, `boolean`, `number`, `object`, `string`, or `void`.
```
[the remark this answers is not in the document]
gspencergoog: Fixed

specification/v0_10/json/client_capabilities.json
```diff
@@ -46,6 +46,12 @@
           "type": "string",
           "enum": ["string", "number", "boolean", "array", "object", "any", "void"],
           "description": "The type of value this function returns."
+        },
+        "callableFrom": {
+          "type": "string",
+          "enum": ["clientOnly", "remoteOnly", "clientOrRemote"],
```
jacobsimionato: Can't this also be called from either server or client? Add clientOrRemote?

specification/v0_10/catalogs/basic/catalog.json
```diff
@@ -994,9 +1103,6 @@
           },
           "required": ["value", "other"],
           "unevaluatedProperties": false
-        },
-        "returnType": {
```
jacobsimionato, line 998: Now this function is not declaring its return type anywhere, so how can client libraries validate that the return type matches the requirement?

specification/v0_10/json/client_capabilities.json
jacobsimionato: This declares a function definition, but looking closer, it seems like this schema is actually not validated anywhere.

The basic catalog is supposed to conform to the Catalog schema below which includes this, but it seems like we don't actually validate that anywhere, because I see returnType is required here, yet it's not declared in the basic catalog JSON document.

I think probably what we really need is for each FunctionDefinition object to include a FunctionCall schema (referenced from common_types) inside which contains the name and args, and also extra metadata like ReturnType and CallableFrom that is used by the client for validation and the LLM to understand the call signature. Then, anyFunction etc in basic_catalog needs to combine together the inner FunctionCall schemas, rather than the outer FunctionDefinitions.

Hmmm maybe we're hitting the edge of what JSON schema validation can get us...

tmp/evaluation_v0_10_features.md
jacobsimionato, line 1: Remove this temp file?

specification/v0_10/json/client_capabilities.json
gspencergoog: > Can't this also be called from either server or client? Add clientOrRemote?

"clientOnly" means it can only be called on the client. "remoteOnly" means it can only be called on the server, and "clientOrRemote" means it is available on both.

I'm not sure I understand your comment, though.  `clientOrRemote` is already available.

specification/v0_10/catalogs/basic/catalog.json
gspencergoog: Right, sorry, got too aggressive removing that.

specification/v0_10/json/client_capabilities.json
gspencergoog: Yes, good point. Okay, I've modified the schema so it is suitable to use for validation and added validation tests for the basic and minimal catalogs to make sure that they pass validation against Catalog.

When we implement this for the core libraries, do you think we should add automatic validation for advertised catalogs?

I think JSON validation still good enough, or at least a part of a functional solution, but you're right, we're getting close to needing more.

specification/v0_10/catalogs/minimal/catalog.json
```diff
@@ -164,8 +164,7 @@
           },
           "required": ["value"],
           "unevaluatedProperties": false
-        },
```
jacobsimionato, line 167: I think this should be reinstated

specification/v0_10/json/client_capabilities.json
jacobsimionato: Ummm, I don't know what I was thinking here. I guess I just didn't read it properly??

specification/v0_10/json/server_to_client.json
```diff
@@ -159,17 +159,9 @@
           "allOf": [
```
jacobsimionato, line 159: I think allOf now only has one child, so you can move the reference to FunctionCall a step higher.

specification/v0_10/json/client_capabilities.json
```diff
@@ -25,36 +25,85 @@
...
-      "required": ["name", "parameters", "returnType"],
-      "additionalProperties": false
+      "required": ["type", "properties", "required"]
+    },
+    "FunctionDefinition": {
+      "type": "object",
+      "description": "Describes a function's validation schema and interface metadata.",
+      "allOf": [
+        {
+          "$ref": "#/$defs/FunctionCallValidationSchema"
```
jacobsimionato, line 76: Actually, how about just inlining FunctionCallValidationSchema into FunctionDefinition rather than having it separate. 

I think I suggested this structure, but I was confused before. I hadn't realised that FunctionCallValidationSchema and FunctionCall are different things, because the first one is for a function declaration (e.g. the types of arguments) and the second is for a function call.

tmp/refactor_function_call_pr_description.md
jacobsimionato, line 1: Remove tmp file!

jacobsimionato (approved):

PR #1522, opened 2026-06-02
feat(v0_10): Add optional instructions field to Catalog schema and remove rules.txt

### LINKED ISSUE #649: add instructions to a catalog definition
so the freetext instructions is customization along with the catalog

gspencergoog:
Add an `instructions` field to the `Catalog` definition in the v0.10 client capabilities schema. This allows embedding catalog-specific instructions or design guidelines directly within the catalog schema itself.

Inline revised contents of the old basic catalog `rules.txt` file into the new `instructions` field in the basic `catalog.json` schema and delete the separate `rules.txt` file.

Refactor the evaluation harness to load catalog rules directly from the schema instead of reading the deprecated `basic_catalog_rules.txt` file.

Fixes #649


specification/v0_10/eval/src/index.ts
```diff
@@ -353,15 +353,7 @@ async function main() {
...
-  const catalogRulesPath = path.join(__dirname, '../../json/basic_catalog_rules.txt');
-  let catalogRules: string | undefined;
-  if (fs.existsSync(catalogRulesPath)) {
-    catalogRules = fs.readFileSync(catalogRulesPath, 'utf-8');
-  } else {
-    logger.warn(
-      `Catalog rules file not found at ${catalogRulesPath}. Proceeding without specific catalog rules.`,
-    );
-  }
+  const catalogRules = schemas['catalogs/basic/catalog.json']?.instructions;
```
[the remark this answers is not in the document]
gspencergoog: This is OK.  Instructions are optional (the rules.txt file should have been too).

specification/v0_10/catalogs/basic/catalog.json
```diff
@@ -4,6 +4,7 @@
   "title": "A2UI Basic Catalog",
   "description": "Unified catalog of basic A2UI components and functions.",
   "catalogId": "https://a2ui.org/specification/v0_10/catalogs/basic/catalog.json",
+  "instructions": "**REQUIRED PROPERTIES:** You MUST include ALL required properties for every component, even if they are inside a template or will be bound to data.",
```
jacobsimionato: I get the desire to show how this field can be used, but I think instructions like this which would apply to any catalog should be included in the system prompt embedded in our SDKs, rather than in each catalog separately.

Maybe this could have the stuff about how to use images etc?

specification/v0_10/json/client_capabilities.json
```diff
@@ -55,10 +55,30 @@
       "type": "object",
       "description": "A collection of component and function definitions.",
       "properties": {
+        "$schema": {
```
jacobsimionato: Should this be a constant referring to this schema?

specification/v0_10/json/client_capabilities.json
```diff
@@ -55,10 +55,30 @@
...
+          "type": "string",
+          "description": "The JSON Schema meta-schema URI."
+        },
+        "$id": {
```
jacobsimionato: I think the catalog is actually not a pure schema - it's a JSON object which contains schemas within it, so I don't think it needs this. It has a catalogId already.

specification/v0_10/catalogs/basic/catalog.json
gspencergoog: >I get the desire to show how this field can be used, but I think instructions like this which would apply to any catalog should be included in the system prompt embedded in our SDKs, rather than in each catalog separately.

Yeah, makes sense.  Maybe I can say something about using Row and Column for layout?  It needs to be something that is specific to the catalog, but not specific to one component.

> Maybe this could have the stuff about how to use images etc?

No, because that should be just part of the `Image` component description.

specification/v0_10/json/client_capabilities.json
gspencergoog: Okay, removed.

specification/v0_10/json/client_capabilities.json
gspencergoog: No, it should just be removed.

specification/v0_10/catalogs/basic/catalog.json
gspencergoog: I replaced it with a comment about using Row and Column.

specification/v0_10/docs/a2ui_protocol.md
```diff
@@ -370,6 +370,16 @@ This structure is designed to be both flexible and strictly validated.
...
 
+#### Catalog structure
+
+Every catalog follows the standard `Catalog` object definition:
+
+- **catalogId** (string, required): A unique identifier URI for this catalog.
+- **instructions** (string, optional): Markdown-formatted design principles, rules, or developer guidelines specific to this catalog. These rules guide LLMs when generating UI layouts under this catalog.
+- **components** (object, optional): A map of supported UI components, where each key is the component type (e.g., `Text`) and its value is its JSON Schema definition.
+- **functions** (array, optional): A list of client-side validation or utility functions supported by the catalog.
+- **theme** (object, optional): A schema defining the catalog's customizable visual properties.
```
jacobsimionato: We are replacing this with configuration, right? Perhaps omit it here and then add it back in that PR?

jacobsimionato (approved):
This looks great to me. How about we get @wrenj or @jgindin  to take a look because I know they were pushing in a different direction, so it'd be great to take this opportunity to get aligned.

specification/v0_10/docs/a2ui_protocol.md
gspencergoog: I already merged it.  I'll fix it here.  It's called `surfaceProperties` now.

specification/v0_10/docs/a2ui_protocol.md
```diff
@@ -370,6 +370,16 @@ This structure is designed to be both flexible and strictly validated.
 
 The set of available UI components and functions is defined in a **Catalog**. The basic catalog is defined in [`catalogs/basic/catalog.json`]. While the Basic Catalog is useful for starting out, most production applications will define their own catalog to reflect their specific design system. The server must generate messages that conform to the catalog understood by the client.
 
+#### Catalog structure
```
jgindin, line 440: I know we haven't been doing this, but if we could reformat this (and other markdown files) to have a "reasonable" line length (80? 100 chars?) it would make it far easier to read.

specification/v0_10/catalogs/basic/catalog.json
```diff
@@ -4,6 +4,7 @@
   "title": "A2UI Basic Catalog",
   "description": "Unified catalog of basic A2UI components and functions.",
   "catalogId": "https://a2ui.org/specification/v0_10/catalogs/basic/catalog.json",
+  "instructions": "For layout, use the Row and Column components to organize other components.",
```
jgindin: I still don't think it's reasonable to expect that a full set of instructions for how to use the components in a design system can be expressed in a single string like this.

I don't think this needs to be a URL -- why can't it just be the name of a file which sits next  to this catalog just like the old `rules.txt` did?  

Maybe there's a syntax that, if the user has some very short instructions, they actually could be inline here? But I think that probably just makes everything a bit more complicated.

wrenj: we could do this bundling of separate md file to the catalog as part of assemble_catalog, thats where we inline other refs

https://github.com/a2ui-project/a2ui/tree/main/tools/build_catalog

specification/v0_10/docs/a2ui_protocol.md
gspencergoog: Sure.

specification/v0_10/catalogs/basic/catalog.json
gspencergoog: Okay, then how about we just abandon this change, and instead just rename the `rules.txt` to be `catalog_rules.md`?

gspencergoog:
Okay, I've updated this to rename `rules.txt` to `instructions.md`, and updated the "instructions" field in the catalog definition to be the relative URI to that file.  Sound good?

specification/v0_10/catalogs/basic/catalog.json
```diff
@@ -4,6 +4,7 @@
   "title": "A2UI Basic Catalog",
   "description": "Unified catalog of basic A2UI components and functions.",
   "catalogId": "https://a2ui.org/specification/v0_10/catalogs/basic/catalog.json",
+  "instructions": "instructions.md",
```
wrenj: maybe "./instructions.md" for clarity?

specification/v0_10/json/client_capabilities.json
```diff
@@ -101,13 +101,24 @@
...
+        },
+        "description": {
+          "type": "string",
+          "description": "A human-readable description of the catalog."
+        },
         "catalogId": {
           "type": "string",
           "description": "Unique identifier for this catalog."
         },
+        "instructions": {
```
wrenj, line 117: instructions_path ? or instructions_uri for clarity?

specification/v0_10/catalogs/basic/catalog.json
wrenj: although not needed technically

specification/v0_10/catalogs/basic/catalog.json
```diff
@@ -4,6 +4,7 @@
   "title": "A2UI Basic Catalog",
   "description": "Unified catalog of basic A2UI components and functions.",
```
wrenj, line 5: reminder update the sdk to handle it

specification/v0_10/json/client_capabilities.json
gspencergoog: I wanted to leave this as just "instructions" because we might change the format in the future (to inline it, or to make it something other than a URI).

specification/v0_10/catalogs/basic/catalog.json
gspencergoog: For now none of the v1.0 changes are implemented in the SDK, so we have to implement all of them once it shifts to "candidate" mode.  But yes, good to have the reminder.

specification/v0_10/catalogs/basic/catalog.json
gspencergoog: Good idea.  Done.

PR #1524, opened 2026-06-02
feat(spec): Add optional initialState to CreateSurfaceMessage in v0.10

### LINKED ISSUE #1239: createSurface that includes initial components and data model
Make the common case simple and let users specify the initial components and data model on createSurface

gspencergoog:
# Description

This pull request fixes https://github.com/google/A2UI/issues/1239 by adding an optional `initialState` property to the `createSurface` message in the v0.10 specification. This allows sending the initial component tree and data model directly during surface creation, reducing network round-trips and latency.

## Technical details

To prevent duplicating properties (such as repeating the `surfaceId`) and avoid unnecessary nested fields, the message schemas were refactored.

### Schema refactoring (`server_to_client.json`)

*   Extracted the inner payload schema of `updateComponents.components` into a reusable `#/$defs/ComponentsList` definition.
*   Inlined the `UpdateComponents` schema definition directly inside `UpdateComponentsMessage.properties.updateComponents`, referencing `#/$defs/ComponentsList` for the components array.
*   Defined `initialState.components` as a direct reference to `#/$defs/ComponentsList` (an array of component objects), removing the need for a nested `surfaceId`.
*   Defined `initialState.dataModel` as a plain JSON object (allowing `additionalProperties: true`) representing the initial root state of the data model directly, without any `path`/`value` nesting or redundant `surfaceId`.
*   Inlined the `UpdateDataModel` schema definition directly inside `UpdateDataModelMessage.properties.updateDataModel`.
*   Deleted the separate `#/$defs/UpdateComponents`, `#/$defs/UpdateDataModel`, and `#/$defs/DataModelUpdate` definitions to keep the schema compact and clean.

### Custom validation (`validator.ts`)

*   Updated `validateCreateSurface` to include `initialState`, `theme`, and `sendDataModel` in the allowed keys.
*   Updated `validateCustom` to validate `initialState.components` (via `validateComponentsList`) if provided in `createSurface`.
*   Enabled tracking of root component existence from nested `initialState.components`.
*   Added safety guards using `Array.isArray` and object validation to prevent type error crashes when parsing malformed component inputs.

### Testing and documentation

*   Created a new validation test suite in `test/cases/initial_state_validation.json` with test cases verifying correct schema validation for various valid and invalid `initialState` structures.
*   Added regression test cases covering malformed nested components (e.g. non-array formats, array containing null values) to ensure they fail validation gracefully.
*   Updated `docs/a2ui_protocol.md` to document the new `initialState` properties and added a complete JSON example illustrating a `CreateSurfaceMessage` using both components and a data model.

### DIFF
--- a/specification/v0_10/json/server_to_client.json
+++ b/specification/v0_10/json/server_to_client.json
@@ -39,20 +39,13 @@
               "type": "boolean",
               "description": "If true, the client will send the full data model of this surface in the metadata of every A2A message sent to the server that created the surface. Defaults to false."
             },
-            "initialState": {
+            "components": {
+              "$ref": "#/$defs/ComponentsList"
+            },
+            "dataModel": {
               "type": "object",
-              "description": "Optional initial state containing components and/or data model.",
-              "properties": {
-                "components": {
-                  "$ref": "#/$defs/ComponentsList"
-                },
-                "dataModel": {
-                  "type": "object",
-                  "description": "The initial root data model object for the surface.",
-                  "additionalProperties": true
-                }
-              },
-              "additionalProperties": false
+              "description": "The initial root data model object for the surface.",
+              "additionalProperties": true
             }
           },
           "required": ["surfaceId", "catalogId"],
--- a/specification/v0_10/json/server_to_client.json
+++ b/specification/v0_10/json/server_to_client.json
@@ -62,7 +62,7 @@
     },
     "UpdateComponents": {
       "type": "object",
-      "description": "Updates a surface with a new set of components. This message can be sent multiple times to update the component tree of an existing surface. One of the components in one of the components lists MUST have an 'id' of 'root' to serve as the root of the component tree. The createSurface message MUST have been previously sent with the 'catalogId' that is in this message.",
+      "description": "The payload representing a collection of UI components to render or update on a surface.",
       "properties": {
         "surfaceId": {
           "type": "string",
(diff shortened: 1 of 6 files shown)


samples/client/flutter/restaurant_finder/app/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -1,18 +1,17 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      https://www.apache.org/licenses/LICENSE-2.0
+//     http://www.apache.org/licenses/LICENSE-2.0
```
jacobsimionato, line 7: Is this intentional to remove the https?

specification/v0_10/json/server_to_client.json
```diff
@@ -38,6 +38,21 @@
             "sendDataModel": {
               "type": "boolean",
               "description": "If true, the client will send the full data model of this surface in the metadata of every A2A message sent to the server that created the surface. Defaults to false."
+            },
+            "initialState": {
+              "type": "object",
+              "description": "Optional initial state containing components and/or data model.",
+              "properties": {
+                "components": {
```
jacobsimionato: I personally think we could put `components` and `dataModel` at the top level rather than wrapping with initialState. I think developers will expect that data fields included in a `create...` call are the initial state.

jacobsimionato (approved):

samples/client/flutter/restaurant_finder/app/macos/Flutter/GeneratedPluginRegistrant.swift
gspencergoog: No... That was just the license fixing script "fixing" things, I thought.  But https makes more sense, and most of the other files are https.  Reverted and fixed other instances.

specification/v0_10/json/server_to_client.json
gspencergoog: Hmm. Yeah, that makes sense.  I actually didn't like the name `initialState` that much anyhow (just couldn't think of a better name).

PR #1525, opened 2026-06-03
refactor(spec)!: rename `theme` to `surfaceProperties` and remove `primaryColor`

### LINKED ISSUE #1118: Consider renaming `theme` on createSurface to `configuration` and remove `primaryColor` from basic catalog
I'm wondering if we should remove agent-driven theming because:
- We haven't seen great uptake so far
- It complicates implementing Components, when there is programmatic styling configuration and JSON theming configuration
- Creates a confusing mental model where Components are not really independent from the Catalog that owns them, because their implementation could be coupled to the Catalog's theme structure.
- Other competitors in this space don't have this concept

I think having some type of immutable surface-level config is still helpful, potentially just to include the agent name, logo etc.

## Proposal

1. Rename `theme` to `configuration` but keep the way it works exactly the same. We will just be allowing catalog designers to specify the structure of some surface-specific information which could be anything, but we aren't suggesting that there *should* be surface-level theming.

2. Remove `primaryColor` from the basic catalog configuration, but preserve `iconUrl` and `agentDisplayName`, which are really configuration parameters anyway.

This still leaves the door open for people to create catalogs which have surface-level agent-driven theming if they want to, but we background that direction and simplify the basic catalog.

gspencergoog:
# Description

This PR addresses issue #1118 by renaming the theme concept to configuration and removing the primaryColor property in version 0.10.

BREAKING-CHANGE: The 'theme' property in CreateSurfaceMessage is now 'surfaceProperties'. The 'primaryColor' property is no longer supported in the basic and minimal catalogs.

## Description

The term "theme" has been changed to "surfaceProperties" across all v0.10 schemas, catalogs, and documentation. The "primaryColor" property has been removed. This change decouples the visual styling attributes from the core surface layout and identity configuration parameters.

## Detailed changes

* Renamed the "theme" object within the Catalog definition to "surfaceProperties" in `client_capabilities.json`.
* Renamed the "theme" property of `createSurface` to "surfaceProperties" in `server_to_client.json`, updating references to point to `catalog.json#/$defs/surfaceProperties`.
* Renamed `$defs/theme` to `$defs/surfaceProperties` in `basic/catalog.json`, `minimal/catalog.json`, and `testing_catalog.json`.
* Removed the `primaryColor` property and its schema validation rules from those catalog configuration definitions.
* Renamed `theme_validation.json` to `configuration_validation.json`.
* Updated test cases to validate surfaceProperties parameters (`agentDisplayName` and `iconUrl`) and removed assertions for `primaryColor`.
* Updated references from "theme" to "surfaceProperties" in `a2ui_protocol.md` and removed the table entry for `primaryColor`.
* Added entry to `evolution_guide.md` tracking this surfaceProperties refactoring.

Fixes https://github.com/google/A2UI/issues/1118


specification/v0_10/json/server_to_client.json
```diff
@@ -31,9 +31,9 @@
               "description": "A string that uniquely identifies this catalog. It is recommended to prefix this with an internet domain that you own, to avoid conflicts e.g. mycompany.com:somecatalog'.",
               "type": "string"
             },
-            "theme": {
-              "$ref": "catalog.json#/$defs/theme",
-              "description": "Initial theme parameters for the surface (e.g., {'primaryColor': '#FF0000'}). These must validate against the 'theme' schema defined in the catalog."
+            "configuration": {
```
jacobsimionato: I'm now wondering if we should call this `surfaceConfiguration` to make it clear that it's surface-scoped? I think generic "configuration" feels especially funny in `catalog.json`, where we just have `"configuration": <some schema>` and it's not obvious that it's a surface level configuration schema.

I don't feel super strongly though, LGTM and you make the call.

jacobsimionato (approved):

specification/v0_10/json/server_to_client.json
gspencergoog: How about `surfaceProperties`?  They're defined in the catalog schema as to what they are allowed to be, and they are properties of the surface.  Configuration does feel a bit too generic.  I also considered `surfaceMetadata`, but it's too "meta": the properties are concrete properties of the surface, not just any old associated data.

PR #1526, opened 2026-06-03
Fix format script to work for Flutter packages

gspencergoog:
# Description

Currently, the format script uses `dart pub get` to get the packages, but that doesn't work if the dart package is a Flutter package.  This fixes that.


josemontespg (approved):
Ah I see the flutter formatter is needed. I did the reverse in https://github.com/google/A2UI/pull/1494.

LGTM! 🚀

ditman:
Yeah, I remember gemini pointing this out, actually!

PR #1529, opened 2026-06-03
refactor(react): use declarative HTML for known Text variants

### LINKED ISSUE #1516: Markdown rendering in Text component corrupts plain-text content in structured UI (buttons, labels, prices)
## Description
When a MarkdownRenderer is injected via MarkdownContext, every Text component's content is parsed as Markdown — including short, plain-text labels inside buttons, price tags, phone numbers, and other structured UI elements. Characters that are valid Markdown syntax (+, -, *, #, >, `, _) get silently misinterpreted.
## Reproduction
```
{
  "component": "Row",
  "children": [
    { "component": "Button", "child": { "component": "Text", "text": "-" } },
    { "component": "Text", "text": "1" },
    { "component": "Button", "child": { "component": "Text", "text": "+" } }
  ]
}
```
Image

```
{ "component": "Text", "text": "* tax included" }
```

Image

But without markdown, variant won't render correctly either:
```
{ "component": "Text", "text": "Xi'an Famous Foods" }
```
Image

## Question
Currently, `handleVariant()` prepends Markdown syntax (e.g. `# ` for `h1`) to the text content and relies on the injected `MarkdownRenderer` to produce the correct HTML. This creates a catch-22:

- **With** `MarkdownRenderer` injected: variants render correctly, but plain-text characters like `+`, `-`, `*`, `#` are silently corrupted.
- **Without** `MarkdownRenderer`: plain text is safe, but `variant="h1"` literally displays `# Hello` on screen because `handleVariant()` has already prepended the `#` prefix.

The variant implementation doesn't need to be coupled to the Markdown pipeline — CSS alone (font-size, font-weight, etc.) is sufficient to differentiate heading levels. Should the current `variant`-based approach be decoupled from Markdown rendering, rather than moving entirely to Markdown as proposed in #999?

chenman9226:
# Description

Refactors React v0.9 Text component to render known variants (h1–h5, caption) using declarative HTML elements instead of prepending Markdown syntax (`# `, `## `, `*...*`) and relying on the injected MarkdownRenderer. This aligns the React renderer with the Angular fix in #1502.

Markdown rendering is still used for unknown/default variants (like 'body').

* Fixes: #1516

## Pre-launch Checklist

- [x] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/react/src/v0_9/catalog/basic/components/Text.tsx
+++ b/renderers/react/src/v0_9/catalog/basic/components/Text.tsx
@@ -25,6 +25,7 @@ import styles from './Text.module.css';
 /** Variants rendered with declarative HTML instead of the Markdown pipeline. */
 const NON_MARKDOWN_VARIANTS = new Set(['h1', 'h2', 'h3', 'h4', 'h5', 'caption']);
 
+/** Renders text through the Markdown pipeline for unknown/default variants. */
 function MarkdownText({
   text,
   className,
@@ -37,13 +38,40 @@ function MarkdownText({
   const renderedHtml = useMarkdown(text);
   const classes = renderedHtml === null ? `${className} no-markdown-renderer` : className;
   const contentProps =
-    renderedHtml !== null
-      ? {dangerouslySetInnerHTML: {__html: renderedHtml}}
-      : {children: text};
+    renderedHtml !== null ? {dangerouslySetInnerHTML: {__html: renderedHtml}} : {children: text};
 
   return <div className={classes} style={style} {...contentProps} />;
 }
 
+/** Renders known variants (h1–h5, caption) as native HTML elements without Markdown. */
+function NonMarkdownText({
+  text,
+  variant,
+  className,
+  style,
+}: {
+  text: string;
+  variant: string;
+  className: string;
+  style: React.CSSProperties;
+}) {
+  const isCaption = variant === 'caption';
+  const HeadingTag = isCaption ? 'em' : (variant as 'h1' | 'h2' | 'h3' | 'h4' | 'h5');
+  if (isCaption) {
+    return (
+      <span className={className} style={style}>
+        <HeadingTag>{text}</HeadingTag>
+      </span>
+    );
+  }
+  return (
+    <div className={className} style={style}>
+      <HeadingTag>{text}</HeadingTag>
+    </div>
+  );
+}
+
+/** Renders text content, using the Markdown pipeline for rich text or native HTML elements for known typographic variants. */
 export const Text = createComponentImplementation(TextApi, ({props}) => {
   useBasicCatalogStyles();
   const text = typeof props.text === 'string' ? props.text : String(props.text ?? '');
@@ -56,22 +84,8 @@ export const Text = createComponentImplementation(TextApi, ({props}) => {
 
   if (variant && NON_MARKDOWN_VARIANTS.has(variant)) {
     const isCaption = variant === 'caption';
-    const className = [styles.a2uiText, isCaption ? styles.a2uiCaption : variant || 'body'].join(
-      ' ',
-    );
-    const HeadingTag = isCaption ? 'em' : (variant as 'h1' | 'h2' | 'h3' | 'h4' | 'h5');
-    if (isCaption) {
-      return (
-        <span className={className} style={style}>
-          <HeadingTag>{text}</HeadingTag>
-        </span>
-      );
-    }
-    return (
-      <div className={className} style={style}>
-        <HeadingTag>{text}</HeadingTag>
-      </div>
-    );
+    const className = [styles.a2uiText, isCaption ? styles.a2uiCaption : variant].join(' ');
+    return <NonMarkdownText text={text} variant={variant} className={className} style={style} />;
   }
 
   const className = [styles.a2uiText, variant || 'body'].join(' ');
--- a/renderers/react/src/v0_9/catalog/basic/components/Text.tsx
+++ b/renderers/react/src/v0_9/catalog/basic/components/Text.tsx
@@ -22,54 +22,58 @@ import {useMarkdown} from '../hooks/useMarkdown';
 // Import CSS Module
 import styles from './Text.module.css';
 
-/**
- * Wraps the plain text with appropriate Markdown syntax based on the requested variant.
- *
- * @param text The plain text to be wrapped.
- * @param variant The typography variant (e.g., 'h1', 'caption').
- * @returns The text wrapped in Markdown syntax.
- */
-const handleVariant = (text: string, variant?: string): string => {
-  switch (variant) {
-    case 'h1':
-      return `# ${text}`;
-    case 'h2':
-      return `## ${text}`;
-    case 'h3':
-      return `### ${text}`;
-    case 'h4':
-      return `#### ${text}`;
-    case 'h5':
-      return `##### ${text}`;
-    case 'caption':
-      return `*${text}*`;
-    default:
-      return text;
-  }
-};
+/** Variants rendered with declarative HTML instead of the Markdown pipeline. */
+const NON_MARKDOWN_VARIANTS = new Set(['h1', 'h2', 'h3', 'h4', 'h5', 'caption']);
+
+function MarkdownText({
+  text,
+  className,
+  style,
+}: {
+  text: string;
+  className: string;
+  style: React.CSSProperties;
+}) {
+  const renderedHtml = useMarkdown(text);
+  const classes = renderedHtml === null ? `${className} no-markdown-renderer` : className;
+  const contentProps =
+    renderedHtml !== null
+      ? {dangerouslySetInnerHTML: {__html: renderedHtml}}
+      : {children: text};
+
+  return <div className={classes} style={style} {...contentProps} />;
+}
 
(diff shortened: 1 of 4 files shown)


renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,46 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
     renderedHtml !== null
       ? {dangerouslySetInnerHTML: {__html: renderedHtml}}
-      : {children: markdownText};
+      : {children: text};
 
-  if (isCaption) {
-    return <span className={classes.join(' ')} style={style} {...contentProps} />;
-  }
   return <div className={classes.join(' ')} style={style} {...contentProps} />;
 });
```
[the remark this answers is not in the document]
chenman9226: The previous implementation also called `useMarkdown` unconditionally for all variants (including h1–h5 and caption), so this PR does not introduce any new overhead. Extracting a `MarkdownText` sub-component would be a further optimization, but I'd prefer to keep this change minimal

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,46 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
-      return `#### ${text}`;
-    case 'h5':
-      return `##### ${text}`;
-    case 'caption':
-      return `*${text}*`;
-    default:
-      return text;
-  }
-};
+const NON_MARKDOWN_VARIANTS = new Set(['h1', 'h2', 'h3', 'h4', 'h5', 'caption']);
```
ditman, line 26: Please, add a JSDoc comment above this constant to explain what it's used for, something like: 'These variants are directly rendered with HTML markup, instead of being handled by the markdown processor' or similar!

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,46 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
 
 export const Text = createComponentImplementation(TextApi, ({props}) => {
   useBasicCatalogStyles();
   const text = typeof props.text === 'string' ? props.text : String(props.text ?? '');
-  const markdownText = handleVariant(text, props.variant);
-  const renderedHtml = useMarkdown(markdownText);
+  const variant = props.variant;
+  const isKnownVariant = variant !== undefined && NON_MARKDOWN_VARIANTS.has(variant);
```
ditman: We're not saving a lot by checking `variant !== undefined` before asking the set if it `has` it IMO. This could be simplified to just the `has` check.

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,46 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
+
+  const renderedHtml = useMarkdown(text);
```
ditman: Do not `renderHtml(text)` until it's needed, move this to under the `if(isKnownVariant)` (right when it's needed)

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,46 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
   const style: React.CSSProperties = {
     ...getBaseLeafStyle(),
     ...getWeightStyle(props.weight),
   };
 
-  const isCaption = props.variant === 'caption';
-  const classes = [styles.a2uiText, isCaption ? styles.a2uiCaption : props.variant || 'body'];
+  const isCaption = variant === 'caption';
+  const classes = [styles.a2uiText, isCaption ? styles.a2uiCaption : variant || 'body'];
```
ditman: These seem to only be used inside `if(isKnownVariant)`, can you move these two consts to inside the if block?

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,46 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
   }
   const contentProps =
     renderedHtml !== null
       ? {dangerouslySetInnerHTML: {__html: renderedHtml}}
-      : {children: markdownText};
+      : {children: text};
 
-  if (isCaption) {
-    return <span className={classes.join(' ')} style={style} {...contentProps} />;
-  }
```
ditman, line 73: This is great, I need to backport this to Lit as well!

renderers/react/tests/v0_9/catalog-components.test.tsx
```diff
@@ -68,7 +68,7 @@ describe('Basic Catalog Components', () => {
       const {view} = renderA2uiComponent(Text, 't1', {text: 'Title', variant: 'h1'});
       const h1 = view.container.querySelector('div.h1');
       expect(h1).not.toBeNull();
-      expect(h1?.textContent).toBe('# Title');
+      expect(h1?.textContent).toBe('Title');
```
ditman, line 71: Chef's kiss!

ditman (changes requested):
Thanks for the contribution, this is great! Can you please also update the CHANGELOG.md, so users of the @a2ui/react package are aware of this change? Thanks!

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,58 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
+        <span className={className} style={style}>
+          <HeadingTag>{text}</HeadingTag>
+        </span>
+      );
+    }
+    return (
+      <div className={className} style={style}>
+        <HeadingTag>{text}</HeadingTag>
+      </div>
+    );
```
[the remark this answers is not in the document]
ditman: Nope, I'm not a fan of dynamically setting tags like this. For h1-h5 makes sense because the code would explode in complexity, but for the return statement the current approach is fine.

renderers/react/tests/v0_9/catalog-components.test.tsx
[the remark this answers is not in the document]
ditman: This is a good call. Also we could just update the querySelector to search for `div.h1 h1` or similar

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,58 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
 
-  const isCaption = props.variant === 'caption';
-  const classes = [styles.a2uiText, isCaption ? styles.a2uiCaption : props.variant || 'body'];
-  if (renderedHtml === null) {
-    classes.push('no-markdown-renderer');
+  if (variant && NON_MARKDOWN_VARIANTS.has(variant)) {
+    const isCaption = variant === 'caption';
+    const className = [styles.a2uiText, isCaption ? styles.a2uiCaption : variant || 'body'].join(
+      ' ',
+    );
```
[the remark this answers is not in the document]
ditman: If tsc is happy with this, this may be a nice change.

ditman:
@chenman9226 prettier is complaining about this PR; do you mind running `scripts/fix_format.sh` in the repo so it reformats the files it wants? Thanks!

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,54 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
-    case 'caption':
-      return `*${text}*`;
-    default:
-      return text;
-  }
-};
+/** Variants rendered with declarative HTML instead of the Markdown pipeline. */
+const NON_MARKDOWN_VARIANTS = new Set(['h1', 'h2', 'h3', 'h4', 'h5', 'caption']);
+
+function MarkdownText({
```
ditman, line 29: Please, add a small doc decribing this new function.

renderers/react/src/v0_9/catalog/basic/components/Text.tsx
```diff
@@ -22,54 +22,54 @@ import {useMarkdown} from '../hooks/useMarkdown';
...
+        <span className={className} style={style}>
+          <HeadingTag>{text}</HeadingTag>
+        </span>
+      );
+    }
+    return (
+      <div className={className} style={style}>
+        <HeadingTag>{text}</HeadingTag>
+      </div>
+    );
```
ditman: Now that you introduced `MarkdownText`, it may be nice to have a `NonMarkdownText` component that does this, so the Text method ends up just doing:

```javascript
if (variant && NON_MARKDOWN_VARIANTS.has(...)) {
  return NonMarkdownText(...);
}
return MarkdownText(...);
```

maybe? (if you end up doing this, remember to add jsdoc to it too!)

ditman (commented):
Small nit, please add some documentation to the MarkdownText component that you added! (Also to `Text` while you're at it!)

The current documentation is not good, so let's improve it little by little!

Thanks for your contribution! I'll stop nitpicking now! :)

ditman (approved):
Let's go with this! Thanks for the fix!

PR #1530, opened 2026-06-03
build: ignore and remove `GeneratedPluginRegistrant` files

gspencergoog:
# Description

Ignore generated iOS and macOS GeneratedPluginRegistrant files in root `.gitignore`. Remove the tracked `GeneratedPluginRegistrant.swift` file under `restaurant_finder/app/macos/Flutter/` to keep the repository clean from generated build artifacts.

This was causing problems with the fix_formatting.sh and fix_licenses.sh scripts because the fix_licenses.sh script would add a license, and then the fix_formatting.sh script would cause the generated registrants to be regenerated, removing the license again.  We don't need these files checked in anyhow.


josemontespg (approved):
Makes sense, thanks!

PR #1534, opened 2026-06-03
Updates the PartConverter to support versioning.

suyangw-g:
1. Adds a version parameter to the PartConverter constructor, defaulting to "0.8", the version used by Gemini Enterprise and most existing agents.
2. Propagates the version to calls to create_a2ui_part and parse_response_to_parts.

### DIFF
--- a/agent_sdks/python/src/a2ui/adk/a2a/part_converter.py
+++ b/agent_sdks/python/src/a2ui/adk/a2a/part_converter.py
@@ -68,7 +68,7 @@ class A2uiPartConverter:
       a2ui_catalog: A2uiCatalog,
       bypass_tool_check: bool = False,
       fallback_text: Optional[str] = None,
-      version: str = "0.8",
+      version: str = constants.VERSION_0_8,
   ):
     self._catalog = a2ui_catalog
     self._bypass_tool_check = bypass_tool_check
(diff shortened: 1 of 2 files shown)


josemontespg (approved):
I am not super familiar with this Python code, but the change looks good to me

josemontespg:
Looks like the CI signals are not passing, PTAL

agent_sdks/python/src/a2ui/adk/a2a/part_converter.py
```diff
@@ -55,17 +55,25 @@ class A2uiPartConverter:
   This converter handles both tool-based A2UI (via `send_a2ui_json_to_client`)
   and text-based A2UI (via A2UI delimiter tags). It uses the provided
   catalog to validate and fix JSON payloads.
+
+  Args:
+      a2ui_catalog: The A2UI catalog.
+      bypass_tool_check: If True, bypass tool validation.
+      fallback_text: Optional text to fall back on if parsing fails.
+      version: The A2UI version to use. Defaults to "0.8".
```
nan-yu, line 63: It feels odd to default to an old version, but I guess that's the best way to avoid introducing breaking changes.

nan-yu (approved):

PR #1543, opened 2026-06-04
spec(v0_10): specify that surfaceId must be globally unique per client session

### LINKED ISSUE #1275: figure out how to handle surfaceid collisions
we need to either require surface ids be globally unique, or give orchestrators some logic to handle collisions 

current issue would be
agenta creates surface "mysurface"
agentb creates surface "mysurface" (it would have no idea another "mysurface" exists since agents can only see surfaces they create)

one option would be for orchestrator to append the agent name and hash it, although this has issues as its hard to debug an id that changes between hops

maybe we just say surfaceids need to be globally unique (and hope 3Ps actually do that)

gspencergoog:
# Description

Updates the protocol specification document, schema definitions, and evolution guide to define `surfaceId` as globally unique per renderer instance's lifetime. Updates the TS validator to verify active surfaces, enforce that duplicate surface creations without prior deletion are errors, and validate that messages are only sent to active surfaces. Adds a new unit test suite `test_validator.ts` for these validation rules.

Fixes https://github.com/a2ui-project/a2ui/issues/1275

### DIFF
--- a/specification/v0_10/docs/a2ui_protocol.md
+++ b/specification/v0_10/docs/a2ui_protocol.md
@@ -175,7 +175,7 @@ The envelope defines several message types, and every message streamed by the se
 
 This message signals the client to create a new surface and begin rendering it. A surface must be created before any `updateComponents` or `updateDataModel` messages can be sent to it. While typically achieved by the agent sending a `createSurface` message, an agent may skip this if it knows the surface has already been created (e.g., by another agent). Once a surface is created, its `surfaceId` and `catalogId` are fixed; to reconfigure them, the surface must be deleted and recreated.
 
-It is an error to try to create a surface with a `surfaceId` that already exists without first deleting it; `surfaceId` must be globally unique for the renderer instance's lifetime. Orchestrators with subagents are empowered to manage surface IDs as they wish to prevent conflicts (e.g., prefixing the subagent's name to the `surfaceId` or requiring subagents to use UUIDs).
+It is an error to try to create a surface with a `surfaceId` that already exists without first deleting it; `surfaceId` must be globally unique for the renderer instance's lifetime. Orchestrators with subagents are empowered to manage surface IDs as needed to prevent conflicts (e.g., prefixing the subagent's name to the `surfaceId` or requiring subagents to use UUIDs).
 
 One of the components in one of the component lists MUST have an `id` of `root` to serve as the root of the component tree.
 
--- a/specification/v0_10/docs/a2ui_protocol.md
+++ b/specification/v0_10/docs/a2ui_protocol.md
@@ -175,7 +175,7 @@ The envelope defines several message types, and every message streamed by the se
 
 This message signals the client to create a new surface and begin rendering it. A surface must be created before any `updateComponents` or `updateDataModel` messages can be sent to it. While typically achieved by the agent sending a `createSurface` message, an agent may skip this if it knows the surface has already been created (e.g., by another agent). Once a surface is created, its `surfaceId` and `catalogId` are fixed; to reconfigure them, the surface must be deleted and recreated.
 
-It is an error to try to create a surface with a `surfaceId` that already exists without first deleting it. Orchestrators with subagents are empowered to manage surface IDs as they wish to prevent conflicts (eg, prefixing the subagent's name to the `surfaceId`).
+It is an error to try to create a surface with a `surfaceId` that already exists without first deleting it. Orchestrators with subagents are empowered to manage surface IDs as they wish to prevent conflicts (eg, requiring UUIDs or prefixing the subagent's name to the `surfaceId or requiring subagents to use UUIDs).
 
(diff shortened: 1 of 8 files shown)


specification/v0_10/eval/src/validator.ts
```diff
@@ -272,6 +273,11 @@ export class Validator {
             `updateComponents message received for surface '${surfaceId}' before createSurface message.`,
           );
         }
+        if (surfaceId && !activeSurfaces.has(surfaceId)) {
+          errors.push(
+            `updateComponents message received for inactive or deleted surface '${surfaceId}'.`,
+          );
+        }
```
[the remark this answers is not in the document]
gspencergoog: Updated the custom validator so that the error for an inactive/deleted surface is only reported if the surface was previously created.

specification/v0_10/eval/src/validator.ts
```diff
@@ -309,8 +321,28 @@ export class Validator {
...
+        if (surfaceId && !createdSurfaces.has(surfaceId)) {
+          errors.push(
+            `updateDataModel message received for surface '${surfaceId}' before createSurface message.`,
+          );
+        }
+        if (surfaceId && !activeSurfaces.has(surfaceId)) {
+          errors.push(
+            `updateDataModel message received for inactive or deleted surface '${surfaceId}'.`,
+          );
+        }
```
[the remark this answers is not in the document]
gspencergoog: Added an else if block for updateDataModel validation so only the most specific error is reported if the surface does not exist.

specification/v0_10/eval/src/test_validator.ts
```diff
@@ -0,0 +1,313 @@
...
+                text: 'Hello',
+              },
+            ],
+          },
+        },
+      ],
+      expectedErrors: [
+        "updateComponents message received for surface 'surf3' before createSurface message.",
+        "updateComponents message received for inactive or deleted surface 'surf3'.",
+      ],
```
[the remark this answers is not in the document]
gspencergoog: Updated the test case updateComponents before createSurface to expect only the single, non-redundant error message.

specification/v0_10/docs/a2ui_protocol.md
```diff
@@ -173,11 +173,11 @@ The envelope defines several message types, and every message streamed by the se
...
+This message signals the client to create a new surface and begin rendering it. A surface must be created before any `updateComponents` or `updateDataModel` messages can be sent to it. While typically achieved by the agent sending a `createSurface` message, an agent may skip this if it knows the surface has already been created (e.g., by another agent). Once a surface is created, its `surfaceId` and `catalogId` are fixed; to reconfigure them, the surface must be deleted and recreated. It is an error to try to create a surface with the same ID twice without first deleting it; `surfaceId` must be globally unique per client session. One of the components in one of the component lists MUST have an `id` of `root` to serve as the root of the component tree.
```
jacobsimionato: how about:

"globally unique per a2ui MessageProcessor" or "globally unique per a2ui context"

And if you change the language here, also change it below to be consistent.

I know we haven't really defined the concept of A2UI context yet, but I think the term "session" suggests that the ID space is kind of temporaral, e.g. if I come back to an app tomorrow, I have a fresh ID space and can reuse the same ID. But really, many apps will store data persistently, so the ID space would be about conversation, or possibly even more granular than that. E.g. when rendering an a2UI response from an MCP server, I'd advocate for developers creating a separate MessageProcessor for each surface, to ensure complete isolation of surfaces, so in this case the same surface ID can be used multiple times in one conversation.

specification/v0_10/docs/evolution_guide.md
```diff
@@ -41,7 +41,7 @@ Version 0.10 differs from 0.9 in the following ways:
 
 ### 2.7. Processing Rules
 
-- <TBD>
+- Explicitly specified that `surfaceId` must be globally unique per client session. Creating a surface with an ID that already exists (without first deleting it) is an error.
```
jacobsimionato: Optional: We already said this and required it in v0.9, though not as clearly or emphatically, so I don't think we need to highlight this change.

specification/v0_10/docs/a2ui_protocol.md
gspencergoog: Updated all occurrences of 'client session' to 'A2UI MessageProcessor' in `a2ui_protocol.md`, `client_to_server.json`, and `server_to_client.json` to be precise and consistent.

specification/v0_10/docs/evolution_guide.md
gspencergoog: Removed this point from the evolution guide since it was already a requirement in v0.9.

specification/v0_10/docs/a2ui_protocol.md
gspencergoog: Correction: In the final version, I went with "globally unique for the renderer's lifetime" to make it transport/implementation-agnostic.

jacobsimionato (approved):

PR #1562, opened 2026-06-05
refactor(angular): remove (basic) suffix from v0.8 basic tests

josemontespg:
refactor(angular): remove (basic) suffix from v0.8 basic tests

This PR is stacked on top of #1497. It should be reviewed/merged after #1497.

It removes the redundant '(basic)' suffix from the 28 basic v0.8 test files (both
describe block and loadExample calls), as the minimal catalog is no longer
used in renderers.

Also:
- Remove the suffix logic from generate-examples.mjs for v0.8.
- Remove the fallback logic from test_utils.ts.

TAG=agy
CONV=9137c307-af0d-4380-8e89-52054dd1fcb0


josemontespg:
Obsolete

PR #1572, opened 2026-06-08
feat(spec): Add built-in @index function for list templates

### LINKED ISSUE #1558: feat(spec): Add a way to get the index of a templated component in a list.
# Feature request: Add a built-in `index` function for list templates

## Problem statement

Currently, there is no way to retrieve the index of the current item when rendering a dynamic list from a template. If a template needs to display the position of an item (for example, "Item 1", "Item 2"), the index must be explicitly added to every item in the data model. This introduces UI presentation logic into the data model and increases payload sizes.

## Proposed solution

We propose adding a built-in client-side function named `index`. When called within a list template scope, this function returns the 0-based index of the current item.

The function can support an optional `offset` argument (e.g., `index(offset: 1)`) to make 1-based indexing easier.

## Example payload

Here is how the `index` function would be used in an A2UI payload.

### Component update (`updateComponents`)

```json
{
  "version": "v0.10",
  "updateComponents": {
    "surfaceId": "todo_surface",
    "components": [
      {
        "id": "todo-list",
        "component": "List",
        "children": {
          "componentId": "todo-item-template",
          "path": "/todos"
        }
      },
      {
        "id": "todo-item-template",
        "component": "Row",
        "children": ["todo-index", "todo-title"]
      },
      {
        "id": "todo-index",
        "component": "Text",
        "text": {
          "call": "formatString",
          "args": {
            "value": "#${index(offset: 1)}"
          }
        }
      },
      {
        "id": "todo-title",
        "component": "Text",
        "text": {
          "path": "title"
        }
      }
    ]
  }
}
```

### Data model update (`updateDataModel`)

The data model does not need to store the index of the items:

```json
{
  "version": "v0.10",
  "updateDataModel": {
    "surfaceId": "todo_surface",
    "path": "/todos",
    "value": [
      { "title": "Buy groceries" },
      { "title": "Clean the kitchen" }
    ]
  }
}
```

gspencergoog:
# Description

Introduces the `@index` client-side built-in function to the A2UI v0.10 protocol specification, allowing dynamic retrieval of the 0-based iteration index during list template rendering.

Key architectural highlights:
- Universal Availability: Embedded `@index` directly into `common_types.json` under `FunctionCall`, ensuring universal inheritance across all active catalogs (Basic, Minimal, and third-party custom catalogs).
- System Namespace Standard: Reserved the `@` prefix for core system context evaluations to permanently prevent collisions with custom catalog utilities.
- Strict Scoping: Restricts `@index` availability strictly to Evaluation Contexts operating within a Collection Scope (template instantiation). Calling outside of template iteration results in an invalid evaluation / error.
- Offset Support: Supports an optional `offset` argument (e.g., `@index(offset: 1)`) for easy 1-based indexing.

Fixes: https://github.com/a2ui-project/a2ui/issues/1558

### DIFF
--- a/specification/v0_10/docs/a2ui_protocol.md
+++ b/specification/v0_10/docs/a2ui_protocol.md
@@ -820,8 +820,11 @@ The [`catalogs/basic/catalog.json`] provides the baseline set of components and 
 
 ### Functions
 
+> **System Namespace Rule (`@` Prefix)**: Function names beginning with `@` (e.g., `@index`) represent universal system context evaluations available across all catalogs. Custom catalogs MUST NOT define functions prefixed with `@`.
+
 | Function           | Description                                                              |
 | :----------------- | :----------------------------------------------------------------------- |
+| **@index**         | Returns the 0-based index of the current item during template rendering. |
 | **required**       | Checks that the value is not null, undefined, or empty.                  |
 | **regex**          | Checks that the value matches a regular expression string.               |
 | **length**         | Checks string length constraints.                                        |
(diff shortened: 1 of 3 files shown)


josemontespg (commented):
Using @ is a nice way to reduce the chance of collision. Should we add @ to all of them? And keep the original name for backwards compatibility

josemontespg (approved):

gspencergoog:
> Using @ is a nice way to reduce the chance of collision. Should we add @ to all of them? And keep the original name for backwards compatibility

Add `@` to all of which?  The function names in the basic catalog?  No, this was to be able to avoid collisions between the "built-in" functions like this index function, and any catalog-defined functions.  That way the catalog defined functions don't have any chance to collide with the built-in functions, because they aren't allowed to have `@` in their names.

PR #1575, opened 2026-06-08
Initialize a2ui_core SDK with schema definitions and basic UI catalog components

nan-yu:
# Description

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/agent_sdks/python/a2ui_core/scripts/generate_schemas.py
+++ b/agent_sdks/python/a2ui_core/scripts/generate_schemas.py
@@ -1,0 +1,865 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import os
+import re
+from typing import Any, Dict, List
+
+# Base directories
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+SPEC_VERSION = "v0_9"
+SPEC_VERSION_DOT = SPEC_VERSION.replace("_", ".")
+SPEC_DIR = os.path.abspath(
+    os.path.join(SCRIPT_DIR, "../../../../specification", SPEC_VERSION)
+)
+JSON_DIR = "json"
+CATALOGS_DIR = "catalogs"
+
+# Input file paths
+COMMON_TYPES_PATH = os.path.join(SPEC_DIR, JSON_DIR, "common_types.json")
+CLIENT_CAPABILITIES_PATH = os.path.join(SPEC_DIR, JSON_DIR, "client_capabilities.json")
+CLIENT_TO_SERVER_PATH = os.path.join(SPEC_DIR, JSON_DIR, "client_to_server.json")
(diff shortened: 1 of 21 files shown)


agent_sdks/python/a2ui_core/scripts/generate_schemas.py
jacobsimionato, line 1: Overall, this approach seems fantastic and we should adopt it in our other libraries.

Remind me again why we couldn't use something like https://pypi.org/project/json-schema-to-pydantic/ ? It'd be amazing if we didn't need to maintain so much special cased logic here. But also, we don't change the schema very often, so it's not a huge deal.

agent_sdks/python/a2ui_core/scripts/generate_schemas.py
```diff
@@ -0,0 +1,865 @@
...
+    output.append(compile_object_def("ActionFunctionCallWrapper", func_wrapper_spec))
+    output.append("Action = Union[ActionEventWrapper, ActionFunctionCallWrapper]\n")
+
+    # 8. Generate ComponentCommon
+    output.append(compile_object_def("ComponentCommon", defs["ComponentCommon"]))
+
+    return "\n".join(output)
+
+
+def generate_basic_catalog_components(
```
jacobsimionato, line 411: Future idea: It could be nice to generalize this to generate_catalog_components at some point, so we can point it at any catalog file that adheres to our structure.

jacobsimionato (approved):

agent_sdks/python/a2ui_core/scripts/generate_schemas.py
nan-yu: I tried https://pypi.org/project/datamodel-code-generator/ instead of https://pypi.org/project/json-schema-to-pydantic/. 

https://docs.google.com/document/d/11PYbSt0ThCjYabW8YN6fU2goJu44z-CmOzG49hPw5ew/edit?resourcekey=0-jN1YnS8nl6r7N7HOhVNXKg&tab=t.2fcdpw7eay5q#bookmark=id.r6fnse63e7y2 summarizes some issues I encountered during the generation process.

I can give `json-schema-to-pydantic` a quick shot to see whether it is a better fit.

nan-yu: I ran into the following issues with `json-schema-to-pydantic`:
- **Rejection of Cross-File / Multi-Document References ($ref)**: json-schema-to-pydantic strictly enforces intra-document references (#/). Even when converting remote URLs to local relative file paths (common_types.json#/..., catalog.json#/...), it immediately throws ReferenceError: Only local references (#/...) are supported. It lacks a multi-document resolver capable of linking our separated, modular specifications.
- **Custom Catalog DSL Incompatibility**: Our component catalog (`basic/catalog.json`) groups specifications inside custom top-level container keys `("components": { ... })` rather than standard JSON Schema `"properties"`. Generic parsers do not recognize this DSL layout, treating the catalog as an empty schema object.
- **Discriminated Union Optimization (O(1) Routing)**: We require O(1) routing via Pydantic's `Annotated[Union[...], Field(..., discriminator="component")]` for high-frequency UI payload parsing. Off-the-shelf tools fail to synthesize optimized discriminated unions across heterogeneous component schemas, generating slow sequential try-except evaluators.
- **Symmetrical Inheritance & Singularization**: We need components to cleanly inherit from `CatalogComponentCommon` without duplicating inherited base fields (`id`, `weight`), and resolve inline JSON object definitions into clean semantic helper classes (e.g., `OptionItem` instead of generic `Option` or DynamicModel wrappers).

I also updated the doc to include that.

agent_sdks/python/a2ui_core/scripts/generate_schemas.py
nan-yu: I created an issue to track the idea: https://github.com/a2ui-project/a2ui/issues/1592.

PR #1577, opened 2026-06-08
Python a2ui_core: Implement catalog with components validation

nan-yu:
## Summary of Changes

This pull request implements a comprehensive validation layer for the A2UI core library. It establishes a structured approach to catalog management, allowing for both statically defined Pydantic models and dynamic JSON schema-based validation. The changes include new utilities for checking component integrity, analyzing graph topology, and ensuring that component references and function calls conform to the defined catalog specifications. This ensures higher reliability for A2UI protocol payloads.

### Highlights

* **Catalog Validation Framework**: Introduced a robust validation framework for A2UI catalogs, including base classes for catalog definitions and concrete implementations for both Pydantic-based models and dynamic JSON schema validation.
* **Integrity and Topology Analysis**: Added comprehensive integrity checkers and topology analyzers to ensure component graphs are valid, acyclic, and reachable, with support for custom reference field definitions.
* **Dynamic Validation Support**: Implemented `JsonCatalog` to support server-side validation using JSON Schema (Draft 2020-12), enabling dynamic checks without requiring pre-compiled Pydantic models.
* **Dependency Updates**: Added `jsonschema` and `referencing` as new dependencies to support the dynamic JSON schema validation requirements.

<details>
<summary><b>Activity</b></summary>

* Pull request created by nan-yu.
* Multiple failed attempts to trigger automated review and summary generation.
* User requested summary generation via '/gemini summary'.
</details>

### DIFF
--- a/agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
+++ b/agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
@@ -1,0 +1,53 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Dict, List, Optional, Set, Tuple
+from ..schema.constants import DEFAULT_SINGLE_REF_FIELDS, DEFAULT_LIST_REF_FIELDS, ROOT_ID, SPEC_BASE_URL
+
+
+class Catalog:
+    """Base abstract Catalog definition for framework-agnostic A2UI catalog integrations."""
+
+    def __init__(
+        self,
+        version: str,
+        catalog_id: str,
+        custom_single_refs: Optional[List[str]] = None,
+        custom_list_refs: Optional[List[str]] = None,
+    ):
+        if not version:
+            raise ValueError("version must be provided.")
+        if not catalog_id:
+            raise ValueError("catalog_id must be provided.")
+
+        self.version = version
+        self.catalog_id = catalog_id
+        self.single_refs = set(custom_single_refs or DEFAULT_SINGLE_REF_FIELDS)
+        self.list_refs = set(custom_list_refs or DEFAULT_LIST_REF_FIELDS)
+
+    def validate_components(self, comp_payload: List[Dict[str, Any]]) -> None:
+        """Validates a list of component payloads conforming to the catalog's schemas."""
+        raise NotImplementedError("Subclasses must implement validate_components()")
+
+    def validate_function(self, func_name: str, args: Dict[str, Any]) -> None:
+        """Validates that function arguments conform to the catalog's schema for this function."""
+        raise NotImplementedError("Subclasses must implement validate_function()")
+
+    def validate_theme(self, theme_payload: Dict[str, Any]) -> None:
+        """Validates that theme properties conform to the catalog's theme schema."""
+        raise NotImplementedError("Subclasses must implement validate_theme()")
+
+    def extract_ref_fields(self) -> Dict[str, Tuple[Set[str], Set[str]]]:
+        """Inspects and retrieves the topological reference pointer map for the active catalog components."""
+        raise NotImplementedError("Subclasses must implement extract_ref_fields()")
--- a/agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
+++ b/agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
@@ -21,17 +21,17 @@ class Catalog:
 
     def __init__(
         self,
-        version: str,
+        spec_version: str,
         catalog_id: str,
         custom_single_refs: Optional[List[str]] = None,
         custom_list_refs: Optional[List[str]] = None,
     ):
-        if not version:
-            raise ValueError("version must be provided.")
+        if not spec_version:
+            raise ValueError("A2UI specification version must be provided.")
         if not catalog_id:
             raise ValueError("catalog_id must be provided.")
 
-        self.version = version
+        self.spec_version = spec_version
         self.catalog_id = catalog_id
         self.single_refs = set(custom_single_refs or DEFAULT_SINGLE_REF_FIELDS)
         self.list_refs = set(custom_list_refs or DEFAULT_LIST_REF_FIELDS)
--- a/agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
+++ b/agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
@@ -15,8 +15,8 @@
 from typing import Any, Dict, List, Optional, Set, Tuple
 
 
-class Catalog:
-    """Base abstract Catalog definition for framework-agnostic A2UI catalog integrations."""
+class CatalogApi:
+    """Base abstract Catalog API definition representing the schemas of an A2UI Catalog."""
 
     def __init__(
         self,
@@ -31,7 +31,51 @@ class Catalog:
         self.spec_version = spec_version
         self.catalog_id = catalog_id
 
+    def get_component_schema(self, comp_type: str) -> Optional[Dict[str, Any]]:
+        """Retrieves the JSON Schema representing a component's properties."""
+        raise NotImplementedError("Subclasses must implement get_component_schema()")
+
+    def get_function_schema(self, func_name: str) -> Optional[Dict[str, Any]]:
+        """Retrieves the JSON Schema representing a function's arguments."""
+        raise NotImplementedError("Subclasses must implement get_function_schema()")
+
+    def get_theme_schema(self) -> Optional[Dict[str, Any]]:
+        """Retrieves the JSON Schema representing the catalog's theme."""
+        raise NotImplementedError("Subclasses must implement get_theme_schema()")
 
     def extract_ref_fields(self) -> Dict[str, Tuple[Set[str], Set[str]]]:
         """Inspects and retrieves the topological reference pointer map for the active catalog components."""
         raise NotImplementedError("Subclasses must implement extract_ref_fields()")
+
+
+class CatalogImplementation(CatalogApi):
+    """Abstract Catalog subclass that extends CatalogApi with concrete runtime implementations.
+
+    In addition to schemas, a CatalogImplementation provides executable function
+    invokers and component model classes, serving client-side rendering and
+    local function evaluation.
+    """
+
+    def get_component_class(self, comp_type: str) -> Optional[Any]:
+        """Retrieves the concrete model class representing a component."""
+        raise NotImplementedError("Subclasses must implement get_component_class()")
+
+    def get_function_class(self, func_name: str) -> Optional[Any]:
+        """Retrieves the concrete model class representing a function's schema."""
+        raise NotImplementedError("Subclasses must implement get_function_class()")
+
+    def get_function_implementation(self, func_name: str) -> Optional[Any]:
+        """Retrieves the concrete FunctionImplementation object for a function."""
+        raise NotImplementedError(
+            "Subclasses must implement get_function_implementation()"
+        )
+
+    def invoke_function(
+        self,
+        name: str,
+        args: Dict[str, Any],
+        context: Any = None,
+        abort_signal: Optional[Any] = None,
+    ) -> Any:
+        """Executes a catalog function dynamically."""
+        raise NotImplementedError("Subclasses must implement invoke_function()")
--- a/agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
+++ b/agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from typing import Any, Dict, List, Optional, Set, Tuple
-from ..schema.constants import DEFAULT_SINGLE_REF_FIELDS, DEFAULT_LIST_REF_FIELDS, ROOT_ID, SPEC_BASE_URL
 
 
 class Catalog:
@@ -23,8 +22,6 @@ class Catalog:
         self,
         spec_version: str,
         catalog_id: str,
-        custom_single_refs: Optional[List[str]] = None,
-        custom_list_refs: Optional[List[str]] = None,
     ):
         if not spec_version:
             raise ValueError("A2UI specification version must be provided.")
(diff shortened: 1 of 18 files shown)


agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
```diff
@@ -0,0 +1,53 @@
...
+from typing import Any, Dict, List, Optional, Set, Tuple
+from ..schema.constants import DEFAULT_SINGLE_REF_FIELDS, DEFAULT_LIST_REF_FIELDS, ROOT_ID, SPEC_BASE_URL
+
+
+class Catalog:
+    """Base abstract Catalog definition for framework-agnostic A2UI catalog integrations."""
+
+    def __init__(
+        self,
+        version: str,
```
jacobsimionato: How about `specification_version` here, to make it clear what the version is of?

agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
```diff
@@ -0,0 +1,53 @@
...
+        catalog_id: str,
+        custom_single_refs: Optional[List[str]] = None,
```
jacobsimionato: Let's remove *both* custom_single_refs *and* custom_list_refs.

Special casing behavior based on field name was a mistake. We should instead detect reference fields in  a catalog.json because they refer to `ComponentId` or `ChildList` schemas in common_types. We should ideally find some way to flag these in the Pydantic representation of the schema which does not rely on magic key names.

See https://github.com/a2ui-project/a2ui/issues/1528 for progress on this issue in the web codebases.

agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
```diff
@@ -0,0 +1,53 @@
...
+            raise ValueError("version must be provided.")
+        if not catalog_id:
+            raise ValueError("catalog_id must be provided.")
+
+        self.version = version
+        self.catalog_id = catalog_id
+        self.single_refs = set(custom_single_refs or DEFAULT_SINGLE_REF_FIELDS)
+        self.list_refs = set(custom_list_refs or DEFAULT_LIST_REF_FIELDS)
+
+    def validate_components(self, comp_payload: List[Dict[str, Any]]) -> None:
```
jacobsimionato: Not blocking: How about just exposing this logic on the Validator object, for better separation of concerns?

I wonder if that will make it easy for a reader to see what is part of the Catalog representation, and what is built on top of it. And if we choose to not implement a full validator suite for client languages, there will be a clear boundary between components.

Overall, I think you should own the design of the validation APIs - this is just feedback. It'd be helpful to add a design doc for the validation logic somewhere, so we can replicate it across other languages.

agent_sdks/python/a2ui_core/src/a2ui/core/validating/validator.py
```diff
@@ -0,0 +1,197 @@
...
+                    errors.append(e)
+        if errors:
+            err_msg = "\n".join(str(err) for err in errors)
+            raise A2uiValidatorError(err_msg)
+
+    def validate(
+        self,
+        catalog: Catalog,
+        a2ui_payload: Union[Dict[str, Any], List[Any]],
+        root_id: Optional[str] = None,
```
jacobsimionato: I think this should always be "root" right, for v0.9+?

agent_sdks/python/a2ui_core/src/a2ui/core/validating/validator.py
```diff
@@ -0,0 +1,197 @@
...
+            path_str = ".".join(loc_parts)
+            msg = err.get("msg", "Validation failed")
+            formatted_errors.append(f"{path_str}: {msg}")
+        return formatted_errors
+
+    def validate_components(
+        self,
+        catalog: Catalog,
+        components: List[Dict[str, Any]],
+        strict_integrity: bool = False,
```
jacobsimionato: Not blocking: It's not clear what strict_integrity means - would it be possible to use a more specific flag name (s)? E.g. allow_orphan_components etc. Perhaps even this should accept a ValidationConfig object, so that we can expand the API surface without adding too many args here?

agent_sdks/python/a2ui_core/src/a2ui/core/validating/validator.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+class A2uiValidatorError(ValueError):
+    """Exception raised when an A2UI Catalog payload validation fails."""
+
+
+class A2uiValidator:
+    """Validates the A2UI JSON payload against catalog schemas and checks for layout integrity."""
+
+    def validate_structure(self, messages: List[Dict[str, Any]]) -> None:
```
jacobsimionato: Not blocking: I'm not super clear on what "structure" means here as a developer.

What about just exposing a single "validate" method which returns a set of errors, then configuration can be used to enable/disable specific features?

That way, we also limit the size of the API surface we need to support going forward.

agent_sdks/python/a2ui_core/src/a2ui/core/validating/validator.py
```diff
@@ -0,0 +1,197 @@
...
+                                catalog,
+                                components,
+                                strict_integrity,
+                                root_id=resolved_root_id,
+                            )
+                except Exception as e:
+                    errors.append(e)
+
+        if errors:
+            raise A2uiValidatorError("\n".join(str(err) for err in errors))
```
jacobsimionato, line 191: Not blocking: I wonder if it might be worthwhile throwing an object here which some parent ValidationFailure which contains an array of ValidationErrors that can iterated over and understood etc. ValidationErrors could just be a string for now. That might make it easier to implement deterministic fixer logic etc, or give developers flexibility around how the format the errors if/when they pass them back to an LLM.

Alternatively this could be an API which returns a result?

agent_sdks/python/a2ui_core/src/a2ui/core/catalog/model_catalog.py
```diff
@@ -0,0 +1,239 @@
...
+from pydantic import BaseModel
+from .catalog import Catalog
+from ..schema.constants import (
+    DEFAULT_SINGLE_REF_FIELDS,
+    DEFAULT_LIST_REF_FIELDS,
+)
+from .functions import FunctionImplementation
+
+
+class ModelCatalog(Catalog):
```
jacobsimionato: Not blocking: I found it hard to understand the practical difference between ModelCatalog and JsonCatalog. Is it basically that ModelCatalog accepts FunctionObjects, and so it can expose FunctionInvoker?

I'm wondering if its possible to share more of the code between these somehow, because conceptually, I would expect a Catalog to behave the same whether it's instantiated from JSON or programmatically.

Maybe the conceptual distinction is between CatalogApi and CatalogImplementation which extends CatalogApi. Catalog API has the schema of Components and Functions, and CatalogImplementation also has function implementations (and component rendering implementations for client languages).

jacobsimionato (commented):
Hey I left a bunch of comments, but I'm conscious you have a long chain, so happy to defer most of this. The validation APIs are new, so it's natural for us to iterate on them.

I would be super keen to remove the custom_single_refs etc if possible before submission!

agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
nan-yu: That sounds good. `JsonCatalog` already had methods to detect reference fields that refer to `ComponentId` or `ChildList` schemas in common_types, but `ModelCatalog` did not. Initially, `custom_single_refs` and `custom_list_refs` served as a fallback guardrail for custom component models that didn't strictly follow those common types. I agree that this should be addressed in the schema and catalog definitions themselves rather than having the SDK provide an ad-hoc defense layer.

I have removed both `custom_single_refs` and `custom_list_refs` and refactored `ModelCatalog`. Specifically, we introduced helper marker classes (`ComponentReference`, `SingleReference`, and `ListReference`) into the Python schema definitions. `ModelCatalog` now leverages pure Python reflection to recursively inspect Pydantic V2 model fields and detect subclasses of these reference markers. This detects all single and list reference fields (including nested models) without relying on magic property key names or magic type string matching.

agent_sdks/python/a2ui_core/src/a2ui/core/catalog/model_catalog.py
nan-yu: Thanks for great feedback! CatalogApi and CatalogImplementation exactly describes the difference between ModelCatalog and JsonCatalog. Basically, ModelCatalog serves as client-side rendering class with implementations, while JsonCatalog focuses on the Catalog API definition on the agent side, which doesn't require client-side implementation.

I refactored the code to adopt your exact conceptual hierarchy and separation of concerns.

agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
nan-yu: That sounds good. I moved the validation logic to a CatalogValidator class, so Catalog only represents the definition, while validation logic all resides in the validating package.

I'll come up with a doc for the validation APIs later.

agent_sdks/python/a2ui_core/src/a2ui/core/validating/validator.py
nan-yu: Correct. The core A2uiValidator is supposed to only support v0.9+, while `analyze_topology` and `validate_component_integrity` are also needed by the v0.8 validator. I removed the root_id from A2uiValidator.

agent_sdks/python/a2ui_core/src/a2ui/core/validating/validator.py
nan-yu: Agreed that "structure" can be a bit ambiguous. We need this standalone method when messages are ingested in client renderers before a catalog is known. Here is the lifecycle on how messages are processed by the MessageProcessor.
1. Phase 1 (Protocol routing): MessageProcessor verify the overall correctness, including checks for 
   - the spec version is correct
   - the payload is an object
   - safety recursion limits are not exceeded
   - valid core messages branches (createSurface, updateComponents, and etc) exits
   
   At this stage, the catalog has not been determined or loaded yet. 

2. Phase 2 (Catalog validation). Only after the createSurface message is processed, the corresponding catalog will be resolved based on the catalogId in the createSurface message. Then we can validate components (schema validity, integrity and topology), functions, theme.

With that being said, we still need to expose the method separately. Otherwise, we have to defer the overall validation after a createSurface message is processed.

Regarding the method name, I renamed to `validate_protocol_envelope` to make it obvious that it verifies the wire protocol envelope without needing a catalog.

agent_sdks/python/a2ui_core/src/a2ui/core/validating/validator.py
nan-yu: Refactored with a ValidationConfig object and consolidated the param names.

agent_sdks/python/a2ui_core/src/a2ui/core/validating/validator.py
nan-yu: We can handle the validation errors in a unified way later on.

agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def get_function_schema(self, func_name: str) -> Optional[Dict[str, Any]]:
+        """Retrieves the JSON Schema representing a function's arguments."""
+        raise NotImplementedError("Subclasses must implement get_function_schema()")
+
+    def get_theme_schema(self) -> Optional[Dict[str, Any]]:
+        """Retrieves the JSON Schema representing the catalog's theme."""
+        raise NotImplementedError("Subclasses must implement get_theme_schema()")
+
+    def extract_ref_fields(self) -> Dict[str, Tuple[Set[str], Set[str]]]:
```
jacobsimionato, line 46: Maybe this function could move to the validation package rather than being in the Catalog models themselves. That way we have good separation of concerns where the Catalog just represents the Catalog data, and all manipulation or analysis of that data happens separately.

agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
```diff
@@ -0,0 +1,81 @@
...
+    ):
+        if not spec_version:
+            raise ValueError("A2UI specification version must be provided.")
+        if not catalog_id:
+            raise ValueError("catalog_id must be provided.")
+
+        self.spec_version = spec_version
+        self.catalog_id = catalog_id
+
+    def get_component_schema(self, comp_type: str) -> Optional[Dict[str, Any]]:
```
jacobsimionato, line 34: I think it would be helpful to have a top-level Component and Function classes (or perhaps ComponentApi, FunctionApi, FunctionImplementation) classes in the design, so that then Catalog can have a neat API that just exposes Component, and then each Component can optionally have a schema and class etc. That way, we can add additional properties to Component or Function more easily.

agent_sdks/python/a2ui_core/src/a2ui/core/validating/catalog_validator.py
```diff
@@ -0,0 +1,292 @@
...
+        if not func_class:
+            raise ValueError(f"Unknown function: {func_name}")
+        if hasattr(func_class, "model_fields") and "call" in func_class.model_fields:
+            payload = {"call": func_name, "args": args}
+            func_class.model_validate(payload)
+        else:
+            func_class.model_validate(args)
+
+
+class JsonCatalogValidator(CatalogValidator):
```
jacobsimionato, line 134: It seems like ModelCatalogValidator and JsonCatalogValidator are basically the same. My understanding is that Pydantic models can be converted to JSON schemas. So how about we write validation logic once against the JSON, and do validation of pydantic-backed models etc by converting them to JSON first?

jacobsimionato (approved):
I think there are some architectural improvements that would be possible here to make this logic more flexible and more consistent with web_core. E.g.
- *Avoid the branching between the JSON and Model Catalog variants*, so that there is one single way of interacting with Catalogs and Components. That way, there can be multiple ways to create catalogs (by loading JSON files, programmatically, by converting some other format to a Catalog object), and then all logic that processes catalogs (e.g. validation) can be written once and work anywhere. I also think there's a possibility we might need to deal with hybrid catalogs where some Components are loaded from JSON (e.g. basic catalog components?) and others are defined programmatically (e.g. some custom component).
- *Add first-class Component object and Function objects, to help developers mix and match components and functions. This should reduce the API surface on Catalog directly.
- *Unify validation logic* so it doesn't need to branch between model and JSON formats, seeing as that leaks implementation details of how the Catalog object was created. 

But, there is a chain of PRs that I want to unblock. So I figure the fastest way is to approve here and try to get the others submitted too, and then we can always iterate later.

I filed a bug with a detailed proposal that we can always try to implement later. I'm curious about your thoughts on that - https://github.com/a2ui-project/a2ui/issues/1606

WDYT?

agent_sdks/python/a2ui_core/src/a2ui/core/catalog/catalog.py
nan-yu: This method is also used in the streaming parser, while it is also used for validation. It makes sense to move it to the validation package. As you suggested, we can refactor it in a follow up PR.

nan-yu:
Submitting the changes in this PR. Comments will be resolved in a separate PR.

PR #1579, opened 2026-06-08
Python a2ui_core: introduce reactive Signal and ComponentNode primitives and add a text-based layout renderer for testing

nan-yu:
# Description

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md


agent_sdks/python/a2ui_core/src/a2ui/core/state/surface_model.py
```diff
@@ -63,11 +87,308 @@ def dispatch_error(self, error: Dict[str, Any]) -> None:
         err_payload["surfaceId"] = self.id
         self.on_error.emit(err_payload)
 
+    def get_or_create_node(self, component_id: str, data_path: str) -> ComponentNode:
```
jacobsimionato: Does the caller need to call this to lazily kick off node tree creation? I assume it won't work if there are no FunctionImplementations to resolve function values

I'm now wondering if it makes sense for the node tree to be managed by a separate object e.g. NodeGraph which can be instantiated only with a Catalog that has function implementations. If we split it out, it might be easier to iterate on the design, too.

agent_sdks/python/a2ui_core/tests/text_renderer.py
jacobsimionato, line 1: Interesting! This could be great for tests. Maybe the output of it can be formalized, and we can have YAML files that include examples of what  the node graph should be after certain operations.

jacobsimionato (approved):

agent_sdks/python/a2ui_core/src/a2ui/core/state/surface_model.py
nan-yu: Yes, it makes sense to split the reactive node tree out into a separate NodeGraph class and keeps SurfaceModel as a pure data container.

The node graph needs to reactively monitor component and data updates, so it is instantiated with a SurfaceModel instead of a Catalog.

agent_sdks/python/a2ui_core/tests/text_renderer.py
nan-yu: Good idea. I added a recursive `to_dict()` method to `ComponentNode` and exposed it on `NodeGraph`. We can now dump the structure using standard YAML formatting to assert structure layout parity directly in unit tests.
```python
node_dict = node_graph.to_dict()
yaml_str = yaml.dump(node_dict, sort_keys=False)

# Compare with target YAML block
assert yaml_str.strip() == expected_yaml.strip()
```

I've added a test case demonstrating this under test_node_graph_yaml_serialization in `tests/test_node_graph.py`.

PR #1582, opened 2026-06-09
Python a2ui_agent: Refactor a2ui_agent to depend on a2ui_core

nan-yu:
# Description

## Pre-launch Checklist

- [ ] I signed the [CLA].

[CHANGELOG]: ../CHANGELOG.md


agent_sdks/python/a2ui_core/src/a2ui/core/state/data_model.py
```diff
@@ -0,0 +1,175 @@
...
+        """Checks if a JSON Pointer path physically exists in the data model."""
+        tokens = self.parse_pointer(path)
+        if not tokens:
+            return True
+
+        current = self._data
+        for token in tokens:
+            if isinstance(current, dict) and token in current:
+                current = current[token]
+            elif isinstance(current, list) and NUMERIC_PATTERN.match(token):
```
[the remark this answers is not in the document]
nan-yu: can be fixed later

agent_sdks/python/a2ui_agent/src/a2ui/schema/catalog.py
jacobsimionato, line 64: I wonder if this class should be replaced with the new Catalog object from Core, or a function that returns a Catalog object (to defer async file system loads if necessary). That way, Catalog is the once source of truth, and developers can create them from a JSON file, or programmatically.

I think a workflow that is critical is to define a Catalog programmatically from a set of components and functions, and then use the Schema manager etc to create a prompt from it and parse the results. Right now, it seems like we're using the new Catalog abstraction for validation, but it's not yet possible to start with a Catalog object and then generate prompts from it. Ideally, the core Catalog representation should be the source of truth for the whole app.

This can be done in a future PR!

jacobsimionato, line 145: Maybe we can rename this to InferenceCatalog or something at some point? And I think it should be instantiated from a Catalog!

jacobsimionato (approved):
It looks like this PR goes a long way to removing logic from the existing agent SDK and having it depend on core instead which is great.

Do you plan to keep tweaking it after this? I think it might be possible to do some more cleanups and deduplication! But this seems like a big step towards the architecture we want, and it's nice that it seems like all the existing APIs are preserved.

agent_sdks/python/a2ui_agent/src/a2ui/schema/catalog.py
nan-yu: The renaming will be a breaking change. I'll wait until I'm back to start the migration ball.

nan-yu: Sure, the catalog class can be further refactored to depend more on the Core catalog. I'll follow up.

gspencergoog:
Looks like this broke the e2e tests.  https://github.com/a2ui-project/a2ui/issues/1648. The evals failure is unrelated, and already fixed.  I'll take a look at why.

PR #1590, opened 2026-06-09
feat(v1_0): inline catalog instructions and remove instructions.md

gspencergoog:
# Description

This addresses feedback from PR #1522 discussing the drawbacks of URI references inside catalog definitions sent to LLMs.

- Modify the Catalog schema definition in `client_capabilities.json` to accept plain Markdown strings for instructions instead of requiring relative file URIs.
- Inline the basic catalog layout instructions directly into `catalog.json` and delete the external `instructions.md` file entirely.
- Update the evaluation harness and protocol documentation to reflect that catalog instructions are inlined Markdown strings.


jgindin (commented):
What about the changes to `tools/build_catalog/assemble_catalog.py` ?

gspencergoog:
> What about the changes to `tools/build_catalog/assemble_catalog.py` ?

Good point, I was just focusing on the spec.  I've updated the `assemble_catalog.py` script to take zero or more `--instructions` file path arguments that will be read, concatenated and added to the instructions string field in the catalog.

jacobsimionato (approved):

PR #1597, opened 2026-06-09
feat(web_core): export basic_catalog in v0.9

josemontespg:
Exposes the basic catalog components under the v0.9 exports of @a2ui/web_core.

This is needed for 1P consumers that want to extend or compose the basic catalog.

### DIFF
--- a/renderers/web_core/src/v0_9/index.ts
+++ b/renderers/web_core/src/v0_9/index.ts
@@ -35,6 +35,7 @@ export * from './state/surface-components-model.js';
 export * from './state/surface-group-model.js';
 export * from './state/surface-model.js';
 export * from './errors.js';
+export * from './basic_catalog/index.js';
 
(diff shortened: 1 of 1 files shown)


renderers/web_core/src/v0_9/index.ts
```diff
@@ -35,6 +35,7 @@ export * from './state/surface-components-model.js';
 export * from './state/surface-group-model.js';
 export * from './state/surface-model.js';
 export * from './errors.js';
+export * from './basic_catalog/index.js';
```
ditman, line 38: I'm not a fan of `export *`, this makes it harder to control the real API surface of the package. I think this file should be very specific about what parts of /basic_catalog are being exported. Can we create an issue to reconsider making these barrel files more explicit?

ditman (approved):

PR #1598, opened 2026-06-09
style: auto-fix lint and formatting errors across monorepo

gspencergoog:
# Description

Fixes the Typescript lint issues found by `yarn lint:all`.

### DIFF
--- a/renderers/angular/a2ui_explorer/src/app/demo.component.ts
+++ b/renderers/angular/a2ui_explorer/src/app/demo.component.ts
@@ -718,7 +718,7 @@ export class DemoComponent implements OnInit, OnDestroy {
     try {
       const parsed = JSON.parse(this.currentCreateSurfaceMessageJson);
       this.currentCreateSurfaceMessageJson = JSON.stringify(parsed, null, 2);
-    } catch (e) {
+    } catch {
       // Ignore if invalid, don't format
     }
   }
@@ -751,7 +751,7 @@ export class DemoComponent implements OnInit, OnDestroy {
     try {
       const parsed = JSON.parse(this.currentDataModelJson);
       this.currentDataModelJson = JSON.stringify(parsed, null, 2);
-    } catch (e) {
+    } catch {
       // Ignore if invalid, don't format
     }
   }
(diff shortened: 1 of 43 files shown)


josemontespg:
How does this interact with https://github.com/a2ui-project/a2ui/blob/main/scripts/fix_format.sh ? Should we update the script?

gspencergoog:
> How does this interact with [`main`/scripts/fix_format.sh](https://github.com/a2ui-project/a2ui/blob/main/scripts/fix_format.sh?rgh-link-date=2026-06-09T22%3A58%3A25Z) ? Should we update the script?

Yes, I think?  I think eventually we should probably turn on formatting lint checks and fold the format fixing into the "yarn lint:fix:all" target.  What do you think?

gspencergoog:
> > How does this interact with [`main`/scripts/fix_format.sh](https://github.com/a2ui-project/a2ui/blob/main/scripts/fix_format.sh?rgh-link-date=2026-06-09T22%3A58%3A25Z) ? Should we update the script?
> 
> Yes, I think? I think eventually we should probably turn on formatting lint checks and fold the format fixing into the "yarn lint:fix:all" target. What do you think?

Actually, maybe we should go the other way and make all the formatting lints be warnings, and only use `prettier` for the code formatting.  That seems like a better path, really.  And maybe we should fold the formatting script into yarn targets so that it's integrated into the monorepo and you can just do "yarn format:all" and don't need to know about the special script.

EDIT: I looked into it, and it seems this is the right path because turning on formatting lint warnings makes linting a lot slower, and causes lots of squigglies in IDEs that are just formatting issues.  Better to keep them separate passes.

josemontespg:
Makes sense to keep them separate for now. Can you check that after formatting and linting the repo, both scripts agree on the style? I'd like to avoid the two fighting back and forth.

gspencergoog:
> Makes sense to keep them separate for now. Can you check that after formatting and linting the repo, both scripts agree on the style? I'd like to avoid the two fighting back and forth.

Actually, if you look at the next PR, I'm going to have the `fix_format.sh`  use a `yarn format:all` target which runs prettier and so they'll both use the same mechanism.

josemontespg (approved):

josemontespg:
Nice, thank you for doing this! LGTM 🚀

PR #1602, opened 2026-06-10
[eval] Introduce inference strategies

jiahaog:
See go/a2ui-eval-strategies for discussion. TL;DR: We want to experiment and evaluate N different ways of generating A2UI. Before this PR, we only can do so in one way, now we have scaffolding to do multiple ways.

- Introduce an additional dimension of inference strategies. It includes `direct` (the current) and `subagent_tool`, which is a naive tool call implementation just to illustrate having different inference strategies
  - Surprisingly the tool call impl is actually slower than the `direct` impl. This is surprising but is not the goal of the PR, I'll investigate it separately.
  - To run multiple strategies programatically, introduce a `main.py` that uses a `eval_set`. This changes the invocation command. Instead of using `uv run inspect eval tasks.py <model params>`, use `uv run main.py` which programatically runs both evals.
- The interface for a strategy is just a [`list[Solver]`](https://inspect.aisi.org.uk/solvers.html) from Inspect AI. A solver is just a way of composing python code that conditionally calls the model. Inspect AI provides a `get_model` abstraction over all the models so we can follow up and introduce a new dimension of models (will do in a follow up).
- Introduce `--sanity` to iterate on the evals
- Add a skill for inspect ai to make iterating on the evals easier


jiahaog (commented):
resolved comments

eval/main.py
```diff
@@ -0,0 +1,59 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+import traceback
+import argparse
+from inspect_ai import eval_set
```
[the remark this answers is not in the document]
jiahaog: https://inspect.aisi.org.uk/eval-sets.html.md

eval/tasks.py
```diff
@@ -86,10 +92,7 @@ async def score(state, target):  # pylint: disable=unused-argument
 
     return Task(
         dataset=dataset,
-        solver=[
-            a2ui_system_prompt(SCHEMA_PATH, CATALOG_PATH),
-            measured_generate()
-        ],
+        solver=get_solver(strategy, SCHEMA_PATH, CATALOG_PATH),
```
jacobsimionato, line 94: Can the schema path actually change? Should that just be hardcoded in at a lower level?

jacobsimionato (approved):
Nice!

eval/tasks.py
jiahaog: Not now, but I plan to follow up with datasets with different catalogs and schemas.

jiahaog:
@gspencergoog going to merge this first. Please feel free to leave comments and I'll be happy to follow up!

gspencergoog:
It looks like this broke the automated evals: https://github.com/a2ui-project/a2ui/issues/1649

gspencergoog:
Actually, I think it was broken before, but maybe running the unit tests now updates the uv.lock file.  Anyhow, I have a fix: https://github.com/a2ui-project/a2ui/pull/1652

Issue #1614, opened 2026-06-11
Python code in this repo should be typechecked


jiahaog:
We should use something like mypy

jiahaog:
@nan-yu would something like this be desirable? I can try to add these :)

nan-yu:
Yes, @josemontespg brought up the same thing in google3. I was planning to enable the strict type check after re-architecturing the agent SDK work to avoid merge conflicts in my PR chain.

I saw you have a WIP PR already. Feel free to get it in if it blocks your work as I will be out for 2 weeks starting 6/13.

jiahaog:
No worries, it can wait and your chain is definitely more important! My wip pr is mostly just a few prompts anyway :)

gspencergoog:
A2UI Issue Triage: Enabling Python typechecking (mypy) is planned and tracked as P2.

cc @jiahaog Since the blocker period for the agent SDK re-architecture is ending, please let us know if you want to proceed with your WIP PR.

jiahaog:
Oh, I thought we wanted to wait for Nan to finish her wip refactor since adding the types could add many conflicts?

gspencergoog:
> Oh, I thought we wanted to wait for Nan to finish her wip refactor since adding the types could add many conflicts?

Sorry, I was just triaging and gave it a P2.  If she's not done, then yeah, let's wait until after she comes back.

nan-yu:
All the WIP PRs have been merged into master. We can enable the strict type check now.

ditman:
@nan-yu is there anything else that needs to happen here? I'm coming from #1975. Also removing Jia Hao so this shows up in our triaging again

nan-yu:
This issue is good to be closed as fixed!

Just a heads-up on #1975—it's a bit different for the Python SDK. Some of our methods are still using dynamic types like `dict[str, Any]` instead of strongly typed classes like `A2uiClientCapabilities`. We'll need to run an audit of dynamic types in the Python SDK soon and update them to use strong types.

ditman:
Thanks for taking a look @nan-yu!

PR #1615, opened 2026-06-11
chore(python): enforce strict type-checking across a2ui_core

jiahaog:
Towards #1614.


jiahaog:
AI comment:

I have implemented all the suggested changes and pushed a new commit:
1. Relaxed `context` from `Optional[Dict[str, Any]]` to `Optional[Any]` to support custom context objects.
2. Tightened the `execute` callback parameter in `create_function_implementation` and updated the `self.functions` dictionary to strictly expect `Dict[str, Union[FunctionImplementation, FunctionApi]]`.
3. Aligned the `CoercedFunctionImplementation.execute` signature with the base class to satisfy the Liskov Substitution Principle.

All MyPy type-checks continue to pass with `strict=true`. The comments are resolved.

PR #1653, opened 2026-06-15
Update auto-assignee lists in auto-assignment.mjs

selamw1:
Update auto-assignee lists in auto-assignment.mjs

Issue auto-assignment changed to user name: 'Varun-S10'


yjbanov (approved):
lgtm

PR #1655, opened 2026-06-15
chore: Apply transformations to make code compatible with 1P

josemontespg:
This PR applies safe, style-only, or best-practice transformations to the codebase, reducing the number of transformations needed when importing to 1P. It covers adding the `override` keyword, replacing external imports with relative ones, and explicitly typing TextField. The transformation related to stripping the `accessor` keyword was skipped as it is not an inherently backwards-compatible/style change.

Next, I'll add enforcing lint enforcing so that we don't regress on this https://github.com/a2ui-project/a2ui/issues/1667


renderers/lit/src/v0_9/surface/render-a2ui-node.ts
```diff
@@ -43,6 +42,7 @@ export function renderA2uiNode(context: ComponentContext, catalog: Catalog<LitCo
     return nothing;
   }
 
-  const tag = unsafeStatic(implementation.tagName);
-  return html`<${tag} .context=${context}></${tag}>`;
+  const element = document.createElement(implementation.tagName);
+  (element as any).context = context;
+  return element;
```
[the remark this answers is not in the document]
josemontespg: Agent: Just undo this change.

scripts/fix_format.sh
```diff
@@ -28,16 +28,16 @@ echo "Running Prettier formatting for Node/Web assets..."
 if [ -f ".yarn/install-state.gz" ]; then
   # Local Node environment already installed; invoke standard script targets
   if [ "$CHECK_ONLY" = true ]; then
-    yarn format:check:all
+    corepack yarn format:check:all
```
ditman: Interesting, why not `corepack enable` at the top of the block?

renderers/lit/src/v0_9/surface/render-a2ui-node.ts
```diff
@@ -17,7 +17,7 @@
 import {nothing} from 'lit';
 import {html, unsafeStatic} from 'lit/static-html.js';
```
ditman, line 18: It seems this PR does not really address `unsafeStatic` in any way in the open source version of the repo. Could you update the description to match what it's doing?

ditman (approved):
Minor comment about the PR description

renderers/lit/src/v0_9/surface/render-a2ui-node.ts
josemontespg: Good catch! I attempted this but was harder than I thought, and then reverted it without updating the description. The description is up to date now.

josemontespg:
Ah, good catch! I've updated `fix_format.sh` to run `corepack enable` once at the top of the block so we don't have to repeat `corepack yarn` for each individual command.

PR #1658, opened 2026-06-15
feat(angular): export MarkdownRenderer in public API of v0.8

josemontespg:
Export the MarkdownRenderer symbol in the public api file for angular v0.8 and v0.9. This improves 1P compatibility when dependents import MarkdownRenderer directly.

No changes to v0.9, because it already exports MarkdownRenderer

### DIFF
--- a/renderers/angular/src/v0_8/data/index.ts
+++ b/renderers/angular/src/v0_8/data/index.ts
@@ -16,4 +16,4 @@
 
 export * from './processor';
 export * from './types';
-export {provideMarkdownRenderer} from './markdown';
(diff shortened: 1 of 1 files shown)


ditman:
v0.9 is exporting `MarkdownRenderer` from here:

https://github.com/a2ui-project/a2ui/blob/15427b333e2f720979593e43157f1e2479341db2/renderers/angular/src/v0_9/public-api.ts#L34

(It's harder to see because it's an `export *`)

ditman (approved):
This needs an update to the CHANGELOG because it'll affect the published API of the package

josemontespg:
Agent: Create a PR to add this to the appropriate CHANGELOG file

PR #1659, opened 2026-06-15
style: run prettier and apply formatting findings

josemontespg:
Automated prettier formatting run.


josemontespg:
I thought this would have affected more files. No need to merge this.

PR #1668, opened 2026-06-15
spec(v1.0)!: Remove redundant heading variants from Text component

gspencergoog:
## Summary

This pull request removes the redundant `h1`, `h2`, `h3`, `h4`, and `h5` heading variants from the `Text` component's `variant` field in both the basic and minimal catalogs for the v1.0 specification. Heading variants are redundant because the `Text` component natively supports Markdown formatting, and having both options can lead to confusing and conflicting UI states.

## Changes

- **Schema Updates**:
  - Removed `"h1"`, `"h2"`, `"h3"`, `"h4"`, and `"h5"` from the allowed `enum` values for the `Text` component's `variant` property in `specification/v1_0/catalogs/basic/catalog.json` and `specification/v1_0/catalogs/minimal/catalog.json`.
- **Example Catalog Payloads**:
  - Updated all JSON example files in `specification/v1_0/catalogs/basic/examples/` and `specification/v1_0/catalogs/minimal/examples/` (38 files total) to use Markdown formatting instead of `h1`-`h5` variants. For dynamic function call outputs, the `variant` field was removed.
- **Tests**:
  - Updated `specification/v1_0/test/cases/text_variants.json` to replace the valid `h1` test case with `caption`, and added a test case asserting that `h1` is now invalid.
  - Updated `specification/v1_0/test/cases/contact_form_example.jsonl` to remove the redundant `variant: h2` configuration from the title Text component.
- **Documentation**:
  - Updated `specification/v1_0/docs/basic_catalog_implementation_guide.md` to remove documentation regarding heading variant sizes and mappings.
  - Updated the example JSONL payload in `specification/v1_0/docs/a2ui_protocol.md` to remove the redundant `variant` property.

## Impact & Risks

- **Breaking Change**: Heading variants `h1` through `h5` are no longer accepted. Existing payloads that rely on `"variant": "h1"` (etc.) will fail validation under A2UI v1.0. Integrations should update payloads to use Markdown (e.g., prefixing text with `# `, `## `, etc.) or omit the `variant` field (defaulting to standard body text).

## Testing

1. Run the specification test runner from the root of the repository:
   ```bash
   python3 specification/v1_0/test/run_tests.py
   ```
2. Confirm that all 125 test cases pass, validating the catalog integrity, identifier rules, and test JSON cases.


specification/v1_0/catalogs/basic/examples/01_flight-status.json
```diff
@@ -77,22 +76,19 @@
...
+            "text": "## \u2192"
           },
           {
             "id": "destination",
             "component": "Text",
             "text": {
               "path": "/destination"
-            },
-            "variant": "h2"
+            }
```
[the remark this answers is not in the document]
gspencergoog: Great suggestion. Prepending the '## ' heading to the dynamic origin and destination values using 'formatString' restores the original header styling hierarchy.

specification/v1_0/catalogs/basic/examples/04_weather-current.json
```diff
@@ -51,16 +50,14 @@
...
           },
           {
             "id": "location",
             "component": "Text",
             "text": {
               "path": "/location"
-            },
-            "variant": "h3"
+            }
           },
```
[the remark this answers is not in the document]
gspencergoog: Done. Pre-pended heading hashes to the 'temp-low' and 'location' values using 'formatString' to restore the correct visual hierarchy.

specification/v1_0/catalogs/basic/examples/19_software-purchase.json
```diff
@@ -134,8 +130,7 @@
               "args": {
                 "value": "${formatCurrency(value: ${/total}, currency: 'USD')}/year"
               }
-            },
-            "variant": "h2"
+            }
```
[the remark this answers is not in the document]
gspencergoog: This one is OK: #### is a pretty small header anyhow, and this is just an example.

specification/v1_0/catalogs/basic/examples/01_flight-status.json
```diff
@@ -76,23 +75,26 @@
             "id": "origin",
             "component": "Text",
             "text": {
-              "path": "/origin"
-            },
-            "variant": "h2"
+              "call": "formatString",
+              "args": {
+                "value": "## ${/origin}"
```
jacobsimionato, line 80: Ahhh! I hadn't contemplated that formatString would become so much more important with this change, so you can inject static markdown into dynamic strings. I think it's okay though!

specification/v1_0/catalogs/minimal/catalog.json
```diff
@@ -19,7 +19,7 @@
             },
             "variant": {
               "type": "string",
-              "enum": ["h1", "h2", "h3", "h4", "h5", "caption", "body"]
+              "enum": ["caption", "body"]
```
jacobsimionato, line 22: Hmmm it feels weird to me that we're using the "variant" property of text only to distinguish between "caption" and "body". I wonder if "primary" (typically black or white) and "secondary" (typically grey) might be more generic? Or even to remove the "variant" property entirely?

I don't feel strongly though. I think we will end up creating a new, finer grained transport-focused catalog at some point.

jacobsimionato (approved):

specification/v1_0/catalogs/minimal/catalog.json
gspencergoog: I thought about this too, but the reason I kept it was that "caption" makes the text smaller than body text, and there's no way to do that with Markdown.

specification/v1_0/catalogs/basic/catalog.json
```diff
@@ -28,7 +28,7 @@
             "variant": {
               "type": "string",
               "description": "A hint for the base text style.",
-              "enum": ["h1", "h2", "h3", "h4", "h5", "caption", "body"],
+              "enum": ["caption", "body"],
```
ditman, line 31: Isn't this backwards breaking with v0.9? IIRC we generate the values of "variant" from here, so now the code that we have that handles non-existing variants will be using non-existing enum values? (It might work in JS though)

gspencergoog: No, because this only applies to v1.0 adapters, which will be backward compatible with v0.9. If they receive a message with `"version": "v0.9"` and the old values of variant, they will render them as they do now.  If they receive a `"version": "v1.0"` message, if they contain h1-h5 they will send back an error.

We'll have to address this when we implement the v1.0 adapters.

ditman: Ah yeah, we can distinguish by the version to handle this differently for v1.0

PR #1678, opened 2026-06-16
feat(express): Introduce A2UI Express compiler, draft proposals, and evaluations

gspencergoog:
## Summary
This pull request introduces A2UI Express, an experimental, compact domain-specific language (DSL) that allows agents to express UI layouts with minimal token usage. It provides a complete compilation/decompilation pipeline, integrates it into the evaluations strategy suite, and documents the specification draft under proposals.

## Evaluation Results
Below is a summary of the evaluations benchmark run (47 samples) comparing the layout generation strategies using the lightweight `google/gemini-3.1-flash-lite` model:

| Strategy | `a2ui_scorer` Accuracy (Syntax) | `measured_model_graded_qa` Accuracy (Semantics) | Avg Latency | Median Latency | Output Tokens | Total Tokens |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| **`direct`** (Raw JSON) | 87.23% | 93.62% | 4.99s (baseline) | 3.57s (baseline) | 35,022 (baseline) | 527,882 (baseline) |
| **`express`** (A2UI Express DSL) | **91.49%** | **84.04%** | **1.09s (78% reduction)** | **1.06s (70% reduction)** | **9,912 (72% reduction)** | **232,748 (56% reduction)** |

By compiling into inline components and inline data models, the **Express** strategy matches or exceeds the direct strategy's syntactic validity while reducing generation latencies by **70% (over 3.3x faster)** and saving **56% of total generation tokens** (and **72% of output tokens**).

### Python Agent SDK
* **`agent_sdks/python/a2ui_agent/src/a2ui/express/`**: Added the core package including:
  * `compiler.py`: Compiles flat Express DSL statements into standard A2UI wire JSON. Includes robust exception handling and positional argument validation.
    * Added support for bare `$` path token resolving to relative root JSONPointer `{"path": ""}`.
    * Added recursive nested check compilation (e.g. `?and([?required, ?email])`).
    * Added automatic unrolling of inline component constructor instantiations (e.g. `Row([Text("Soup")])` -> `_inline_1`).
    * Added lexical scanner rules to ignore Python-style `#` and JS-style `//` comments outside string literals.
  * `decompiler.py`: Translates standard wire JSON layouts back into the flat DSL representation, with safety guards on missing properties.
  * `prompt_generator.py`: Generates LLM system instruction prompt contracts containing component signatures, enums, static property indicators, and nested schema requirements dynamically.
  * `schema_helper.py`: Crawls catalog JSON schemas to resolve positional parameter bounds and properties, with type checks to skip boolean schemas inside `allOf` loops.
* **`agent_sdks/python/a2ui_core/`**: Improved core validation routines to make them schema-driven:
  * `node_graph.py` & `integrity_checker.py`: Decoupled structural and static checks from hardcoded catalog definitions, resolving constraints dynamically from catalog schemas.
* **Feature Flag Control**: Added environment variable switches:
  * `A2UI_EXPRESS_ENABLED`: If not set to `true`, importing `a2ui.express` raises an `ImportError`.
  * `A2UI_VERSION_1_0`: Gates the new dynamic JSON schema validator for v1.0. Automatically enabled if `A2UI_EXPRESS_ENABLED` is true.
* **SDK Tests**: Added comprehensive test cases in `tests/express/test_express.py` covering DSL tokenization, parsing, expression nesting, map variable inlining, comments, and validator gating. Added `tests/express/test_cli_tools.py` to test all 4 CLI wrapper scripts under `specification/proposals/express/`, achieving **90% test coverage** for the CLI utility suite.

### Specification & Proposals
* **`specification/proposals/express/`**: Relocated and formalized the post-1.0 draft specification (`a2ui_express.md`, `evolve_express.md`, and sample configs) to prevent polluting the ratified `v1_0` baseline.
* **DSL Examples**: Relocated all `.a2ui` evaluation layout examples to `specification/proposals/express/examples/`.

### Evaluations Suite
* **`eval/a2ui_eval/strategies/express.py`**: Added the `express` evaluation solver that runs and merges layout accuracy scores. Express compiler results generate standard A2UI v1.0 JSON payloads directly.
* **`eval/tasks.py` / `main.py`**: Shifted the core baseline evaluation task to run against **A2UI v0.9.1** (renamed to `a2ui_v0_9_1_eval`), while keeping the `express` strategy target isolated to A2UI v1.0.
* **Evaluation Datasets**: Introduced a new translated dataset `eval/datasets/v1_0_prompts.yaml` that adapts prompt texts and target descriptions to use v1.0 terminology (inline components, `longText` instead of `multiline`, and omission of `returnType` properties). This aligns target outputs with strategy versions for clean, accurate LLM grading.

## Impact & Risks
* **No Breaking Changes**: The compilation pipeline is isolated and disabled by default under the `A2UI_EXPRESS_ENABLED` environment switch.
* **No Regression**: The existing baseline evaluation runs (`direct`, `subagent_tool`) are preserved as defaults.

## Testing
* **Unit Tests**: Run `uv run pytest` inside `agent_sdks/python/a2ui_agent/` folder.
* **Evaluations Run**: To run the new strategy in the benchmark runner, run `uv run main.py --strategies express` inside the `eval/` folder.


agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,657 @@
...
+        if var_name.startswith("$"):
+          data_path_assignments[var_name] = parsed_val
+        else:
+          raw_symbols[var_name] = parsed_val
+      except Exception:
+        if var_name.startswith("$"):
+          data_path_assignments[var_name] = None
+        else:
+          raw_symbols[var_name] = {"call": "Text", "args": ["Loading..."]}
+
```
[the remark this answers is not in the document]
gspencergoog: Good catch. I replaced the string-based splitting of statements with full statement tokenization. The compiler now tokenizes the statement first and inspects the prefix tokens to distinguish assignments from standalone function calls, supporting statements that contain '=' characters inside string literals.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,657 @@
...
+    elif kind == "STRING":
+      val = val[1:-1].replace('\\"', '"')
+    elif kind == "NUMBER":
+      val = float(val) if "." in val else int(val)
+    elif kind == "BOOLEAN":
+      val = val == "true"
+    elif kind == "NULL":
+      val = None
+    tokens.append((kind, val))
+  return tokens
```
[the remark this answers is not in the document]
gspencergoog: Added tracking of the scanner's end match index to catch unmatched character gaps. A SyntaxError is now raised if any unexpected characters are encountered.

eval/a2ui_eval/strategies/express.py
```diff
@@ -0,0 +1,219 @@
...
+            except Exception:
+                # If JSON parsing failed, fall back to compiling the string as DSL
+                pass
+
+        # 2. Extract DSL content inside <a2ui> tags if present, or clean markdown blocks
+        dsl_content = completion
+        if "<a2ui>" in completion:
+            start_idx = completion.find("<a2ui>") + len("<a2ui>")
+            end_idx = completion.find("</a2ui>")
+            dsl_content = completion[start_idx:end_idx].strip()
```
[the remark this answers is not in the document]
gspencergoog: Fixed slicing bounds calculation. Slicing now correctly checks if end_idx != -1 to avoid dropping the final character of completion strings when the closing tag is missing.

eval/a2ui_eval/strategies/express.py
```diff
@@ -0,0 +1,219 @@
...
+                            if "updateComponents" in msg and isinstance(msg["updateComponents"], dict):
+                                uc = msg["updateComponents"]
+                                if "dsl" in uc and isinstance(uc["dsl"], str):
+                                    inner_dsl = uc["dsl"].strip()
+                                    if "<a2ui>" in inner_dsl:
+                                        s_idx = inner_dsl.find("<a2ui>") + len("<a2ui>")
+                                        e_idx = inner_dsl.find("</a2ui>")
+                                        inner_dsl = inner_dsl[s_idx:e_idx].strip()
+                                    compiled_inner = compiler.compile(inner_dsl, surface_id=surface_id)
+                                    uc.pop("dsl")
```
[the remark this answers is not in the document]
gspencergoog: Fixed the inline DSL slicing bounds checking as suggested.

agent_sdks/python/a2ui_agent/src/a2ui/express/prompt_generator.py
```diff
@@ -0,0 +1,207 @@
...
+              desc_suffix = f" - {desc}" if desc else ""
+              sub_keys.append(f"    * {sub_k}{desc_suffix}")
+            prop_details.append(f"  - {p}: Map with keys:\n" + "\n".join(sub_keys))
+          elif p_schema.get("type") == "array" and "items" in p_schema:
+            items_schema = p_schema["items"]
+            if (
+                isinstance(items_schema, dict)
+                and items_schema.get("type") == "object"
+                and "properties" in items_schema
+            ):
```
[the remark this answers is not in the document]
gspencergoog: Added isinstance checks for dictionary schemas to prevent AttributeErrors if boolean schemas are defined in the component catalogs.

agent_sdks/python/a2ui_agent/src/a2ui/express/compiler.py
```diff
@@ -0,0 +1,673 @@
...
+          parsed_val = parser.parse_expression()
+          if var_name.startswith("$"):
+            data_path_assignments[var_name] = parsed_val
+          else:
+            raw_symbols[var_name] = parsed_val
+        except Exception:
+          if var_name.startswith("$"):
+            data_path_assignments[var_name] = None
+          else:
+            raw_symbols[var_name] = {"call": "Text", "args": ["Loading..."]}
```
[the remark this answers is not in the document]
gspencergoog: Good point. I updated the exception handling in the statement compilation loop to propagate a descriptive ValueError wrapping the original parse error.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,673 @@
...
+              f"Inline component instantiations are disallowed: {fn_name}(...). You"
+              " must assign it to a variable on its own line and reference that"
+              " variable instead."
+          )
+
+        # Is it a reserved Template signature?
+        if fn_name == "_template":
+          path_val = self._compile_value(fn_args[0], raw_symbols, is_action)
+          comp_id_val = self._compile_value(fn_args[1], raw_symbols, is_action)
+          return {"path": path_val["path"], "componentId": comp_id_val}
```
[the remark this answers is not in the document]
gspencergoog: Fixed. Added validation to ensure the _template helper is called with exactly 2 parameters (path and component template ID).

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,673 @@
...
+
+        # Is it a reserved Event signature?
+        if fn_name == "Event":
+          event_name = fn_args[0] if len(fn_args) > 0 else ""
+          context_map = fn_args[1] if len(fn_args) > 1 else {}
+          compiled_context = {}
+          for k, v in context_map.items():
+            compiled_context[k] = self._compile_value(v, raw_symbols, is_action)
+          return {"event": {"name": event_name, "context": compiled_context}}
```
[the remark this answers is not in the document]
gspencergoog: Exhaustively handled both dictionary and list layouts for compiled context arguments within the Event helper logic.

agent_sdks/python/a2ui_agent/src/a2ui/express/decompiler.py
```diff
@@ -0,0 +1,289 @@
...
+            for idx in range(start_idx, len(check_props)):
+              p = check_props[idx]
+              if p in check_args:
+                explicit_args_reprs.append(
+                    self._decompile_value(check_args[p], comp_ids)
+                )
+
+            if message and message != f"{check_name.capitalize()} check failed":
+              escaped_msg = message.replace('"', '\\"')
+              explicit_args_reprs.append(f'"{escaped_msg}"')
```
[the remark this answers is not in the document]
gspencergoog: Corrected bounds checking. Added a safety check to ensure check_name is not None/empty before capitalizing.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/schema_helper.py
```diff
@@ -0,0 +1,188 @@
...
+    if not schema:
+      return None
+
+    sub_schemas = [schema]
+    if "allOf" in schema:
+      sub_schemas.extend(schema["allOf"])
+
+    for sub in sub_schemas:
+      if "properties" in sub and property_name in sub["properties"]:
+        return sub["properties"][property_name]
```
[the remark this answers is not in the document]
gspencergoog: Added safety checks to ignore boolean sub-schemas inside allOf loops, preventing TypeErrors.

specification/proposals/express/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+Once the local Gemma 4 server is active, you can run the end-to-end compilation and validation pipeline.
+
+### Running local inference and validation
+
+To test local generation and compiler correctness in a single command, run the inference script with the `--mlx` flag:
+
+```bash
+# Navigate to the express directory
+cd main/specification/v1_0/express
```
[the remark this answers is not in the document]
gspencergoog: Corrected the outdated specification directory path inside the README documentation.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,834 @@
...
+    for line in dsl_text.splitlines():
+      trimmed = line.strip()
+      if "<a2ui>" in trimmed:
+        inside_a2ui = True
+        continue
+      if "</a2ui>" in trimmed:
+        inside_a2ui = False
+        continue
+      if inside_a2ui:
+        lines.append(line)
```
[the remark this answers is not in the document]
gspencergoog: Applied suggestion.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/schema_helper.py
```diff
@@ -0,0 +1,233 @@
...
+          ref = sub["$ref"]
+          if "Checkable" in ref:
+            is_checkable = True
+        if "properties" in sub:
+          props.update(sub["properties"])
+          for pk, pv in sub["properties"].items():
+            if isinstance(pv, dict) and "enum" in pv:
+              self.component_property_enums[(name, pk)] = pv["enum"]
+        if "required" in sub:
+          reqs.extend(sub["required"])
```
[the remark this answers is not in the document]
gspencergoog: Applied suggestion.

agent_sdks/python/a2ui_agent/src/a2ui/express/decompiler.py
```diff
@@ -0,0 +1,302 @@
...
+          if len(compiled_checks_list) == 1:
+            args_reprs.append(compiled_checks_list[0])
+          else:
+            args_reprs.append(f"[{', '.join(compiled_checks_list)}]")
+          continue
+
+        # Map other regular properties
+        if prop_name in c:
+          val = c[prop_name]
+          args_reprs.append(self._decompile_value(val, comp_ids))
```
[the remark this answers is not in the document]
gspencergoog: I have added the `is_ref` parameter to `_decompile_value`. Now, string values are only decompiled as variable references when they are known to represent component references, preventing collision with plain string values.

agent_sdks/python/a2ui_agent/src/a2ui/express/decompiler.py
```diff
@@ -0,0 +1,302 @@
...
+      escaped = val.replace('"', '\\"')
+      return f'"{escaped}"'
+
+    if isinstance(val, bool):
+      return "true" if val else "false"
+
+    if val is None:
+      return "_"
+
+    return str(val)
```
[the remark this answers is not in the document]
gspencergoog: I have added the `is_ref` parameter to `_decompile_value`. Now, string values are only decompiled as variable references when they are known to represent component references, preventing collision with plain string values.

eval/bin/translate_dataset.py
```diff
@@ -0,0 +1,113 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import yaml
+import os
+
+source_path = "/Users/gspencer/code/a2ui/a2ui_express/eval/datasets/v0_9_prompts.yaml"
+dest_path = "/Users/gspencer/code/a2ui/a2ui_express/eval/datasets/v1_0_prompts.yaml"
```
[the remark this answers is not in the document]
gspencergoog: I have updated the script to resolve dataset paths dynamically relative to the script's directory, removing the absolute path dependency.

eval/a2ui_eval/strategies/express.py
```diff
@@ -0,0 +1,152 @@
...
+# limitations under the License.
+
+import json
+import os
+os.environ["A2UI_EXPRESS_ENABLED"] = "true"
+from inspect_ai.solver import Solver, solver, TaskState, Generate
+from inspect_ai.model import ChatMessageSystem, ModelOutput, ChatCompletionChoice, ChatMessageAssistant, ChatMessageUser
+from a2ui.express.prompt_generator import ExpressPromptGenerator
+from a2ui.express.compiler import ExpressCompiler
+from ..shared.utils import measured_generate
```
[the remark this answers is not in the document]
gspencergoog: I have moved the `re` import to the top of the file to adhere to PEP 8 import grouping standards.

eval/a2ui_eval/strategies/express.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+@solver
+def compile_express_dsl(catalog_path: str) -> Solver:
+    """Solver to compile generated A2UI Express DSL back to standard JSON."""
+    compiler = ExpressCompiler(catalog_path)
+
+    # Initialize the catalog schema validator for parsing
+    from a2ui.schema.manager import A2uiSchemaManager
+    from a2ui.schema.catalog import CatalogConfig
```
jiahaog: Shouldn't be inline imports

eval/a2ui_eval/strategies/express.py
```diff
@@ -0,0 +1,152 @@
...
+                content = re.sub(
+                    r"(?i)generate a? 'createSurface' (?:message )?and a? 'updateComponents' (?:message)?",
+                    "generate A2UI Express DSL",
+                    content
+                )
+                content = re.sub(
+                    r"(?i)generate 'createSurface' and 'updateComponents' messages?",
+                    "generate A2UI Express DSL",
+                    content
+                )
```
jiahaog: This seems like a hack. Let's update the dataset to not mention "JSON"?

eval/a2ui_eval/strategies/express.py
```diff
@@ -0,0 +1,152 @@
...
+                message.content = (
+                    "Translate the following request into A2UI Express DSL wrapped inside <a2ui> and </a2ui> sentinels:\n\n"
+                    + content
+                    + "\n\nREMINDER: You must output ONLY A2UI Express DSL wrapped in <a2ui> and </a2ui> sentinels. "
+                    "Do NOT output JSON or <a2ui-json> blocks under any circumstances. Directly generating JSON will fail compilation."
+                )
+        print("MESSAGES PATH:", [(m.role, m.content[:40]) for m in state.messages])
```
jiahaog: Remove?

eval/a2ui_eval/strategies/express.py
```diff
@@ -0,0 +1,152 @@
...
+                    r"(?i)generate a? 'createSurface' (?:message )?and a? 'updateComponents' (?:message)?",
+                    "generate A2UI Express DSL",
+                    content
+                )
+                content = re.sub(
+                    r"(?i)generate 'createSurface' and 'updateComponents' messages?",
+                    "generate A2UI Express DSL",
+                    content
+                )
+                message.content = (
```
jiahaog: Can we try putting this in the `generator.generate_prompt()`? Is it because we're using a lite model and that's why it isn't working great as part of the system prompt?

eval/a2ui_eval/strategies/express.py
```diff
@@ -0,0 +1,152 @@
...
+        try:
+            # 1. Compile DSL to JSON
+            compiled_json = compiler.compile(dsl_content, surface_id=surface_id)
+            
+            # 2. Validate using catalog schema validator (runs integrity checker too)
+            validator.validate([compiled_json])
+
+            # 3. Successful compilation and validation: format and yield
+            messages = [compiled_json]
+            formatted = f"<a2ui-json>\n{json.dumps(messages, indent=2)}\n</a2ui-json>"
```
jiahaog: Doesn't seem right that there's so much parsing at this level. Should we have a helper with the same signature as  `parse_respone`:

https://github.com/a2ui-project/a2ui/blob/c308ef93a8f74d5f29c916f21a6cdaf4f8a874a4/agent_sdks/python/a2ui_agent/src/a2ui/parser/parser.py#L45

That lives inside the agent SDK? Something like `parse_express_response`

eval/datasets/v1_0_prompts.yaml
```diff
@@ -0,0 +1,968 @@
+U2FsdGVkX1901+Sbyrv+JJtHyPfAmeaIPWC/Gu1bYySZVPDqkav5dfiaFlzCYEkw
```
jiahaog: What's the high level delta here between 0.9 and 1.0?

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+"""A2UI Express parser, compiler, and generator package.
+
+Provides high-performance conversion utilities to compile A2UI Express DSL syntax
+into standard A2UI v1.0 wire JSON messages and vice-versa.
+"""
+
+import os
+
+if os.environ.get("A2UI_EXPRESS_ENABLED", "").lower() not in ("true", "1", "yes"):
```
jiahaog, line 23: Would it also help if we created a `experimental` namespace in the path so it's obvious when importing . Instead of `import ... express`, the call site is `import ....experimental.express`

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
+# A2UI Express technical specification
+
+A2UI Express is a compact, model-optimized declarative syntax designed for dynamic generative user interfaces. It acts as an intermediate, highly compressed representation that on-device large language models generate to describe user interfaces. A host-side compiler parses this syntax and compiles it into standard A2UI v1.0 wire protocol payloads.
+
+## Core design goals
+
+The design of A2UI Express focuses on four main requirements:
+
+- Token footprint reduction. Generative models spend excessive output tokens when producing verbose JSON structures. A2UI Express removes structural keys, brackets, and repeated quotes, reducing output tokens by 55% to 70% compared to native A2UI wire payloads.
+- On-device model optimization. Small local models, such as Gemma 4 E2B and E4B, operate with limited context windows and reasoning budgets. The syntax uses clean positional signatures that fit into prompt contracts without consuming excessive context.
```
jiahaog, line 10: What does "clean positional signatures" refer to?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+Every component definition is assigned to a unique, alphanumeric variable. The compiler uses these variables to resolve parent-child hierarchies. A reserved variable named `root` acts as the primary entry point for the interface tree.
+
+To eliminate syntax errors from complex bracket structures and enable line-oriented streaming compilation, A2UI Express prohibits inline component nesting. Component constructor calls (e.g., `Text(...)`, `Column(...)`) can **only** appear on the right-hand side of a variable assignment (`var = ComponentName(...)`). They **cannot** be passed directly as positional arguments to other components. Instead, you must declare them separately and reference their variable names.
+
+### Core primitive types
+
+The syntax supports three literal primitive types:
+
+- Strings are enclosed in straight double quotes, for example `"Enter your name"`.
```
jiahaog: What about escaping?

eval/a2ui_eval/strategies/express.py
```diff
@@ -0,0 +1,152 @@
...
+from inspect_ai.solver import Solver, solver, TaskState, Generate
+from inspect_ai.model import ChatMessageSystem, ModelOutput, ChatCompletionChoice, ChatMessageAssistant, ChatMessageUser
+from a2ui.express.prompt_generator import ExpressPromptGenerator
+from a2ui.express.compiler import ExpressCompiler
+from ..shared.utils import measured_generate
+
+@solver
+def a2ui_express_prompt(catalog_path: str) -> Solver:
+    """Solver to inject A2UI Express prompt contract instructions."""
+    generator = ExpressPromptGenerator(catalog_path)
```
jacobsimionato: Next, let's combine ExpressPromptGenerator and ExpressCompiler into one ExpressInferenceFormat which implements AbstractInferenceFormat, so that people can implement other inference formats and reuse this inspect_ai strategy.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/decompiler.py
```diff
@@ -0,0 +1,305 @@
...
+      for k, v in current.items():
+        recurse(v, f"{path}/{k}")
+    else:
+      results.append((path, current))
+
+  recurse(data_dict, "")
+  return results
+
+
+class ExpressDecompiler:
```
jacobsimionato, line 72: Yes! Love the idea of having the reverse direction too, so we can make sure that the model also understands sees content through the DSL

specification/proposals/express/examples/02_email-compose.a2ui
jacobsimionato, line 1: Do these examples need to be checked in? Could we generate them on demand using the decompiler from the canonical A2UI examples?

specification/proposals/express/leaderboard.json
jacobsimionato, line 1: Amazing! Should this be checked in though?

jacobsimionato (commented):
Exciting!!!

jacobsimionato:
Are you intending to land this PR as-is? Any chance of breaking it up a bit, to make it easier to do thorough reviewing?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+    "args": {
+      "url": "https://example.com"
+    }
+  }
+}
+```
+
+## Compilation pipeline
+
+The compilation pipeline runs on the host application. It takes the plain text stream of A2UI Express, processes it, and emits a standard A2UI v1.0 JSON payload.
```
jiahaog: > The compilation pipeline runs on the host application.

Do we need to opine on this? I thought the compiler runs (and is an implementation detail of the agent). If it runs on the client, we have more problems like dealing with versioning

specification/proposals/express/scripts/run_decompiler.py
```diff
@@ -0,0 +1,132 @@
...
+            "python",
+            "a2ui_agent",
+            "src",
+        )
+    ),
+)
+from a2ui.express.decompiler import ExpressDecompiler
+
+
+def decompile_example(example_path: str, catalog_path: str) -> str:
```
jiahaog, line 46: Is this just an internal tool? Not sure if there's any use case for a developer to use it?

jiahaog:
Some more general comments:

- For the benchmarks we should use at least use a non-lite model. It's unlikely that a lite model will be sufficient in the "direct" strategy since the agent needs to do domain specific things, in addition to generating output as A2UI. I'll send some changes to allow us to benchmark multiple providers besides just Gemini too.
- Is your intention for express to be an implementation detail of the agent SDK? If not, perhaps we should start with that first (also see my inline comment on where the compiler should run)

specification/proposals/express/leaderboard.json
gspencergoog: No, that is an artifact of the improvement loop I was running.  I'll remove it.

specification/proposals/express/scripts/run_decompiler.py
gspencergoog: These are just internal.  I needed them in order to be able to try out the various parts, and to convert existing examples, etc.  I can delete them or move them someplace where it is more obvious that they are just accessories for our own use.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/decompiler.py
gspencergoog: This was partly to make sure that I could "round trip" things to make sure that the express format could support everything that A2UI can.  It was also helpful in converting examples in the catalog to use in the prompt.

specification/proposals/express/examples/02_email-compose.a2ui
gspencergoog: We could generate them, yes.  That's how I produced them.

specification/proposals/express/a2ui_express.md
gspencergoog: It just means that the DSL uses argument positions without named arguments. "Clean" is because they don't have the named arguments, I guess.

gspencergoog:
> Are you intending to land this PR as-is? Any chance of breaking it up a bit, to make it easier to do thorough reviewing?

No, I mainly wanted to get your (and Jia Hao's) impressions on it.  I'll split it up today into at least two PRs, one with the proposal directory and docs, and one with the Python agent implementation and evals, taking into account both of your review comments.

I was maybe a little overexcited to share the results and it needs some more pruning and cleaning.

gspencergoog:
> Some more general comments:
> 
> * For the benchmarks we should use at least use a non-lite model. It's unlikely that a lite model will be sufficient in the "direct" strategy since the agent needs to do domain specific things, in addition to generating output as A2UI. I'll send some changes to allow us to benchmark multiple providers besides just Gemini too.

Yes, I agree.  I did run the direct model on Lite too, so it's at least comparable, but you're right that there isn't a lot more reasoning overhead left there after building the UIs.  I mainly wanted to see 1) that a Lite model could handle the reasoning needed (showing that it wasn't too reasoning heavy), and 2) to see how fast we could go if latency were the only driver.

It works well in Flash 3.5, but is of course slower.  I found that limiting the thinking budget helped a bit there.

> * Is your intention for express to be an implementation detail of the agent SDK? If not, perhaps we should start with that first (also see my inline comment on where the compiler should run)

Yes, that is my intention.  The rendering clients should have no idea that Express was ever involved, and just see regular A2UI.

specification/proposals/express/examples/08_user-profile.a2ui
```diff
@@ -0,0 +1,27 @@
...
+$/username = "@sarahchen"
+root = Card(main-column)
+main-column = Column([header, info, bio, stats-row, follow-btn], _, "center")
+header = Image($/avatar, _, "cover", "avatar")
+info = Column([name, username], _, "center")
+name = Text($/name)
+username = Text($/username, "caption")
+bio = Text($/bio, "body")
+stats-row = Row([followers-col, following-col, posts-col], "spaceAround")
+followers-col = Column([followers-count, followers-label], _, "center")
```
yjbanov: Are we sure we want to support names that contain "-" and other special characters? In the future, when models become more advanced, we might want to introduce features that would take advantage of special characters, but allowing them in identifier names could cause ambiguities in the grammar.

specification/proposals/express/examples/35_markdown-text.a2ui
```diff
@@ -0,0 +1,11 @@
+root = Card(main-column)
+main-column = Column([title-text, markdown-content], _, "stretch")
+title-text = Text("### Markdown Rendering")
+markdown-content = Text("# Heading 1
```
yjbanov: Another " can almost certainly appear inside a multiline string like this. How do we ensure we can still parse it correctly?

specification/proposals/express/examples/32_advanced-form-validator.a2ui
```diff
@@ -0,0 +1,14 @@
+$/formData/agree = false
+$/formData/email = ""
+$/formData/phone = ""
+$/formData/zip = ""
+$/now = "2025-12-15T12:00:00Z"
+root = Card(main-column)
+main-column = Column([welcome-text, email-field, phone-field, zip-field, terms-checkbox, submit-btn], _, "stretch")
+welcome-text = Text(formatString("Hello! Today is ${formatDate(value: ${/now}, format: 'EEEE, MMMM d')}."))
```
yjbanov: Do we need `formatString`? Dart and JS can format strings without it. Just the `${` syntax is enough. It's also fewer tokens this way.

gspencergoog: `formatString` is implemented in the catalog.  This makes it so that a catalog author can implement whatever string formatting they want to use (e.g. you could implement a `printf` function and use `printf` formatting instead).  It also makes it a lot easier to implement a renderer, since the string formatting isn't part of the specification, and implementing `formatString` is actually pretty involved (it's reactive, and so generates new strings whenever the interpolated values, including interpolated function calls, change, for instance).

We provide an implementation in the core libraries that catalog writers can leverage so they can just include it.

I can see the argument for including it in the language, but if we do that for Express, it needs to be included in A2UI as well, and implemented in all renderers.

specification/proposals/express/examples/08_user-profile.a2ui
gspencergoog: Good point.  We probably should limit it to the recommendations in https://www.unicode.org/reports/tr31/ so that we can allow Unicode characters in identifiers (I recently added this to A2UI as well).

specification/proposals/express/examples/35_markdown-text.a2ui
gspencergoog: "Naaah, that'll
"never" happen."  :-)

Sure, we should specify quoting rules.  I really like the """ rules that other languages like Dart have.  I want to make sure we can keep the prompt small though, so it's easy to describe, but I don't want to open the door to something like just saying "Use Python string quoting rules" because that opens the door to supporting lots of other syntax (raw string quoting, multiple character escape formats, misinterpreting conflicting string interpolation formats, etc.).  Maybe there's a way we can say it that is concise but doesn't mention a language.

e.g. "For quoting strings, surround with a double quote.  If a double quote appears in the string, then use """ around the string. Individual double quotes may also be escaped with a preceding backslash. For strings that are raw strings interpreted literally, precede the string with an r: r" or r""" ".

Unlike other languages, one of the driving criteria here is that its rules are able to be concisely described in English prose.

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
+# A2UI Express technical specification
+
+A2UI Express is a compact, model-optimized declarative syntax designed for dynamic generative user interfaces. It acts as an intermediate, highly compressed representation that on-device large language models generate to describe user interfaces. A host-side compiler parses this syntax and compiles it into standard A2UI v1.0 wire protocol payloads.
+
+## Core design goals
+
+The design of A2UI Express focuses on four main requirements:
+
+- Token footprint reduction. Generative models spend excessive output tokens when producing verbose JSON structures. A2UI Express removes structural keys, brackets, and repeated quotes, reducing output tokens by 55% to 70% compared to native A2UI wire payloads.
```
munificent, line 9: It's surprising that this reduces token size so much. In theory, a directly nested syntax should be *more* concise than what's proposed here because it avoids repeating path components, as in:

```
foo/bar/baz/bang/a = 1
foo/bar/baz/bang/b = 2
foo/bar/baz/bang/c = 3
foo/bar/baz/bang/d = 4
foo/bar/baz/bang/e = 5
// 55 tokens ("token" in the PL sense, not LLM sense)

// Versus:
foo(
  bar(
    baz(
      bang(
        a = 1
        b = 2
        c = 3
        d = 4
        e = 5
      )
    )
  )
)
// 27 tokens
```

So is the improvement here really from the flattening, or from not having a nesting syntax that has a lot of other unnecessary boilerplate like quoted key names, argument names, comma separators, etc.?

If so, perhaps it would be beneficial for this notation to allow nesting too?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+- On-device model optimization. Small local models, such as Gemma 4 E2B and E4B, operate with limited context windows and reasoning budgets. The syntax uses clean positional signatures that fit into prompt contracts without consuming excessive context.
+- Streaming compatibility. The line-oriented grammar allows the client host to parse and build the component hierarchy line-by-line, enabling progressive rendering of the interface before the model finishes its output.
```
munificent, line 11: One of the examples below looks like:

```
root = Card(main-column)
main-column = Column([icon, title, description, actions], _, "center")
icon = Icon($/icon)
title = Text($/title, "h3")
...
```

Lines here often refer to names declared on later lines. That implies that we can't always process lines as they come in, unless the system can gracefully handle references to unknown entities.

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+```
+variable_name = ComponentName(argument1, argument2, ...)
+```
+
+The syntax rules define how types, structures, data paths, and validation actions are expressed.
+
+### Variable declarations
+
+Every component definition is assigned to a unique, alphanumeric variable. The compiler uses these variables to resolve parent-child hierarchies. A reserved variable named `root` acts as the primary entry point for the interface tree.
```
munificent: What happens if a user refers to `root`? What does this do:

```
root = Row([root])
```

Or is it "write-only" in some way?

Related: how are cyclic references handled?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+To eliminate syntax errors from complex bracket structures and enable line-oriented streaming compilation, A2UI Express prohibits inline component nesting. Component constructor calls (e.g., `Text(...)`, `Column(...)`) can **only** appear on the right-hand side of a variable assignment (`var = ComponentName(...)`). They **cannot** be passed directly as positional arguments to other components. Instead, you must declare them separately and reference their variable names.
```
munificent, line 41: For what it's worth, I've seen various little DSLs and hobby languages over the years try to stake a claim like this in the name of simplicity (or because their authors aren't comfortable writing a full expression parser) and most usually end up dialing it back over time. It becomes really annoying if you can't do *any* computation in a nested expression.

If someone wants to do:

```
root = Framed(app-frame-thickness + (is-android ? android-frame-adjust : ios-frame-adjust) + 4)
```

Do you really want them to have to write something like:

```
a = is-android ? android-frame-adjust : ios-frame-adjust
b = app-frame-thickness + a
c = b + 4
root = Framed(c)
```

If this DSL is _really_ only for authoring component trees, it's probably fine. But you do have literal values and even lists. Presumably it will be useful to add numbers, concatenate strings, or append to lists. Having to hoist all of that out to separate named declarations could get really annoying.

Though if this code will never be written or read by a human... 🤷

specification/proposals/express/a2ui_express.md
munificent: And newlines in strings.

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Core primitive types
+
+The syntax supports three literal primitive types:
+
+- Strings are enclosed in straight double quotes, for example `"Enter your name"`.
+- Numbers are written as plain integers or decimals, for example `42` or `3.14`.
```
munificent: Is `-123` a negative number literal, or a `123` literal with a unary `-` applied to it?

(The distinction can matter if you ever find yourself supporting `.` calls on numbers because then it affects what, say, `-1.abs` means.)

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+```
+breedList = List(_template($/breeds, breedTemplate), "horizontal")
+```
+
+The helper starts with an underscore `_` to clearly distinguish it from components in custom catalogs and prevent naming conflicts.
+
+### Data binding and reactive paths
+
+To connect component properties to the application data model, properties accept bound paths prefixed with the `$` symbol:
```
munificent, line 70: Does the language also support string interpolation using `$`? If so, it might be less confusing to use a different syntax here.

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+- Absolute paths start with a forward slash after the prefix, for example `$/user/email`. These paths resolve from the root of the shared data model.
+- Relative paths omit the slash, for example `$lastName`. These resolve within list iteration contexts.
```
munificent, line 73: Do you think most path references will be from the root or from within lists? If the former, it might be better to have a syntax that makes absolute paths shorter. The syntax here is biased towards relative paths.

Alternatively, you could have all paths be absolute from some named root and then use some built-in identifiers for referring to the actual data model root object, the surrounding list context, etc.

Can list contexts ever nest? If so, how do you specify which one a relative path is relative to?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Data model population
+
+To populate or initialize values within the shared data model directly from the generated output, A2UI Express supports data model assignments. A statement with a left-hand side that represents an absolute data path will populate that path in the surface's `dataModel`:
+
+```
+$/path/to/key = value_expression
+```
```
munificent, line 81: What happens if you try to add a property to something that is already a primitive?

```
$/foo = "a string"
$/foo/bar = "stuff a property onto a string?"
```

What does that do?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+### Data binding and reactive paths
+
+To connect component properties to the application data model, properties accept bound paths prefixed with the `$` symbol:
+
+- Absolute paths start with a forward slash after the prefix, for example `$/user/email`. These paths resolve from the root of the shared data model.
+- Relative paths omit the slash, for example `$lastName`. These resolve within list iteration contexts.
+
+### Data model population
+
+To populate or initialize values within the shared data model directly from the generated output, A2UI Express supports data model assignments. A statement with a left-hand side that represents an absolute data path will populate that path in the surface's `dataModel`:
```
munificent, line 77: Are these declarations or assignments? Can you assign to the same path multiple times?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+```
+$/path/to/key = value_expression
+```
+
+Where `value_expression` is any valid literal primitive, array, or map. For example:
```
munificent, line 83: This is the first time you've mentioned maps. Are they only usable in data model population assignments, or can they be used in component definitions too?

Are there different expression grammars for the two contexts?

What expressions are allowed as the elements of lists, and the key/value pairs of maps? Are map keys always bare identifiers? If so, are those identifiers treated as strings (like JS), or references to named variables (like Dart)?

Can collections be arbitrarily nested? Are these two lines equivalent?

```
$/foo = {bar: {baz: 123}}
$/foo/bar/baz = 123
```

If so, how are merges handled?

```
$/foo/a = 1
$/foo = {b: 2}
```

Here, do we end up with `foo` having value `{a: 1, b: 2}` or just `{b: 2}`?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+  }
+}
+```
+
+### Nested function calls and actions
+
+To support catalog flexibility and avoid hardcoding specific formatting or action helpers, A2UI Express uses standard nested function call syntax.
+
+- Client functions are written as `<FunctionName>(<args>)`, matching the exact function names registered in the loaded catalog.
+- If the client catalog contains a text formatting helper (such as `formatString`), it is called explicitly: `welcomeText = Text(formatString("Welcome, ${/user/firstName}!"))`. This prevents failures if a client catalog uses a different naming convention for interpolation.
```
munificent, line 114: Is string interpolation part of the language, or handled by a text formatting helper? If the latter, how do you avoid that helper having to effectively reimplement the whole parser/compiler/resolver?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+- Local actions use this same signature to trigger behaviors, for example `openUrl("https://example.com")`. The compiler maps these to standard client function actions.
+- Server events use a reserved `Event` signature to declare backend actions, for example `Event("save_deal", {rep: $/form/rep})`.
```
munificent, line 116: So it seems like map literals can be used as expressions basically anywhere? If so, you probably want to add them to `### Core primitive types`.

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Validation and logic expressions
+
+Validation checks are defined using the `?` prefix. If a component expects validation rules, the compiler converts these expressions into standard client-side functions:
```
munificent: I don't have enough context to know what "validation" means here. But if the leading `?` is just syntactic sugar for calling a function with that name, does it carry its weight?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+- Simple checks are written with the function name, for example `?required`.
+- Parameterized checks accept arguments in parentheses, for example `?regex("^[0-9]{5}$", "Must be a valid zip code")`.
+- Multiple checks are grouped in lists: `[?required, ?email]`.
```
munificent, line 124: So the system implicitly understands that a list containing validation checks is itself a validation check? What about:

```
[?required, "oops, not a validation check"]
```

Would it make more sense to do:

```
?[required, email]
```

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+    A("Plain text stream") --> B("Lexer and line parser")
+    B --> C("Abstract syntax tree")
+    C --> D("Schema mapper")
+    D --> E("AST flattener")
+    E --> F("Standard A2UI v1.0 JSON")
+```
+
+### Line parsing and tokenization
+
+The compiler reads the input text line-by-line. It discards empty lines and parses assignments into tokens.
```
munificent, line 183: You do say an assignment can span multiple lines, so it's probably clearer to say that separate top-level assignments or standalone operations may be executed before later ones are parsed. Is that the intent here?

Are there comments?

specification/proposals/express/a2ui_express.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Error recovery and micro-refinement loops
+
+If the compiler encounters a syntax error or catalog schema mismatch during parsing, it triggers a structured error recovery workflow:
+
+1. Isolation. The compiler flags the invalid line, discards that sub-branch of the AST, and continues parsing the remaining lines to avoid collapsing the user interface.
```
munificent, line 189: What if the offending line is defining some name that is referred to elsewhere on other lines?

munificent (commented):
@yjbanov asked me to take a look with an eye to language design stuff. I have very little context on the overall problem being solved and I'm definitely not an expert at what kinds of code LLMs are good at writing and reading, so take all of this with a very large grain of salt.

Most of my comments are probably fine but I figured it's better to bring it up then potentially miss a problem.

Overall, this makes sense to me. Unlike a syntax where the entire thing is one monolithic tree which isn't meaningfully parsable until you have the whole thing, it gives you a way to break the code into smaller separately handle-able units.

That does raise questions around forward declarations and how names are resolved and managed. Name resolution in general is pretty vague here and is something you'll likely want to be pedantic about. It's a part of language design that has a lot of sharp edges.

Also, an explicit grammar in something like EBNF would be nice to see. I know it's not something that everyone loves, but the exercise of writing it will force you to answer a lot of things that might otherwise be left implicit and then become subtle parser bugs. (For example, it's not clear from the spec here if component definitions can have map literals as arguments or not. Can lists have trailing commas? Be empty? Be empty except for just a comma?)

I'm always excited to see people doing novel language design to approach a problem a different way! :D

eval/a2ui_eval/strategies/express.py
gspencergoog: Moved the imports of A2uiSchemaManager and CatalogConfig to the top of the file to clean up the function-level scope.

eval/a2ui_eval/strategies/express.py
gspencergoog: I will refactor the prompts in v1_0_prompts.yaml to be format-agnostic (removing JSON/createSurface terminology). This allows us to delete the regex rewriting hacks in express.py completely while keeping the dataset reusable for future non-Express v1.0 strategies.

eval/a2ui_eval/strategies/express.py
gspencergoog: Removed the debug print statement.

eval/a2ui_eval/strategies/express.py
gspencergoog: It wasn't because of the Lite model: I actually did most of the improvement loop using 3.5 Flash. Now that the dataset prompts are format-agnostic, I can remove the user prompt wrapping/reminder hack from express.py and rely entirely on the system instructions. I will verify that this does not negatively impact formatting accuracy on the lite model.

eval/a2ui_eval/strategies/express.py
gspencergoog: I will introduce a new parser helper parse_express_response in the SDK that has a matching signature to the standard parse_response. This extracts the DSL block inside <a2ui> tags, compiles it to JSON, and returns standard ResponsePart structures. I'll refactor the evaluation strategy to use this cleaner helper.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/__init__.py
gspencergoog: Good suggestion. Moving it to a2ui.experimental.express makes the status clear from the import statement. I'll restructure the directories and update imports.

specification/proposals/express/scripts/run_decompiler.py
gspencergoog: I will move the development helper scripts (run_*) to a scripts/ subfolder under specification/proposals/express/ so they don't clutter the main proposal directory, while keeping them available for developers and test verification.

eval/a2ui_eval/strategies/express.py
gspencergoog: Good idea. I'll leave that for another PR though.

eval/datasets/v1_0_prompts.yaml
gspencergoog: Mostly, it removes the format-specific terminology (e.g. "Make a surfaceUpdate message with...") and replaces it with agnostic wording ("Make a layout with...").  There were also some formatting things (moved to `foo: |-` for the multi-line YAML strings to get rid of the escaping, just for readability).  One or two of the example prompts were pretty vague, which made them hard to score, so I made them more concrete.

specification/proposals/express/a2ui_express.md
gspencergoog: Embedded newlines are allowed, they just have to be properly closed.
I added more rules about quoting and escaping too:

- **Strings** are represented in two formats:
  - **Standard Strings**: Enclosed in single double quotes (e.g., `"Enter your name"`) or triple double quotes (e.g., `"""Line 1\nLine 2"""`). Standard strings support common escape sequences: `\n` (newline), `\t` (tab), `\\` (backslash), and `\"` (double quote). Embedded newlines are allowed.
  - **Raw Strings**: Prefaced by `r` (e.g., `r"^[a-zA-Z]+$"` or `r"""Raw multi-line content"""`). In raw strings, no escape sequences are processed, and backslashes are interpreted as literal characters. This is particularly useful for validation regex patterns containing backslashes. Embedded newlines are allowed.

specification/proposals/express/a2ui_express.md
gspencergoog: It does.  I think "host application" is the agent here.  I'll reword it.

specification/proposals/express/examples/08_user-profile.a2ui
gspencergoog: Done.

gspencergoog:
> @yjbanov asked me to take a look with an eye to language design stuff. I have very little context on the overall problem being solved and I'm definitely not an expert at what kinds of code LLMs are good at writing and reading, so take all of this with a very large grain of salt.

Thank you for taking a look!  I really appreciate the feedback.

> Overall, this makes sense to me. Unlike a syntax where the entire thing is one monolithic tree which isn't meaningfully parsable until you have the whole thing, it gives you a way to break the code into smaller separately handle-able units.

Yes, we want to be able to stream the data, make corrections, and recover from missing pieces.
 
> That does raise questions around forward declarations and how names are resolved and managed. Name resolution in general is pretty vague here and is something you'll likely want to be pedantic about. It's a part of language design that has a lot of sharp edges.

Okay, point well taken.  That makes a lot of sense.  I'll see if I can lock that down.

> Also, an explicit grammar in something like EBNF would be nice to see. I know it's not something that everyone loves, but the exercise of writing it will force you to answer a lot of things that might otherwise be left implicit and then become subtle parser bugs. (For example, it's not clear from the spec here if component definitions can have map literals as arguments or not. Can lists have trailing commas? Be empty? Be empty except for just a comma?)

Also a great idea.  We won't be giving it to the LLM because that's too verbose, but we need it for the compiler and it would help formalize the language.

> I'm always excited to see people doing novel language design to approach a problem a different way! :D

Thanks!  This one is weird because

1) it's not meant to be authored, or even read, by a human
2) it's description needs to be easily, accurately, and completely describable in a minimum of English words in a prompt. and
3) it's throw-away code: we're not going to compile it to machine code, optimize it, etc. and
4) we're actively trying to stay away from making a Turing complete language so that we can avoid all the security and complexity implications of that.

specification/proposals/express/a2ui_express.md
gspencergoog: The other place will just ignore the undefined value. It might not render because of that, but it would just wait until there's a valid value there, which could come in a an error correction update.

specification/proposals/express/a2ui_express.md
gspencergoog: There were two reasons I stayed away from this kind of nesting:

1) if we want to "stream" user interfaces, nesting like this is hard, since we have to wait until the end, or do some auto-closing magic in order to do intermediate states.  If I can use adjacency lists to separate the components from the list, then it makes streaming much nicer: we ignore symbols that we don't recognize yet, and fill them in when they are defined.  This lets the LLM be sloppy about ordering the definitions, and

2) the LLM then doesn't need to keep track of all the nesting paren levels, which it is not terribly good at.

specification/proposals/express/a2ui_express.md
gspencergoog: Yes, this is on purpose to let the LLM control streaming behavior. The A2UI renderers already handle this by ignoring symbols that they don't recognize until they are defined, and also caching symbol definitions that aren't yet connected to anything until they get used.  This lets us stream in a Column with identifiers for the children, and then fill in the children as the come in, or vice versa and have them all pop in at once when the Column definition arrives.

specification/proposals/express/a2ui_express.md
gspencergoog: It is an error to have circular references.

The renderers will throw an error back to the agent if they catch circular references. This is a somewhat bad design in that in order to catch these (algorithmically) on the server before they hit the client, the server has to keep track of everything the client has seen.  On the other hand, the renderer is the final say as to the actual state of things, so maybe it's appropriate there.

specification/proposals/express/a2ui_express.md
gspencergoog: Well, exactly, it will not be written by a human. Which is weird.

The prohibition there is actually there more to keep an LLM from writing an entire tree in one expression, so that streaming works better. It forces it to split it up into a bunch of lines that can be evaluated as they come in.

specification/proposals/express/a2ui_express.md
gspencergoog: Very good question. It is a negative number literal.  I'm not planning to provide for arithmetic expressions, but if we do, I can see that could change things too.  Would it be better to parse the unary "-"?

specification/proposals/express/a2ui_express.md
gspencergoog: No.  String interpolation is handled by a separate client side function call (`formatString`) that is implemented in the catalog, so it's not part of the specification.  This is mainly to make it so that if you want to implement a stripped down renderer (say, on a watch), you don't need to implement string interpolation to render A2UI.

The `formatString` function in the basic catalog (the implementation of which is reusable from our core library) is pretty complex though: because it plugs into a reactive UI, it produces a stream, and any changes in the interpolated values (which can be other streams from other functions or data model values) will produce new stream updates.

specification/proposals/express/a2ui_express.md
gspencergoog: It depends a lot on the use case.  For the templates (called "list iteration contexts" here, e.g. data-driven lists of cards, where the cards are templates), the paths are all relative to the location of the data, and that's fairly common.

But if you're driving a lot of disparate values from a larger data model, then a lot of them could be absolute.

The "nesting" is only possible through the templating mechanism at the moment, so outside of that, everything is absolute.

specification/proposals/express/a2ui_express.md
gspencergoog: It will throw an error because it's a type mismatch.  If `$/foo` were an object, though, it would add or overwrite a "bar" member.

specification/proposals/express/a2ui_express.md
gspencergoog: Assignments, and yes, you can assign multiple times an overwrite. The data model is basically a javascript/JSON model.

specification/proposals/express/a2ui_express.md
gspencergoog: They're usable in data model assignments, but they can also (if the component takes a map as an argument) be used as literals in the argument.

e.g. if your Form component takes a map to construct the form, it could take one:

```
my_form = Form({"Name": "Greg", "GitHub": "gspencergoog"})
```

> Can collections be arbitrarily nested? Are these two lines equivalent?

Yes, those are equivalent.

Merges are additive, so you end up with `{a: 1, b:2}`.

specification/proposals/express/a2ui_express.md
gspencergoog: String interpolation isn't part of the language. (see my other answer about the `stringFormat` function).

specification/proposals/express/a2ui_express.md
gspencergoog: Yeah, good point, I'll do that.

specification/proposals/express/a2ui_express.md
gspencergoog: Validation here is in the context of "form validation", in the sense of wanting to check that a text field contains an email address, for instance.  It is handled by defining a client side function to do the validation and return a boolean.  Any function that returns a boolean can be used as a validation function.

The `?` is just syntactic sugar for calling a boolean-returning function that takes an implicit first "value" first argument.

To be honest, I haven't really thought this part through that well. I think probably it could just be a regular function call syntax, but that does have slightly higher (LLM) token size (not much though).

Maybe it should instead be something like 

```
username = TextInput($/form/username, [required(_), regex(_, "^[0-9]{5}$")])
```
 and we can sub in the value for the `_`. Right now that would be: 

```
username = TextInput($/form/username, [?required, ?regex("^[0-9]{5}$")])
```

specification/proposals/express/a2ui_express.md
gspencergoog: > So the system implicitly understands that a list containing validation checks is itself a validation check? What about:
> 
> ```
> [?required, "oops, not a validation check"]
> ```

This would fail because the string isn't a boolean.

> Would it make more sense to do:
> 
> ```
> ?[required, email]
> ```

No, the values don't have to be functions, they could also be from the data model.

specification/proposals/express/a2ui_express.md
gspencergoog: Yes, that's the intent.

There aren't explicitly comments, but I do actually ignore both `#` and `//` end of line comments in the parser because the LLM sometimes adds them anyhow.  We don't want to mention or "allow" them because they just take up tokens we're not going to use.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/decompiler.py
jacobsimionato: I think this is also important to handle conversation history, assuming it is stored in the more standardized, stable A2UI format. If the agent is communicating back and forth with A2UI, we want it to always read and write A2UI in the same format, to avoid confusion (and maximise efficiency!). So I think we need the decompiler for this.

Steps for an inference in a multi-turn conversation:
1. Convert all A2UI messages in existing conversation history to Express format (required decompiler)
2. Convert Catalog to express format
3. Prompt agent with system prompt in express format, and conversation history also using express format.
4. Parse express output back to A2UI (requires compiler)
5. Persist conversation history in A2UI format.

specification/proposals/express/examples/32_advanced-form-validator.a2ui
jacobsimionato: I agree with just using our existing formatString function call approach for the first cut of express, to keep it catalog-agnostic.

But in the future, I think it'd be interesting to pursue these optimizations which make the format catalog-specific in service of greater performance gains.

We might be able to generalize this - JSON render has a concept of "directives" which is something like this - https://json-render.dev/docs/directives.

So perhaps you create a generic ExpressInferenceFormat and then install a directive that transforms `${/number}` to `formatString(/number)` etc

chenman9226:
One thing I’m curious about: since Express drops many of the JSON/property keys, have you observed any impact on generation quality from losing those semantic cues? For example, keys like `label`, `value`, `children`, `action`, or `validation` seem like they might help the model understand which argument is which, especially for less common components or longer positional signatures.

I can see the token-efficiency benefit, so I’m mostly wondering whether this showed up in practice, or whether the shorter format generally outweighed the loss of those semantic anchors.

gspencergoog:
> One thing I’m curious about: since Express drops many of the JSON/property keys, have you observed any impact on generation quality from losing those semantic cues? For example, keys like `label`, `value`, `children`, `action`, or `validation` seem like they might help the model understand which argument is which, especially for less common components or longer positional signatures.
> 
> I can see the token-efficiency benefit, so I’m mostly wondering whether this showed up in practice, or whether the shorter format generally outweighed the loss of those semantic anchors.

The thing I thought might be a problem, but doesn't appear to be, is that LLMs aren't great at counting, so I thought that positional parameters would be the issue (getting them in the wrong order, or inserting something between them).  

The lack of property keys also doesn't seem to affect quality as long as the descriptions from the JSON schema in the catalog are included. For example, if the catalog item for TextField is converted to this:

```md
• TextField(label, value?, placeholder?, variant? (static only), weight? (static only), checks? (static only))
  - label: The text label for the input field.
  - value: The value of the text field.
  - placeholder: The placeholder text for the input field.
  - variant: The type of input field to display. Must be one of: 'longText', 'number', 'shortText', 'obscured'
  - weight: The relative weight of this component within a Row or Column. This is similar to the CSS 'flex-grow' property. Note: this may ONLY be set when the component is a direct descendant of a Row or Column.
```

Then the LLM has enough context to be able to decide what each argument means when it writes the output.  We may have to play with how the catalogs are converted to prompts to minimize the tokens in the prompt, but we will need to include the entire descriptions that are supplied by the catalog developers because they can include important instructions for how to use the components.

In fact, for a while I wasn't including any of the descriptions at all (which saved a lot of input tokens!), and as long as the parameter was something well named and intuitive, the LLM seemed to be able to extrapolate.  If anything was vague or unconventional then it started to break down, however.

gspencergoog:
FYI, I created a stack of three PRs that I split this PR into:

- https://github.com/a2ui-project/a2ui/pull/1718
  - The core library and validation layers for the A2UI v1.0 and Express compiler outputs. It introduces support for v1.0 schemas and enhances component reference resolution for nested slot structures.
- https://github.com/gspencergoog/A2UI/pull/3
  - The A2UI Express technical specification, introducing the A2UI Express format. It includes the compiler, decompiler, schema helper, and parser modules, as well as a EBNF format description of the format.
- https://github.com/gspencergoog/A2UI/pull/4
  - Integrates the A2UI Express compiler and prompt strategies into the Inspect-ai evaluation suite. It updates the scorers, solvers, and task configurations to support v1.0 and Express DSL targets.

The last two are diffs on the first one and each other to make a stacked set of PRs.  The last two are PRs on my fork, not on the main repo, until we roll the first one into the repo, and then I'll change their target to the main repo and land them too.

I converted this PR back to a draft and will close it once the other PRs land.

gspencergoog:
Closing this because all of the split-out components of this PR have now landed.

specification/proposals/express/a2ui_express.md
munificent: If you intend to support unary negation expressions like `-foo`, then I would parse `-123` also as a unary negate followed by `123`. If you don't otherwise have unary negate expressions, then I'd have the lexer lex `-123` as a single token. Note that the choice here affects whether `-   123` is valid or not.

gspencergoog: I think I might go with a unary expression not because we actually want to support them (at least right now), but because it would be more forgiving if the LLM happened to write something as `- 123`.

PR #1714, opened 2026-06-22
feat(web-core): allow signal implementations to be swapped out

crisbeto:
Currently `web_core` uses Preact signals and it's up to the individual renderers to translate them to their own flavor of signals. At least for Angular, this is problematic for the following reasons:
1. It uses `effect` to synchronize a Preact signal and an Angular signal. Signals are meant to be synchronous, whereas effects are asynchronous and run on a scheduler. Using `effect` in this case can cause the Angular signal to miss some incoming values.
2. Angular signals can be used anywhere, whereas `toAngularSignal` needs to be run in an injection context in order to get a handle on `DestroyRef` and `NgZone`. This limits the places from which A2UI APIs can be invoked.
3. It relies on `NgZone` and we generally want to steer users towards zoneless.

### DIFF
--- a/renderers/angular/src/v0_9/core/utils.ts
+++ b/renderers/angular/src/v0_9/core/utils.ts
@@ -15,46 +15,52 @@
  */
 
 import {DestroyRef, Signal, signal as angularSignal} from '@angular/core';
-import {Signal as PreactSignal, effect, signal as preactSignal} from '@a2ui/web_core/v0_9';
+import {
+  Signal as CoreSignal,
+  effect,
+  signal as preactSignal,
+  getValue,
+  peekValue,
+} from '@a2ui/web_core/v0_9';
+import {NgZone} from '@angular/core';
+
 export {preactSignal};
 
 /**
- * Bridges a Preact Signal (from A2UI web_core) to a reactive Angular Signal.
+ * Bridges a signal from A2UI web_core to an Angular Signal.
  *
  * This utility handles the lifecycle mapping between Preact and Angular,
  * ensuring that updates from the A2UI data model are propagated correctly
  * to Angular's change detection, and resources are cleaned up when the
  * component is destroyed.
  *
- * @param preactSignal The source Preact Signal.
+ * @param coreSignal The source Preact Signal.
  * @param destroyRef Angular DestroyRef for lifecycle management.
  * @param ngZone Optional NgZone to ensure updates run within the Angular zone
  *               (necessary for correct change detection in OnPush components).
  * @returns A read-only Angular Signal.
  */
-import {NgZone} from '@angular/core';
-
 export function toAngularSignal<T>(
-  preactSignal: PreactSignal<T>,
+  coreSignal: CoreSignal<T>,
   destroyRef: DestroyRef,
   ngZone?: NgZone,
 ): Signal<T> {
-  const s = angularSignal(preactSignal.peek());
+  const s = angularSignal(peekValue(coreSignal));
 
   const dispose = effect(() => {
+    const value = getValue(coreSignal);
+
     if (ngZone) {
-      ngZone.run(() => s.set(preactSignal.value));
+      ngZone.run(() => s.set(value));
     } else {
-      s.set(preactSignal.value);
+      s.set(value);
     }
   });
 
   destroyRef.onDestroy(() => {
     dispose();
     // Some signals returned by DataContext.resolveSignal have a custom unsubscribe for AbortControllers
-    if ((preactSignal as any).unsubscribe) {
-      (preactSignal as any).unsubscribe();
-    }
+    coreSignal.unsubscribe?.();
   });
 
   return s.asReadonly();
--- a/renderers/web_core/src/v0_9/index.ts
+++ b/renderers/web_core/src/v0_9/index.ts
@@ -37,7 +37,19 @@ export * from './state/surface-model.js';
 export * from './errors.js';
 export * from './basic_catalog/index.js';
 
-export {effect, Signal, signal, computed} from '@preact/signals-core';
+export {
+  type Signal,
+  effect,
+  signal,
+  computed,
+  getValue,
+  peekValue,
+  batchWrite,
+  isSignal,
+  setValue,
+  setSignalImplementation,
+  type SignalImplementations,
+} from './reactivity/signals.js';
 
 import A2uiMessageSchemaRaw from './schemas/server_to_client.json' with {type: 'json'};
 
--- a/renderers/web_core/src/v0_9/reactivity/signals.test.ts
+++ b/renderers/web_core/src/v0_9/reactivity/signals.test.ts
@@ -15,131 +15,94 @@
  */
 
 import assert from 'node:assert';
-import {describe, it} from 'node:test';
-import {Signal as PSignal, computed as pComputed, effect as pEffect} from '@preact/signals-core';
+import {after, describe, it} from 'node:test';
 import {
   signal as aSignal,
   computed as aComputed,
   Signal as ASignal,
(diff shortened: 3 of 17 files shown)


renderers/web_core/src/v0_9/reactivity/signals.ts
```diff
@@ -14,60 +14,90 @@
...
-   */
-  unwrap<T>(val: SignalKinds<T>[K]): T;
-
-  /**
-   * Sets the value of the provided framework signal.
-   */
-  set<T>(signal: WritableSignalKinds<T>[K], value: T): void;
+export interface Signal<T = any> {
+  __signalBrand?: T; // Marker that prevents any value from being assigned as a signal.
+}
```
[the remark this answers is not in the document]
crisbeto: This is intentional so it's easier to cast to this value.

renderers/web_core/src/v0_9/rendering/data-context.ts
```diff
@@ -190,7 +198,7 @@ export class DataContext {
       if (Object.keys(argSignals).length === 0) {
         const abortController = new AbortController();
         const result = this.evaluateFunctionReactive<V>(call.call, {}, abortController.signal);
-        const sig = result instanceof Signal ? result : signal(result);
+        const sig = isSignal(result) ? result : signal(result as V);
```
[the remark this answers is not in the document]
crisbeto: This is matching the previous behavior.

renderers/web_core/src/v0_9/reactivity/signals.ts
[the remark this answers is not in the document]
josemontespg: Can you share an example of how __signalBrand is intended to be used? If it is not used at runtime, please clarify this in the comment

renderers/web_core/src/v0_9/reactivity/signals.ts
```diff
@@ -14,60 +14,98 @@
...
+  effect as preactEffect,
+  batch as preactBatch,
+  Signal as PreactSignal,
+  Computed as PreactComputed,
+} from '@preact/signals-core';
 
-// This <T>, while unused, is required to pass through to a given Signal impl.
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-export interface WritableSignalKinds<T> {}
+export interface Signal<T = any> {
```
josemontespg: can this use `T = unknown` ? For better type safety.

renderers/web_core/src/v0_9/reactivity/signals.ts
```diff
@@ -14,60 +14,98 @@
...
-   * Check if an arbitrary object is a framework signal.
-   */
-  isSignal(val: unknown): val is SignalKinds<any>[K];
+setSignalImplementation(PREACT_SIGNAL_IMPLEMENTATION);
 
-  /**
-   * Wrap the value in a signal.
-   */
-  wrap<T>(val: T): WritableSignalKinds<T>[K];
+export function setSignalImplementation(impl: SignalImplementations): void {
```
josemontespg, line 76: Please add a jsdoc comment here saying that this is used to override the default Preact implementation of Signals, and should be called early during app load.

renderers/angular/src/v0_9/core/utils.ts
```diff
@@ -15,45 +15,53 @@
...
+      s.set(value);
     }
   });
 
   destroyRef.onDestroy(() => {
     dispose();
     // Some signals returned by DataContext.resolveSignal have a custom unsubscribe for AbortControllers
-    if ((preactSignal as any).unsubscribe) {
-      (preactSignal as any).unsubscribe();
+    if ((coreSignal as any).unsubscribe) {
```
josemontespg: Can we add `unsubscribe` to the `CoreSignal` interface? to avoid having to cast here

renderers/web_core/src/v0_9/reactivity/signals.test.ts
```diff
@@ -15,131 +15,94 @@
...
+            cleanupRegisterFn(cleanup);
+          }
+        });
+        return () => e.destroy();
+      },
+      getValue: <T>(s: Signal<T>) => (s as ASignal<T>)(),
+      setValue: <T>(s: Signal<T>, v: T) => (s as AWritableSignal<T>).set(v),
+      peekValue: <T>(s: Signal<T>) => auntracked(() => (s as ASignal<T>)()),
+      batchWrite: (batchFn: () => void) => {
+        batchFn();
```
josemontespg, line 91: This implementation looks useful beyond the test. I think we can refactor the angular renderer to use this (in a follow up PR)

renderers/web_core/src/v0_9/index.ts
```diff
@@ -37,7 +37,17 @@ export * from './state/surface-model.js';
...
+  type Signal,
+  effect,
+  signal,
+  computed,
+  getValue,
+  peekValue,
+  batchWrite,
+  isSignal,
+  setValue,
+} from './reactivity/signals.js';
```
josemontespg, line 52: Should we export * here? `setSignalImplementation`, `SignalImplementations` are key missing exports

josemontespg (changes requested):
Amazing refactor to add an abstraction layer to Signals, allowing renderers to override the Signals used in web_core, for better compatibility with Angular and potentially more. Thanks!

renderers/angular/src/v0_9/core/utils.ts
crisbeto: Ah good idea. Added.

renderers/web_core/src/v0_9/reactivity/signals.test.ts
crisbeto: Yep, I think the Angular version will end up being thing.

renderers/web_core/src/v0_9/reactivity/signals.ts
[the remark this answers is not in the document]
crisbeto: I expanded the comment.

renderers/web_core/src/v0_9/index.ts
crisbeto: I didn't want to expose things like the default Preact implementation. Also AFAIK the barrel exports can affect tree shaking.

crisbeto (commented):
I've addressed the feedback.

renderers/web_core/src/v0_9/index.ts
josemontespg: Sounds good, thanks for adding those!

renderers/web_core/src/v0_9/reactivity/signals.ts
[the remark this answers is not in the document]
josemontespg: Sorry, I still don't understand how this can prevent `let s: Signal = {} as Signal`. Can you explain how this __signalBrand works?

josemontespg (approved):
Please add a note under "Unreleased" to the  /renderers/web_core/CHANGELOG.md, with the addition of `setSignalImplementation`

renderers/web_core/src/v0_9/reactivity/signals.ts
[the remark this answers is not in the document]
crisbeto: It ensures that there's at least one property so TS can distinguish it from any other object. Here's an example: https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgMrAOYjgGwHID2AQlHCACbIDeAvgFCiSyIrpa4DqwYAFiWZSp1kI5AH0xAI1IUAXMjIBPANx16dHBDAL5bbPmIzyqzdsm7M+rr34VVdOMgC81GqsnPXyoA

PR #1722, opened 2026-06-22
Release: bump package versions for angular and web_core

josemontespg:
I ran these commands:

`renderers/scripts/increment_version.mjs web_core`
`renderers/scripts/increment_version.mjs angular`

Also, updated the Changelogs to reflect the new versions.

### DIFF
--- a/renderers/web_core/package.json
+++ b/renderers/web_core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/web_core",
-  "version": "0.10.7",
+  "version": "0.10.3",
   "description": "A2UI Core Library",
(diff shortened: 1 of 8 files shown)


renderers/web_core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/web_core",
-  "version": "0.10.2",
+  "version": "0.10.7",
```
josemontespg: I'm not sure why this resolved the next version to .7

ditman (approved):
Thanks for releasing this! It seems that since our integration testing passes without any changes these are both minor releases. Let's go!

PR #1726, opened 2026-06-23
feat(express): Implement A2UI Express compiler, decompiler and parser

gspencergoog:
## Summary
This PR implements the A2UI Express technical specification, introducing A2UI Express — a highly compressed, model-optimized declarative syntax (DSL) for generative user interfaces. It provides a complete end-to-end Python implementation, including an ANTLR-based compiler, a decompiler, a schema-based system prompt generator, helper scripts, and comprehensive test suites.

This is a refined, standalone extraction of the A2UI Express compiler/decompiler portions originally proposed in PR #1678, incorporating automated parser generation, strict validation, and extensive test coverage.

## Changes
* **Build System & Code Generation**:
  * Added Hatch build hook in [pack_specs_hook.py](file:///Users/gspencer/code/a2ui/main/agent_sdks/python/a2ui_agent/pack_specs_hook.py) to automatically compile the ANTLR grammar [Express.g4](file:///Users/gspencer/code/a2ui/main/agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/Express.g4) into Python3 source files at build-time.
  * The build hook handles target case-insensitive file renaming to clean snake_case (`express_lexer.py`, `express_parser.py`, `express_visitor.py`), relative import post-processing, and automatic formatting of generated code with `pyink`.
  * Updated [pyproject.toml](file:///Users/gspencer/code/a2ui/main/agent_sdks/python/a2ui_agent/pyproject.toml) to include `antlr4-python3-runtime` as a runtime dependency, and `antlr4-tools` in the build system requirements.
* **Compiler & Parser** (`a2ui.experimental.express`):
  * Implemented an ANTLR-based parsing pipeline using `Express.g4` to parse line-oriented declarative layout files.
  * The `ExpressCompiler` compiles the AST directly into standard A2UI v1.0 JSON payloads (with dynamic positional parameter resolution and variable flattening).
  * Supports rich string types: standard strings, raw strings (`r"..."`), raw multiline strings (`r"""..."""`), and escaped carriage returns.
  * Integrates a partial parser mode supporting streaming recovery for incomplete layouts.
  * Incorporates strict enum validation for component properties, raising `ValueError` on mismatch rather than silently ignoring invalid values.
* **Decompiler**:
  * Implemented `ExpressDecompiler` to convert standard A2UI v1.0 JSON payloads back into the highly compact, line-oriented Express DSL.
* **Schema Helper & Prompt Generator**:
  * Implemented `CatalogSchemaHelper` to parse catalog schema definitions.
  * Implemented `ExpressPromptGenerator` to compile active catalog schemas into positional signatures used to prompt generative models.
* **Evaluation & Testing Scripts**:
  * Added `run_inference.py` to evaluate the A2UI Express prompt contract by converting JSON examples to Express DSL via Gemini/Ollama/MLX models and validating the round-trip compilation.
  * Added `recreate_dsl_examples.py` to programmatically regenerate the dynamic markdown documentation.
* **Documentation & Examples**:
  * Added comprehensive layout examples under `specification/proposals/express/examples/*.a2ui` (36 files) along with their corresponding compiled JSON targets.
  * Created [README.md](file:///Users/gspencer/code/a2ui/main/specification/proposals/express/README.md) and [a2ui_express.md](file:///Users/gspencer/code/a2ui/main/specification/proposals/express/a2ui_express.md) detailing the DSL grammar, compiler mechanics, and usage.
  * Created [express_dsl_examples.md](file:///Users/gspencer/code/a2ui/main/specification/proposals/express/express_dsl_examples.md) detailing the active system prompt contract and compiled weather forecast examples.

## Impact & Risks
* The feature is fully experimental, contained in the `a2ui.experimental.express` namespace, and gated behind the `A2UI_EXPRESS_ENABLED=true` environment variable.
* There is no impact on stable production paths or other existing SDK modules.
* Build-time code generation introduces a dependency on `antlr4` (via `antlr4-tools` and `antlr4-python3-runtime`) during development/builds, which is automatically resolved by Hatch and standard pip/uv environments.

## Testing
* Added 44 robust unit tests under `agent_sdks/python/a2ui_agent/tests/express/` including:
  * [test_compiler.py](file:///Users/gspencer/code/a2ui/main/agent_sdks/python/a2ui_agent/tests/express/test_compiler.py): Verifies parser correctness, token parsing, raw string handling, and carriage return unescaping.
  * [test_decompiler.py](file:///Users/gspencer/code/a2ui/main/agent_sdks/python/a2ui_agent/tests/express/test_decompiler.py): Validates round-trip integrity (JSON -> Express -> JSON).
  * [test_integration.py](file:///Users/gspencer/code/a2ui/main/agent_sdks/python/a2ui_agent/tests/express/test_integration.py): Tests the compiler against all 36 catalog layout examples.
  * [test_cli_tools.py](file:///Users/gspencer/code/a2ui/main/agent_sdks/python/a2ui_agent/tests/express/test_cli_tools.py): Tests script interfaces and prompt generation.
* The tests can be executed via the standard Dart/Python test runners (e.g. `uv run pytest`).


agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,943 @@
...
+      k_tok = self.consume()
+      self.consume("COLON")
+      v = self.parse_expression()
+      res[k_tok[1]] = v
+      while self.peek() and self.peek()[0] == "COMMA":
+        self.consume("COMMA")
+        if self.peek() and self.peek()[0] == "RBRACE":
+          break
+        next_tok = self.peek()
+        if next_tok[0] not in ("IDENTIFIER", "STRING"):
```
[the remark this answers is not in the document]
gspencergoog: Updated _parse_map to check if the next token is None (representing an abrupt end of input after a comma) and raise a descriptive SyntaxError.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/decompiler.py
```diff
@@ -0,0 +1,362 @@
...
+def _decompile_string(val: str) -> str:
+  """Formats a string literal using the cleanest/most readable representation."""
+  has_newline = "\n" in val or "\r" in val
+  has_tab = "\t" in val
+  has_quote = '"' in val
+  has_backslash = "\\" in val
+
+  # 1. Use triple-quotes for multi-line or strings containing double quotes
+  if has_quote or has_newline:
+    if '"""' not in val:
```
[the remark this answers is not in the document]
gspencergoog: Updated `_decompile_string` to avoid triple quotes when the string ends with a double quote. This forces the decompiler to fall back to double-quoted escaping and prevents syntax errors from trailing quote collisions.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/decompiler.py
```diff
@@ -0,0 +1,362 @@
...
+      if is_ref and val in comp_ids:
+        return val
+      # Otherwise quote as string literal
+      return _decompile_string(val)
+
+    if isinstance(val, bool):
+      return "true" if val else "false"
+
+    if val is None:
+      return "_"
```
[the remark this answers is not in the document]
gspencergoog: Updated `_decompile_value` to return `null` instead of `_` when the value is `None`. This preserves explicit null values during round-trips.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/schema_helper.py
```diff
@@ -0,0 +1,235 @@
...
+          continue
+        if "$ref" in sub:
+          ref = sub["$ref"]
+          if "Checkable" in ref:
+            is_checkable = True
+        if "properties" in sub:
+          props.update(sub["properties"])
+          for pk, pv in sub["properties"].items():
+            if isinstance(pv, dict) and "enum" in pv:
+              self.component_property_enums[(name, pk)] = pv["enum"]
```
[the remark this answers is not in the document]
gspencergoog: Added the recursive `_find_enum` helper to extract enums from nested `oneOf`, `anyOf`, or `allOf` schemas. This ensures nested enums are correctly validated.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,945 @@
...
+      return "\n"
+    if char == "t":
+      return "\t"
+    if char == "\\":
+      return "\\"
+    if char == '"':
+      return '"'
+    return seq
+
+  return re.sub(r"\\(.)", repl, val)
```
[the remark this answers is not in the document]
gspencergoog: I'm not sure this is correct.  It is searching for a backslash followed by a character, but it isn't intended to find a backslash followed by an *actual* newline (it's intended to find a backslash followed by an "n" or "t" character, for example).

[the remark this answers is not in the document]
gspencergoog: Declined. As you noted, the function is only intended to resolve standard escape sequences (\n, \t, \\, \") and treat others as literal. Since . excludes newlines, \\(.) correctly leaves \<newline> sequences untouched. Changing the regex to match newlines would be redundant since they would still be returned unchanged by the helper.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,945 @@
...
+              isinstance(mapped_val, dict)
+              and "path" in mapped_val
+              and "componentId" not in mapped_val
+          ):
+            raise ValueError(
+                f"Property '{prop_name}' of component '{comp_name}' does not support"
+                " dynamic data bindings (paths). You must provide a static value/array"
+                " instead."
+            )
+          if isinstance(mapped_val, list) and schema_expects_option_objects(
```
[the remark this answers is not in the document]
gspencergoog: Added a recursive `has_databinding` helper to ensure that any nested structures (like lists or dicts) passed to a property that disallows data bindings are thoroughly validated, preventing nested data bindings from bypassing the check.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/decompiler.py
```diff
@@ -0,0 +1,362 @@
...
+  if has_backslash and not has_newline and not has_tab and not has_quote:
+    return f'r"{val}"'
+
+  # 3. Fall back to standard double-quoted string with escapes
+  escaped = (
+      val.replace("\\", "\\\\")
+      .replace('"', '\\"')
+      .replace("\n", "\\n")
+      .replace("\t", "\\t")
+  )
```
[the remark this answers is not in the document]
gspencergoog: Updated `_decompile_string` to escape carriage return characters (\r to \\r) when falling back to standard double-quoted strings. This prevents literal carriage returns from breaking tokenization.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,953 @@
...
+      return "\n"
+    if char == "t":
+      return "\t"
+    if char == "\\":
+      return "\\"
+    if char == '"':
+      return '"'
+    return seq
+
+  return re.sub(r"\\(.)", repl, val)
```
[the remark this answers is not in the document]
gspencergoog: Updated `_unescape_string` in `compiler.py` to support resolving escaped carriage returns (\r to \r) and updated the escape sequence regex pattern from `\\(.)` to `\\([\s\S])` to ensure all escape sequences (including newlines) are matched. This allows carriage returns and other complex characters to cleanly round-trip compile and decompile without losing formatting. Also rewrote the decompiler carriage return escaping test to be an elegant end-to-end round-trip test.

specification/v1_0/catalogs/basic/examples/01_flight-status.json
```diff
@@ -18,34 +18,34 @@
           {
             "id": "root",
             "component": "Card",
-            "child": "main-column"
+            "child": "main_column"
```
jiahaog, line 21: Not sure I follow, are these changes related to the PR?

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,955 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Compilation engine for A2UI Express.
+
+Tokenizes, lexes, and parses A2UI Express plain-text statements into a clean
+AST, compiling it directly into standard A2UI v1.0 JSON messages.
+
+================================================================================
+A2UI EXPRESS EBNF GRAMMAR
```
jiahaog: Suggest that we formalize this in a separate file and generate a parser from this grammar with a library instead of implementing it by "hand" in parser.py? That will save us potentially thousands of LOC to maintain and test.

Also since we might need to port the parser across multiple languages too. IIUC tree sitter or something can generate parsers for many different target languages.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,955 @@
...
+  keys = clean_path.split("/")
+  current = d
+  for key in keys[:-1]:
+    if key not in current or not isinstance(current[key], dict):
+      current[key] = {}
+    current = current[key]
+  current[keys[-1]] = val
+
+
+def schema_allows_databinding(schema: Any) -> bool:
```
jacobsimionato: Does this and most of the other functions and constants below need to be public? Shall we mark them all as private unless they need to be exposed, to limit our public API surface? I believe nothing in Python is completely private, but people respect the underscore, right?

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/schema_helper.py
```diff
@@ -0,0 +1,249 @@
...
+
+Provides dynamic schema crawling to identify component properties, logical function
+signatures, and requirements directly from standard catalog JSON schemas.
+"""
+
+import json
+from typing import Optional
+
+
+class CatalogSchemaHelper:
```
jacobsimionato, line 25: This is cool - I can see this ending up in the core library eventually.

agent_sdks/python/a2ui_agent/tests/express/test_compiler.py
jacobsimionato, line 1: I think this and the compiler test would be easier to read if they were expressed as folders of example inputs/outputs or examples of invalid inputs and corresponding errors. I think we'll want this soon anyway, so we can have language-agnostic conformance tests. I know this makes the test extremely strict, e.g. it will end up testing whitespace etc, but I think that could actually be a good thing to force all the compiler/decompiler implementations to just work exactly the same.

Happy to skip this now if you want to submit and iterate though! There are clearly tests here, so it shouldn't be hard for an agent to convert them to conformance tests, or at least data-driven tests.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,955 @@
...
+
+  Returns:
+      A list of token tuples matching (TokenKind, TokenValue).
+  """
+  tok_regex = "|".join(f"(?P<{name}>{pattern})" for name, pattern in TOKEN_SPEC)
+  tokens = []
+  last_end = 0
+  for mo in re.finditer(tok_regex, text):
+    if mo.start() != last_end:
+      raise SyntaxError(f"Unexpected character: {text[last_end:mo.start()]!r}")
```
jacobsimionato: Are the exact errors that the compiler should throw formalized in the express spec? I think that'd be worth doing, and having conformance tests for them, to make sure that if we build and test error correction infra and prompts, they are portable across platforms.

specification/v1_0/catalogs/basic/catalog.json
```diff
@@ -4,7 +4,7 @@
...
...,\n          \"checks\": [\n            {\"call\": \"required\"}\n          ]\n        }\n      ],\n      \"dataModel\": {\n        \"form\": {\n          \"rep\": \"John Doe\",\n          \"value\": 1500.00\n        }\n      }\n    }\n  }\n]\n```\n\nExample 2: Dynamic list with templates\n```json\n[\n  {\n    \"version\": \"v1.0\",\n    \"createSurface\": {\n      \"surfaceId\": \"main\",\n      \"components\": [\n        {\n          \"id\": \"root\",\n          \"component\": \"Card\",\n          \"child\": \"breedList\"\n        },\n        {\n          \"id\": \"breedList\",\n          \"component\": \"List\",\n          \"children\": {\n            \"path\": \"/breeds\",\n            \"componentId\": \"breedTemplate\"\n          },\n          \"direction\": \"horizontal\"\n        },\n        {\n          \"id\": \"breedTemplate\",\n          \"component\": \"Image\",\n          \"url\": {\"path\": \"url\"}\n        }\n      ],\n      \"dataModel\": {\n        \"breeds\": [\n          {\n            \"url\": \"https://example.com/poodle.jpg\"\n          },\n          {\n            \"url\": \"https://example.com/lab.jpg\"\n          }\n        ]\n      }\n    }\n  }\n]\n```",
```
jacobsimionato, line 7: I wonder if the example here is actually sort of harmful, because alternative inference format prompt generators inline it into their prompt, it could confuse the LLM by being in in the vanilla A2UI protocol format.

Happy to fix that in v1.1 or something though!

samples/client/flutter/restaurant_finder/e2e_test/test/infra_test.dart
```diff
@@ -30,9 +30,13 @@ void main() {
     print('Joke from AI:\n\n$result\n\n');
   });
 
-  test('test can start restaurant_finder', () async {
-    final restaurantFinderClient = TestRestaurantFinderClient();
-    addTearDown(restaurantFinderClient.dispose);
-    await restaurantFinderClient.startAndVerify();
-  }, timeout: const Timeout(Duration(minutes: 5)));
+  test(
```
jiahaog, line 33: We can revert this, I fixed this in a separate PR so CI on master is green.

specification/v1_0/catalogs/basic/catalog.json
gspencergoog: Hmm. Yeah, that's a problem.  Right now, I check for markdown code blocks and decompile them if they are json, but that's fragile.  I think we talked about having an "examples" field in addition to an "instructions" field, and that sounds like the way to go.  Then the format agnostic, but catalog related, prompt instructions can go here, but examples go in the examples field, and they can be decompiled.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
gspencergoog: No, but they should be.  We have platform agnostic conformance tests already, but they don't have any error checking. I'll tackle this in a later PR.

agent_sdks/python/a2ui_agent/tests/express/test_compiler.py
gspencergoog: Yeah, I see what you mean.  Let's put that in another PR, this one is hard enough to review.  We should do this, though.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
gspencergoog: Good point.  I made these private:

- `schema_allows_databinding`
- `schema_expects_option_objects`
- `is_check_expression`
- `TOKEN_SPEC`
- `tokenize`
- `TokenParser`

specification/v1_0/catalogs/basic/examples/01_flight-status.json
gspencergoog: Sort of.  I can factor it out if you want.  I defined identifiers more strictly, so that "-" wasn't a valid identifier character anymore, because if we ever want to have expressions (we don't yet, but we might), then allowing a minus sign in identifiers seems like a bad idea.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
gspencergoog: Yes, this is a good idea, I was thinking about that too.  Can I do that in a later PR?

samples/client/flutter/restaurant_finder/e2e_test/test/infra_test.dart
gspencergoog: Yes, thanks.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
gspencergoog: Actually, this wasn't too bad, so I did it in this PR.  I now generate the parser from an Antlr (.g4) file.  Antlr can generate parsers for all the languages we care about, including Dart.

agent_sdks/python/a2ui_agent/src/a2ui/experimental/express/compiler.py
```diff
@@ -0,0 +1,640 @@
...
+          compiled_context = {}
+          if isinstance(raw_context, dict):
+            compiled_context.update(raw_context)
+          elif isinstance(raw_context, list):
+            for item in raw_context:
+              if isinstance(item, dict):
+                compiled_context.update(item)
+          return {"event": {"name": compiled_event_name, "context": compiled_context}}
+
+        # Is it a regular catalog function?
```
jacobsimionato, line 600: These question comments are fun!

specification/proposals/express/examples/01_flight-status.a2ui
jacobsimionato, line 1: I have a moderate preference for deleting all these example files but keeping the script to generate them as necessary, e.g. for integration tests. Perhaps just keep 1 or 2 of these as a kind of documentation, so people can see what the format looks like at a glance?

Motivation: It would be great if people can add additional samples to the specification and not need to run scripts to keep all the other variants in sync. If we force every sample app and tool which wants to use the full set of samples to derive them from a single source of truth during the build process, then we ensure everything always uses the full set of samples.

specification/proposals/express/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+Install the `mlx-vlm` package globally:
+
+```bash
+uv tool install --force mlx-vlm
+```
+
+This registers standard CLI utilities in your path, including `mlx_vlm.server`.
+
+### Starting the Gemma 4 completions server
```
jacobsimionato: Is this, and the run_inference stuff below still relevant?

jacobsimionato (approved):

specification/proposals/express/README.md
gspencergoog: Yeah, I removed the local inference info.

specification/proposals/express/examples/01_flight-status.a2ui
gspencergoog: I agree.  I'd like to do that too, but let's do that in another PR too.

jiahaog (approved):
Nice!

PR #1733, opened 2026-06-23
fix(angular): use Angular signals directly

crisbeto:
Reworks the Angular renderer to use Angular signals directly based on the APIs from #1714, instead of having to bridge Angular and Preact signals.


renderers/angular/src/v0_9/core/reactivity.ts
```diff
@@ -0,0 +1,90 @@
...
+  isAttached = true;
+
+  // Clean up the implementation since attempting to use a destroyed injector will throw.
+  injector.get(DestroyRef).onDestroy(() => {
+    if (isAttached) {
+      setSignalImplementation(_PRIVATE_DEFAULT_SIGNAL_IMPLEMENTATION);
+      isAttached = false;
+    }
+  });
+}
```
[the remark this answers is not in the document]
crisbeto: This isn't a concern, the environment injector would be shared between all apps.

renderers/angular/src/v0_9/core/reactivity.ts
```diff
@@ -0,0 +1,90 @@
...
+    batchWrite: (batchFn: () => void) => {
+      // Angular signals batch automatically. We just execute the function.
+      batchFn();
+    },
+    isSignal: (val: unknown): val is Signal<unknown> => isSignal(val),
+    getValue: <T>(sig: Signal<T>) => (sig as AngularSignal<T>)(),
+    setValue: <T>(s: Signal<T>, value: T) => {
+      const sig = s as AngularWritableSignal<T>;
+
+      if (typeof sig.set === 'function') {
```
josemontespg: We can use `isWritableSignal` here. I'm thinking that we should throw in case it is not a WritableSignal, because it means that something is very broken

renderers/angular/src/v0_9/core/component-binder.service.ts
```diff
@@ -104,43 +113,45 @@ export class ComponentBinder {
...
       }
 
-      const angSig = toAngularSignal(preactSig as any, this.destroyRef, this.ngZone);
+      if (valueSignal?.unsubscribe) {
+        this.destroyRef.onDestroy(() => valueSignal.unsubscribe!());
+      }
 
       bound[key] = {
-        value: angSig,
+        value: valueSignal as AngularSignal<unknown>,
```
josemontespg, line 126: Optional: What are your thoughts on having a helper assertAngularSignal in reactivity.ts, that does an `isSignal` check and throws when not passing? And returns the correct type.

The error message could include to  call `initializeAngularReactivity` if expecting Angular signals.

This way the assumption that this code can expect Angular signals to be returned by web_core is declared and checked in code.

renderers/angular/src/v0_9/core/function_binding.spec.ts
```diff
@@ -48,28 +50,29 @@ describe('Function Bindings', () => {
       };
 
       // 1. Resolve Signal
-      const resSig = context.resolveSignal<number>(callValue as any);
-
-      // 2. Convert to Angular Signal
-      const angSig = toAngularSignal(resSig, mockDestroyRef);
+      const sig = context.resolveSignal<number>(callValue as any) as Signal<number>;
 
       // 3. Initial state
```
josemontespg: Nit: Please update numbering of comments. Same below.

renderers/angular/src/v0_9/core/function_binding.spec.ts
```diff
@@ -48,28 +50,29 @@ describe('Function Bindings', () => {
...
+      TestBed.tick();
 
       // 5. Verify reactive updates
-      expect(angSig()).toBe(6);
+      expect(sig()).toBe(6);
 
       // 6. Update again to confirm reactive stream remains healthy
       dataModel.set('/inputValue', 10);
-      expect(angSig()).toBe(11);
+      TestBed.tick();
```
josemontespg, line 68: Do you know why this tick is needed? I was hoping that with Angular signals the propagation of values would be more robust and not depend on Zone or have timing quirks.

Can you check if this test works without the tick? Same below.

renderers/angular/src/v0_9/core/function_binding.spec.ts
```diff
@@ -48,28 +50,29 @@ describe('Function Bindings', () => {
...
+
+      // TestBed.flushEffects();
```
josemontespg: Nit: remove commented code.

renderers/angular/src/v0_9/core/reactivity.ts
```diff
@@ -0,0 +1,90 @@
...
+export function initializeAngularReactivity(injector: EnvironmentInjector): void {
+  if (isAttached) {
+    return;
+  }
+
+  const angularSignalImpl: SignalImplementations = {
+    signal: <T>(initialValue: T) => signal(initialValue) as Signal<T>,
+    computed: <T>(computeFn: () => T) => computed(computeFn) as Signal<T>,
+    effect: (effectFn: () => void | (() => void)) => {
+      return untracked(() => {
```
josemontespg, line 57: Most of the implementations are trivial, but this one is a bit more involved. Could you add tests checking this behavior please?

renderers/angular/src/v0_9/core/reactivity.ts
```diff
@@ -0,0 +1,90 @@
...
+    peekValue: <T>(sig: Signal<T>) => untracked(() => (sig as AngularSignal<T>)()),
+  };
+
+  setSignalImplementation(angularSignalImpl);
+  isAttached = true;
+
+  // Clean up the implementation since attempting to use a destroyed injector will throw.
+  injector.get(DestroyRef).onDestroy(() => {
+    if (isAttached) {
+      setSignalImplementation(_PRIVATE_DEFAULT_SIGNAL_IMPLEMENTATION);
```
josemontespg, line 99: This is risky. There might have been multiple calls to initializeAngularReactivity, and with the current implementation the first call to `initializeAngularReactivity` wins. If the first one happened with an injector that is destroyed, then the rest of the app would stop receiving Angular Signals from web_core.

Currently this is fine because all calls to initializeAngularReactivity happen in places that are not destroyed, like in services that are provided in root. But that might not continue to be true in the future as the code evolves.

Is there a way we can make sure that after initializeAngularReactivity is called, web_core is guaranteed to always return Angular Signals forever?

renderers/web_core/src/v0_9/reactivity/signals.ts
```diff
@@ -51,7 +51,7 @@ let setValueImpl: SignalImplementations['setValue'];
 let peekValueImpl: SignalImplementations['peekValue'];
 
 /** Default signal implementations. Exported only for testing purposes. */
```
josemontespg, line 53: Now this is used during runtime too. If intended to be used only for testing, consider creating a util function in renderers/web_core/src/v0_9/test/test-utils.ts or similar, that clears any overrides to signal implementations.

josemontespg (changes requested):
Thank you for working on this! I carefully reviewed this code and led me to add more comments than usual, sorry about that! This bit of code is key to the angular renderer and I want to make sure that we are using patterns that can be sustained over time. Happy to chat off-PR to expedite discussions if you want.

renderers/angular/src/v0_9/core/component-binder.service.ts
crisbeto: Yeah I think we might need it. The asserts that I had to sprinkle in were a bit clunky...

renderers/angular/src/v0_9/core/function_binding.spec.ts
crisbeto: I think it's because the data model uses an effect under the hood that needs to be flushed.

renderers/angular/src/v0_9/core/reactivity.ts
crisbeto: Hmm, good question. Maybe we should include a `variant` string in there? Then the Angular renderer can check if it's still dealing with Angular signals.

renderers/web_core/src/v0_9/reactivity/signals.ts
crisbeto: Do you have any preference here? I'm wondering if it makes sense to pass null in to reset it.

renderers/angular/src/v0_9/core/reactivity.ts
crisbeto: I simplified it a little bit. I'm not sure how much there is to test since we're basically just forwarding things to the `effect` API.

josemontespg: Now it is more direct, thanks for the refactor! Ok to skip testing now IMO

renderers/angular/src/v0_9/core/reactivity.ts
josemontespg: Ideally the renderer would not have to check which type of signal is returned every time. I see that the reason this is not trivial is that we need to handle injectors with different lifecycles.

What if we make sure that the injector used is the root injector? Then we can throw early when calling initializeAngularReactivity with a child injector.

This is what AI came up with:

```typescript
import { Injector, inject, InjectFlags } from '@angular/core';

export function isRootInjector(currentInjector: Injector): boolean {
  // Look up the parent injector by skipping the current one
  const parentInjector = currentInjector.get(Injector, null, { skipSelf: true, optional: true });
  
  // If there is no parent injector, this IS the root injector
  return parentInjector === null;
}
```

renderers/angular/src/v0_9/core/function_binding.spec.ts
josemontespg: I see, thanks for checking

renderers/web_core/src/v0_9/reactivity/signals.ts
josemontespg: I think we can modify reactivity.ts so that resetting this during runtime is not needed.

josemontespg (changes requested):

renderers/angular/src/v0_9/core/reactivity.ts
crisbeto: We're already enforcing that it's a top-level injector through the types though. The only time the `EnvironmentInjector` is destroyed is in-between unit tests.

renderers/web_core/src/v0_9/reactivity/signals.ts
crisbeto: I think we'd have to in order to support tests though. In an app the `EnvironmentInjector` that is passed into `initializeAngularReactivity` is a singleton, but in tests it gets destroyed and re-created. That's why I ended up resetting it.

renderers/angular/src/v0_9/core/reactivity.ts
josemontespg: Oh I see, we are using `EnvironmentInjector` as the param type. Makes sense.

renderers/angular/src/v0_9/core/reactivity.ts
```diff
@@ -0,0 +1,101 @@
...
+    peekValue: <T>(sig: Signal<T>) => {
+      assertAngularSignal(sig);
+      return untracked(() => sig() as T);
+    },
+  };
+
+  setSignalImplementation(angularSignalImpl);
+  isAttached = true;
+
+  // Clean up the implementation since attempting to use a destroyed injector will throw.
```
josemontespg, line 94: Please add a comment here explaining that in practice this only happens in tests, because during normal app runtime, the injector is root and never destroyed.

renderers/web_core/src/v0_9/reactivity/signals.ts
josemontespg: I understand now, the ondestroy code in reactivity.ts is there only to support tests. Makes sense. We can keep this as is then.

josemontespg (approved):
Approved with minor request to add a comment 🎉

Thanks for working on this!

PR #1734, opened 2026-06-23
feat(swift_core): implement SchemaRegistry and KeywordRegistry

pinieb:
# Description

- Add `SchemaRegistry` as a shared reference type to cache compiled 
  `SchemaNode`s, enabling synchronous `$ref` resolution and preventing 
  infinite CoW loops during deep evaluation.
- Add `KeywordRegistry` and `EvaluatorFactory` to decouple keyword 
  parsing from the core engine, allowing runtime extensibility.
- Stub `SchemaCompiler` to resolve dependency graph for applicators.
- Add test coverage for node caching and custom keyword injection.

## Pre-launch Checklist

For this PR:

- [X] I updated/added relevant documentation.
- [X] My code changes (if any) have tests.
- [X] If my branch is on fork, I have verified that [scripts/e2e_test.sh](../scripts/e2e_test.sh) passes.
  - Used `run_tests.sh` for Swift.

[CHANGELOG]: ../CHANGELOG.md


jacobsimionato (commented):
Hey I can't remember if I asked this before, but did you consider using an existing Swift JSON schema library? Gemini tells me that this one is well-regarded: https://github.com/ajevans99/swift-json-schema

I know that would require importing an existing library to Google's monorepo, but it does mean less code to own for us. For other languages, we have generally used existing libraries.

It might be easy enough to migrate later if we want to, though?

Otherwise, this looks good to me though, though I didn't pick over it too closely seeing as I believe the JSON schema specification provides a lot of structure. I added Greg to take a look, because he created an equivalent library for the Dart ecosystem.

pinieb:
I had considered using that library a while back, but it may be worth reexamining now.

My reservations before were centered around the build time impact of macros and introducing key dependencies into the library that would be hard to swap out later.

https://github.com/ajevans99/swift-json-schema has a particularly nice set of macros that would let clients easily define both their schema and a truly strongly-typed Swift struct to parse instances into in one go. It would be something of a shame to use the core validation logic from that repo but not the macros.

Being careful about using that library behind a protocol would make me feel better about having it as a key dependency we don't directly control, and could also provide a nice escape valve for clients that are more sensitive to having macros in their build graph.

I'll do a little prototyping today for wrapper protocols and report back!

PR #1736, opened 2026-06-23
feat(swift_core): implement SchemaCompiler for AST translation

pinieb:
# Description

- Fully implement `SchemaCompiler` to translate `JSONValue` into `SchemaNode`s.
- Add support for boolean schemas (`true` = always valid, `false` = always invalid).
- Implement structural keyword interception (`$id`, `$defs`, `$dynamicAnchor`).
- Ensure `$id` properly updates the base URI and resets the JSON pointer scope.
- Ensure `$defs` are recursively compiled and cached in the `SchemaRegistry` 
  for synchronous `$ref` lookup during evaluation.
- Add comprehensive compiler test coverage for scope resets and definition caching.

## Pre-launch Checklist

For this PR:

- [X] I updated/added relevant documentation.
- [X] My code changes (if any) have tests.
- [X] If my branch is on fork, I have verified that [scripts/e2e_test.sh](../scripts/e2e_test.sh) passes.
  - Ran `run_tests.sh` for Swift

[CHANGELOG]: ../CHANGELOG.md


swift/core/JSONSchema/README.md
```diff
@@ -0,0 +1,85 @@
...
+struct MinLengthEvaluator: KeywordEvaluator {
+  let minLength: Int
+
+  func evaluate(instance: JSONValue, context: ValidationContext) -> ValidationResult {
+    guard case .string(let text) = instance else {
+      // Ignore validation if the type is not a string (standard JSON Schema behavior).
+      return .success()
+    }
+
+    if text.count >= minLength {
```
gspencergoog, line 49: This example uses `text.count` to validate the length of a string.

In Swift, `text.count` returns the number of `Character`s (extended grapheme clusters), which may not match the number of Unicode code points. The JSON Schema spec says: *"The length of a string instance is defined as the number of its Unicode codepoints."* For example, the emoji "👨‍👩‍👧‍👦" is a single extended grapheme cluster (`text.count == 1`), but consists of 7 Unicode code points. Under JSON Schema rules, its length is 7.

To be spec-compliant, string length validators should use `text.unicodeScalars.count`.

swift/core/JSONSchema/Sources/SchemaCompiler.swift
```diff
@@ -0,0 +1,146 @@
...
+          evaluators.append(evaluator)
+        }
+      }
+
+      let node = SchemaNode(
+        identity: currentIdentity,
+        dynamicAnchor: parsedAnchor,
+        evaluators: evaluators
+      )
+      schemaRegistry.register(node, for: currentIdentity)
```
gspencergoog, line 117: When a subschema defines an `$id` keyword, `currentIdentity` is updated to the new base URI, and the compiled `SchemaNode` is registered under that new URI. However, according to the JSON Schema specification, the schema must also remain resolvable via its original lexical path (the `identity` parameter passed into `compile()`). If you do not register it under the lexical URI, any local `$ref` pointer referencing this schema by its document path (e.g. `#/properties/child` or `#/$defs/mySchema`) will fail to resolve during evaluation.

swift/core/JSONSchema/Sources/SchemaIdentity.swift
```diff
@@ -0,0 +1,95 @@
...
+  ///   - baseURI: The base URI.
+  ///   - pointer: The JSON Pointer.
+  public init?(baseURI: String, pointer: JSONPointer = JSONPointer()) {
+    let parts = baseURI.components(separatedBy: "#")
+    self.baseURI = parts.first ?? ""
+
+    if parts.count > 1 {
+      let fragment = "#" + parts[1...].joined(separator: "#")
+      guard let basePointer = JSONPointer(stringRepresentation: fragment) else {
+        return nil
```
gspencergoog, line 52: In this code, if the `baseURI` contains a fragment (e.g. `https://example.com/schema#invalid-fragment`), you attempt to parse it as a `JSONPointer`. If parsing fails (since it doesn't start with `/`), the initializer silently returns `nil` at line 52.

When called during compilation, this could cause the compiler to fail with a generic error or crash due to forced unwrapping of `SchemaIdentity` elsewhere.

Consider adding a specific compiler warning or throwing a descriptive error when a base URI with an invalid fragment is encountered, rather than silently failing initialization.

gspencergoog:
I think I have the same comment as Jacob did on your other PR: is there a standard Swift library for this?  It would be good not to have to maintain a JSON Schema validator if there's a well maintained one out there already.

PR #1768, opened 2026-06-25
feat: introduce Spec-Driven Development (SDD) blueprints, skills, and codebase metadata

jacobsimionato:
# Spec-Driven Development (SDD) Foundation

This PR introduces the foundational framework for **Spec-Driven Development (SDD)** in the A2UI project. It formalizes language-agnostic module and feature blueprints, provides codebase-specific metadata tracking via local codebase blueprints, and adds specialized AI agent skills to automate execution workflows.

## Description of Changes

1. **Spec-Driven Development Documentation**:
   - Added `development_docs/spec_driven_development.md` which outlines A2UI's SDD methodology, its goals, the life cycle of blueprints, required vs optional features, and the responsibilities of AI agents and human reviewers.

2. **Language-Agnostic Module Blueprints**:
   - Relocated and restructured specifications to the central `blueprints/modules/` directory with frontmatter metadata:
     - `specification/v1_0/docs/core_sdk_spec.md` -> `blueprints/modules/a2ui_core.blueprint.md`
     - `specification/v1_0/docs/framework_adapter_spec.md` -> `blueprints/modules/a2ui_framework_adapter.blueprint.md`
     - `agent_sdks/agent_sdk_guide.md` -> `blueprints/modules/a2ui_inference.blueprint.md`

3. **Codebase Blueprints (Local Metadata)**:
   - Added `codebase.blueprint.md` trackers to the roots of all implementation directories across Python, Web (React, Lit, Angular, Web Core), Swift, and Flutter. These files associate the codebase with its central module blueprint, track implemented features, and note codebase-specific overrides and local run/test commands.

4. **AI Agent Skills**:
   - Added `.agents/skills/a2ui-blueprint-navigator/SKILL.md`: Guiding AI agents in discovering, auditing, and querying relations between central blueprints and physical directories.
   - Added `.agents/skills/implement_new_sdks_for_client_language/SKILL.md`: Providing detailed, phased execution plans for building new Core SDKs and Framework Adapters from scratch.

## Rationale

As A2UI scales across multiple languages (TypeScript, Python, Swift, Kotlin, Dart) and frameworks, maintaining behavioral consistency, tracking feature parity, and onboarding new contributors is a massive challenge.

By adopting SDD:
- **Human efforts** shift to creating and reviewing specifications (the *blueprints*).
- **AI agents** are empowered to implement features cleanly and port features across different platforms by following structured, non-overlapping workflows.
- **Codebase Blueprints** provide an auditable index of what is implemented, enabling automated parity checks and continuous compliance validation.

Since these changes are purely structural documentation, metadata, and agent skill configurations, there is no runnable runtime code changed.

1. **Verify directory structure**:
   Ensure `/blueprints/modules/` contains the three primary module blueprints and that they have the appropriate YAML frontmatter.
2. **Review codebase blueprints**:
   Inspect `codebase.blueprint.md` in any of the active SDK or renderer roots (e.g. `renderers/react/codebase.blueprint.md`) to see how they reference their parent module and document native commands.
3. **Verify Git alignment**:
   No staging or compilation errors should be present across any targets.


specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+    FB[Feature Blueprint] -->|Merges into| MB[Module Blueprint]
+    FB -->|Guides implementation of| CB[Codebase Blueprint]
+    MB -->|Specifies architecture of| CB
+    CB -->|Documents| C[Codebase Implementation]
+```
+
+Under this model:
+
+1. **Module Blueprints** define the high-level architecture, core interfaces, and behavior of repository modules (e.g., `a2ui_core`, `a2ui_inference`, `a2ui_react`, etc.).
+2. **Feature Blueprints** describe specific feature additions or behavioral changes to a module.
```
gspencergoog, line 42: So, these are basically functionality diffs on the module blueprint?

And once they are "merged" conceptually with the module, the feature blueprints go away?

.agents/skills/a2ui-blueprint-maintenance/SKILL.md
```diff
@@ -0,0 +1,47 @@
...
+
+## **1. Lifecycle & Archiving Recipes**
+
+### **Recipe A: Archiving a Completed Required Feature**
+
+When a required feature is fully adopted and implemented across all active codebases:
+
+1.  **Verify Parity**: Ensure that all codebases listed in the module's `associated_codebases` have successfully implemented the feature (listed in their `codebase.blueprint.md`).
+2.  **Merge Specification**: Integrate the detailed requirements and behavior of the feature directly into the main text of the **Module Blueprint** (e.g. `blueprints/modules/a2ui_core.blueprint.md`). Add the feature's name to the module's `included_features` or similar list.
+3.  **Archive the Feature Blueprint**: Move the feature blueprint file from `blueprints/features/` to the `blueprints/features/archived/` folder using git:
```
gspencergoog: If we have git, why do we need an archive?  I worry that the archive serves as a source of confusion for the agents: they do some keyword search and find it in an archived feature, and make the mistake of assuming it is still valid.

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+
+# **Document types**
+
+## **Feature blueprint**
+
+### **Required vs optional features**
+
+A **Required Feature Blueprint** describes a feature that is expected to be implemented in all codebases for a specific module. When a required feature blueprint is added, the associated module blueprint is updated at the same time to completely include the information required to implement the feature. The purpose of preserving the required feature blueprint in version control is to help agents implement the feature in existing module codebases based on previous versions of the module blueprint.
+
+An **Optional Feature Blueprint** describes a feature that is not baked into the module blueprint and is not expected to be implemented in all codebases. It allows platforms to support experimental or framework-specific features without forcing compliance across all SDKs.
```
gspencergoog, line 72: What about proposed features that aren't even up to the level of optional?  For example, ones that haven't been ratified as a part of the feature set even as optional inclusions, but we'd like to try them out and see how/if they can be ported to the platforms.

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+
+Specification-driven development is still useful for these features, but the specifications don’t always need to be checked into the file system. They could be checked in as blueprints, or written in Github issues, or exist as temporary files on a developer’s file system.
+
+# **Document types**
+
+## **Feature blueprint**
+
+### **Required vs optional features**
+
+A **Required Feature Blueprint** describes a feature that is expected to be implemented in all codebases for a specific module. When a required feature blueprint is added, the associated module blueprint is updated at the same time to completely include the information required to implement the feature. The purpose of preserving the required feature blueprint in version control is to help agents implement the feature in existing module codebases based on previous versions of the module blueprint.
```
gspencergoog: How do we deal with the intervening implementation period and verify that all required features eventually get implemented properly on all platforms?  It seems like you need an intervening designation for when the required feature is specified, but not yet implemented everywhere.

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+## **Technical Decisions & Overrides**
+* **Dynamic Binders**: Unlike static languages, we utilize TypeScript runtime reflection to implement **Generic Binders**. This automatically resolves `DynamicString` properties into static props at the adapter layer, avoiding manual boilerplate for custom components.
+* **Event System**: To handle React's synthetic event loop, we throttle input change actions (e.g., `TextField` keystrokes) in the adapter before committing them to the local `DataModel` to optimize performance.
+```
+
+# **Developer journeys**
+
+This section explains what steps will be taken by developers and agents to perform common tasks using spec-driven development.
+
+## **Specify a new required feature**
```
gspencergoog, line 208: Does this only happen after a feature has been added as "optional" in a previous process?  The definition of an optional feature seems to imply that optional features are the only ones which can be *not* implemented on a platform.  So if you go straight to "required", you'd have to include the implementation (or at least the codebase blueprint) of the feature for all platforms in the same PR.

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+
+## **Promote an optional feature to be required**
+
+1. Update module blueprint based on the feature blueprint, and add it to the module’s “included_features”. (coding agent)
+2. Send PR for review.
+
+## **Implement an optional or required feature in a codebase**
+
+1. Verify that the codebase does not already contain the feature
+2. Create a temporary design describing in detail how the feature should be implemented in the specific codebase, taking into account the feature blueprint, the codebase blueprint, and the actual codebase code. This file should not be checked in. **(human input required)**
```
gspencergoog, line 227: What if the implementation requires multiple steps/PRs/authors?  Shouldn't the temporary designs exist for continuity?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+3. Use the temporary design to implement the feature
+4. Update the codebase blueprint to add the feature to the "implemented_features" list and include any codebase-specific decisions that were made as part of the feature implementation.
+5. Send PR for review.
+
+## **Implement all the features necessary to bring a codebase “up to date”**
+
+1. Read the codebase blueprint and module blueprint and identify all the required features in the module that are not in the codebase.
+2. Implement each feature in chronological order, based on their blueprints, following the steps to implement a feature above.
```
gspencergoog, line 235: Why chronological?  What if a feature depends on another one, but was proposed first?  I'm not sure we need to specify an order.

Also, what counts as the "date" for the feature?  There's not a date in the metadata, and the date of creation might be unclear, what with merges and squashing and timezones and all that.  Or it could be created empty, the other blueprints in the set worked on, and then the empty one filled in.  Is it the creation date then, or the last modification date?  It also seems like you can have multiple features in the same feature blueprint, so sorting those seems problematic.

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+
+1. **Module Blueprints** define the high-level architecture, core interfaces, and behavior of repository modules (e.g., `a2ui_core`, `a2ui_inference`, `a2ui_react`, etc.).
+2. **Feature Blueprints** describe specific feature additions or behavioral changes to a module.
+3. **Codebase Blueprints** reside within each platform's implementation folder, tracking implemented features and documenting local design decisions or deviations.
+
+Formal blueprints are primarily used for major features, e.g. the addition of a new public API, behavior, protocol version or architectural change. By having these specifications to be checked into the codebase, we create a natural opportunity for review of feature designs before they are implemented. We also ensure that when we add a feature to one SDK codebase, we can easily port it to other languages.
+
+### **Smaller features**
+
+Smaller features can still be implemented ad-hoc with no formal blueprint. These small feature typically:
```
gspencergoog, line 49: I worry that over time this will result in big discrepancies in the blueprints and in the implementations.

If you don't have a way to update the blueprints when these small feature changes cause a divergence, then over time the codebase and design will grow apart, like two coral reefs that start with the same shape.

Maybe for these we require some kind of logging that helps bridge the gap (like a CHANGELOG, but skipping the blueprinted features)?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+
+## **Feature blueprint**
+
+### **Required vs optional features**
+
+A **Required Feature Blueprint** describes a feature that is expected to be implemented in all codebases for a specific module. When a required feature blueprint is added, the associated module blueprint is updated at the same time to completely include the information required to implement the feature. The purpose of preserving the required feature blueprint in version control is to help agents implement the feature in existing module codebases based on previous versions of the module blueprint.
+
+An **Optional Feature Blueprint** describes a feature that is not baked into the module blueprint and is not expected to be implemented in all codebases. It allows platforms to support experimental or framework-specific features without forcing compliance across all SDKs.
+
+- **Decoupled Lifecycle**: Optional feature blueprints exist as standalone specification files in `blueprints/features/` and are not merged into the base Module Blueprint.
```
gspencergoog, line 74: How are optional features promoted to required features? Is it treated as deleting the optional feature and adding the required one?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+Every codebase must contain a `codebase.blueprint.md` file in its root directory. This file maps the concrete implementation back to the language-agnostic module blueprint, tracking its feature support and documenting local engineering decisions.
+
+### **codebase.blueprint.md Structure (Example)**
+
+Below is an example codebase blueprint for a React renderer codebase (`a2ui_react`):
+
+```
+---
+associated_module: a2ui_react
+implemented_features:
```
gspencergoog: Is this a list of the features that have actually been implemented in the codebase, or is it a list of the features from the module blueprint that are implemented in this codebase blueprint?

How can I tell what work is left to implement this codebase blueprint, either from the module -> codebase blueprints, or from the codebase blueprint -> code?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+3. Use the temporary design to implement the feature
+4. Update the codebase blueprint to add the feature to the "implemented_features" list and include any codebase-specific decisions that were made as part of the feature implementation.
+5. Send PR for review.
+
+## **Implement all the features necessary to bring a codebase “up to date”**
+
+1. Read the codebase blueprint and module blueprint and identify all the required features in the module that are not in the codebase.
+2. Implement each feature in chronological order, based on their blueprints, following the steps to implement a feature above.
+3. Consult the module blueprint and verify that the codebase now matches it, making minor changes to the codebase as necessary to make it as consistent as possible to the module blueprint.
+4. Update the codebase blueprint to add the feature to the "implemented_features" list.
```
gspencergoog, line 237: If this is only added after the feature is implemented, then what lists the things that are not yet implemented in the codebase blueprint, so that it can be verified that all features in the module are at least designed and specified in the codebase blueprints?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+* **JSON Parsing Suite**: Validates envelope compliance against `server_to_client.json` and throws `A2uiValidationError` on schema violations.
+* **Data Model Binding Suite**: Verifies absolute and relative pointer resolution, type coercion rules, and reactive signal emission.
+* **Path to Mock Data**: `specification/v1_0/test/conformance/a2ui_core/`
+```
+
+## **Codebase blueprint**
+
+A **Codebase** is a concrete, language-specific or framework-specific implementation of a module. Examples of codebases in this repository include `renderers/web_core` (TypeScript implementation of `a2ui_core`), `renderers/react` (React implementation of `a2ui_react`), and `agent_sdks/kotlin` (Kotlin implementation of `a2ui_inference` and `a2ui_core`).
+
+Every codebase must contain a `codebase.blueprint.md` file in its root directory. This file maps the concrete implementation back to the language-agnostic module blueprint, tracking its feature support and documenting local engineering decisions.
```
gspencergoog: This implies that codebases cannot be nested, and it might be helpful to split up the blueprint into submodules, both for organization and for comprehension.

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+
+Below is an example codebase blueprint for a React renderer codebase (`a2ui_react`):
+
+```
+---
+associated_module: a2ui_react
+implemented_features:
+  - dynamic_theming
+  - call_function_rpc
+protocol_compliance: v1.0
```
gspencergoog: So what happens if we want designs for more than one protocol version?  What if we start having multiple protocols?  Maybe call this "a2ui_protocol_version"?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+# **Developer journeys**
+
+This section explains what steps will be taken by developers and agents to perform common tasks using spec-driven development.
+
+## **Specify a new required feature**
+
+1. Create required feature blueprint **(significant human input required)**
+2. Update module blueprint based on the feature blueprint, to ensure the module blueprint fully specifies the feature and how to implement it. Add the feature name to the module’s “included features”. (coding agent)
```
gspencergoog, line 211: This seems redundant: there's already a document with the feature in it, and now the contents of the module blueprint contain that information as well, just in an integrated form.  I worry about divergence over time or even just simply having incomplete or conflicting information in the two documents from the beginning, with no clear source of truth.  Why not just update the module blueprint to point to the feature?  I feel like module blueprints should basically be a table of contents with high level descriptions, like a SKILL.md file and its references.

Also, how do I track a sentence in the module blueprint back to a particular feature blueprint?  What happens if I split one feature into two? Or merge two?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+3. Send PR for review. You can also include an implementation in a codebase in the same PR if you want \- see “Implement an optional or required feature in a codebase”.
+
+## **Specify a new optional feature**
+
+1. Create optional feature blueprint **(significant human input required)**
+2. Send PR for review. You can also include an implementation in a codebase in the same PR if you want \- see “Implement an optional or required feature in a codebase”.
+
+## **Promote an optional feature to be required**
+
+1. Update module blueprint based on the feature blueprint, and add it to the module’s “included_features”. (coding agent)
```
gspencergoog, line 221: Don't you also need to update the optional feature's blueprint to mark it as required?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+
+```
+---
+name: a2ui_core
+code_location:
+  - renderers/web_core
+  - agent_sdks/kotlin/core
+protocol_version: v1.0
+included_features:
+  - bidirectional_rpc
```
gspencergoog, line 153: How do I identify which parts of the document correspond to which included feature?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+- **Discovery**: Codebases that implement an optional feature must list it in their `codebase.blueprint.md` under `implemented_features` to let clients and agents know it is supported.
+
+### **my_feature.blueprint.md structure (Example)**
+
+Every feature blueprint must follow a standardized Markdown structure with YAML frontmatter. Below is an example feature blueprint for a dynamic theming feature, written in a language-agnostic way:
+
+```
+---
+feature_name: dynamic_theming
+module_blueprints:
```
gspencergoog: What does this list imply: a dependency relationship, or a membership relationship?  Do these module blueprints need to exist before this feature can be specified?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+  - a2ui_core
+  - a2ui_react
+  - a2ui_lit
+  - a2ui_angular
+required: false
```
gspencergoog: Required by which modules?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+A **Module Blueprint** describes an entire architectural module in a language-agnostic way. It serves as the primary source of truth for building a new codebase from scratch or verifying the correctness of an existing one.
+
+### **a2ui_core.blueprint.md Structure (Example)**
+
+Every module blueprint must follow a standardized Markdown structure with YAML frontmatter. Below is an example module blueprint for the core state layer (`a2ui_core`):
+
+```
+---
+name: a2ui_core
+code_location:
```
gspencergoog: Can more than one module blueprint contain the same code location?  What about common utilities, are they listed here or not?

specification/proposals/spec_driven_development.md
```diff
@@ -0,0 +1,290 @@
...
+  - renderers/web_core
+  - agent_sdks/kotlin/core
+protocol_version: v1.0
```
gspencergoog: What if a module supports more than one protocol version, or only certain ones?

specification/proposals/spec_driven_development.md
jacobsimionato: Yes, they are functionality diffs.

The mental model I was imagining is:
- Module blueprint: Primarily used for creating brand new codebases for a particular module e.g. a Rust version of the libraries
- Feature blueprint: Primarily used for implementing new features across one or more codebases
   - Required feature blueprint: Gets merged into the module blueprint *immediately* because new codebases should include these features from the start. The required feature blueprints never become completely redundant though, because there will always be stale codebases (e.g. those we don't own) who may use the feature blueprints to update their codebases.
   - Optional feature blueprint: Never gets merged into the module blueprint, because not every new codebases wants this feature.
   
So I imagine existing codebases will primarily use feature blueprints to implement new functionality, and therefore module blueprints do not need to be closely in sync with them. Module blueprints are primarily used to implement new codebases, which is the motivation for merging new required features into them immediately.

Happy to change this model based on the experience and workflows in the team - WDYT?

.agents/skills/a2ui-blueprint-maintenance/SKILL.md
jacobsimionato: I don't feel super strongly about this. I originally proposed we delete features after they are implemented in all codebases we own, but updated the plan to keep an archive at HEAD based on @ditman 's feedback.

My slight preference is to keep an archive at head because I'm persuaded by @ditman 's argument that it can be tricky to find old blueprints in the git history. Also, there are codebases that could and should use this system outside our monorepo and ownership (e.g. Kotlin renderer in AOSP, 3P vue renderers etc), and so it's non-trivial to judge when a feature is implemented "everywhere". I'd love to be able to tell Vue maintainers "Point your agents at this date-sorted folder of feature blueprints and implement them one-by-one", rather than "get your agent to poll our git history at various points in time to recreate a complete list of feature blueprints to implement".

Re confusing agents: I agree this is a potential issue, and I hope that our module related skills can allow agents to manage this carefully.

specification/proposals/spec_driven_development.md
jacobsimionato: Yeah, good point. I tried not to solve every feature lifecycle management problem in this design, and instead just distinguish between features that are merged into module blueprints and those that aren't.

Required = merged into module blueprint
Optional = not merged into module blueprint

Within this system, you can add a note to an optional feature blueprint saying "this is experimental and is not recommended to be implemented beyond Python" etc. At this stage, I expect humans will still decide which features to implement in our codebases when based on knowledge of project context, but then once they ask agents to implement them, the blueprint system will kick in to make sure the correct context is used. If in the future, we switch to a style where a pool of managed agents automatically notice and implement blueprints (wow, probably less than 12 months away at this rate 😬 ), maybe we can add more structure here?

WDYT?

specification/proposals/spec_driven_development.md
jacobsimionato: I was imagining that this system could scale outside our monorepo (e.g. to Kotlin that is hosted separately) and other random open source repos. So that makes the concept of "everywhere" murky. We could define "everywhere" as everything in our monorepo for convenience, but I'm not sure that that's in the spirit of the project.

So I was imagining that rather than us centrally defining when a feature is "fully" implemented, this is a job for owners of each codebase to monitor when their codebase is up to date or not. They monitor the "included features" in the relevant module blueprint and the "implemented_features" in their codebase blueprint. When there is a diff, that indicates that there is work they should do.

Re verifying the features are implemented properly - that is a job for checklists and conformance tests I think.

specification/proposals/spec_driven_development.md
jacobsimionato: I updated a2ui-blueprint-maintenance skill to properly describe this.

- Update the blueprint to mark it as required
- Merge feature into relevant module blueprints

specification/proposals/spec_driven_development.md
jacobsimionato: Actually, how about we remove this parameter, because that implies we know all relevant locations, which could be in different repos etc. Instead, let's let those repos refer back these blueprints, but not have a two-way link.

specification/proposals/spec_driven_development.md
jacobsimionato: I removed protocol version from here. I think it's unnecessarily complex and not exactly related to features.

Instead, I think we might have features like `1_0_specification_support` or `remove_0_8_support` which could represent these changes. If we want to use the codebase blueprint as a handy place to store a list of supported versions, we can add this later, but it's not required to set up blueprints.

specification/proposals/spec_driven_development.md
jacobsimionato: Why do you need to do that? If you want to implement a feature mentioned here in an existing codebase, you should use the feature name to find the relevant module blueprint and then use that as an implementation guide.

The module blueprint is primarily intended for new codebases, and to act as an overall guide to module architecture.

I want to avoid module blueprints being an enumeration of features, because I think a series of incremental designs that build on each other will be hard to understand over time, vs a flattened snapshot of what the entire system should look like with all required features.

specification/proposals/spec_driven_development.md
jacobsimionato: I agree that we'll probably need to handle nested codebases eventually, through a hierarchy of module and/or codebase blueprints. I propose starting with a flat structure for simplicity, then adding this later.

specification/proposals/spec_driven_development.md
jacobsimionato: I removed this field :-D

specification/proposals/spec_driven_development.md
jacobsimionato: This is a list of features that have been implemented in the codebase and codebase blueprint. If the feature is split into multiple PRs, I think it makes sense to only add it here once it's complete.

The codebase blueprint is a snapshot, used to provide context about past decisions to future maintainers, so it should be updated to reflect codebase-specific architectural choices *as the feature is implemented*.

There should be no diff in features between codebase blueprint and code - they should be updated together. If a feature is partially completed, the implementer should track progress themselves and update the `implemented_features` field after completion. If we introduce agent swarms which start stepping on each other's toes, we can add a feature status field later, to mark features as "in progress" etc?

To find the diff in features between module blueprint and codebase, you can compare `implemented_features` and `included_features`.

specification/proposals/spec_driven_development.md
jacobsimionato: The previous text was misleading. Required feature do not need to be implemented on every platform, so you can create them from scratch with no optional phase.

specification/proposals/spec_driven_development.md
jacobsimionato: The primary purpose of module blueprints is to allow new codebases to be created.

We could have a system focused mostly on feature blueprints, where they are the long-lived source of truth for the feature design. But I worry that this will be hard to maintain and require a lot of restructuring and merging/splitting to compact the design and keep it readable. E.g if a module blueprint just points to 20 different features that were added at various times, it's confusing for an agent to navigate inconsistencies in the feature design, e.g. because the features started to override each other, or some are older and designed against older versions of the codebase.

So I was hoping for a lifecycle like:
1. Spec an optional feature.
2. Do initial implementation and iterate on feature spec and codebase implementation for one language
3. Promote to required, and merge into module blueprint for future reference
4. Implement in handful of known codebases
5. Archive the feature spec
6. Maintain the module spec going forward

specification/proposals/spec_driven_development.md
jacobsimionato: Perhaps! But maybe we can solve this separately as an additional change? Perhaps each codebase has a list of designs or something.

specification/proposals/spec_driven_development.md
jacobsimionato: Yes! Updated.

specification/proposals/spec_driven_development.md
jacobsimionato: The list of implemented features in the codebase blueprint should be updated *as each feature is implemented*. I removed this step from here, because it's within the steps for each feature listed above.

specification/proposals/spec_driven_development.md
jacobsimionato: This is required vs optional. I'm updating it to be like an enum instead of a bool, which might be more clear.

specification/proposals/spec_driven_development.md
jacobsimionato: I updated the proposal and skills to be more clear about this, and to add a "dependencies" field to the feature specs to allow pointing out directly dependent features so that agents don't miss them.

Overview:
- The date is just rough date created. The main reason to have it is to make it easy to sort the list of feature spec files once it gets long, to find the most recent and relevant ones. Timezones etc won't make enough of a difference to matter.
-  I added a concept of feature dependencies to solve that problem separately - each feature can now list its direct dependencies.

specification/proposals/spec_driven_development.md
sarashed731-prog: Bitcoin transactions from 2010

jacobsimionato: > Bitcoin transactions from 2010

Lol!!

.agents/skills/a2ui-blueprint-maintenance/SKILL.md
gspencergoog: Yeah, and I guess if we have an archive, we can label it in the `.agents/agents.md` and in a `README.md` in the folder and the agents will use it appropriately.  Okay, I'm convinced.

specification/proposals/spec_driven_development.md
gspencergoog: Yeah, that sounds right to me.  Let's start with that scheme and evolve if we need to.

specification/proposals/spec_driven_development.md
gspencergoog: Yes, I'd agree that the "has been implemented" bit needs to be flipped on the implementation side. I guess I was mainly thinking of how we enforce "required" and the latency in that process. I think the answer is "we don't", because we can't.  But we (or the developers working on the implementation) can tell if a codebase is noncompliant because it hasn't implemented all the features on the list. If they keep a list of feature IDs somewhere, that does make their job easier.  Maybe.  Unless we retroactively modify a feature and it means something different than it did when it was implemented.

Which I think boils down to: The job of the blueprint system is to make it clear what is to be built, not to enforce that it has been.

gspencergoog: Do individual features need semver numbers that get incremented when they change?

specification/proposals/spec_driven_development.md
gspencergoog: Yeah, I would see the codebase set as something similar to the blueprints, only for designs, and it's a codebase specific design for each blueprint feature.  If I was maintaining a codebase with an agent, I'd want that record, so that various agents from various developers will all have the same context about the design instead of relying on the memory mechanisms of the various agents.

specification/proposals/spec_driven_development.md
gspencergoog: Okay.  We're accepting that (because we don't control all platforms) that as soon as we add a required feature, all platforms are out of compliance until they implement the new feature.  Which is fine, I think, but we maybe need to be very careful about semver for the blueprints or something, so we can talk about a codebase being in compliance to a particular version of the blueprints (which is different from the version of the protocol).

specification/proposals/spec_driven_development.md
gspencergoog: What is the purpose of the sorting? Is that just so humans know what changed recently, or is it for the agents?  Because if it's the agents, I'd say all they need is a git hash from the last time they made a pass, and they won't care about the actual order.

specification/proposals/spec_driven_development.md
gspencergoog: What do you think about the logging idea?

specification/proposals/spec_driven_development.md
gspencergoog: And then delete the feature blueprint (or move it to the archive)?  Because otherwise you have to keep them in sync.

specification/proposals/spec_driven_development.md
gspencergoog: Maybe this is just terminology: when you say "codebase blueprint", do you mean the codebase design docs that you talked about elsewhere, or is that different?  I like the thought that there's a codebase specific design doc that lives next to the code that talks about how the source module blueprint was implemented and what design decisions were made at that time for this codebase.

specification/proposals/spec_driven_development.md
gspencergoog: Sounds good.

specification/proposals/spec_driven_development.md
gspencergoog: Yeah, works as an MVP and we can iterate.

specification/proposals/spec_driven_development.md
gspencergoog: > The primary purpose of module blueprints is to allow new codebases to be created.

Ahh, I think this clarifies a lot of this for me.  I was thinking of them as the long lived source of truth.

> 3. Promote to required, and merge into module blueprint for future reference
> 4. Implement in handful of known codebases
> 5. Archive the feature spec

I was thinking that it would go 3, 5, 4 here so that there is never a time when the feature information appears in two places, because someone/something is going to forget to update one of them, and we have to then define what the source of truth is.

How about this: if you want to do it in this order, how about step three just adds a link from the module blueprint to the feature blueprint, and when you do step five, then you replace the link with the feature spec?

specification/proposals/spec_driven_development.md
gspencergoog: > Why do you need to do that?

In your [lifecycle](https://github.com/a2ui-project/a2ui/pull/1768#discussion_r3495814344), this would happen if an agent decided during step four to modify the text in the module spec and needed to know which feature blueprint file to also modify in an identical way. Or actually, vice versa too (modified the feature blueprint and needs to now update it in the module doc).  I can see wanting to modify the description if we find out something new during initial implementation.

specification/proposals/spec_driven_development.md
gspencergoog: Ahh, makes sense.

specification/proposals/spec_driven_development.md
gspencergoog: Yeah, that makes sense.

specification/proposals/spec_driven_development.md
gspencergoog: I make a comment above, but what about having a version for each feature?  Then we could bump the version number independent of the name.  Then the implemented features list would be a map of feature name to supported version of that feature.

.agents/skills/a2ui-blueprint-maintenance/SKILL.md
```diff
@@ -0,0 +1,56 @@
...
+    python3 scripts/validate_blueprints.py
+    ```
+
+### **Recipe B: Promoting an Optional Feature to Required**
+
+When an optional feature is promoted to be required:
+
+1.  **Merge Specification**: Integrate the detailed requirements and behavior of the feature directly into the main text of all affected **Module Blueprints** (e.g., `blueprints/modules/a2ui_core.blueprint.md`). Add the feature's name to each module's `included_features` or similar list.
+2.  **Update Feature Blueprint**: Open the feature blueprint file (e.g., `blueprints/features/YYYY_MM_DD_my_feature.blueprint.md`) and update its YAML frontmatter to set `type: required`.
+3.  **Do Not Archive Yet**: Keep the feature blueprint active in `blueprints/features/` (do **not** move it to the `archived/` folder yet). Because we tolerate incremental development, other codebases/SDKs may still need to implement it. It will remain in `blueprints/features/` until it has been implemented in all associated codebases, at which point it can be archived using **Recipe A**.
```
gspencergoog: I don't think we can know when all the codebases implement it.  Why not just archive the file now that it's part of the module blueprint?
I worry that even if we did know, that implementing it in the last one might not trigger this cleanup step.

Even if we only trigger this on when all of *our* code implements it, what does this do to a partner that is using our spec driven development process to keep their implementation up to date, and they no longer see the separate feature doc because we're done with ours and they're not done with theirs?  I think it would be better to treat all of them the same way: once it's required it's in the module blueprint, and then we all implement it (incrementally).

blueprints/codebases/agent_sdks/python/a2ui_agent/codebase.blueprint.md
```diff
@@ -0,0 +1,29 @@
+---
+associated_module: a2ui_inference
+implemented_features: []
+local_development:
+  test_command: 'uv run pytest'
```
gspencergoog, line 7: I know this is fairly generic, but I have the sneaking suspicion that this is redundant with the build files already in the monorepo and could get out of date with what the CI runs.  Could we just trust the agent to know how to run tests, etc.?  I have my own agent rules about running tests, formatting and linting when I commit a change, and it always seems to know what to run for the various codebases.

blueprints/codebases/renderers/angular/codebase.blueprint.md
```diff
@@ -0,0 +1,27 @@
+---
+associated_module: a2ui_framework_adapter
+implemented_features: []
```
gspencergoog, line 5: Can we populate these lists yet? This means that if I told the LLM to implement missing features in angular, it would try and implement them all, right?

specification/proposals/spec_driven_development.md
jacobsimionato: My take is that the simpler option is to think of features as patches that get applied, which are mostly immutable. Perhaps "feature" isn't the right name, perhaps it should be "work unit". So, if you want to make significant changes to my_feature after it has been implemented in some codebases, you should instead just write another feature called `my_feature_extensions` describes the changes that can be applied on top of `my_feature`. In cases where change our mind about a design, I think we could also do stuff like have `my_feature_extended` describe a shortcut to implement the extended feature from scratch, and add a reference in `my_feature` recommending what to do.

I bet this model isn't quite right, but let's try it out and find out what works and what doesn't!

specification/proposals/spec_driven_development.md
jacobsimionato: Ohhh yeah I quite like your proposal - I see that this avoids having two parallel sources of truth for a feature which is great. And maybe we can move to archive features relatively fast, so that there's an opportunity to keep refactoring the module blueprint to keep it readable.

jacobsimionato: While I think we should have a team discussion at some point, it'd be worth trying se first, because that will give us useful data.

I'll update the plan based on your suggestion and we can try it out.

specification/proposals/spec_driven_development.md
jacobsimionato: My hope was that if we're doing an initial implementation to derisk, we'd keep the feature marked as optional, so that it's not merged into the module blueprint, then we can just modify it however we want.

I guess a related question is: If the feature is implemented in a couple of codebases already, and then we want to modify it, how will we find the relevant code in those codebases to update them? In my original proposal, I was trying to position required features as immutable patches, to solve that problem. If you want to change F1, you create a new F2 describing the changes. Then codebases with F1 can just have F2 applied, and all other codebases can get F1 + F2.

specification/proposals/spec_driven_development.md
jacobsimionato: This is a membership relationship - a list of the modules that the feature impacts.

jacobsimionato:
Hi @gspencergoog, I've updated the PR with several architectural refinements based on recent feedback:

### Summary of Updates

1. **Confined All Blueprint Assets to `blueprints/`**:
   - **Agent Opt-In Directive**: Added `blueprints/README.md` instructing AI agents to ignore the `blueprints/` directory by default unless explicitly instructed to work with blueprints/SDD.
   - **Isolated SDD Skills**: Relocated SDD skills (`a2ui-blueprint-navigator`, `a2ui-create-feature-blueprint`, `a2ui-implement-feature-from-blueprint`, `a2ui-blueprint-maintenance`) from `.agents/skills/` into `blueprints/skills/`. Added `blueprints/link_skills.sh` to allow developers to locally symlink these skills into `.agents/skills/` when needed (ignored in `.gitignore`).
   - **Relocated Codebase Blueprints**: Moved all existing `codebase.blueprint.md` files from their codebase roots into `blueprints/codebases/<relative_path>/codebase.blueprint.md` with an explicit `codebase_path` field.
   - **Validator Script**: Moved the validation script to `blueprints/validate_blueprints.py` and updated CI (`validate_blueprints.yml`).

2. **Commit-Hash Based Module Compliance**:
   - Replaced feature-name compliance tracking with `module_blueprint_commit`, storing the git commit hash of the associated Module Blueprint that the codebase is compliant with.
   - Set `module_blueprint_commit: null` across all codebase blueprints for now so compliance can be verified and marked explicitly.
   - Updated SDD skills with git diffing recipes (`git diff <module_blueprint_commit>..HEAD ...`) for checking compliance and identifying diffs.

3. **Clean Feature Blueprint Lifecycle & Non-Dated Naming**:
   - Feature blueprints are named `feature_name.blueprint.md` without date prefixes.
   - Features begin unmerged in `blueprints/features/feature_name.blueprint.md`. Once merged into a Module Blueprint, the feature file is moved to `blueprints/features/archived/feature_name.blueprint.md`.
   - Removed `included_features:` from module blueprints (archived features are implicitly included).

Could you please have another look? Thanks!

gspencergoog (approved):
This looks great!

PR #1795, opened 2026-06-30
fix(renderers): normalize DateTimeInput placeholder color in Safari

### LINKED ISSUE #792: DateTimeInput: placeholder color differs between date and time input types
## Summary

The `DateTimeInput` component renders with different placeholder text colors for `type="date"` vs `type="time"`. This could be due to browser default styling.

## Context

spotted it during React vs Lit visual parity testing in Safari.

gspencergoog:
## Summary
This PR normalizes the placeholder text color for the `DateTimeInput` component in Safari and other WebKit-based browsers. 

## Changes
* Added WebKit-specific pseudo-element selectors (`::-webkit-datetime-edit` and `::-webkit-datetime-edit-fields-wrapper`) to the input styles of `DateTimeInput` across all three renderers (Lit, React, Angular) and both protocol versions (v0.8, v0.9).
* For v0.9 components, configured the pseudo-elements to use the theme CSS variable `var(--a2ui-datetimeinput-color, var(--a2ui-color-on-input, #333))`.
* For React v0.9, implemented a scoped `<style>` block using `React.useId()` to target the WebKit pseudo-elements, as inline styles cannot target pseudo-elements.
* For v0.8 components, set `color: #333` on the input element and the WebKit pseudo-elements.

## Impact & risks
This is a styling-only change that resolves a visual inconsistency in Safari. There are no breaking changes or logic risks.

## Testing
* Verified that the project builds successfully (`yarn build:all`).
* Verified that all unit and integration tests pass (`yarn test:all`).
* Verified that the linter passes without errors (`yarn lint:all`).

Resolves #792


renderers/react/src/v0_9/catalog/basic/components/DateTimeInput.tsx
```diff
@@ -74,6 +75,12 @@ export const DateTimeInput = createComponentImplementation(DateTimeInputApi, ({p
         gap: 'var(--a2ui-spacing-xs, 0.25rem)',
       }}
     >
+      <style>{`
+        .${className}::-webkit-datetime-edit,
+        .${className}::-webkit-datetime-edit-fields-wrapper {
+          color: var(--a2ui-datetimeinput-color, var(--a2ui-color-on-input, #333));
+        }
+      `}</style>
```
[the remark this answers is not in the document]
gspencergoog: Thanks for the feedback. I have removed the inline <style> tag from the component's render function and replaced it with a static WebKit-specific style rule injected once at the module level, targeting a global .a2ui-datetime-input-webkit class.

ditman (changes requested):
We need to update the CHANGELOG for all the renderer packages touched here, since this code needs to be published. Please add a changelog entry (maybe linking to this PR) to the Unreleased section of each changelog!

Also, add the className you added in the react code to the same input on the other renderers if possible, that way we can keep the selectors consistent.

One last thing, we have some tests that verify that certain elements exist on the page, for example we have a test to assert that there's a date selector here:

* https://github.com/a2ui-project/a2ui/blob/main/renderers/lit/a2ui_explorer/tests/v0_9/07_task-card.test.ts#L57

It'd be nice if that selector well, searched for the date input and not the checkbox (I just noticed this), but also used the new classname you're adding.

There's another test that has a date selector, but it seems it's not being exercised there. See:

* https://github.com/a2ui-project/a2ui/blob/main/renderers/lit/a2ui_explorer/tests/v0_9/30_live-invitation-builder.test.ts

(Note that those same tests exist in React and Angular as well!)

gspencergoog:
> We need to update the CHANGELOG for all the renderer packages touched here, since this code needs to be published. Please add a changelog entry (maybe linking to this PR) to the Unreleased section of each changelog!
> 
> Also, add the className you added in the react code to the same input on the other renderers if possible, that way we can keep the selectors consistent.
> 
> One last thing, we have some tests that verify that certain elements exist on the page, for example we have a test to assert that there's a date selector here:
> 
> * [`main`/renderers/lit/a2ui_explorer/tests/v0_9/07_task-card.test.ts#L57](https://github.com/a2ui-project/a2ui/blob/main/renderers/lit/a2ui_explorer/tests/v0_9/07_task-card.test.ts#L57)
> 
> It'd be nice if that selector well, searched for the date input and not the checkbox (I just noticed this), but also used the new classname you're adding.
> 
> There's another test that has a date selector, but it seems it's not being exercised there. See:
> 
> * [`main`/renderers/lit/a2ui_explorer/tests/v0_9/30_live-invitation-builder.test.ts](https://github.com/a2ui-project/a2ui/blob/main/renderers/lit/a2ui_explorer/tests/v0_9/30_live-invitation-builder.test.ts)
> 
> (Note that those same tests exist in React and Angular as well!)

Oh, good ideas! Okay, I've done all of those. PTAL

renderers/angular/src/v0_9/catalog/basic/date-time-input.component.ts
```diff
@@ -50,15 +50,15 @@ import {DateTimeInputApi} from '@a2ui/web_core/v0_9/basic_catalog';
...
+            class="a2ui-date-time-input a2ui-datetime-input-webkit"
           />
         }
         @if (enableTime()) {
           <input
             type="time"
             [value]="timeValue()"
             (change)="handleTimeChange($event)"
-            class="a2ui-date-time-input"
+            class="a2ui-date-time-input a2ui-datetime-input-webkit"
```
ditman: One quick thing: Do we need to add `a2ui-datetime-input-webkit` having the `a2ui-date-time-input` class already? Can we just use `a2ui-date-time-input` across all renderers?

renderers/lit/src/v0_9/catalogs/basic/components/DateTimeInput.ts
```diff
@@ -99,6 +103,7 @@ export class A2uiDateTimeInputElement extends BasicCatalogA2uiLitElement<typeof
     return html`
       ${props.label ? html`<label>${props.label}</label>` : nothing}
       <input
+        class="a2ui-datetime-input-webkit"
```
ditman: `a2ui-date-time-input` here as well maybe?

renderers/angular/src/v0_9/catalog/basic/date-time-input.component.ts
gspencergoog: Yes, that is much cleaner. I have updated the Angular component to use the existing `a2ui-date-time-input` class for both styling and tests.

renderers/lit/src/v0_9/catalogs/basic/components/DateTimeInput.ts
gspencergoog: Standardized the Lit component to use `a2ui-date-time-input` and updated the corresponding test assertions.

ditman (approved):
Yep, thank you very much for the fixes! I'm on-call next, so I can release this early next week!

PR #1803, opened 2026-06-30
Fix script to avoid adding label 'triage: flag' twice and to label only external PRs.

polina-c:
Contributes to: https://github.com/a2ui-project/a2ui/issues/1802


scripts/triage.mjs
```diff
@@ -185,23 +215,40 @@ export default async function issueTriage({github, context}) {
...
-  // Fetch comments concurrently to avoid a slow, rate-limit-prone serial loop.
-  const itemsWithComments = await Promise.all(
-    openItems.map(async item => ({
-      item,
-      comments: await fetchComments({github, owner, repo}, item),
-    })),
-  );
+  // Fetch comments in bounded concurrent batches to avoid a slow serial loop
+  // without flooding the API.
+  const itemsWithComments = await mapInBatches(openItems, async item => ({
```
ditman, line 220: Why do we need to fetch the comments? Doesn't the issue have a last updated timestamp to determine when it's stale?

scripts/triage.mjs
```diff
@@ -150,25 +157,48 @@ async function postComment({github, owner, repo}, number, body) {
...
-  });
-  return data;
+  try {
+    const {data} = await github.rest.issues.listComments({
+      owner,
+      repo,
+      issue_number: item.number,
+      sort: 'created',
+      direction: 'desc',
+      per_page: 100,
```
ditman, line 195: Are we fetching the last 100 comments? Why such a huge number? Wouldn't something like 10 be enough for this?

ditman (approved):
I've never seen this script, but it seems to be fetching way more data than needed?

scripts/triage.mjs
polina-c: because this bot activity should not be excluded

scripts/triage.mjs
polina-c: some comments can be bot's comments

PR #1816, opened 2026-07-01
Enable typechecks for a2ui_core

jiahaog:
Enable typechecks for a2ui_core

This PR migrates the `a2ui_core` package to strict type checking using `mypy`. 

### Key Fixes:
- Fixed a bug in `MessageProcessor.get_client_capabilities` where programmatically created catalogs (which have no JSON schema and return `None` for `catalog_schema`) were exporting `None` into the `inlineCatalogs` list, violating the protocol.

### Behavioral Breaking Changes:
- **Strict Input Validation in `MessageProcessor`**: Methods handling incoming messages (e.g. `_process_create_surface`) now strictly validate that `surfaceId` is present and is a string. If it is invalid, `_process_create_surface` raises `ValueError` immediately at runtime (previously it would fail later with obscure errors). Other processing methods will return early (noop) instead of processing if `surfaceId` is invalid.

### Other Notable Changes:
- **NodeGraph Refactor**: Removed dynamic monkey-patching of `_binder` on `ComponentNode`. Binders are now tracked internally by `NodeGraph` mapping.
- **PEP 561 Compliance**: Moved `py.typed` to the correct location (`a2ui/core/py.typed`).
- **PEP 484 Re-exports**: Standardized schema imports/exports in `schema/__init__.py` using explicit `as` aliases.

TAG=agy
CONV=f0f6116c-47f0-466b-af69-e8d4ecabedd6


agent_sdks/python/a2ui_core/scripts/generate_schemas.py
```diff
@@ -620,7 +632,9 @@ def generate_server_to_client(s2c_data: Dict[str, Any]) -> tuple[str, List[str]]
             f', alias="{envelope_key}"' if snake_envelope != envelope_key else ""
         )
         output.append(f"class {mname}(StrictBaseModel):")
-        output.append(f"    version: Literal[SPEC_VERSION] = SPEC_VERSION")
+        output.append(
+            f'    version: Literal["{SPEC_VERSION_DOT}"] = "{SPEC_VERSION_DOT}"'
```
nan-yu: Why do we want to resolve the constant here instead of using a variable?

nan-yu (approved):
LGTM with one minor question

agent_sdks/python/a2ui_core/scripts/generate_schemas.py
jiahaog: Previously this was because of mypy limitations. 

> Specifically, mypy does not support using Final variables that are imported from other modules inside Literal[...] type annotations.

I poked around a bit more here and it seems like we can still use the constant if we introduce a type alias; did so in constants.py. Thanks!

PR #1817, opened 2026-07-01
Clarify specifications for auto-flagging script.

polina-c:


scripts/triage.mjs
```diff
@@ -35,8 +35,12 @@
 // bot's own label edits never reset the clock. A PR is "stale" when no internal
 // member has commented after the external author's last comment for more than a
 // day.
+//
+// The job prints to console what items are flagged/unflagged and why. To see the
+// history of runs see:
+// https://github.com/a2ui-project/a2ui/actions/workflows/triage.yml
```
yjbanov, line 44: nit: a more useful link would be to the list of issues labeled with `triage: flag`, which seems to be the actual output of the script.

yjbanov (approved):
lgtm

scripts/triage.mjs
yjbanov: (I'd keep the link to the history too though)

polina-c: added, thanks

PR #1804, opened 2026-06-30
Consolidate repository dependencies and lockfiles into a single workspace-level uv.lock

nan-yu:
## Summary of Changes

This pull request migrates the repository to a unified uv workspace architecture. By centralizing dependency management and lockfiles, the changes aim to reduce conflicts and simplify the maintenance of shared packages across the various sub-projects and samples within the repository.

### Highlights

* **Workspace Consolidation**: Consolidated multiple individual dependency lockfiles into a single workspace-level uv.lock file to improve dependency management and consistency across the project.
* **Dependency Management**: Updated individual project pyproject.toml files to use workspace-level dependencies, ensuring all members resolve to compatible versions.
* **Configuration**: Added a root pyproject.toml to define the uv workspace structure and centralized test configuration.

<details>
<summary><b>Ignored Files</b></summary>

*   **Ignored by pattern:** `.github/workflows/**` (1)
    *   .github/workflows/python_ci.yml
</details>

<details>
<summary><b>Activity</b></summary>

* The pull request was created by nan-yu.
* Gemini Code Assist flagged a dependency version conflict for 'a2ui-agent-sdk' in several project files.
* Gemini Code Assist recommended adding missing test paths to the root configuration.
* Gemini Code Assist advised against using 'packages = ["."]' in Hatchling configuration to avoid including unnecessary files in the wheel.
</details>

Fixes https://github.com/a2ui-project/a2ui/issues/1800

PR #1824, opened 2026-07-01
Migrate docs workflow to use uv and local python setup action

### LINKED ISSUE #1823: Bots are failing at main branch.
Failure details:

PR: https://github.com/a2ui-project/a2ui/pull/1822

Failure: https://github.com/a2ui-project/a2ui/actions/runs/28550576375/job/84646537957?pr=1822

nan-yu:
Fixes https://github.com/a2ui-project/a2ui/issues/1823

## Cause of the Issue

The introduction of a workspace-level pyproject.toml configuration at the root of the repository in #1804 configured global pytest options (like testpaths and import options) for the whole project. In the Docs CI workflow, pytest was run under the global python environment without virtual environment encapsulation, which conflicted with the workspace layout and dependency structure.

## Solution

Update `.github/workflows/docs.yml` to align with other Python workflows.

- Replaced the generic `actions/setup-python` step with the repository's local `./.github/actions/setup-python` composite action to enable uv caching.
- Removed the manual pip caching step as setup-uv manages dependency caching automatically.
- Configured `uv venv` and `uv pip` to install documentation dependencies safely within a virtual environment.
- Updated the test validation, admonition conversion, and mkdocs build/deployment commands to execute via `uv run` to guarantee that they run inside the controlled virtual environment.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -66,23 +66,17 @@ jobs:
           git config --global user.email 41898282+github-actions[bot]@users.noreply.github.com
 
       - name: Setup Python
-        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6
+        uses: ./.github/actions/setup-python
         with:
-          python-version: 3.13
-
-      - name: Restore pip cache
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
-        with:
-          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-docs.txt') }}
-          path: ~/.cache/pip
-          restore-keys: |
-            ${{ runner.os }}-pip-
+          python-version: "3.13"
 
       - name: Install documentation dependencies
-        run: pip install -r requirements-docs.txt
+        run: |
+          uv venv
+          uv pip install -r requirements-docs.txt
 
       - name: Validate documentation conversion script
-        run: pip install pytest && pytest docs/scripts/test_convert_docs.py
+        run: uv run pytest docs/scripts/test_convert_docs.py
 
       - name: Copy schemas to documentation
         run: |
@@ -95,11 +89,11 @@ jobs:
           done
 
       - name: Convert Admonitions in Documentation
-        run: python docs/scripts/convert_docs.py
+        run: uv run python docs/scripts/convert_docs.py
 
       - name: Build Documentation (PR Check)
         if: github.event_name == 'pull_request'
-        run: mkdocs build
+        run: uv run mkdocs build
 
       # Pre-submit only: verify that links in every Markdown file resolve,
       # catching broken internal paths and dead external URLs. Placeholder URLs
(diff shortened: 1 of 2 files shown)


nan-yu:
A successful run of the docs CI has confirmed the fix worked: https://github.com/a2ui-project/a2ui/actions/runs/28551927359/job/84650859914?pr=1824.

README.md
```diff
@@ -1,3 +1,5 @@
+do not merge!
+
```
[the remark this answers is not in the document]
nan-yu: This for testing purpose to trigger the docs CI. 

https://github.com/a2ui-project/a2ui/actions/runs/28551927359/job/84650859914?pr=1824 has confirmed the fix is working and the changes are reverted.

lychee.toml
```diff
@@ -52,6 +52,8 @@ exclude = [
   "flutter/genui/blob/main/examples/e2e/",
   # Flaky. See: https://github.com/a2ui-project/a2ui/issues/1747
   "https://ojs\\.aaai\\.org",
+  # Placeholder / illustrative URLs from dependencies
+  "https://via\\.placeholder\\.com",
```
ditman: Why is this needed?

.github/workflows/docs.yml
```diff
@@ -29,6 +29,7 @@ on:
     paths:
```
ditman, line 29: Oh, we probably also want the lychee.toml file here, since it's configuration for mkdocs anyway, it's similar to the mkdocs.yml IMO?

lychee.toml
nan-yu: It is from one of the dependencies, but since I have excluded Python virtual environment directories in `exclude_path`, I think it is good enough.

.github/workflows/docs.yml
nan-yu: Make sense. It is already included on pull-requests. For consistency, we should also include it on pushes.

lychee.toml
nan-yu: https://github.com/a2ui-project/a2ui/actions/runs/28552615991/job/84653058252?pr=1824 confirmed it's working without this. Removed.

.github/workflows/docs.yml
nan-yu: Done

ditman (approved):
Thanks for fixing this so quickly! PS: Remember to check in `main` after this lands, because I suspect the 2nd half of the docs workflow should trigger (the deployment bits that are now being skipped)

nan-yu:
> Thanks for fixing this so quickly! PS: Remember to check in `main` after this lands, because I suspect the 2nd half of the docs workflow should trigger (the deployment bits that are now being skipped)

The validation step succeeded, but the deployment failed in https://github.com/a2ui-project/a2ui/actions/runs/28552952425/job/84654113350. I'm not clear if it is a new issue. I'll follow up.

ditman:
> the deployment failed in https://github.com/a2ui-project/a2ui/actions/runs/28552952425/job/84654113350. I'm not clear if it is a new issue. 

@nan-yu No, I think I've seen that failure before, for example here: https://github.com/a2ui-project/a2ui/actions/runs/28134560696/job/83318359009

The error seems to be `fatal: could not read Username for 'https://github.com/': No such device or address` maybe? ([link](https://github.com/a2ui-project/a2ui/actions/runs/28134560696/job/83318359009#step:13:62))

PR #1910, opened 2026-07-02
Consolidate python CI jobs

jiahaog:
Now that we use a uv workspace (https://github.com/a2ui-project/a2ui/pull/1814) we can simplify this quite a lot.

TAG=agy

CONV=14bfb759-2e1b-4e01-bd26-758c333ce23d


nan-yu (approved):
Thanks for cleaning it up! It looks much cleaner.

PR #1931, opened 2026-07-06
fix(@a2ui/markdown-it): widen web_core peer dep to >=0.9.2

### LINKED ISSUE #1930: @a2ui/markdown-it peer dep on web_core is pinned to ^0.9.2, conflicts with web_core 0.10.x
Running `npm install` with `@a2ui/react@0.10.1` and `@a2ui/web_core@^0.10` produces a peer dependency warning:

```
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: @a2ui/markdown-it@0.0.4
npm warn Found: @a2ui/web_core@0.10.3
npm warn Could not resolve dependency:
npm warn peer @a2ui/web_core@"^0.9.2" from @a2ui/markdown-it@0.0.4
```

`@a2ui/react@0.10.1` depends on `@a2ui/markdown-it@0.0.4`, but `@a2ui/markdown-it@0.0.4`'s peer dep is still pinned to `^0.9.2`. Since `web_core` is now at `0.10.3` (latest), any project pulling in `@a2ui/react@^0.10` alongside `@a2ui/web_core@^0.10` will hit this on every install.

npm resolves it with an override so it doesn't hard-break installs, but it's noise in CI logs and it'll bite anyone running with `--legacy-peer-deps=false`.

**Versions:**
- `@a2ui/web_core`: `0.10.3` (latest)
- `@a2ui/react`: `0.10.1`
- `@a2ui/markdown-it`: `0.0.4`

**Expected:** peer dep range on `@a2ui/web_core` widened to `^0.9.2 || ^0.10` (or `>=0.9.2 <0.11`) in `@a2ui/markdown-it`.

Happy to send a PR if it helps — the fix should just be a one-line bump to `peerDependencies` in `renderers/markdown/markdown-it/package.json`.

eugenelim:
Fixes #1930.

`@a2ui/markdown-it@0.0.4` was published when `web_core` was at `0.9.2`, so `prepare-publish.mjs` baked in `^0.9.2` as the peer dep. Since then `web_core` shipped `0.10.x`, and any project that pulls in `@a2ui/react@^0.10` alongside `@a2ui/web_core@^0.10` gets an `ERESOLVE` override warning on every install:

```
npm warn ERESOLVE overriding peer dependency
npm warn peer @a2ui/web_core@"^0.9.2" from @a2ui/markdown-it@0.0.4
```

**The fix:** change `peerDependencies["@a2ui/web_core"]` from `workspace:*` (which `prepare-publish.mjs` resolves to `^<current-version>` at publish time) to an explicit `>=0.9.2`. This bypasses the workspace replacement and produces a wide range in the published package — no more peer conflict when `web_core` bumps again. Bumps to `0.0.5`.

The `devDependencies` entry is left as `workspace:*` so local builds continue to use the monorepo copy.

### DIFF
--- a/renderers/markdown/markdown-it/package.json
+++ b/renderers/markdown/markdown-it/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/markdown-it",
-  "version": "0.0.4",
+  "version": "0.0.5",
   "description": "A Markdown renderer using markdown-it and dompurify.",
   "repository": {
     "directory": "renderers/markdown/markdown-it",
@@ -61,7 +61,7 @@
     "wireit": "^0.15.0-pre.2"
   },
   "peerDependencies": {
-    "@a2ui/web_core": "workspace:*"
+    "@a2ui/web_core": ">=0.9.2"
   },
   "wireit": {
     "build": {


ditman:
Thanks for the contribution! I didn't notice this! We republished `markdown-it` so it now works with 0.10.something. We also made markdown-it 0.1.0 so we can roll these changes more easily without forcing users to take a completely new version all the time, following semver. Thanks for the contribution though, keep them coming!

PR #1938, opened 2026-07-07
Release TypeScript packages: @a2ui/web_core v0.10.4, @a2ui/angular v0.10.3, and @a2ui/markdown-it v0.1.0

### LINKED ISSUE #1766: @a2ui/markdown-it@0.0.4 peer dependency blocks @a2ui/web_core@0.10.1 install
We are upgrading our code to use:

- @a2ui/react@0.10.1
- @a2ui/web_core@0.10.1

However, npm install/ci fails because @a2ui/react@0.10.1 depends on:

- @a2ui/**web_core@^0.10.1**
- @a2ui/markdown-it@^0.0.4

But the latest available @a2ui/markdown-it@0.0.4 still declares this peer dependency:

- @a2ui/**web_core@^0.9.2**

This causes npm peer resolution to fail with ERESOLVE when running plain npm ci.

Can you please publish a compatible @a2ui/markdown-it version whose peer dependency allows @a2ui/web_core@^0.10.1?

Example expected peer dependency:

{
  "peerDependencies": {
    "@a2ui/web_core": "^0.10.1"
  }
}

Until that is available, we are temporarily using project-level legacy-peer-deps=true to unblock PR builds, but we would like to remove that workaround once the package metadata is fixed.

jacobsimionato:
Automated version bump and changelog preparation for `@a2ui/web_core` v0.10.4, `@a2ui/angular` v0.10.3, and `@a2ui/markdown-it` v0.1.0.

#### `@a2ui/markdown-it` (v0.1.0)
- Promotes `@a2ui/markdown-it` from `0.0.4` to `0.1.0`. By graduating out of `0.0.x`, future patch updates (`0.1.1+`) are treated as non-breaking maintenance releases in NPM/Yarn SemVer.
- Widens `@a2ui/web_core` peer dependency range from `workspace:*` to `^0.9.2 || ^0.10.0`. Fixes peer dependency conflicts (`ERESOLVE`) when installing alongside `@a2ui/web_core@0.10.x` while preventing compatibility issues with future major breaking releases (like v1.0.0).
- Fixes #1766 and #1930.

#### `@a2ui/web_core` (v0.10.4)
- Support JSON Pointer escaping (RFC 6901) in `DataModel` (#1796).

#### `@a2ui/angular` (v0.10.3)
- Export Angular test utilities under `@a2ui/angular/testing` and secondary entry point `@a2ui/angular/v0_9/testing` (#1737).

### Verification
- Ran `./renderers/scripts/increment_version.mjs` for `web_core`, `markdown-it`, and `angular` to update packages and dependent `yarn.lock` entries.
- Ran unit tests across `@a2ui/web_core` and `@a2ui/markdown-it`.
- Ran full production monorepo build verification via `yarn build:all`.


renderers/markdown/markdown-it/package.json
```diff
@@ -61,7 +61,7 @@
     "wireit": "^0.15.0-pre.2"
   },
   "peerDependencies": {
-    "@a2ui/web_core": "workspace:*"
+    "@a2ui/web_core": ">=0.9.2"
```
ditman: Please, revert this, we won't be able to use workspace dependencies to develop this package locally

renderers/markdown/markdown-it/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/markdown-it",
-  "version": "0.0.4",
+  "version": "0.0.5",
```
ditman: I was thinking of calling this `0.1.0`. That way we can do similar maintenance releases like this one without rolling a major version.

renderers/markdown/markdown-it/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Unreleased
```
ditman, line 1: Thanks for adding the Unreleased section to this package! It's old!

ditman (commented):
There's a small problem with our prepare to publish script that will cause it to substitute the version of web_core in markdown-it to the very latest.

I've been thinking that we may need to make the dependency resolution script be a little bit more lenient, for example resolving to the current major version of the package (so 0.10.0 instead of 0.10.4 in this case)

That way we can release within the current major version without needing to touch anything else. When we do a major release of the packages, we should verify that markdown-it continues working and release a patch version of it so it picks up the latest versions.

Not sure what dependency problems we can introduce by not pinning to the absolute latest, though. Maybe weirdness when building off the artifact registry?

jacobsimionato:
Thanks @ditman! Really appreciate the feedback and insights. Here is a breakdown of how we addressed your suggestions in this PR:

### 1. Promoting `@a2ui/markdown-it` to `v0.1.0` (Implemented)
We completely agree and have updated this PR to bump `@a2ui/markdown-it` to **`0.1.0`** instead of `0.0.5`.
- **Rationale**: In NPM/Yarn SemVer (`node-semver`), any patch bump on a `0.0.x` package is treated as a breaking major version (`^0.0.4` means `>=0.0.4 <0.0.5`). By graduating `@a2ui/markdown-it` to `0.1.0`, future maintenance updates (`0.1.1+`) will be non-breaking (`^0.1.0` means `>=0.1.0 <0.2.0`), allowing consumers to receive patch updates seamlessly.

### 2. How we handled peer dependency ranges & `prepare-publish.mjs` substitution
Regarding `prepare-publish.mjs` substituting `web_core` to `^0.10.4`:
- **Why it won't substitute here**: In `renderers/markdown/markdown-it/package.json`, we declared an explicit bounded SemVer string for `peerDependencies`:
  ```json
  "peerDependencies": {
    "@a2ui/web_core": "^0.9.2 || ^0.10.0"
  }
  ```
  Because `prepare-publish.mjs` only substitutes versions starting with `workspace:` or `file:`, it leaves explicit SemVer ranges untouched when generating `dist/package.json`.
- **Monorepo development preservation**: We retained `"devDependencies": { "@a2ui/web_core": "workspace:*" }`, which ensures local monorepo builds (`tsc -b`, `vitest`) continue to link directly against the local workspace copy.
- **Why we prefer explicit ranges over modifying `prepare-publish.mjs` for peer deps**: Explicit bounded ranges clearly communicate broad peer compatibility (`0.9.x` and `0.10.x`) without relying on build-time script transformations. That said, making `prepare-publish.mjs` more lenient for regular `workspace:^` dependencies (e.g. resolving to base minor `^0.10.0`) is a great idea to explore as an architectural follow-up!

renderers/markdown/markdown-it/package.json
jacobsimionato: My understanding is that it will still work, because the workspace config will still result in the local dependency being used for web core.

renderers/markdown/markdown-it/package.json
jacobsimionato: That sounds great. Done.

josemontespg (approved):

PR #1939, opened 2026-07-07
Angular Renderer Resilience & Reactivity Updates

vikramsanthalia:
This pull request fixes critical reliability bugs and change-detection rendering skips in the Angular Renderer (`ComponentHostComponent`):

1. **Fix subscription memory leak:** Resolves a pattern where the component-host subscribes to `onSurfaceCreated` and attempts to unsubscribe self-referentially within the synchronous emission path, which evaluates to `undefined` and leaks listeners.
2. **Asynchronous surface resolution:** In progressive rendering or streaming environments, component host nodes can be mounted in the DOM before the surface itself is registered. Added a subscription to `onSurfaceCreated` to automatically bootstrap the component when the surface registers.
3. **Optimize reactivity (Signals upgrade):** Upgraded `componentType` and `props` to Angular Signals. In strict `OnPush` applications, asynchronous updates inside RxJS callbacks were not recognized by the change detector, leaving components unrendered.

## Pre-launch Checklist

- [x] I signed the CLA https://cla.developers.google.com/.

### DIFF
--- a/renderers/angular/src/v0_9/core/component-host.component.ts
+++ b/renderers/angular/src/v0_9/core/component-host.component.ts
@@ -16,13 +16,14 @@
 
 import {
   ChangeDetectionStrategy,
-  ChangeDetectorRef,
   Component,
   DestroyRef,
   Type,
   inject,
   input,
   effect,
+  signal,
+  NgZone,
 } from '@angular/core';
 import {NgComponentOutlet} from '@angular/common';
 import {ComponentContext, ComponentModel, SurfaceModel, Subscription} from '@a2ui/web_core/v0_9';
@@ -48,12 +49,12 @@ import {BoundProperty} from './types';
     style: 'display: contents;',
   },
   template: `
-    @if (componentType) {
+    @if (componentType()) {
       <ng-container
         *ngComponentOutlet="
-          componentType;
+          componentType()!;
           inputs: {
-            'props': props,
+            'props': props(),
             'surfaceId': surfaceId(),
             'componentId': resolvedComponentId,
             'dataContextPath': resolvedDataContextPath,
@@ -74,10 +75,10 @@ export class ComponentHostComponent {
   private readonly rendererService = inject(A2uiRendererService);
   private readonly binder = inject(ComponentBinder);
   private readonly destroyRef = inject(DestroyRef);
-  private readonly cdr = inject(ChangeDetectorRef);
+  private readonly ngZone = inject(NgZone);
 
-  protected componentType: Type<unknown> | null = null;
-  protected props: Record<string, BoundProperty> = {};
+  protected readonly componentType = signal<Type<unknown> | null>(null);
+  protected readonly props = signal<Record<string, BoundProperty>>({});
   private context?: ComponentContext;
 
   protected resolvedComponentId: string = '';
@@ -85,17 +86,22 @@ export class ComponentHostComponent {
 
   private propsSub?: Subscription;
   private createSub?: Subscription;
+  private surfaceSub?: Subscription;
 
   constructor() {
     effect(() => {
       const key = this.componentKey();
       const surfaceId = this.surfaceId();
-      this.setupComponent(key, surfaceId);
+      if (key && surfaceId) {
+        this.resetState();
+        this.setupComponent(key, surfaceId);
+      }
     });
 
     this.destroyRef.onDestroy(() => {
       this.propsSub?.unsubscribe();
       this.createSub?.unsubscribe();
+      this.surfaceSub?.unsubscribe();
     });
   }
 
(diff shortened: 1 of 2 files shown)


renderers/angular/src/v0_9/core/component-host.component.ts
```diff
@@ -154,21 +178,28 @@ export class ComponentHostComponent {
...
+      this.cdr.detectChanges();
     });
 
     this.cdr.markForCheck();
+    this.cdr.detectChanges();
+    Promise.resolve().then(() => {
+      if (this.destroyed) return;
+      this.cdr.markForCheck();
+      this.cdr.detectChanges();
+    });
```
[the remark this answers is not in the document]
vikramsanthalia: Done. Refactored to inject NgZone and wrapped the asynchronous stream subscription updates in ngZone.run(), removing all manual change detection and Promise.resolve blocks.

renderers/angular/src/v0_9/core/component-host.component.ts
```diff
@@ -105,7 +111,28 @@ export class ComponentHostComponent {
     const surface = this.rendererService.surfaceGroup?.getSurface(surfaceId);
 
     if (!surface) {
-      console.warn(`Surface ${surfaceId} not found`);
+      console.warn(`Surface ${surfaceId} not found. Waiting for it...`);
+      this.surfaceSub?.unsubscribe();
+      let unsubscribed = false;
+      const sub = this.rendererService.surfaceGroup?.onSurfaceCreated?.subscribe(s => {
```
josemontespg, line 117: The subscription handling got a bit messy, how about something like this?

    if (!surface) {
      console.warn(`Surface ${surfaceId} not found. Waiting for it...`);
      this.surfaceSub?.unsubscribe();
      this.surfaceSub = this.rendererService.surfaceGroup?.onSurfaceCreated?.subscribe((s) => {
        if (s.id === surfaceId) {
          this.surfaceSub?.unsubscribe();
          this.surfaceSub = undefined;
          this.ngZone.run(() => {
            this.setupComponent(key, surfaceId);
          });
        }
      });
      return;
    }

renderers/angular/src/v0_9/core/component-host.component.ts
```diff
@@ -105,7 +111,28 @@ export class ComponentHostComponent {
...
+        if (s.id === surfaceId) {
+          unsubscribed = true;
+          if (this.surfaceSub) {
+            this.surfaceSub.unsubscribe();
+            this.surfaceSub = undefined;
+          }
+          this.ngZone.run(() => {
```
josemontespg, line 124: Please add a test that fails when this is not ran inside ngZone, ideally not just checking that it was called but actually checking that the bug you were experiencing no longer happens. This will help this fix to not be undone unintentionally by future code updates.

josemontespg (commented):
The changes make sense to improve reliability of Surface lookups during component tree creation. Left some comments about code style and testing

renderers/angular/src/v0_9/core/component-host.component.ts
vikramsanthalia: Thanks for the suggestion!

The reason we have the unsubscribed boolean flag tracker is to handle synchronous emission of onSurfaceCreated (which is common in mocked unit testing environments).

In JavaScript, if the subscription callback is triggered synchronously inside the .subscribe() call itself, the assignment this.surfaceSub = ... hasn't resolved yet during callback execution (it is still undefined). Thus, calling this.surfaceSub?.unsubscribe() inside the synchronous callback execution does nothing, and the subscription is permanently leaked after subscribe() returns and binds the reference.

The unsubscribed flag tracks whether the callback has already fired, and ensures we safely trigger sub.unsubscribe() right after the assignment returns, preventing the leak.

renderers/angular/src/v0_9/core/component-host.component.ts
vikramsanthalia: Added! I've introduced two unit tests in component-host.component.spec.ts that specifically verify that both setupComponent (on async surface creation) and the properties binder trigger inside NgZone.run().

To do this deterministically in a unit testing context (where TestBed uses NoopNgZone by default and does not enforce real zone transitions), we mock the async triggers and spy on the injected NgZone.run method to ensure it intercepts and delegates the execution. If we remove the ngZone.run() wrappers from the component, these specs fail consistently.

vikramsanthalia:
Hi @josemontespg,

I have resolved the conflicts by rebasing the branch on top of `main` (pulling in your Angular signals updates). 

I've also:
1. Replied to your review comments explaining the synchronous subscription emission edge case.
2. Added the requested unit tests in `component-host.component.spec.ts` that verify that both surface creation and properties updates run inside the Angular zone via `NgZone.run()`.

The PR checks are now green. Could you please take a look and merge the PR? Thanks!

renderers/angular/src/v0_9/core/component-host.component.ts
```diff
@@ -48,15 +49,15 @@ import {BoundProperty} from './types';
...
+    @if (componentType()) {
       <ng-container
         *ngComponentOutlet="
-          componentType;
+          componentType()!;
           inputs: {
-            'props': props,
-            'surfaceId': surfaceId(),
-            'componentId': resolvedComponentId,
-            'dataContextPath': resolvedDataContextPath,
```
josemontespg, line 59: We should keep the quotes, these are important for the Closure compiler with property renaming

josemontespg (approved):

renderers/angular/src/v0_9/core/component-host.component.ts
vikramsanthalia: Done. Restored the single quotes on the inputs keys in the component template to preserve compatibility with the Closure compiler's property renaming.

PR #1941, opened 2026-07-07
fix(parser): safely handle non-string inputs in has_a2ui_parts

janandrzejnowakowski:
**Summary**
Guards has_a2ui_parts against non-string inputs so that inspecting arbitrary tool/function responses (e.g., integers, dicts, booleans, or None) does not raise a TypeError.

**Root Cause & Error**
When an agent or tool emits a FunctionResponse part whose result is a non-string data type (such as an integer return value like {"result": 6}), passing that value into has_a2ui_parts(content) raises:

Traceback (most recent call last):
  File ".../a2ui_agent/src/a2ui/adk/a2a/part_converter.py", line 120, in convert
    if has_a2ui_parts(result):
  File ".../a2ui_agent/src/a2ui/parser/parser.py", line 30, in has_a2ui_parts
    return A2UI_OPEN_TAG in content and A2UI_CLOSE_TAG in content
TypeError: argument of type 'int' is not iterable

**Fix**
Updated has_a2ui_parts(content: Any) -> bool to return False early if content is not a string (if not isinstance(content, str): return False).
Added unit test coverage for string, integer, dictionary, and None inputs in test_parser.py.

jnowakowski@noj:~/src/A2UI/scripts$ ./e2e_test.sh
...
01:42 +4: All tests passed!


polina-c:
Thank you for your contribution. I am sorry it took so long for our team to review your PR. Unfortunately the branch has conflicts with the main branch.  If you believe the issue is not resolved yet, can you, please, resolve the conflicts? We will review it again after you resolve conflicts.

janandrzejnowakowski:
Hi - the problem is still there, so I resolved the merge conflicts. However, I think it does not fix all the problems now that there are multiple separate parsers that all seem to have the same problem. I would need to test / investigate before I am sure though.

agent_sdks/python/a2ui_agent/tests/parser/test_parser.py
nan-yu, line 1: We have the tests in the conformance test spec: https://github.com/a2ui-project/a2ui/blob/main/conformance/agent/parser.yaml#L115-L131. Would you mind adding missing test case there? Thanks!

PR #1942, opened 2026-07-07
feat(web_core): export UserAction as ClientEventUserAction in v0.8

josemontespg:
Exports the UserAction symbol as ClientEventUserAction from client-event.js in v0.8 web_core.

This is needed because there is a usage of this in 1P, and it makes sense to export this type.


sarashed731-prog (changes requested):
Debug

gspencergoog (approved):

josemontespg:
Thanks for the quick review!

renderers/web_core/src/v0_8/types/types.ts
```diff
@@ -17,6 +17,7 @@
 export {
   type ClientToServerMessage as A2UIClientEventMessage,
   type ClientCapabilitiesDynamic,
+  type UserAction as ClientEventUserAction,
```
ditman, line 20: Is this not a breaking change for current v0.8 users? We should also export UserAction without renaming for this to be backwards compatible.

ditman: Ah no, because UserAction was not exported previously. Sorry!

PR #1945, opened 2026-07-07
feat: align MCP app v0.9 bridge, specs, pong improvements, and data d…

sugoi-yuzuru:
# Description

## Overview
This pull request introduces the MCP Apps component specification for A2UI V0.9, and updates the McpApp component and the Pong game demo to the introduced A2UI v0.9 specification. The changes include adding the component specification document, implementing host-guest message routing, updating the Pong application logic, and preventing cyclic updates.

See the updated Pong Demo here https://screencast.googleplex.com/cast/NjE3MDM1NzE3MzA1OTU4NHxkYTRhMjhjNy01Yw

## Key changes

 - Added specification: Added `mcp-apps-component_spec.md` to define security boundaries, message schemas, and catalog properties for the McpApp component.
 - Component properties: Updated McpAppSchema in catalog.ts to replace the older `content` field with `htmlContent` (which was conflicting with a reserved property name in A2UI v0.9), and added `allowedFunctions` and data properties.
 - Double-iframe host logic: Updated `mcp-app.ts` to process sizing requests, handle client-side function invocations, and route data model updates.
 - Deduplication: Added structural equality validation using stringified JSON comparisons to break infinite state update loops between host and the embedded app.

 - Scoreboard commentary: Updated pong-scoreboard.ts to display generated game commentary to demonstrate the Allowed Tool Calls from the embedded MCP App. 
 - Score tracking: Modified pong_app.html to run score tracking using the A2UI V0.9's Tow-way data binding feature (instead of ToolCall routing we did in A2UI V0.8). 
 - Game winner: Implemented a UI overlay that announces the winner of the game in the A2UI catalog to demonstrate the Allowed Function Calls from the embedded MCP App. This is integration with the new FunctionCall feature on A2UI V0.9.

The sections below describe how the specifications are demonstrated in the Pong game demo.

*Dynamic resizing*
The specification allows sandboxed applications to request window dimension changes. The host implements layout constraints including clamping, throttling, and change thresholds.

In the demo, once 
pong_app.html
 finishes initializing and completes the handshake, it transmits a ui/notifications/size-change request to resize the wrapper to 728 by 502 pixels. The host component receives this and resizes the iframe and parent container styles. This keeps the canvas fully visible without displaying scrollbars.

*Two-way local data binding*
The specification defines data synchronization between the sandboxed application and the host data model. This path is configured via data.path in the catalog.

In the demo, this is shown in two ways:

 - Upward state updates; When a player or the CPU scores,  pong_app.html updates its local score values and posts a ui/notifications/data-model-change message containing the new scores. The host updates /pong_state on the client, which updates the native pong-scoreboard.ts component. Because this update occurs client-side, score displays update immediately without remote agent roundtrips.
 - Downward state updates; When a match ends, the user can click "Play Again" in the victory modal. This triggers host-side logic that resets /pong_state scores to zero. The host detects this change and pushes a ui/notifications/data-model-update notification to the iframe. The game script processes this update to clear its internal score variables, reset the ball, and resume play.

*Infinite event loop prevention*
Both the host component and the game script check incoming values against the last processed state. If the stringified JSON matches, the update is dropped. This prevents the host and guest from continuously re-triggering updates.

*Local client-side function execution*
The specification allows the sandboxed application to trigger pre-registered host functions if permitted by allowedFunctions.

In the demo, when a score reaches three points, the game script calls the host function showWinnerModal. The host validates permission and evaluates the function. The function creates and displays an HTML dialog stating the winner. Clicking the restart button inside the modal updates the host data model to reset the scores.

*Tool call routing*
The specification permits sandboxed applications to request remote tool executions defined by the agent.

In the demo, whenever a point is scored or the match restarts, the game script requests the commentate_pong tool. The host forwards this action to the agent backend. The agent calls the Gemini model to produce a short comment under 15 words. The agent then writes the comment back to /pong_state/commentary, which is updated on the scoreboard.

## Pre-launch Checklist

One time:

- [ ] I signed the [CLA].

For this PR:

[CHANGELOG]: ../CHANGELOG.md


samples/community/agent/adk/mcp_app_proxy/pong_app.html
```diff
@@ -246,11 +248,29 @@
...
+              localPlayerScore = value.player_score;
+            }
+            if (typeof value.cpu_score === 'number') {
+              localCpuScore = value.cpu_score;
+            }
+            // Automatically restart if scores reset to 0
+            if (localPlayerScore === 0 && localCpuScore === 0) {
+              isPaused = false;
+              hideOverlay();
+              resetBall();
```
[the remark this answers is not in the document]
sugoi-yuzuru: Added type checking.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/catalog.ts
```diff
@@ -28,27 +33,105 @@ import {Column} from '@a2ui/angular';
...
+    btn.style.fontSize = '1rem';
+    btn.style.boxShadow = `0 0 15px ${
+      args.winner === 'player' ? 'rgba(0, 242, 255, 0.4)' : 'rgba(255, 0, 255, 0.4)'
+    }`;
+
+    btn.onclick = () => {
+      dialog.close();
+      dialog.remove();
+      // Reset scores in A2UI Data Model
+      context.surface.dataModel.set('/pong_state', {player_score: 0, cpu_score: 0});
```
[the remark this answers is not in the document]
sugoi-yuzuru: Updated.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-app.ts
```diff
@@ -129,6 +140,49 @@ export class McpApp extends CatalogComponent<any> implements OnDestroy, OnInit {
...
+          : 100;
+      const heightDiff =
+        targetHeight !== undefined && this.lastHeight !== undefined
+          ? Math.abs(targetHeight - this.lastHeight)
+          : 100;
+
+      if (targetWidth !== undefined && widthDiff >= 5) {
+        iframeEl.style.width = `${targetWidth}px`;
+        this.lastWidth = targetWidth;
+      }
```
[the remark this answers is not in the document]
sugoi-yuzuru: Done.

samples/community/agent/adk/mcp_app_proxy/tools.py
```diff
@@ -216,10 +186,57 @@ async def get_pong_app_a2ui_json(tool_context: ToolContext):
...
+
+
+async def commentate_pong_game(tool_context: ToolContext, game_event: str):
+  """Generates a witty neon-themed sports commentary or lighthearted trash talk comment based on the game event description, and applies it to the game scoreboard.
+
+  Args:
+      game_event: The description of the game event (e.g. 'Score: Player 1 - CPU 0 (player scored).').
+  """
+  import os
+  from google.genai import Client
```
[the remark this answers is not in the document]
sugoi-yuzuru: Removed.

samples/community/agent/adk/mcp_app_proxy/tools.py
```diff
@@ -216,10 +186,57 @@ async def get_pong_app_a2ui_json(tool_context: ToolContext):
...
+  
+  Keep it strictly under 15 words. Be punchy, energetic, and match the "neon arcade" retro theme!
+  """
+  
+  response = client.models.generate_content(
+      model=gemini_model,
+      contents=prompt,
+  )
+  
+  commentary_text = response.text.strip()
```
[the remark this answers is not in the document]
sugoi-yuzuru: Done.

suyangw-g (approved):

samples/community/agent/adk/mcp_app_proxy/catalogs/0.9/mcp_app_catalog.json
```diff
@@ -59,9 +59,21 @@
...
+            },
+            "allowedFunctions": {
+              "type": "array",
+              "description": "List of local function names the app is allowed to call.",
+              "items": {
+                "type": "string"
+              }
+            },
+            "data": {
+              "$ref": "common_types.json#/$defs/DataBinding",
```
jacobsimionato: Elsewhere, we use DynamicString etc for two-way bindings. I actually don't know what happens if you pass a constant value to it at runtime, but we should just define it as making mutations a no-op.

But can you use `DynamicValue` here to be consistent?

Other background:
- We're trying to limit what types from common_types a catalog JSON file can reference, to restrict the catalog format, so that we can express catalogs in different ways beyond JSON schema, e.g through different DSLs that have more performant inference.
- In some cases, using the data model adds unnecessary complexity to expressing a layout, so the ability to pass some constant values to an MCP app via a literal values in `UpdateComponents` might be handy.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
...
+    Note over App: Initializes MCP App Bridge
+```
+
+# 3. The MCP app bridge runtime and communication contract
+
+All communications between the host, the sandbox proxy, and the embedded app use structured JSON-RPC 2.0 messages defined by the `@modelcontextprotocol/ext-apps/app-bridge` specification.
+
+## 3.1. Handshake lifecycle
+
+To establish communication without race conditions, the following lifecycle is followed:
```
jacobsimionato: I believe this is restating the MCP Apps handshake, right? It might be handy to flag that, so the reader can figure out what this is adding on top of the existing MCP Apps specification.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
...
+    }
+  },
+  "id": "string-or-number"
+}
+```
+
+**Host action**  
+The host checks if the requested tool name is in the component's `allowedTools` list. If it is authorized, the host dispatches the tool call as an A2UI action to the agent backend. If not, it rejects the tool call and returns a JSON-RPC error response.
+
+### B. Reactive state synchronization (`ui/notifications/data-model-change`)
```
jacobsimionato, line 150: I wonder if this should include a subpath param. There is only one data model binding key, which is used to propagate state from host -> app and app -> host, so if both are overwriting the entire state, I think race conditions and clobbering could occur seeing as there is an async jump over the sandbox boundary.

E.g. 
1. Establish MCP calculator app bound to `/calculator_state`
2. Server sends `UpdateDataModel` message to update `/calculator_state/mode` to `scientific`
3. Calculator responds to button press, sending `data-model-change` message to update `/calculator_state/expression_string` to `3`. However, it actually has to copy the entire existing content of `/calculator_state` to update one value, e.g. it is setting it to `{"mode": "basic", "expression_string": "3"}`.
4. Host receives server update, updating `/calculator_state/mode` to `scientific`. It sends `ui/notifications/data-model-update` to calculator with the updated state.
5. Host receives `data-model-change` setting `/calculator_state` back to `{"mode": "basic", "expression_string": "3"}`, clobbering its state
6. Calculator receives `ui/notifications/data-model-update` with `{"mode": "scientific", "expression_string": ""}`, clobbering its state

I dunno, WDYT? Maybe I'm nit picking! Another option would be to have two params - one intended for host->app state and another for app->host.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
...
+       "permissions": ["string"]
+     }
+   }
+   ```
+3. **Sandbox initialization (`sandbox-init`):** The proxy loads the inner iframe with the application HTML and dispatches a `sandbox-init` message to the inner frame to signal that the connection is active.
+
+## 3.2. Outgoing messages (Embedded app to host)
+
+### A. Tool call execution (`tools/call`)
+
```
jacobsimionato, line 128: This is also standard, right?

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
...
+    "arguments": {
+      "key": "value"
+    }
+  },
+  "id": "string-or-number"
+}
+```
+
+**Host action**  
+The host checks if the requested tool name is in the component's `allowedTools` list. If it is authorized, the host dispatches the tool call as an A2UI action to the agent backend. If not, it rejects the tool call and returns a JSON-RPC error response.
```
jacobsimionato, line 148: (sorry if I asked this before) is this necessary given that these RPCs will be forwarded to the host anyway, and it can always reject them there?

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
jacobsimionato, line 150: I think this and function-call are where you start proposing extensions to the core MCP App protocol. It'd be great to make it clear what the extensions are in contrast to what is standard.

I think your design does an awesome job of reusing MCP apps protocol and extending it in an elegant way that preserves the spirit of both protocols.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
...
+    "width": "number",
+    "height": "number"
+  }
+}
+```
+
+**Host action**  
+The host updates the dimensions of the wrapper element to match the requested size.
+
+### E. Logging notification (`ui/notifications/logging`)
```
jacobsimionato: This isn't in MCP app AFAICT and it's kind of orthogonal to A2UI. Do you think we need it here? Perhaps this ends up being a separate extension to MCP apps?

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
...
+    }
+  },
+  "id": "string-or-number"
+}
+```
+
+**Host action**  
+The host checks if the target function is listed in the component's `allowedFunctions` list. If verified, it evaluates the function using the A2UI client catalog engine and returns the result (or error) to the app.
+
+### D. Frame resize request (`ui/notifications/size-change`)
```
jacobsimionato: This is directly from MCP apps, right? I wonder if for these parts, you could even just list that they are supported, and direct to the [MCP apps specification doc](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx) as the source of truth for what they mean.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
...
+# 5. Rendering setup and security controls
+
+Running untrusted code requires isolation controls to prevent data access and sandbox escape.
+
+## 5.1. Content decoding
+
+The host receives the `htmlContent` property from the catalog definition:
+
+1. It extracts the raw HTML string.
+2. If the string starts with `url_encoded:`, it decodes the content using `decodeURIComponent`.
```
jacobsimionato, line 382: Is this a standard approach? It seems a bit magical and confusing to me. Could we just support either always URL encoded, or never URL encoded? Or even have alternative input properties for each option (though that would require them to be optional I know).

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
...
+```html
+<meta
+  http-equiv="Content-Security-Policy"
+  content="default-src 'self' 'unsafe-inline' 'unsafe-eval' data:; connect-src 'none';"
+/>
+```
+
+This blocks connection protocols like fetch, XHR, WebSockets, and Server-Sent Events, forcing all communication to be routed through the `AppBridge` channel to the host.
+
+### Security self-test
```
jacobsimionato: Is this a standard approach suggested by ISE etc? I worry about it creating false confidence unless it's a practice that has been shown by security experts be reliable.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
...
+
+1. **Clamping:** Height must be clamped between 100px and 2000px; width must be clamped between 200px and 3000px.
+2. **Throttling:** Consecutive size changes must be throttled to a maximum of one redraw per 100 milliseconds.
+3. **Threshold Gate:** Size changes of less than 5 pixels are ignored.
+
+# 6. Double-iframe proxy blueprint
+
+This section provides the reference implementation for the sandbox proxy page (`sandbox.html` and `sandbox.ts`).
+
+## 6.1. sandbox.html
```
jacobsimionato: Would it be possible to reuse the MCP apps SDK for this (and the TS below), and add our own small library on top of it?

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
```diff
@@ -0,0 +1,500 @@
+# A2UI MCP App Component Specification (v0.9)
```
jacobsimionato, line 1: So fantastic to have a spec for this!! Thank you for adding it!

jacobsimionato (commented):
Hey this looks great to me in general - most of my comments are about the way it's expressed and more subtle technical details.

samples/community/agent/adk/mcp_app_proxy/catalogs/0.9/mcp_app_catalog.json
sugoi-yuzuru: Makes sense. Updated to `DynamicValue`. Thanks!

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: Good eye!!

I have revised this place to clarify that we are aligning with MCP Apps specification, and also extended the discussion to the actual 2 phased handshake (first for the outer shell and the next for the embedded app), to better clarify what is expected to run here.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: Yes. Added a link to call out the alignment with standard.

I'll see if I can make appendix and move all MCP Standard entries to the bottom with references.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: It mostly depends on how much trust-worthy the embedded app is. 

From the back-end's POV, the request will be coming from the host application and will assume the source is trust-worthy. However, if the request is originating from the embedded app, there may be some calls that the backend might not want to blindly execute (e.g., `deleteAccount` or `readFileFromLocal` etc)

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: I have updated to use `notifications/message` which is the latest MCP Apps standard. 

https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: Yes. I have updated with link.  https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#container-dimensions

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: So this is actually tricky. 

MCP Apps naturally are served from MCP server. As a result, native MCP Apps are fetched from a client directly from a MCP server's resource endpoint (or Tool endpoint). This means the raw HTML doesn't go through LLM's context which is great. 

On the other hand, when MCP Apps are inserted within an A2UI payload, they can be included in the Agent's LLM dialog. In such case, the raw HTML is terrible due to the HTML tags and excessive new lines. URL encoding helps make them into a single "mystery string" to maintain its composure. 

I may be able to study a bit more from Liad's change and more examples about MCP Apps on how I might be able to get the HTML blob out-of-LLM-context, but as of now, this was the solution I had in place.

sugoi-yuzuru: Assuming we want to make A2UI as simple as Client UI <-> A2UI Agent, I need to figure out how I might be able to make MCP Apps A2UI component to asynchronously fetch MCP App resource from A2UI Agent without contaminating the LLM Context. 

I will work on this  as a follow up since it is pretty critical.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: Good point. I have omitted this section for now.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
jacobsimionato: Hey I agree this is a critical problem!

The architecture I was actually hoping to get to with A2UI is to have a transformer between the actual persisted message context, and any LLM inferences, so that we can store the HTML string in its most basic form, but then omit context as necessary to hide it from the LLM, without needing to introduce a more complex RPC protocol with async fetching etc.

There are multiple reasons we might want to transform A2UI content between the message history and LLM:
- To expose A2UI via a more token efficient format - see https://docs.google.com/document/d/1mH2DJ_jZZA2vZ9EQlWaU4v8_xCAcmqrox_LlElmJ3ws/edit?tab=t.0
- For surfaces that have been updated incrementally, to flatten them to a snapshot of current state
- To include ephemeral state of surfaces pulled from the client
- To omit aspects of the A2UI content which are not relevant to future inferences, to save tokens. E.g. protocol details like catalog ID, surface ID, component IDs even (perhaps we could use a nested format for reading).

Seeing as we need all this translation stuff anyway, maybe we can have a concept of "plugins" that allow you to add additional catalog-specific transformations, or they can even be included as part of the catalog implementation, or annotated on the catalog API in the future "hiddenFromHistory = true" etc.

WDYT?

sugoi-yuzuru: I love the idea of having an API to omit certain parts of A2UI payload from the context history, and very happy to onboard on that!

I think there are ways I can leverage architectures learned from MCP to mitigate the problem for MCP Apps and IFrame, but would be great to have a solution overarching the A2UI as a whole :) 

That being said, is there something I can add to the current PR? Might I add a Note or a comment on forthcoming changes that might influence this specification?

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: I realized that this implementation is not very good as a "Spec" doc. I have updated to point to the MCP App's repo to reference the recommended sandbox implementation and specs so that the point of reference is consolidated to the source of truth.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: Yay!

It is a working progress, but hope it reduces friction for other frameworks to build MCP Apps component that can offer MCP App developers frictionless integration with A2UI agent across different frameworks.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/mcp-apps-component_spec.md
sugoi-yuzuru: I didn't think about that so thank you for pointing me to this. 

I have extended to accept a more surgical data update by targeting a `subpath` for the data.
It adds a bit of complexity to the deep-equality checks for preventing loop, but this is contained within the MCP App component. 

I have also added a toggle switch to prevent echoing when the data updates from App -> Host, and the Host -> App broadcast is auto-triggered, that the event (downstream) propagation is terminated if the data update was originated form the App. 

See https://github.com/a2ui-project/a2ui/pull/1945/changes/cc9b4f629f0780566965e971a1fbdbcabd3bc10d commit.

PR #1948, opened 2026-07-07
chore(github): add standard issue templates for bug reports, features, and docs

gspencergoog:
# Description

This PR introduces standard GitHub issue templates for:
- Bug Reports: including steps to reproduce, expected behavior, screenshots/logs, and environment details.
- Feature Requests: describing the problem, proposed solution, and alternatives considered.
- Documentation Improvements: detailing the problem page, describing the issue, and suggesting changes.

All templates have been formatted with Prettier.


eval/datasets/prompts.yaml
ditman, line 1: Unrelated change, please revert!

ditman (changes requested):
I don't think we want to remove the prompts.yaml file on this change!

eval/datasets/prompts.yaml
gspencergoog: Okay, I'll move it to another PR, but it's only a formatting change to make the CI pass the format check.

gspencergoog: Looks like Jia Hao already fixed the formatting at HEAD.  Reverted.

gspencergoog: (it also wasn't removing the file, just reformatting)

ditman: You're right! I saw a lot of red and thought it was being removed (x_x). Thanks for reverting that!

ditman (approved):

PR #1949, opened 2026-07-07
fix(samples): make a2ui-in-mcpapps sample MCP Apps spec compliant

liady:
# Description

The `a2ui-in-mcpapps` sample used a spec-inspired dialect of the [MCP Apps protocol (SEP-1865, 2026-01-26)](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx) rather than the spec itself, so it could not interoperate with spec-compliant apps or hosts. This PR fixes the host↔View wire protocol and the template discovery flow:

1. **Standard `tools/call` messages** — the apps invented `ui/<toolName>` methods (squatting on the reserved `ui/` protocol namespace) and the host relayed any unknown `ui/*` verb as a tool call, returning only the bare content array. Views now send standard `{method: "tools/call", params: {name, arguments}}`; the host validates `params.name` against its allowlist, returns the full `CallToolResult` as the JSON-RPC `result`, and sends valid JSON-RPC errors with a `code` field.
2. **Spec-compliant lifecycle** — apps send `ui/initialize` with `protocolVersion`, `clientInfo`, and `appCapabilities`, and emit `ui/notifications/initialized`; the host returns a full `McpUiInitializeResult` (previously the non-spec `hostCapabilities.displayModes`) and, gated on the `initialized` notification, delivers `ui/notifications/tool-input` and `ui/notifications/tool-result` for the instantiating tool call. `ui/ping` becomes standard `ping`.
3. **`ui/notifications/size-changed`** — replaces the non-spec `ui/resize`, reporting both `width` and `height`.
4. **Predeclared `ui://` templates via `_meta.ui`** — the entry tools previously delivered the `ui://` HTML template as an embedded resource in their results, a pattern the spec explicitly rejects (templates must be predeclared and fetched with `resources/read`). The server now declares `_meta.ui.resourceUri` on `get_basic_app`/`get_editor_app`, marks View-callable tools with `_meta.ui.visibility: ["app"]`, returns regular `CallToolResult`s from the entry tools (`get_basic_app` returns the initial counter payload), and serves `resources/read` contents with the required `text/html;profile=mcp-app` mime type. The host discovers the template from `tools/list` `_meta.ui.resourceUri` and derives its View tool allowlist from the declared visibility instead of hardcoding both, and the basic app renders the instantiating tool's result delivered via `ui/notifications/tool-result`.

[...]

PR #1953, opened 2026-07-08
feat(samples): add generic A2UI MCP App renderer in React

liady:
## What

Adds a third micro-app to the `a2ui-in-mcpapps` sample: `server/apps/react/`, a **generic, server-agnostic A2UI renderer** for MCP Apps, selectable in the sample host as "Generic React Renderer".

Unlike the Basic and Editor apps, it contains **zero server-specific logic**. It is pure chrome: the MCP Apps handshake comes from the official `@modelcontextprotocol/ext-apps` SDK (`App` class, including automatic `size-changed` iframe resizing), rendering is `@a2ui/react` (v0.9 basic catalog) driven by a `@a2ui/web_core` `MessageProcessor`, and all content arrives through tool results. Any A2UI-speaking MCP server can serve the built `react.html` as its own `ui://` resource if it follows two conventions:

1. **A2UI payloads travel as marked embedded resources** — tool results (including the entry tool's result, which produces the initial render) carry A2UI messages as `EmbeddedResource` content blocks with mimeType `application/a2ui+json`, each marked with `_meta: {"ui": {"content": {}}}`, and the `ui://` resource serving the renderer declares `_meta.ui.contentMimeTypes: ["application/a2ui+json"]` (on `resources/list` and `resources/read`). This is the MCP Apps [Dynamic View Content](https://github.com/modelcontextprotocol/ext-apps/pull/699) contract: the marker tells hosts to forward the block to the View unmodified and keep it out of model context; the declaration is the reviewable contract of what the View can parse; routing stays on the tool's `_meta.ui.resourceUri`. The renderer still accepts unmarked A2UI blocks as a legacy allowance for servers that predate the marker (`extractA2uiMessages(content, {allowUnmarked: false})` turns it off).
2. **A2UI actions map to tools** — each action `name` matches an app-visible tool name (`_meta.ui.visibility` includes `"app"`), and the action's resolved `context` becomes the tool `arguments`; the response payload is applied incrementally to the live surfaces. (The mapping is overridable: the `GenericA2uiApp` component takes an optional `actionToToolName` prop for servers whose action vocabulary differs from their tool names.)

The sample host negotiates dynamic content by advertising `contentMimeTypes: ["application/a2ui+json"]` (alongside `mimeTypes`) in the `io.modelcontextprotocol/ui` extension capability at `initialize`.

This is intended as the seed of an extractable/reusable artifact: the renderer is a standalone React component (`GenericA2uiApp` in `src/generic-a2ui-app.tsx`, with `main.tsx` only mounting it), so it can be reused either as the prebuilt `react.html` or as a component.

## Changes

- **`server/apps/react/`** — the renderer: `GenericA2uiApp` component + `extract-a2ui-messages.ts` (unit-tested with vitest), built with Vite + `vite-plugin-singlefile` into a self-contained `public/react.html`. No `inline.js` post-processing needed (React has no forced chunking, unlike the Angular apps).
- **`server/`** — v0.9-vocabulary counter payload (`simple_counter_a2ui_v0_9.json`), `ui://react/app` resource, `get_react_app` entry tool (declares the template via `_meta.ui.resourceUri`) and `increase_counter_v0_9` app tool. All A2UI payload blocks are built by an `a2ui_content_block()` helper that sets the `_meta.ui.content` marker, and the `ui://` resources declare `contentMimeTypes` in both `resources/list` and `resources/read`.
- **`client/`** — third option in the app selector; entry-tool lookup becomes a map; null-guard on the relay target when delivering tool input/result; advertises the `io.modelcontextprotocol/ui` extension capability with `mimeTypes` + `contentMimeTypes`. Everything else (tool allowlist, resource URI) already derives from `_meta.ui`.
- **`samples/community/package.json` + `yarn.lock`** — the new app is picked up by the existing `mcp/a2ui-in-mcpapps/server/apps/*` workspace glob; added to `build:web` and the lockfile.
- **READMEs** — sample README documents the two conventions and build "Option C"; `server/apps/README.md` and `server/apps/react/README.md` cover the React build and the component's props.

## Noteworthy implementation detail

`vite.config.ts` sets `resolve.dedupe` for `@a2ui/web_core` / `@preact/signals-core` (and friends). With duplicated copies in `node_modules` (hoisted vs nested), the bundle gets two signal-library instances and data-model updates stop propagating to rendered components — the UI renders once and silently never updates. Worth knowing for anyone who copies this app as a template.

## Testing

- Unit tests for the payload extraction (`yarn test`, vitest, 16 cases: single/array payloads, legacy mime type, non-object JSON, invalid JSON, non-resource blocks, marked vs unmarked `_meta.ui.content` selection, `isViewContentBlock`).
- Verified end-to-end in the sample host (Angular client + Python SSE server): selecting "Generic React Renderer" completes the `ui/initialize` handshake via the ext-apps SDK; the entry tool's embedded v0.9 payload renders the counter card; clicking "Increase counter" relays `tools/call increase_counter_v0_9` through the host and the returned `updateDataModel` patches the counter in place (no surface rebuild).
- Regression: the Basic (v0.8) and Editor apps still load and work.
- `tsc --noEmit` clean; `yarn build:all` emits a single-file `react.html` with no external references.

Builds on #1949 (merged) for the `_meta.ui` template declaration, the standard `tools/call` relay, and the host fix for JSON-RPC id-0 responses (the ext-apps SDK numbers requests from 0).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sugoi-yuzuru (approved):

josemontespg:
Thank you for your contribution. I am sorry it took so long for our team to review your PR. The a2ui team is working on implementing an mcp apps component catalog for cross framework support of mcp apps. This will be an optional catalog independent of the React renderer.

You can follow https://github.com/a2ui-project/a2ui/issues/691 for status updates

PR #1962, opened 2026-07-08
chore(ci): revert #1938. Release just markdown-it v0.1.0.

### LINKED ISSUE #1930: @a2ui/markdown-it peer dep on web_core is pinned to ^0.9.2, conflicts with web_core 0.10.x

josemontespg:
Reverts https://github.com/a2ui-project/a2ui/pull/1938, and prepares `markdown-it` for release so it is compatible with the current version of web_core (^0.10.3), and not the next one.

* Fixes: https://github.com/a2ui-project/a2ui/issues/1930

### DIFF
--- a/renderers/angular/CHANGELOG.md
+++ b/renderers/angular/CHANGELOG.md
@@ -1,7 +1,5 @@
 ## Unreleased
 
-## 0.10.3
-
 - (v0_9) Export Angular test utilities under `@a2ui/angular/testing` and secondary entry point `@a2ui/angular/v0_9/testing`. [#1737](https://github.com/a2ui-project/a2ui/pull/1737)
 
 ## 0.10.2
--- a/renderers/angular/package.json
+++ b/renderers/angular/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@a2ui/angular",
-  "version": "0.10.3",
+  "version": "0.10.2",
   "license": "Apache-2.0",
   "homepage": "https://a2ui.org/",
   "repository": {
--- a/renderers/markdown/markdown-it/CHANGELOG.md
+++ b/renderers/markdown/markdown-it/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+## 0.1.0
+
+- Bump dependencies so the package is compatible with `@a2ui/web_core: ^0.10.3`.
+
 ## 0.0.2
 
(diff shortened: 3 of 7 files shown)


ditman:
This PR fixes https://github.com/a2ui-project/a2ui/issues/1930 because the actual dependency on web_core is created when *releasing* the package by the [`updateInternalDeps` method of the `prepare-publish.mjs` build script](https://github.com/a2ui-project/a2ui/blob/main/renderers/scripts/prepare-publish.mjs#L53-L61).

Once we publish this version of markdown-it, its peer depenency on web_core will be updated to the version of the workspace (and npm): ^0.10.3.

Bumping to v0.1.0 also allows us to do this update behind the scenes later without users having to update their dependencies on `@a2ui/markdown-it: ^0.1.0` (at least until we need to do a truly breaking version of the package, which would follow semver version rules).

ditman (approved):
Let's try this!

PR #1973, opened 2026-07-09
fix(v0_9): resolve v0.9.1 payload compatibility gap (Issue #1749)

### LINKED ISSUE #1749: a2ui/react: v0.9.1 payload compatibility gap in v0_9 runtime
While checking v0.9.1 compatibility in `@a2ui/react@0.10.1`, I noticed a gap between the v0.9.1 spec and the current published runtime/types.

The v0.9.1 evolution guide says v0.9.1 is compatible with v0.9 payloads, and that schemas accept both `"v0.9"` and `"v0.9.1"`.

However, in the current published packages:

- `@a2ui/react@0.10.1` exposes `@a2ui/react/v0_9`, but no `v0_9_1` renderer export.
- `@a2ui/web_core@0.10.1` exposes `@a2ui/web_core/v0_9`, but no `v0_9_1` runtime export.
- The v0_9 server-to-client TypeScript types still narrow message `version` to `"v0.9"` only.
- `MessageProcessor` resolves `createSurface.catalogId` by exact catalog ID match.
- `@a2ui/react@0.10.1` uses the v0.9 Basic catalog ID:
  `https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json`
  but does not appear to also register the v0.9.1 Basic catalog ID:
  `https://a2ui.org/specification/v0_9_1/catalogs/basic/catalog.json`.

As a result, it is unclear whether clients are expected to pass v0.9.1 payloads through the existing v0_9 runtime, or whether there should be official v0.9.1-compatible exports/types/catalog IDs.

Could the expected support path for v0.9.1 payloads in `@a2ui/react` be clarified?

Varun-S10:
# Description

This PR adds support for **v0.9.1** protocol payloads in the `v0_9` runtime across `web_core`, `react`, `lit`, and `angular`. Previously, the runtime rejected v0.9.1 messages because of hard-coded version strings and missing catalog aliases. 

What changed:
- **Schemas & Types**: Updated Zod validation and TypeScript types from `"v0.9"` to `"v0.9" | "v0.9.1"`.
- **Catalog Resolution**: Added an alias check in `MessageProcessor` so surfaces referencing the v0.9.1 basic catalog URL resolve correctly.
- **Package Exports**: Added `./v0_9_1` package subpath exports to `web_core`, `react`, and `lit`.

This change is 100% backward compatible and does not affect any existing v0.9 features.

Fixes #1749 

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].

For this PR:

[CHANGELOG]: ../CHANGELOG.md

### DIFF
--- a/renderers/angular/src/v0_9/catalog/basic/basic-catalog.ts
+++ b/renderers/angular/src/v0_9/catalog/basic/basic-catalog.ts
@@ -96,6 +96,11 @@ export interface BasicCatalogOptions {
   id?: string;
 
   /**
+   * An optional array of alias URLs for the catalog.
+   */
+  aliases?: readonly string[];
+
+  /**
    * An optional locale to configure catalog-level formatting.
    */
   locale?: string;
@@ -148,7 +153,9 @@ export class BasicCatalogBase extends AngularCatalog {
       ...(options.extraComponents ?? []),
     ];
 
-    super(id, components, functions);
+    const aliases =
+      options.aliases ?? ['https://a2ui.org/specification/v0_9_1/catalogs/basic/catalog.json'];
+    super(id, components, functions, undefined, aliases);
   }
 }
 
--- a/renderers/angular/src/v0_9/catalog/basic/basic-catalog.ts
+++ b/renderers/angular/src/v0_9/catalog/basic/basic-catalog.ts
@@ -153,8 +153,9 @@ export class BasicCatalogBase extends AngularCatalog {
       ...(options.extraComponents ?? []),
     ];
 
-    const aliases =
-      options.aliases ?? ['https://a2ui.org/specification/v0_9_1/catalogs/basic/catalog.json'];
+    const aliases = options.aliases ?? [
+      'https://a2ui.org/specification/v0_9_1/catalogs/basic/catalog.json',
+    ];
     super(id, components, functions, undefined, aliases);
   }
 }
--- a/renderers/react/package.json
+++ b/renderers/react/package.json
@@ -55,6 +55,16 @@
         "default": "./dist/v0_9/index.cjs"
       }
     },
+    "./v0_9_1": {
+      "import": {
+        "types": "./dist/v0_9/index.d.ts",
+        "default": "./dist/v0_9/index.js"
+      },
+      "require": {
+        "types": "./dist/v0_9/index.d.cts",
+        "default": "./dist/v0_9/index.cjs"
+      }
+    },
     "./styles": {
       "import": {
         "types": "./dist/styles/index.d.ts",
--- a/renderers/web_core/package.json
+++ b/renderers/web_core/package.json
@@ -38,6 +38,14 @@
       "types": "./dist/src/v0_9/basic_catalog/index.d.ts",
       "default": "./dist/src/v0_9/basic_catalog/index.js"
     },
+    "./v0_9_1": {
+      "types": "./dist/src/v0_9/index.d.ts",
+      "default": "./dist/src/v0_9/index.js"
+    },
+    "./v0_9_1/basic_catalog": {
+      "types": "./dist/src/v0_9/basic_catalog/index.d.ts",
+      "default": "./dist/src/v0_9/basic_catalog/index.js"
+    },
     "./data/*": {
       "types": "./dist/src/v0_8/data/*.d.ts",
       "default": "./dist/src/v0_8/data/*.js"
@@ -80,6 +88,8 @@
         "../../specification/v0_8/json/*.json",
         "../../specification/v0_9/json/*.json",
         "../../specification/v0_9/catalogs/**/*.json",
+        "../../specification/v0_9_1/json/*.json",
+        "../../specification/v0_9_1/catalogs/**/*.json",
         "../../specification/v1_0/json/*.json",
         "../../specification/v1_0/catalogs/**/*.json",
         "scripts/copy-spec.js"
--- a/renderers/web_core/src/v0_9/processing/message-processor.ts
+++ b/renderers/web_core/src/v0_9/processing/message-processor.ts
@@ -73,7 +73,9 @@ export class MessageProcessor<T extends ComponentApi> {
   getClientCapabilities(options?: CapabilitiesOptions): A2uiClientCapabilities {
     const capabilities: A2uiClientCapabilities = {
       'v0.9': {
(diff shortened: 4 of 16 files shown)


renderers/angular/src/v0_9/catalog/basic/basic-catalog.ts
```diff
@@ -95,6 +95,11 @@ export interface BasicCatalogOptions {
    */
   id?: string;
 
+  /**
+   * An optional array of alias URLs for the catalog.
+   */
+  aliases?: readonly string[];
```
gspencergoog, line 101: This isn't correct, since the `v0_9_1` catalog and the `v0_9` catalog are different and aren't interchangable.  They aren't aliases for each other, and need to be treated separately.

renderers/angular/src/v0_9/catalog/basic/basic-catalog.ts
```diff
@@ -95,6 +95,11 @@ export interface BasicCatalogOptions {
    */
   id?: string;
 
+  /**
+   * An optional array of alias URLs for the catalog.
```
gspencergoog, line 99: They're not necessarily URLs.  They are often written that way, but they can be any string.

renderers/web_core/src/v0_9/processing/message-processor.ts
```diff
@@ -73,7 +73,9 @@ export class MessageProcessor<T extends ComponentApi> {
   getClientCapabilities(options?: CapabilitiesOptions): A2uiClientCapabilities {
     const capabilities: A2uiClientCapabilities = {
       'v0.9': {
-        supportedCatalogIds: this.catalogs.map(c => c.id),
+        supportedCatalogIds: this.catalogs.flatMap(c =>
+          c.aliases ? [c.id, ...c.aliases] : [c.id],
+        ),
       },
     };
```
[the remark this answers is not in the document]
gspencergoog: Yes, this would be needed.

renderers/web_core/src/v0_9/schema/verify-schema.test.ts
```diff
@@ -32,7 +32,7 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 // `__dirname` will be `dist/src/v0_9/schema` when run via `node --test dist/**/*.test.js`
-const SPEC_DIR_V0_9 = resolve(__dirname, '../../../../../../specification/v0_9/json');
+const SPEC_DIR_V0_9 = resolve(__dirname, '../../../../../../specification/v0_9_1/json');
```
gspencergoog, line 35: This effectively ignores the v0_9 directory.  We should be testing them independently to make sure they are both still valid.

renderers/web_core/package.json
```diff
@@ -38,6 +38,14 @@
       "types": "./dist/src/v0_9/basic_catalog/index.d.ts",
       "default": "./dist/src/v0_9/basic_catalog/index.js"
     },
+    "./v0_9_1": {
```
gspencergoog, line 41: This treats the `v0_9_1` catalog as if it were the `v0_9` catalog.  They do have differences, and so this should point to the `v0_9_1` catalog.

renderers/angular/src/v0_9/catalog/basic/basic-catalog.ts
```diff
@@ -148,7 +153,10 @@ export class BasicCatalogBase extends AngularCatalog {
       ...(options.extraComponents ?? []),
     ];
 
-    super(id, components, functions);
+    const aliases = options.aliases ?? [
+      'https://a2ui.org/specification/v0_9_1/catalogs/basic/catalog.json',
```
gspencergoog, line 157: This is saying that if you have a v0_9 renderer you can render v0_9_1 content, and while that is often true (they are similar), where they are different, it will fail.

renderers/web_core/package.json
ditman: Why are there differences in the v0.9 and 0.9.1 catalogs? I thought 0.9.1 was only to release the mimetype update?

gspencergoog: Yeah, you're right, that's the only difference.  Hmm. Maybe in this case the parsing of the catalogs is the same, but I'm thinking of it as "what if we release 0.9.2 and it has extra fields/components?"  I don't see that happening for 0.9, but I'm pretty sure it will happen sometime soon, e.g. for v1.1.

I'm wary of just aliasing them like this becoming a pattern.

ditman: I really think we should hide/remove any actual implementation code for v0.9.1, and just make 0.9 accept both 0.9 and 0.9.1. The aliasing of the catalog id should be a one-time only thing because of the way 0.9.1 was made (that I guess can live in the v0.9 directory only). Maybe we can leave instructions for agents not to replicate that pattern later? :)

renderers/react/package.json
```diff
@@ -55,6 +55,16 @@
...
+    "./v0_9_1": {
+      "import": {
+        "types": "./dist/v0_9/index.d.ts",
+        "default": "./dist/v0_9/index.js"
+      },
+      "require": {
+        "types": "./dist/v0_9/index.d.cts",
+        "default": "./dist/v0_9/index.cjs"
+      }
+    },
```
ditman, line 67: /cc @gspencergoog for example, I would *not* recommend adding this alias. People should be importing from v0_9 only.

As we discussed before, IMO the versions of the protocols shouldn't necessarily map to the versions of the renderers.

gspencergoog: Yes, I agree in this case that they are functionally the same unless you're using A2A.  There is one other minor spec change besides the MIME change: in v0.9 the `surfaceId` had to be globally unique for the entire lifetime of the renderer, and in v0.9.1 it was relaxed so that the `surfaceId` only needs to be unique among currently active surfaces.  But I think that's backward compatible by design.

I'm OK with just using 0.9 code for 0.9.1 (and 0.9) everywhere except A2A support, but we should probably restructure the spec itself so that there isn't duplication and so it makes it abundantly clear that this is a special case and not the norm. I'm pretty sure that from now on dot releases will contain additional features.

ditman:
I took a stab at some of this here: https://github.com/a2ui-project/a2ui/pull/1993

ditman:
@Varun-S10 I landed #1993 so I'm going to close this, but maybe you can use your expertise from this PR to verify my fix? Thanks!

PR #1976, opened 2026-07-09
add a2ui_subagent_map class for orchestrator sdk

wrenj:
pulls out common class of orchestrator to make reusable sdk for mapping client/server events in orchestrators


agent_sdks/python/a2ui_agent/src/a2ui/adk/orchestration/a2ui_subagent_map.py
```diff
@@ -93,3 +131,53 @@ async def remove_subagent(
...
+        """Processes a single server-to-client part and updates the subagent map.
+        Raises SurfaceIdAlreadyExistsError if a collision occurs.
+        """
+        if not is_a2ui_part(a2a_part) or not isinstance(a2a_part.root, DataPart):
+            return
+
+        data = a2a_part.root.data
+        if (
+            isinstance(data, dict)
+            and (begin_rendering := data.get(A2UI_BEGIN_RENDERING_KEY))
```
nan-yu: This only works for v0.8. If we plan to move this to the SDK, we would need to support newer versions as well.

wrenj: updated, ptal, tnx

nan-yu (approved):
Formatting failures need to be fixed before submitting.

PR #1977, opened 2026-07-09
chore(formatting): isolate pyink cache and scope by black version

gspencergoog:
## Summary
Removes the `black` override from `pyproject.toml`. This allows the python dependency resolver to naturally install `black==25.12.0`, which matches the exact version requirement of `pyink` version `25.12.0` and resolves formatting compatibility issues under Python 3.13.

## Problem
The `black>=26.5.1` override was previously added to satisfy a Dependabot security alert (CVE-2026-32274) on `black`. However, since `pyink` monkeypatches `black` internals and has a strict internal requirement for `black==25.12.0`, forcing the override broke formatting under Python 3.13. 

Specifically, newer versions of `black` expect class attribute annotations for `lazy_import` on `_python_symbols`, which does not exist in `black` 25.12.0. This mismatch caused a startup crash on Python 3.13:
```
AttributeError: '_python_symbols' object has no attribute 'lazy_import' and no __dict__ for setting new attributes
```

Since the code formatter is a development-only tool and is not deployed to production environments, we can safely remove the override and allow `black` to downgrade back to its matched version `25.12.0`.

## Changes
- Removed `override-dependencies` from `pyproject.toml`.
- Re-synced dependencies, locking `black` back to `25.12.0` in `uv.lock`.

## Impact & Risks
- **Dependabot alert:** The Dependabot warning for `black` version `<26.5.1` will reappear. This is acceptable since it only affects local/CI development formatting runs.
- **No code changes:** Reverts the previous cache-scoping hacks in `scripts/fix_format.sh`, keeping the script completely clean.

## Testing
- Verified that running `./scripts/fix_format.sh` succeeds under Python 3.13 with `black==25.12.0` and `pyink==25.12.0`.


nan-yu (approved):

gspencergoog:
I updated this to just remove the `black` override.  The Dependabot CVE really isn't relevant, since we're not formatting code on a production service (it's only dev-time).

PR #1981, opened 2026-07-10
refactor(formats): introduce InferenceFormat base API

gspencergoog:
> [!IMPORTANT]
> **Stacked PR Series:**
> 1. `#1981` (Foundation & Experiments API) 👈 *This PR*
> 2. [#1982](https://github.com/a2ui-project/a2ui/pull/1982) (Inference Strategies & Eval Solver)
> 3. [#1996](https://github.com/a2ui-project/a2ui/pull/1996) (Streaming Support)

## Summary

This pull request implements a unified strategies framework to support multiple message formats in the Python Agent SDK. It introduces the `InferenceFormat` base class and separates the standard JSON formatting logic from experimental layouts (such as Express DSL and Elemental).

## Changes

* Added the abstract `InferenceFormat` class at `a2ui/inference_format.py` to decouple prompt generation and parsing.
* Added `a2ui/prompt/generator.py` and `a2ui/prompt/__init__.py` to hold the abstract `PromptGenerator` class, and added a deprecated compatibility shim in `a2ui/prompt_generator.py`.
* Moved standard JSON parsing and generation to the `TransportFormat` strategy under `a2ui/inference_formats/transport/`.
* Moved the legacy streaming parser code to `a2ui/inference_formats/transport/streaming.py`.
* Converted `A2uiSchemaManager` (`schema/manager.py`) to a deprecated compatibility wrapper that delegates to `TransportFormat`.
* Created `a2ui/validation/validator.py` containing a unified `A2uiValidator` to consolidate validation, moving the legacy validation logic to `a2ui/validation/validator_v08.py`.
* Updated `A2uiValidator` to gate v1.0 schema validation programmatically using experiments configuration instead of fallback environment variables.
* Added `has_format_content` to the `Parser` interface and deprecated the global helper `has_a2ui_parts`.
* Removed all remaining `A2UI_EXPRESS_ENABLED` and `A2UI_VERSION_1_0` global environment variable gating checks across the SDK and test suites.

## Impact and risks

This change is backward-compatible. By default, the SDK uses the standard JSON strategy (`TransportFormat`), and experimental formats remain opt-in.

## Testing

Run the pytest suite to verify all unit and conformance tests pass:
```bash
uv run pytest
```


agent_sdks/python/a2ui_agent/src/a2ui/a2a/parts.py
```diff
@@ -98,23 +100,34 @@ def parse_response_to_parts(
...
 
         for part in response_parts:
             if part.text:
                 parts.append(Part(root=TextPart(text=part.text)))
 
             if part.a2ui_json:
                 json_data = part.a2ui_json
-                if validator:
+                if validator and isinstance(strategy, JsonInferenceFormat):
                     validator.validate(json_data)
```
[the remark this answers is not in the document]
gspencergoog: Yes, I have removed the JsonInferenceFormat check and enabled validation unconditionally for all strategy output payloads.

agent_sdks/python/a2ui_agent/src/a2ui/a2a/parts.py
```diff
@@ -98,23 +100,34 @@ def parse_response_to_parts(
...
     try:
-        response_parts = parse_response(content)
+        if isinstance(strategy, JsonInferenceFormat):
+            from a2ui.parser.parser import parse_response
+
+            response_parts = parse_response(content)
+        else:
+            if not catalog:
+                raise ValueError("A catalog is required to parse non-JSON formats.")
+            response_parts = strategy.parse_response(content, catalog)
```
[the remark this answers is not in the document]
gspencergoog: Agreed. I have updated the InferenceFormat interface to make catalog optional and removed the concrete type check, invoking it polymorphically.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @abstractmethod
+    def parse_response(
+        self,
+        content: str,
+        catalog: A2uiCatalog,
+        surface_id: str | None = None,
+    ) -> List[ResponsePart]:
+        """Parses model response content and compiles it to standard A2UI JSON payload parts."""
+        pass
```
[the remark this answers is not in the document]
gspencergoog: Updated the abstract base class method signature to make catalog optional.

agent_sdks/python/a2ui_agent/src/a2ui/formats/json_inference_format.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def generate_instructions(self, catalog: A2uiCatalog) -> str:
+        return catalog.render_as_llm_instructions()
+
+    def parse_response(
+        self,
+        content: str,
+        catalog: A2uiCatalog,
+        surface_id: str | None = None,
+    ) -> List[ResponsePart]:
```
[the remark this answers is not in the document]
gspencergoog: Updated JsonInferenceFormat.parse_response to match the base signature.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+
+__all__ = [
+    "InferenceFormat",
+    "InferenceFormatRegistry",
+    "JsonInferenceFormat",
+]
+
+
+class InferenceFormat(ABC):
+    """Abstract interface defining an alternative A2UI payload format."""
```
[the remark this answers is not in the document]
gspencergoog: Declining this because we are resolving this via registry dynamic loading instead of creating a separate file.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @classmethod
+    def get(cls, name: str) -> InferenceFormat:
+        if name not in cls._formats:
+            raise ValueError(f"Unknown inference format: {name}")
+        return cls._formats[name]
+
+    @classproperty
+    def available_formats(cls) -> list[str]:
+        return list(cls._formats.keys())
```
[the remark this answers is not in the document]
gspencergoog: Done. Replaced @classproperty with standard @classmethod and updated the call sites in tests.

agent_sdks/python/a2ui_agent/src/a2ui/a2a/parts.py
```diff
@@ -90,6 +91,8 @@ def parse_response_to_parts(
     validator: Optional[Any] = None,
     fallback_text: Optional[str] = None,
     version: Optional[str] = None,
+    format_strategy: Optional[Any] = None,
```
jacobsimionato: Why do we need this? I would expect A2A to only ever convey true A2UI protocol messages - the inference formats are an implementation detail and should be converted to A2UI protocol as the first step after they come from the LLM.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    ) -> List[ResponsePart]:
+        """Parses model response content and compiles it to standard A2UI JSON payload parts."""
+        pass
+
+    @abstractmethod
+    def decompile(self, val: dict[str, Any], catalog: A2uiCatalog) -> str:
+        """Decompiles standard JSON payload back into this format's string representation."""
+        pass
+
+    def detect_format(self, content: str) -> bool:
```
jacobsimionato: I think this sets up an expectation that formats are detectible, but instead, we should design our system to ensure that the inference format is always explicitly stated where necessary. I worry that introducing this could result in us relying on it unnecessarily.

How about an architecture where every part of the system uses the standard A2UI format by default, but then when you trigger a specific LLM inference, an inference format is determined (hardcoded for now!), and then we adapt the catalog, examples, and any A2UI content in the message history to the inference format, then convert the results back to A2UI. So the choice of inference format has a scope of *just* that single inference, and it's super clear which one was picked, because it's basically a const local variable in that scope. A followup inference in the same conversation could use a different inference format.

I know you're proposing the idea of allowing for dynamic choosing of inference formats which is extremely cool, but I'm worried that it adds unnecessary complexity at this point, given express is a clear winner based on all available evals. And allowing for dynamic picking of inference formats adds some complexity for cases where there is A2UI in the message history. Ideally, that content should be presented to the LLM in the same/similar format to the output format, to avoid confusion. So if the output format is picked dynamically, then either it might not match, or else we need to trigger a whole second inference that transforms the message history to the new format which costs latency and breaks token caching.

I dunno, WDYT?

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        """Returns True if the content contains this format's sentinel/payload tags."""
+        return False
+
+    def extract_surface_ids(self, content: str) -> List[str]:
```
jacobsimionato: Are you sure we need this method? I would imagine that if you want to extract the surface IDs from content, then you can convert it to A2UI then read the IDs from the A2UI content in a standard way.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    def generate_workflow_rules(self, custom_workflow_description: str = "") -> str:
+        """Generates formatting rules and sentinel tag expectations."""
+        pass
+
+    @abstractmethod
+    def generate_instructions(self, catalog: A2uiCatalog) -> str:
+        """Generates component and function signatures for the catalog in this format."""
+        pass
+
+    def transform_examples(
```
jacobsimionato: Can't we implement this as generic functionality inside a PromptGenerator class, which just runs the markdown through 'decompile' etc? I don't see how there would need to be any format-specific logic here, apart from decompile.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    """Abstract interface defining an alternative A2UI payload format."""
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """The format name (e.g., 'json', 'express')."""
+        pass
+
+    @abstractmethod
+    def generate_workflow_rules(self, custom_workflow_description: str = "") -> str:
```
jacobsimionato: I found the naming of this method and the one below a little confusing. How about:

`format_description_prompt_fragment`

`catalog_description_prompt_fragment`

Or just `format_description` / `catalog_description`?

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        pass
+
+    def transform_examples(
+        self, raw_examples_markdown: str, catalog: A2uiCatalog
+    ) -> str:
+        """Transforms examples markdown (e.g., converting JSON blocks to the target format)."""
+        return raw_examples_markdown
+
+    @property
+    def default_surface_id(self) -> str:
```
jacobsimionato: Surely this doesn't need to be different for each format, so we can remove it from here?

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        """The default surface ID used if none is specified."""
+        return "main"
+
+    @abstractmethod
+    def parse_response(
```
jacobsimionato: For a specific inference, I think a few things need to be consistent between prompt and parser, for the inference to work correctly:
- Format choice
- Catalog choice (perhaps multiple catalogs eventually if we adopt wrenj@'s proposal)
- Strategy around surface ID management (e.g. do we prompt the LLM to define this, or do we choose one in advance)

So I'm wondering if it makes sense to design our APIs to represent this, so that mismatch is more difficult.
```
format_factory = ExpressInferenceFormatFactory()

def do_inference():
  catalog = my_catalog;
  format = format_factory.create(catalog, surface_id=ConstantSurfaceId("default"));
  
  prompt = PromptBuilder(format).build()
  processed_message_history = HistoryTransformer.transform(format, message_history);
  
  result = await llm.generate_text(prompt, history);
  output = format.process(result);

class InferenceFormatFactory:
  def create(Catalog catalog, SurfaceIdStrategy surface_id) -> InferenceFormat;

class ExpressInferenceFormatFactory(InferenceFormatFactory); etc
```

That way, you kind of configure the parameters for inference in one place, and then the prompt and parsing flow are forced to be consistent. Each compile/decompile/prompt method has minimal parameters, because the configuration is defined already.

You could do this without InferenceFormatFactory also, and just have a constructor for ExpressInferenceFormat etc that accepts the catalog etc. Less polymorphism, but maybe still nicer? 

I don't feel super strongly, just an idea. Maybe mine is more complex due to more classes etc.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    def detect_format(self, content: str) -> bool:
+        """Returns True if the content contains this format's sentinel/payload tags."""
+        return False
+
+    def extract_surface_ids(self, content: str) -> List[str]:
+        """Extracts any surface IDs declared or defined in the content according to this format."""
+        return []
+
+
+class InferenceFormatRegistry:
```
jacobsimionato: YAGNI? Could we replace this with an array of inference formats? Or just default to express for now and leave the others as purely present for experimentation / eval purposes?

agent_sdks/python/a2ui_agent/src/a2ui/schema/manager.py
```diff
@@ -33,13 +29,15 @@ def __init__(
         self,
         version: str,
         catalogs: Optional[list[CatalogConfig]] = None,
+        format_strategy: Optional[InferenceFormat] = None,
         accepts_inline_catalogs: bool = False,
```
jacobsimionato, line 44: Side comment: I propose we break this class apart  later, but @nan-yu will handle that in her design, so this LGTM for now.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        return raw_examples_markdown
+
+    @property
+    def default_surface_id(self) -> str:
+        """The default surface ID used if none is specified."""
+        return "main"
+
+    @abstractmethod
+    def parse_response(
+        self,
```
jacobsimionato: Another related thought: Maybe for streaming we might want a concept of a stateful parser that processes chunks and remembers what it has seen before. E.g.

```
parser = format.parser()

for chunk in llm.generate_text(prompt):
  parser.addChunk()
```

I guess worry about streaming later?

agent_sdks/python/a2ui_agent/src/a2ui/a2a/parts.py
gspencergoog: We need it because if we asked the agent to produce non-JSON output we have to convert it back to JSON, because that's all the A2A Parts should contain.  I'm agreeing with you here: this is how they get converted before being put into Parts so that the format is consistently JSON.  We also don't want to hardcode this to JSON because if we do, then JSON would be the only allowed inference format for A2A agents.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Yeah, good points, all.  I agree, let's just make it explicit what format is expected.  The format detection was janky anyhow.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: You're right, we don't need it.  I'll remove it.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Yes, that's a good point. I'll have the prompt generator call the format strategy's decompile() method.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Sure.  We'll use  `format_description` and `catalog_description` .

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Good point, I'll promote it.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: No, I like the factory method: it's good for grouping all the things: it makes things more consistent.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: I make the registry so we can dynamically load experimental formats via environment flags without hardcoding their imports to make sure the experimental code wasn't bloating the compiled code.  Also, I think it makes it easier for people to build their own inference formats and plug them in.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Yeah, I hadn't gotten to streaming yet, but it will need to be at least somewhat stateful.  Let's punt that for now, but I plan to tackle streaming soon because I don't want to leave that for too long, since it will likely affect the API.

gspencergoog: I created https://github.com/a2ui-project/a2ui/pull/1985 as a possible streaming solution.

agent_sdks/python/a2ui_agent/src/a2ui/a2a/parts.py
jacobsimionato: Ahhhh I misunderstood the role if this function. I see, it goes all the way from raw LLM output to A2A, hence the need for strategy. Makes sense.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
jacobsimionato: I want to fully understand this, because although this particular class isn't a huge deal, I think it's important that we establish a good pattern for separating experimental code from non-experimental code. We will have more experimental features soon, so I want to figure out if we want to follow this pattern e.g. for templates etc.

So I think our goals are:
- Developer who wants to use express (or elemental or some other strategy) can use it easily, while being clear that it's experimental
- Developer who doesn't want this strategy can not use it, and not incur any performance cost (compilation or runtime) 

I think we can achieve all this by just having application developers pass `ExpressInferenceFormat()` somewhere, right? To me, that seems quite a lot simpler than a model where you need to set an environment flag, and there's a stateful repository to think about. Developers don't need to know the flag name, the format name etc. To build your own format, you just implement the interface and pass it programmatically, rather than having a two step process of registering it, then referring to it via a string key.

This is a strong opinion loosely held! I'm no Python expert and I will defer to the experts!

jacobsimionato: From Gemini:

# Proposal: Simplifying A2UI Format Strategies via Explicit Programmatic Instantiation

  ## Executive Summary

  We propose removing the  InferenceFormatRegistry  class ( a2ui.formats ) and all environment variable flags (
  A2UI_EXPRESS_ENABLED ,  A2UI_ELEMENTAL_ENABLED ). Instead of relying on a string-based global registry and dynamic imports,
  developers will explicitly import and instantiate concrete  InferenceFormat  strategy classes ( ExpressInferenceFormat  or
  ElementalInferenceFormat ) directly in their application code when opting into alternative formats.
  ──────
  ## Proposed Developer Workflow

  ### Standard Usage (Default JSON)

  For developers using standard A2UI JSON, the SDK defaults to  JsonInferenceFormat()  automatically with zero configuration:

    from a2ui.schema.manager import A2uiSchemaManager

    # Defaults automatically to JsonInferenceFormat()
    schema_manager = A2uiSchemaManager(version="0.9.1")

  ### Experimental Usage (Express / Elemental DSL)

  Developers opting into alternative formats explicitly import and instantiate the concrete strategy class and pass it into
  their SDK components:

    from a2ui.schema.manager import A2uiSchemaManager
    # Explicit opt-in from the experimental subpackage:
    from a2ui.experimental.express.format import ExpressInferenceFormat

    express_strategy = ExpressInferenceFormat()
    schema_manager = A2uiSchemaManager(version="0.9.1", format_strategy=express_strategy)
    ──────
  ## Performance Impact (Compilation & Runtime)

  ### 1. For Developers Who Do NOT Use Express

  • Compilation / Import Time Overhead: ZERO. By removing the registry and ensuring core classes never reference  a2ui.
  experimental.* , Python never opens, reads, or compiles any Express or Elemental  .py  files during application startup.
  • Runtime Overhead: ZERO. Applications running standard JSON ( JsonInferenceFormat ) execute without global dictionary
  lookups,  importlib  checks, or environment variable inspections.

  ### 2. For Developers Who DO Use Express

  • Compilation / Import Time: Only files containing the explicit  import a2ui.experimental.express.format  statement incur the
  one-time import cost of the Express parser and compiler modules at process startup.
  • Runtime Overhead: ZERO. Once instantiated,  ExpressInferenceFormat  is passed directly as a polymorphic object (
  InferenceFormat ) to  A2uiSchemaManager  and part converters, executing at native speed without registry indirection.
  ──────
  ## Preventing Accidental Usage of Experimental Features

  Instead of runtime exceptions gated by environment variables, we rely on explicit namespace separation and runtime warnings:

  1. Namespace Isolation ( /experimental  subfolder): Experimental format strategies will never be exposed in or exported from
  a2ui.formats  (the public umbrella namespace). To use Express, developers must explicitly type  from a2ui.experimental.
  express.format import ExpressInferenceFormat . This makes opt-in completely transparent during code reviews and statically
  trackable by linters and IDEs.
  2. Feature Decorators & Runtime Warnings: We can decorate  ExpressInferenceFormat  with the existing  @experimental
  decorator (or emit a one-time  UserWarning  /  logging.warning  upon class instantiation) to ensure developers and CI logs
  are clearly notified when an unstable specification is active in a running process.
  ──────
  ## Preventing Unnecessary Imports & Dependency Leakage

  To guarantee that experimental code is never imported or compiled automatically when not in use, we enforce strict dependency
  directionality:

  • One-Way Dependency Rule: Core library packages ( a2ui.schema ,  a2ui.parser ,  a2ui.a2a ,  a2ui.formats ,  a2ui.adk ) MUST
  NEVER import from or reference  a2ui.experimental.* .
  • Polymorphic Interfaces: Core functions (such as  parse_response_to_parts ) and classes ( A2uiSchemaManager ,
  A2uiPartConverter ) accept only the abstract base class  InferenceFormat  or default to  JsonInferenceFormat() .
  • Elimination of the Registry: Removing  InferenceFormatRegistry._ensure_initialized()  eliminates the only existing bridge
  where non-experimental code imported  a2ui.experimental.express  via  importlib .
  ──────
  ## Pros and Cons of This Approach

  ### Pros

  • Architectural Simplicity: Eliminates stateful global registries,  importlib  dynamic module loading, and environment
  variable checks.
  • Static Type Safety & IDE Support: Direct instantiation gives developers full IDE auto-completion, static type checking (
  mypy / pyright ), and "Go to Definition" support for strategy parameters, which is impossible when passing opaque strings
  like  registry.get("express") .
  • Zero Accidental Leakage: Guaranteed complete isolation between stable core components and experimental parsers.
  • Predictable Bundling: Makes static dependency analysis straightforward for bundling tools (like PyInstaller or  pex ),
  ensuring experimental modules are naturally excluded unless explicitly imported in application code.

  ### Cons

  • No Built-in String/CLI Resolution: If an application developer wants to let end-users switch formats via a command-line
  flag ( --format=express ) or configuration file, the SDK will no longer provide a string-to-class lookup out of the box. The
  application developer must write their own small factory mapping ( {"json": JsonInferenceFormat, "express":
  ExpressInferenceFormat} ).
  • Minor Breaking Change: Any existing internal tests or experimental prototypes currently calling  InferenceFormatRegistry.
  get("express")  alongside  os.environ["A2UI_EXPRESS_ENABLED"] = "true"  will need a small import refactor to import
  ExpressInferenceFormat  directly.

jacobsimionato (approved):

jacobsimionato:
I think we should get @nan-yu to review this too, because she's driving an overhaul of the inference SDK API surface

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Well... What can I say other than "Sure, let's do it that way!". :-)

I'm not super invested in the way I did it: I just wanted to make sure we could have self-created inference methods and to make sure that if people didn't opt in that they wouldn't get their code bloated.

Happy to do it this way.

gspencergoog: Yes, after implementing this, it looks a lot nicer.  Thanks!

agent_sdks/python/a2ui_agent/src/a2ui/a2a/parts.py
nan-yu: What about passing a parser instance to this method? There will be a parser abstract class after the redesign in https://docs.google.com/document/d/17BuULV88IlsUStOFthGMgfeW1XMtLUyQDm3oflgXi-k/edit?resourcekey=0-fltDWdFS4VEdDHSTzq--QA&tab=t.0#bookmark=id.1yfukr4tw9yb.

Each inference strategy will implement a `parse_response` method. The `parse_response_to_parts` will call `parse_response` and convert the parsed result into A2A parts.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,223 @@
...
+
+__all__ = [
+    "InferenceFormat",
+    "InferenceFormatRegistry",
+    "JsonInferenceFormat",
+    "PromptGenerator",
+]
+
+
+class InferenceFormat(ABC):
```
nan-yu: I'm wondering whether we need to introduce a `InferenceFormat` with the new design in https://docs.google.com/document/d/17BuULV88IlsUStOFthGMgfeW1XMtLUyQDm3oflgXi-k/edit?usp=sharing&resourcekey=0-fltDWdFS4VEdDHSTzq--QA.

It could be replaced with InferenceStrategy + Parser.
- `InferenceFormat.format_prompt()` could be handled by `InferenceStrategy.generate_system_prompt()` (e.g. `SchemaPromptGenerator`, `ExpressPromptGenerator`)
- `InferenceFormat.convert_to_json()` could be handled by `Parser.parse_response()` and `Parser.process_chunk()` (e.g. `ExpressParser` + `ExpressCompiler`, `A2uiSchemaParser`)
- `InferenceFormatRegistry` could be handled by concrete strategy packages under `a2ui.strategies.<name>`.

That could help eliminate a parallel abstraction.

gspencergoog: Okay, I've refactored this PR so that it uses a separate `InferenceStrategy` and `*Parser`.

Let me know what you think.

gspencergoog:
I also updated how experiments are enabled, so that it doesn't rely on environment variables at all anymore.  The experimental inference formats both rely on 1.0 features, so they needed different validation, which means that we needed some way to configure experiments on the validator and other components.

I opted for an explicit set of experiment strings in the configuration, which seemed to be the cleanest way to enable them.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
nan-yu: Can this be consolidated with the InferenceStrategy in inference_strategy.py? json_inference_format.py can be moved to schema/inference_strategy.py.

The structure will be
```
src/a2ui/
├── inference_strategy.py 
├── strategies/ 
│   └── schema/       
│       ├── inference_strategy.py
└── experimental/
    └── express/ 
        ├── inference_strategy.py   

```

nan-yu: The parser abstract class doesn't exist yet. We can refactor later separately.

agent_sdks/python/a2ui_agent/src/a2ui/a2a/parts.py
nan-yu: Once we have the parser abstract class (not available yet), we can change it to pass a parser instead of a inference_strategy. The parser implementation (A2uiSchemaParser or JsonParser, ExpressParser) should already have access to catalog, so we don't need to pass the catalog param either. That will help simplify the signature.

gspencergoog: Yes, agreed.  I almost created the abstract parser here, but kept it like this.  Should I go ahead and add the abstract Parser? It doesn't seem too hard, and we can always modify it if it doesn't match what you end up with.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+        """Parses model response content and compiles it to standard A2UI JSON payload parts."""
+        pass
+
+    @abstractmethod
+    def decompile(self, val: dict[str, Any]) -> str:
+        """Decompiles standard JSON payload back into this format's string representation."""
+        pass
+
+    @abstractmethod
+    def wrap_decompiled_blocks(self, blocks: list[str]) -> str:
```
jacobsimionato: Nit: Currently, it seems like parseResponse handles unwrapping *and* compiling, while we separate the decompile/wrap processes in the other direction. I'm wondering if it makes sense to have symmetry between the directions, to simplify the mental model. I also think this might help for certain use cases, e.g. let's say we have a single full screen surface application. The agent never outputs text, only the inference format directly. So to do this, we'd want to call "compile" and "decompile" directly, ignoring the wrap/unwrap part. Currently, you can't do that on this API, because `parse_response` combines the two.

Straw person API:
```py
# Tokenize text to a sequence of text and inference format parts,
# though the inference format parts are still a string for now.
def parse(textAndInferenceFormatContent: str) -> list[ResponsePart] # or unwrap() or tokenize() ?

# Actually compile the inference format string to A2UI messages
def compile(inferenceFormatContent: str) -> list[ServerToClientMessage]

def decompile(messages: list[ServerToClientMessage]) -> str
def wrap(content: str) -> str
```

Another idea:
```py
# Parse string to text and/or A2UI messages. If the isWrapped is false,
# it does not look for separators and tries to parse the entire input as an A2UI block.
def parse(textAndInferenceFormatContent: str, isWrapped: bool) -> list[ResponsePart]

# Decompiles A2UI to inference format, wrapping with separators if 
# "wrap" is true.
def decompile(messages: list[ServerToClientMessage], wrap: bool) -> str
```

I know I'm getting in the weeds here. If you think it's not worth iterating further, I'm happy to merge the code as you have it now!

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+        self.surface_id = surface_id
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """The format name (e.g., 'json', 'express')."""
+        pass
+
+    @abstractmethod
+    def format_description(self, custom_workflow_description: str = "") -> str:
```
jacobsimionato: I think `custom_workflow_description` can be a parameter on a `PromptGenerator` method instead of  `InferenceFormat`. Looking at the implementation in `json_inference_format.py`, it just concatenates a string into the format description, so there's no need for every different format to implement that in parallel.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+        """Generates formatting rules and sentinel tag expectations."""
+        pass
+
+    @abstractmethod
+    def catalog_description(self, include_schema: bool = True) -> str:
```
jacobsimionato: Let's also remove `include_schema` from here. We already have this parameter on `generate_system_prompt` in `PromptGenerator`.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+        """Decompiles standard JSON payload back into this format's string representation."""
+        pass
+
+    @abstractmethod
+    def wrap_decompiled_blocks(self, blocks: list[str]) -> str:
+        """Wraps decompiled code blocks in format-specific tags and code block wrappers."""
+        pass
+
+    @abstractmethod
+    def has_a2ui_parts(self, content: str) -> bool:
```
jacobsimionato: Optional: Rather than having each format implement this specially, how about we just rely on `parse_response` above. If a caller wants to scan for A2UI parts, it can just call `parse_response` and inspect the output to see if there are an A2UI chunks. I know that's likely less efficient, but I'm not sure we need to optimize at that micro level? Also, if we split unwrapping from compiling above, the unwrapping part would be cheap.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
jacobsimionato: Thanks for iterating! Sorry to drop a random design doc in your comments.

nan-yu (approved):

agent_sdks/python/a2ui_agent/src/a2ui/schema/manager.py
gspencergoog: Agreed. This aligns with the downstream strategy/parser package cleanup already in https://github.com/a2ui-project/a2ui/pull/1982

agent_sdks/python/a2ui_agent/src/a2ui/a2a/parts.py
gspencergoog: I defined the base `Parser` interface directly in this PR, and updated the `A2uiPartConverter` signature to accept it instead of the strategy, removing the catalog parameter.  We can revise it later if your design changes.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Good point, OK.  I'll refactor the parser to support separate `unwrap` and `compile` phases.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Yes, I'll remove `custom_workflow_description` from the strategy/format classes and pass it as an argument during prompt generation.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Agreed. I'll clean this up and pass it directly during prompt generation.

agent_sdks/python/a2ui_agent/src/a2ui/formats/__init__.py
gspencergoog: Sure. Once we split the unwrapping step, callers can check for format parts in the unwrapped result. I'll remove the helper.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/streaming.py
```diff
@@ -0,0 +1,1153 @@
...
+        raw_contents = update.get("contents")
+        if raw_contents is not None:
+            contents = self._parse_contents_to_dict(raw_contents)
+        else:
+            # Fallback for old v0.8 flat structure or other layouts
+            contents = {
+                k: v
+                for k, v in update.items()
+                if k not in (SURFACE_ID_KEY, "root", "contents")
+            }
```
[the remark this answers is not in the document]
gspencergoog: Removed the unused `contents` calculation inside the base `update_data_model` method to address the warning.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/streaming.py
```diff
@@ -0,0 +1,1153 @@
...
+        # State for buffering updates until surface is ready
+        self._pending_messages: Dict[str, List[Dict[str, Any]]] = (
+            {}
+        )  # surfaceId -> list of msgs delayed until start message arrives
+        self._buffered_start_message: Optional[Dict[str, Any]] = (
+            None  # The start message to yield before any components
+        )
+        self._topology_dirty = False  # Set to true if components are added out of order
+        self._in_top_level_list = False
+        self._found_valid_json_in_block = False
```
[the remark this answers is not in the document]
gspencergoog: Removed the duplicate initialization of `self._in_top_level_list` in the constructor.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/streaming.py
```diff
@@ -0,0 +1,1153 @@
...
+        """Recursively processes path placeholders and child pruning in one pass."""
+        comp_id = comp.get("id", "unknown")
+
+        # Deduce the component type for better placeholder typing
+        comp_type = (
+            next(iter(comp.get("component", {}).keys()))
+            if comp.get("component") and isinstance(comp.get("component"), dict)
+            else ""
+        )
+
```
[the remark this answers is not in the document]
gspencergoog: Extracted `traverse` to a private class method `_traverse_component_topology` to prevent recreation on every iteration, and removed the unused `comp_type` variable.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/streaming.py
```diff
@@ -0,0 +1,1153 @@
...
+    def _get_placeholder_id(self, child_id: str) -> str:
+        """Returns the ID to use for a missing child placeholder."""
+        return f"loading_{child_id}"
+
+    def _process_component_topology(
+        self,
+        comp: Dict[str, Any],
+        extra_components: List[Dict[str, Any]],
+        inline_resolved: bool = False,
+    ) -> None:
```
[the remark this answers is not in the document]
gspencergoog: Removed the unused `inline_resolved` parameter from `_process_component_topology`.

agent_sdks/python/a2ui_agent/src/a2ui/validation/validator.py
```diff
@@ -223,9 +223,29 @@ def validate(
...
 
-    def __init__(self, catalog: A2uiCatalog):
+    def __init__(
+        self,
+        catalog: A2uiCatalog,
+        experiments: Optional[Union[set[str], frozenset[str]]] = None,
+    ):
         ver = catalog.version
         self.version = ver if isinstance(ver, str) else VERSION_0_8
+        import os
```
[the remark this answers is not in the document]
gspencergoog: Moved `import os` to the top of the validator module per PEP 8 guidelines.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/format.py
```diff
@@ -0,0 +1,260 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+from typing import Any, Optional, Callable, Union
+
+from a2ui.schema.utils import load_from_bundled_resource
+from a2ui.inference_format import InferenceFormat
+from a2ui.schema.constants import *
```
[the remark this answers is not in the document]
gspencergoog: Replaced the wildcard import of constants with explicit, structured imports.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/format.py
```diff
@@ -0,0 +1,274 @@
...
+
+class TransportFormat(InferenceFormat):
+    """Manages standard A2UI JSON schema responses and prompt injection (Transport Format)."""
+
+    def __init__(
+        self,
+        version: str,
+        catalogs: Optional[list[CatalogConfig]] = None,
+        accepts_inline_catalogs: bool = False,
+        schema_modifiers: Optional[
```
nan-yu, line 45: This is basically the StrictnessRemovalTransformer, which mutates the catalog contract between client and agent by removing the additionalProperties field. We should remove it in the new design.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/prompt_generator.py
```diff
@@ -0,0 +1,93 @@
...
+from a2ui.prompt.generator import PromptGenerator
+
+if TYPE_CHECKING:
+    from a2ui.inference_formats.transport.format import TransportFormat
+
+
+class TransportPromptGenerator(PromptGenerator):
+    """Formats standard JSON schema system prompt instructions (Transport Format)."""
+
+    def __init__(self, format_inst: "TransportFormat"):
```
nan-yu, line 27: Why do we need the TransportFormat instance?

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/prompt_generator.py
```diff
@@ -0,0 +1,93 @@
...
+            allowed_components: Optional list of component tags the LLM may use.
+            allowed_messages: Optional list of A2UI message types allowed.
+            include_schema: Whether to include component schemas in the prompt.
+            include_examples: Whether to include few-shot examples.
+            validate_examples: Whether to validate few-shot examples on generation.
+
+        Returns:
+            The complete generated prompt system instruction.
+        """
+        selected_catalog = self._format.get_selected_catalog(
```
nan-yu, line 63: The `get_selected_catalog` method can be extracted out as a helper function in the utils/catalog_resolver.py. 
https://docs.google.com/document/d/17BuULV88IlsUStOFthGMgfeW1XMtLUyQDm3oflgXi-k/edit?resourcekey=0-fltDWdFS4VEdDHSTzq--QA&tab=t.0#bookmark=id.im80bpefqd27.

The intent is to remove the need for the format instance.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/prompt_generator.py
```diff
@@ -0,0 +1,93 @@
...
+            client_ui_capabilities, allowed_components, allowed_messages
+        )
+
+        examples_str = ""
+        if include_examples:
+            examples_str = self._format.load_examples(
```
nan-yu, line 69: This can be messy because the old package supports loading examples from a path, while the new design would like users to provide the example as a list of A2uiMessages directly. 

I think I'm okay with what it is right now and then refactor later in a separate PR.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/prompt_generator.py
```diff
@@ -0,0 +1,93 @@
...
+        role_description: str,
+        workflow_description: str = "",
+        ui_description: str = "",
+        client_ui_capabilities: Optional[dict[str, Any]] = None,
+        allowed_components: Optional[list[str]] = None,
+        allowed_messages: Optional[list[str]] = None,
+        include_schema: bool = False,
+        include_examples: bool = False,
+        validate_examples: bool = False,
+    ) -> str:
```
nan-yu, line 46: Are these parameters mostly for backward compatibility?

agent_sdks/python/a2ui_agent/src/a2ui/parser/parser.py
```diff
@@ -12,78 +12,123 @@
...
-    json_string = json_string.strip()
-    if json_string.startswith("```json"):
-        json_string = json_string[len("```json") :]
-    elif json_string.startswith("```"):
-        json_string = json_string[len("```") :]
-    if json_string.endswith("```"):
-        json_string = json_string[: -len("```")]
-    json_string = json_string.strip()
-    return json_string
+    def parse_response(self, content: str) -> List[ResponsePart]:
```
nan-yu, line 39: The parser_response method was moved as a standalone helper method instead of method in the base parser. See comment in https://docs.google.com/document/d/17BuULV88IlsUStOFthGMgfeW1XMtLUyQDm3oflgXi-k/edit?resourcekey=0-fltDWdFS4VEdDHSTzq--QA&disco=AAACDC4mPpY .

agent_sdks/python/a2ui_agent/src/a2ui/strategies/schema/__init__.py
nan-yu, line 1: This can be deleted

agent_sdks/python/a2ui_agent/src/a2ui/prompt_generator.py
nan-yu, line 1: This can be removed directly as it is only introduced in this PR.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/format.py
gspencergoog: Okay, sounds good, but I'll leave it here for now, right?

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/prompt_generator.py
gspencergoog: Okay, so for now I'll keep the format instance and load the decompiled examples this way?

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/prompt_generator.py
gspencergoog: To get at the selected catalog and decompiled examples.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/format.py
nan-yu: Yes, let's leave it here and resolve it later with other changes.

agent_sdks/python/a2ui_agent/src/a2ui/prompt_generator.py
gspencergoog: Done

agent_sdks/python/a2ui_agent/src/a2ui/strategies/schema/__init__.py
gspencergoog: Done.

agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/transport/prompt_generator.py
nan-yu: We can follow up to get it removed with some helper functions.

agent_sdks/python/a2ui_agent/src/a2ui/parser/parser.py
nan-yu: It got moved back to the Parser class for consistency.

nan-yu (approved):

PR #1990, opened 2026-07-12
Update a2ui_agent blueprint with refined APIs

nan-yu:
# Description

## Pre-launch Checklist

One time:

- [ ] I signed the [CLA].

For this PR:

[CHANGELOG]: ../CHANGELOG.md


blueprints/modules/a2ui_inference.blueprint.md
```diff
@@ -37,89 +128,301 @@ class CatalogConfig:
...
+        catalog_schema: Dict[str, Any],
+        spec_version: str,
+        catalog_id: Optional[str] = None,
+        name: Optional[str] = None,
+        custom_cuttable_keys: Optional[frozenset[str]] = None,
+    ) -> "A2uiCatalog":
+        """Constructs an A2uiCatalog instance from a raw JSON Schema dictionary."""
+        pass
+
+    def with_pruning(self, allowed_components: Optional[List[str]] = None) -> "A2uiCatalog":
```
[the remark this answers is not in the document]
nan-yu: allowed_messages is for the server-to-client schema, not for the catalog, so it is moved outside of the A2uiCatalog class.

nan-yu:
@jacobsimionato This is updated to be in sync with the API design doc. Please take a look when you get a chance.

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1210 @@
...
+    Facade --> Parsers
+    Transformers --> CoreModels
+    Formats --> CoreModels
+    Parsers --> CoreModels
+```
+
+1. **Decoupled Primitive Layer**:
+   - **Catalog Representation**: Directly uses canonical `Catalog` models from `a2ui_core`.
+   - **Catalog Transformers**: Standalone rule sets (`CatalogTransformer`, `ComponentPruningTransformer`, `FunctionPruningTransformer`) for filtering component definitions and function signatures from pristine catalogs.
+   - **Inference Formats**: Strategy facades (`InferenceFormat`, `InferenceFormatFactory`) pairing format-specific prompt generators (`PromptGenerator`) and parsers (`Parser`). Supported strategies include `DirectJsonFormat`, `ExpressFormat`, and `ElementalFormat`.
```
jacobsimionato: Let's just include direct JSON format and express format here.

All other formats can be experimental formats that just exist in Python. I think it's unclear if we'll maintain them indefinitely at this point.

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1210 @@
...
+   - **Prompt Generators**: Format builders consuming transformed catalogs and prompt examples to generate system instruction snippets.
+   - **Parsers**: Response extraction engines performing tag unwrapping (`unwrap`), streaming chunk processing (`process_chunk`), syntax compilation (`compile`) and decompilation (`decompile`).
+   - **Validation Layer**: Leverages core `A2uiValidator` capabilities directly from `a2ui_core`, natively supporting protocol version branching (`v0_8`, `v0_9`, `v0_9_1`, `v1_0`).
+2. **Encapsulated Application Session**:
+   - `CatalogConfig`: Configuration dataclass encapsulating catalog providers (`BundledCatalogProvider`, `FileSystemCatalogProvider`, `InMemoryCatalogProvider`), custom transformers, and examples.
+   - `A2uiSessionManager`: Agent-level lifecycle manager holding supported `CatalogConfigs`, caching pre-negotiated `A2uiSession` instances per renderer capability signature.
+   - `A2uiSession`: Central session facade object unifying multi-catalog capability resolution (`resolve_catalogs`), system prompt snippet rendering, turn-scoped parser creation, and response validation.
```
jacobsimionato: I think the catalog resolution is actually in A2uiSessionManager, right?

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1210 @@
...
+│   ├── session_manager        # A2uiSessionManager class
+│   └── catalog_providers      # Catalog provider classes
+├── inference_format           # Abstract InferenceFormat & InferenceFormatFactory facades
+├── inference_formats/         # Concrete inference format strategy implementations
+│   ├── direct_json/           # Self-contained Direct JSON format package
+│   │   ├── format             # DirectJsonFormat, DirectJsonFormatFactory
+│   │   ├── prompt_generator   # DirectJsonPromptGenerator
+│   │   ├── parser             # DirectJsonParser class
+│   │   └── streaming          # DirectJsonStreamProcessor class
+│   └── experimental/          # Experimental inference format packages
```
jacobsimionato: Let's omit the experimental folder and put express at the top level. We can update Python to match this later. CC @gspencergoog

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1210 @@
...
+│           ├── compiler       # ElementalCompiler class
+│           ├── decompiler     # ElementalDecompiler class
+│           ├── parser         # ElementalParser class
+│           └── prompt_generator # ElementalPromptGenerator class
+├── parser/                    # Common Parser contracts and data structures
+│   ├── parser                 # Abstract Parser base class
+│   └── response_part          # RawResponsePart, TextPart, A2uiPart data structures
+├── prompt/                    # Prompt Generation contracts
+│   └── generator              # Abstract PromptGenerator base class
+├── transformers/              # Catalog and Protocol Transformers
```
jacobsimionato: Nit: catalog_transformers perhaps, seeing as we may have other types of transformers eventually.

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1210 @@
...
+
+#### Response Part Structures
+
+```python
+@dataclass
+class RawResponsePart:
+    """Represents an uncompiled part of an LLM response.
+
+    Attributes:
+        text: The conversational text segment preceding or following sentinel tags. Can be empty.
```
jacobsimionato: If this text can be preceeding *or* following sentinel tags, how does a chat-based agent know what order to display the content in? The LLM could output a response that has multiple A2UI blocks in it, and it should get converted to some wire transport protocol and then rendered in the same order that the LLM intended.

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1210 @@
...
+
+        Returns:
+            List of newly parsed ResponsePart objects.
+        """
+        pass
+```
+
+---
+
+### B. Experimental Express Format (`a2ui.inference_formats.experimental.express`)
```
jacobsimionato: Let's not frame this as experimental and just have it as a top-level format

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1210 @@
...
+
+        Returns:
+            List of newly parsed ResponsePart objects.
+        """
+        pass
+```
+
+---
+
+### C. Experimental Elemental Format (`a2ui.inference_formats.experimental.elemental`)
```
jacobsimionato: Remove this format completely!

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1210 @@
...
+
+# 6. Validate output payload messages against active catalog schemas
+for part in response_parts:
+    if part.a2ui:
+        session.validate_response(part.a2ui)
+```
+
+---
+
+## 6. Rejected Alternatives & Design Iterations
```
jacobsimionato: I think these can be omitted from the blueprint? There are so many different alternatives in a design this large, so I'm not sure if it makes sense to try to list them. If there are specific parts of the implementation that are counterintuitive and we want to carefully steer coding agents to not fall into design pitfalls, I think we are better of mentioning them in the relevant section of the document, rather than at the bottom.

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1210 @@
...
+4. **Dedicated `ExampleManager` Class**: REJECTED. Managing prompt examples does not require a complex manager abstraction. Developers pass example dictionaries directly to `CatalogConfig` or `PromptGenerator`.
+5. **Monolithic `generate_system_prompt` Function with Rigid Flags**: REJECTED. Forced developers into a rigid system prompt layout. Refactored into modular `PromptGenerator` returning format-specific prompt instruction snippets.
+6. **`StrictnessRemovalTransformer`**: REJECTED. Relaxing schema constraints caused contract drift between renderer and agent. Enforces strict schema adherence against negotiated specification contracts.
+7. **`MessagePruningTransformer` as a `CatalogTransformer`**: REJECTED. Message pruning operates on payload envelope types rather than component definitions. Functionality was moved directly into `DirectJsonPromptGenerator.generate(allowed_messages=...)`.
+8. **`InferenceFormatType` Closed Enum Selection**: REJECTED. Python `Enum` classes cannot be extended by third-party developers, locking users into fixed built-in formats. Replaced by extensible `InferenceFormat` strategy instances and factories.
+
+---
+
+## 7. Conformance Test Plan
```
jacobsimionato: Is there a readme explaining how to set up the conformance tests, e.g. the harness etc? It could be good to link from that here.

Maybe if we link to that, then there's no point in summarizing the content here because the agent will read the source of truth.

jacobsimionato (approved):
Yay! I left a couple of other design comments on the Google Doc version of this which you could address if it makes sense to you - I didn't want to repeat them all here. This looks good enough to me as a starting point though! Thank you for iterating!! We can always update this anyway as you do the first implementation of it.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: sounds good. Cleaned up.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: done

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: done

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: updated

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: removed. The idea to add them here was to prevent agent from re-exploring directions that have been rejected.
But the direction is approved in the API design doc. It makes sense to only keep this section in the doc, rather than in the blueprint.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: added link and removed details.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: done

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: It is now in A2uiGenerator when it creates an A2uiPrcoessor instance.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: It is indeed misleading.

In the SDK, `unwrap(content)` and `parse_response(content)` return an ordered list (`List[ResponsePart]`) rather than a single response object. In each `RawResponsePart(text=..., a2ui_raw=...)`: `text` is always the conversational text preceding `a2ui_raw`.
Any trailing text after the final tag becomes its own list item at the end where `a2ui_raw=None`.

For example, if the LLM emits text after an A2UI block:
```
<a2ui-json>[createSurface(...)]</a2ui-json>
Would you also like me to reserve a hotel in Rome? Here are top options:
```

`unwrap()` tokenizes this sequentially into:

```python
[
  # 1. First tag at index 0 -> preceding text is empty ""
  RawResponsePart(text="", a2ui_raw="[createSurface(...)]"),
  # 2. Trailing text after tag -> a2ui_raw is None
  RawResponsePart(text="Would you also like me to reserve a hotel in Rome? Here are top options:", a2ui_raw=None),
]
```

Because `parse_response` returns these in strict chronological order, any wire transport or caller iterating over for part in response_parts: will transmit and render the UI blocks and chat bubbles in the exact order intended by the LLM.

I’ve updated the RawResponsePart docstring in the blueprint to make this explicit:
```
text: The conversational text segment preceding a format block (or trailing text after the last block when a2ui_raw is None). Can be empty.
```

nan-yu:
Addressed all review comments from this PR and from the API Design doc.

blueprints/modules/a2ui_agent.blueprint.md
jacobsimionato: Cool! My personal preference would still be for RawResponsePart to contain only Text *or* A2UI rather than potentially both, because that is a simpler mental model for my brain. What you have here represents exactly the same data though, so just an optional idea.

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1010 @@
...
+   - **Prompt Generators**: Format builders consuming transformed catalogs and prompt examples to generate system instruction snippets.
+   - **Parsers**: Response extraction engines performing tag unwrapping (`unwrap`), streaming chunk processing (`process_chunk`), syntax compilation (`compile`) and decompilation (`decompile`).
+   - **Validation Layer**: Leverages core `A2uiValidator` capabilities directly from `a2ui_core`, natively supporting protocol version branching (`v0_8`, `v0_9`, `v0_9_1`, `v1_0`).
+2. **Encapsulated Application Processor**:
+   - `CatalogConfig`: Configuration dataclass encapsulating catalog providers (`BundledCatalogProvider`, `FileSystemCatalogProvider`, `InMemoryCatalogProvider`), custom transformers, and examples.
+   - `A2uiGenerator`: Agent-level lifecycle manager holding supported `CatalogConfig`s, generating pre-negotiated `A2uiProcessor` instances per renderer capability signature.
+   - `A2uiProcessor`: Central processor facade object unifying multi-catalog capability resolution (`resolve_catalogs`), system prompt snippet rendering, turn-scoped parser creation, and response validation.
```
jacobsimionato: Nit: I know we have discussed this endlessly already, but I can't help pointing out that A2uiProcessor is a very similar name to A2uiMessageProcessor, and I worry this could create confusion. A2uiHandler? A2uiRequestProcessor?

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1010 @@
...
+@dataclass
+class RawResponsePart:
+    """Represents an uncompiled part of an LLM response.
+
+    Attributes:
+        text: The conversational text segment preceding a format block (or trailing text after the last block when a2ui_raw is None). Can be empty.
+        a2ui_raw: The raw uncompiled format content string (e.g., raw XML/DSL/JSON). None if conversational text.
+        is_final: Whether this format-content block is complete/closed (not truncated).
+    """
+    text: str = ""
```
jacobsimionato: Why empty string as the default for `text` rather than making it optional? Conceptually, it seems like sometimes the model will return with just a2ui, so you might expect that this would be null?

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1010 @@
...
+class Parser(ABC):
+    """Abstract base class for response parsers.
+
+    Responsible for tokenizing LLM output streams, unwrapping format tags, and compiling raw format
+    expressions into standard A2UI payload messages.
+    """
+
+    @abstractmethod
+    def wrap(self, blocks: List[RawResponsePart]) -> str:
+        """Wraps multiple decompiled blocks with the format's enclosing tags/markers."""
```
jacobsimionato: Does this also pass through the text content? E.g. if I call `wrap(unwrap(someContent))`, I should get back something very close to what I passed in? The doc string seems to imply that it it only does wrapping, but it also does concatenation too.

Perhaps the docstring could say something like:
"Converts a list of RawResponseParts to a string, adding enclosing tags or markers around each A2UI section"

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1010 @@
...
+        format_factory: Optional[InferenceFormatFactory] = None,
+    ):
+        """Initializes A2uiProcessor, resolving active catalogs and instantiating validator and format strategy.
+
+        Args:
+            catalogs: List of active Catalog instances.
+            examples: Optional dictionary of prompt examples.
+            format_factory: Format factory for instantiating format strategies.
+        """
+        self._catalogs = catalogs
```
jacobsimionato: I think we can probably omit the implementation of functions in most cases, instead just putting a placeholder with a comment explaining what the implementation does, to make it clear which are abstract methods (e.g on Parser) and which are implemented. Agents can always look at the Python implementation etc as a reference when they are working. Otherwise, this implementation can easily get out of sync with the code. I think the APIs are the most important.

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1010 @@
...
+    def examples(self) -> Optional[Dict[str, List[AgentToRendererMessage]]]:
+        """Returns the prompt example mapping."""
+        return self._examples
+
+    @property
+    def prompt_snippet(self) -> str:
+        """Format-specific system prompt instruction snippet."""
+        return self._prompt
+
+    def _validate_examples(self, examples: Optional[Dict[str, List[AgentToRendererMessage]]]) -> None:
```
jacobsimionato: We could perhaps omit private functions here too, or at lest their implementation.

blueprints/modules/a2ui_agent.blueprint.md
```diff
@@ -0,0 +1,1010 @@
...
+```
+
+---
+
+### B. EXPRESS Format (`a2ui.inference_formats.express`)
+
+Compact functional DSL format designed to reduce output token consumption.
+
+```python
+class ExpressFormatFactory(InferenceFormatFactory):
```
jacobsimionato: Seeing as this just implements the interfaces described in detail above, and the A2UIExpress format has it's own detailed specification, I think we can probably omit the code examples here and just say what files should exist and what classes they should implement, then point to the express specification. While there is a lot to implement here, I think it should be easy to ensure the implementations are correct via conformance tests.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: That makes sense! Enforcing mutual exclusion between text and raw A2UI is much cleaner, and it mirrors ResponsePart (which is already a union of TextPart and A2uiPart).

I've updated RawResponsePart to be:

```python
@dataclass
class RawResponsePart:
    part: Union[TextPart, RawA2uiPart]
    is_final: bool = True
```

jacobsimionato (approved):

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: Renamed  to  to avoid any naming confusion with .

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: Refactored  to use a polymorphic structure ( and ), eliminating the optional/default  field.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: Updated the  docstring to clarify that it concatenates conversational text parts while wrapping raw A2UI sections with format sentinel tags.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: Removed concrete method implementations throughout the blueprint, replacing them with abstract signatures and  placeholders.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: Omitted private helper function implementations from the blueprint specification.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: Omitted redundant code examples for Express format, replaced with a concise summary of files/classes and links to the source of truth specification and grammar.

nan-yu: Omitted redundant code examples for Express format, replaced with a concise summary of files/classes and links to the source of truth specification and grammar.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: I renamed from handler to processor for naming consistency between agent and renderer.

Agreed that A2uiProcessor seems too generic. Changed to A2uiRequestProcessor.

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: It was on an outdated version.

I've updated RawResponsePart to be:

```python
@dataclass
class TextPart:
    text: str


@dataclass
class RawAuiPart:
    a2ui_raw: str

@dataclass
class RawResponsePart:
    part: Union[TextPart, RawA2uiPart]
    is_final: bool = True
```

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: removed implementation

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: updated

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: removed

blueprints/modules/a2ui_agent.blueprint.md
nan-yu: updated docstring

PR #1994, opened 2026-07-13
feat(swift): define A2UI v0.9.1 common type schemas with swift-json-schema

pinieb:
# Description
Implements all 14 A2UI v0.9.1 common type JSON Schema definitions using
the swift-json-schema fork. Schemas are defined as a raw JSON document
matching the specification's common_types.json exactly, and registered
into Context via the remoteSchema parameter for $ref resolution.

# Changes:
- Define complete common types document in A2UICommonSchema (all 14
  defs: ComponentId, AccessibilityAttributes, ComponentCommon, ChildList,
  DataBinding, DynamicValue, DynamicString, DynamicNumber, DynamicBoolean,
  DynamicStringList, FunctionCall, CheckRule, Checkable, Action)
- Add A2UISchemaRegistry helper to create a Context pre-populated with
  A2UI schemas
- Add 31 tests covering URI helpers, document structure, schema registry,
  validation of all schema types (positive and negative cases), and
  cross-reference resolution (including recursive Action → FunctionCall
  → DynamicValue chains)
- Remove placeholder A2UIJSONTests (superseded by comprehensive tests)

# Design notes:
- Raw JSON string approach used instead of JSONValue dictionary literals
  to avoid Swift type-checker issues with deeply nested expressions
- FunctionCall's catalog.json oneOf reference omitted; catalog-level
  validation handled at runtime by ComponentCatalog
- Context.remoteSchemaStorage is internal in the fork, so schemas are
  registered via the Context(remoteSchema:) initializer

# Verification:
- swift test passes (34 tests in 3 suites, 0.017s)


swift/core/Sources/A2UIJSON/A2UICommonSchema.swift
```diff
@@ -33,4 +34,295 @@ public enum A2UICommonSchema {
   public static func uri(for name: String) -> String {
     "\(baseURI)#/$defs/\(name)"
   }
+
+  /// The raw JSON string of the complete common types document.
+  ///
+  /// This document contains all 14 A2UI v0.9.1 common type schema
+  /// definitions under the `$defs` key, with cross-references using
+  /// `$ref` URIs relative to `baseURI`.
+  static let rawDocument = """
```
jacobsimionato, line 43: Is there some way we can have some codegen solution that pulls this schema from the `specification` folder, to reduce the duplication here? I think codegen is better than using an asset or something, because we want this available synchronously, for convenience.

After a schema is finalized, we don't typically make any changes to the structure of it, but sometimes we might change the comments on the fields, which can affect inference performance in some cases.

Not a huge deal though. And another option would be to write a test that dumps the schema from Swift and verifies that it matches the central JSON.

jacobsimionato (approved):

swift/core/Sources/A2UIJSON/A2UICommonSchema.swift
jacobsimionato: BTW we can definitely add this later if it's possible - do submit to reduce the length of your PR chain!

pinieb: Definitely! I'll need to do something similar with the example JSON streams in my sample app when I get going on the Basic Catalog, so I'll solve it for both in one shot :)

I've opened https://github.com/a2ui-project/a2ui/issues/2034 to track it!

PR #1999, opened 2026-07-14
feat(swift): add A2UICore error types, JSONValue path extensions, and message models

pinieb:
# Description
Ports the foundational A2UICore types from the prototype: error types,
JSONValue path-based subscripting, and all server/client message models.

# Changes
- Add 4 error types: ClientServerError, GenericError,
  ValidationFailedError, LocalFunctionError (now Sendable)
- Add JSONValue+Path extensions with type-safe accessors and JSON
  Pointer get/set subscripting, adapted for the fork's
  OrderedDictionary-based JSONValue (including .integer case handling)
- Add Resolved protocol and ResolvedAction model with identity-based
  equality
- Add 6 message types: EnvelopeMessage, CreateSurfaceMessage,
  DeleteSurfaceMessage, UpdateComponentsMessage, UpdateDataModelMessage,
  ClientToServerMessage (with v0.9/v0.9.1 version validation)
- Add OrderedJSON and JSONSchema product dependencies to A2UICore target
- Add 50 tests covering type accessors, path operations, message
  decoding/encoding round-trips, and error cases
- Remove placeholder A2UICore.swift stub and placeholder test

# Verification
- swift test passes (84 tests in 5 suites)


swift/core/Sources/A2UICore/Messages/ClientToServerMessage.swift
```diff
@@ -0,0 +1,115 @@
...
+    let version = try container.decode(String.self, forKey: .version)
+    guard version == "v0.9" || version == "v0.9.1" else {
+      throw DecodingError.dataCorruptedError(
+        forKey: .version,
+        in: container,
+        debugDescription: "Unsupported version: \(version)"
+      )
+    }
+
+    if container.contains(.action) {
```
jacobsimionato: The reported `action` payload structure seems to deviate from the standard `client_to_server.json` schema. The Core SDK blueprint (Section 3, "Client-to-Server Events") defines the action payload as a flat object containing `name`, `surfaceId`, `sourceComponentId`, `timestamp`, and `context`, which is also how it is implemented in `web_core`. Here, the code decodes/encodes it with nested keys (`event` or `call`/`args`), which aligns with component action properties rather than the reported event payload.

swift/core/Sources/A2UICore/Messages/EnvelopeMessage.swift
```diff
@@ -0,0 +1,80 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// A container message enclosing one of the supported incoming
+/// server-to-client commands.
+public enum EnvelopeMessage: Codable, Sendable, Equatable {
```
jacobsimionato: The `EnvelopeMessage` type currently does not decode or encode the "version" field at the top level of the JSON payload. While decoding tolerates extra keys, omitting "version" when encoding makes the output invalid according to the official `server_to_client.json` schema, which is strictly validated in `web_core`. It would be safer to encode the version to maintain consistency with the protocol schema.

swift/core/Sources/A2UICore/Extensions/JSONValue+Path.swift
```diff
@@ -0,0 +1,266 @@
...
+        if index <= array.count {
+          if isLastComponent {
+            if let newValue {
+              if index == array.count {
+                array.append(newValue)
+              } else {
+                array[index] = newValue
+              }
+            } else if index < array.count {
+              array.remove(at: index)
```
jacobsimionato: Setting an array index to `nil` currently removes the element and shifts subsequent elements, shrinking the array size. The Core SDK blueprint explicitly specifies under "JSON Pointer Implementation Rules" that setting an array index to undefined must preserve the array length (creating a sparse array). This is also consistent with the `web_core` data model implementation. Removing the element shifts indices and can corrupt other bindings pointing to specific indices.

swift/core/Sources/A2UICore/Extensions/JSONValue+Path.swift
```diff
@@ -0,0 +1,266 @@
...
+        default:
+          return nil
+        }
+      }
+      return currentValue
+    }
+    set {
+      let components = Self.parsePath(path)
+      guard !components.isEmpty else {
+        if let newValue, case .object = newValue {
```
jacobsimionato: Replacing the root node (when path is "/" or empty) is restricted only to `.object` values. For consistency with the `web_core` implementation—which allows replacing the root of the data model with any JSON value type—it would be more flexible to allow setting the root to any valid `JSONValue` (e.g. array or primitive) at this utility layer.

swift/core/Sources/A2UICore/Extensions/JSONValue+Path.swift
```diff
@@ -0,0 +1,266 @@
...
+            components: remainingComponents,
+            newValue: newValue
+          )
+        }
+        return .object(dict)
+      }
+
+    default:
+      if newValue == nil { return node }
+      if let index = Int(key), index == 0 {
```
jacobsimionato: Auto-vivification of arrays currently only triggers if the key is exactly "0". For consistency with standard JSON Pointer auto-vivification rules and the `web_core` implementation (which vivifies arrays for any numeric segment, not just index "0"), standardizing this check to any valid integer string (`Int(key) != nil`) would allow vivifying arrays for any index slot.

jacobsimionato (commented):
Feedback on Swift Core implementation inconsistencies compared to the blueprint and web_core reference implementation.

swift/core/Sources/A2UICore/Messages/ServerToClientMessage.swift
jacobsimionato, line 1: Can we rename this to ServerToClientMessage for consistency with the spec?

swift/core/Sources/A2UICore/Messages/ClientToServerMessage.swift
pinieb: Great catch! I've updated the code and also updated the AGENTS.md file for Swift with some language about referencing the web core implementation for behavior so that they hopefully stay more on track in the future.

jacobsimionato (approved):
Thanks for iterating!!

PR #2005, opened 2026-07-15
feat(swift): add component catalog protocols and resolved model types

pinieb:
# Description

Introduce the component catalog system and resolved model types that
SurfaceViewModel will consume in a future PR.

# Changes:
- Add catalog protocols: LocalFunction, ComponentCatalog (schema lookup,
  theme parsing, function registration), ActionHandling
- Add resolved models: Node (immutable thread-safe component node with
  children accessor), DataBinding (generic two-way binding with
  path/literal identity), ComponentProperties (validated unresolved
  properties), SurfaceTheme (marker protocol)
- Add Resolved conformances for primitives (String, Double, Int, Bool,
  JSONValue) and Array where Element == Node
- Adapt ComponentCatalog and ComponentProperties to use fork's Schema
  type instead of prototype's JSONSchema
- Add 33 tests with concrete test implementations (TestCatalog,
  ConcatFunction, IsEmptyFunction, TestTheme) covering schema lookup,
  theme parsing, function evaluation, node equality, data binding
  get/set, and component properties

# Verification:
- swift test passes (117 tests in 10 suites)


swift/core/Sources/A2UICore/Models/Node.swift
jacobsimionato, line 1: @andrewkolos is centralizing the design on this in https://docs.google.com/document/d/1aA_2ysLn_6NpRet2CtqII90SfPc1wCXZNVnPEmEJZDE/edit?resourcekey=0-sIEWbOROTyd9XRDWwS3tZQ&tab=t.0#heading=h.8ng5v0gkam8a

It'd be great to iterate towards consistency here. I think the biggest difference between this and Andrew's design is that the design has the properties as an observable.

swift/core/Sources/A2UICore/Models/DataBinding.swift
```diff
@@ -0,0 +1,64 @@
...
+// limitations under the License.
+
+import OrderedJSON
+
+/// A thread-safe, generic two-way data binding.
+///
+/// `DataBinding` connects a component property to a value in the data
+/// model via a JSON Pointer path, or wraps a literal value. It provides
+/// get/set closures for reading and writing the bound value.
+public struct DataBinding<Value: Sendable>: Sendable {
```
jacobsimionato, line 22: @andrewkolos do we have anything like this currently?

I think I might prefer it to what we have in the web_core generic binder, but the inconsistency also bothers me.

jacobsimionato (commented):
I think I like this, but let me take another look tomorrow!

swift/core/Sources/A2UICore/Models/Node.swift
pinieb: At least for SwiftUI, the whole Node struct ends up more or less inheriting its observability from further up the chain. Essentially, any change we make to the Node gets immediately reflected in the View. That code will land in a future PR though, so I have nothing to point at in this one.

We may need to add a specific observability wrapper for UIKit, but I'm not sure about that yet

swift/core/Sources/A2UICore/Catalogs/ComponentCatalog.swift
```diff
@@ -0,0 +1,40 @@
...
+/// A registry and factory for components, schemas, and local functions.
+///
+/// Implementations provide component schemas (compiled from JSON Schema
+/// definitions), theme parsing, and local function lookup.
+public protocol ComponentCatalog: Sendable {
+  /// Returns the compiled schema for the given component type.
+  ///
+  /// - Parameter type: The component type identifier.
+  /// - Returns: A `Schema` if registered, otherwise `nil`.
+  func schema(forType type: String) -> Schema?
```
jacobsimionato: This should be based on https://github.com/a2ui-project/a2ui/blob/main/blueprints/modules/a2ui_core.blueprint.md#5-the-catalog-api--functions

Can we call this `component(forName name: String)`? Because functions also have schemas etc. Components have "names" in A2UI lingo - see https://github.com/a2ui-project/a2ui/blob/main/blueprints/modules/a2ui_core.blueprint.md#componentapi

swift/core/Sources/A2UICore/Catalogs/ComponentCatalog.swift
```diff
@@ -0,0 +1,40 @@
...
+  ///
+  /// - Parameter jsonObject: The raw theme JSON value.
+  /// - Returns: A `SurfaceTheme` if parsing succeeds, otherwise `nil`.
+  func makeTheme(jsonObject: JSONValue) -> (any SurfaceTheme)?
+
+  /// Returns a local function implementation.
+  ///
+  /// - Parameter name: The function name.
+  /// - Returns: A `LocalFunction` if registered, otherwise `nil`.
+  func localFunction(for name: String) -> (any LocalFunction)?
```
jacobsimionato: `function(for name: String) -> Function` ?

Or do we need to use `localFunction` to avoid colliding with a reserved word?

swift/core/Sources/A2UICore/Catalogs/LocalFunction.swift
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+import OrderedJSON
+
+/// Abstract definition of local validation or transformation functions.
+///
+/// Implementations are registered in a `ComponentCatalog` and invoked
+/// by the `SurfaceViewModel` when evaluating `FunctionCall` dynamic
+/// values.
+public protocol LocalFunction: Sendable {
```
jacobsimionato: Can this be just `FunctionApi` or `FunctionImplementation`?

It needs to have name, return type, schema on it, for validation purposes and for inference later.

See https://github.com/a2ui-project/a2ui/blob/main/blueprints/modules/a2ui_core.blueprint.md#5-the-catalog-api--functions

swift/core/Sources/A2UICore/Models/Node.swift
```diff
@@ -0,0 +1,54 @@
...
+      }
+    }
+    return true
+  }
+}
+
+extension Node {
+  /// Convenience helper to access resolved child nodes.
+  public var children: [Node] {
+    properties["children"] as? [Node] ?? []
```
jacobsimionato: The specification does not require that `child` or `children` properties have those names. E.g. you could have a "FoodAndDrinkCard" which accepts "foodChildren" and "drinkChildren" arrays. So I think this needs to be designed so that any logic scans through the properties to find children in resolved values.

swift/core/Sources/A2UICore/Models/Resolved+Conformances.swift
jacobsimionato, line 1: Wow, this is neat!

swift/core/Sources/A2UICore/Catalogs/ComponentCatalog.swift
```diff
@@ -0,0 +1,40 @@
...
+  ///
+  /// - Parameter type: The component type identifier.
+  /// - Returns: A `Schema` if registered, otherwise `nil`.
+  func schema(forType type: String) -> Schema?
+
+  /// Parses a raw theme payload into a theme.
+  ///
+  /// - Parameter jsonObject: The raw theme JSON value.
+  /// - Returns: A `SurfaceTheme` if parsing succeeds, otherwise `nil`.
+  func makeTheme(jsonObject: JSONValue) -> (any SurfaceTheme)?
```
jacobsimionato: Shouldn't this be a `theme` accessor?

jacobsimionato (approved):
I do think it's is worth making the Catalog API match the blueprint a little more closely! But LGTM to unblock you once you make those changes.

swift/core/Sources/A2UICore/Catalogs/ComponentCatalog.swift
pinieb: How would you feel about naming it `componentSchema(forName)` instead of just `component(forName)`?

Since we need to split the catalog into a data catalog and a view catalog to prevent creating a deeply-nested tree of `AnyView` instances in SwiftUI, I'd like the naming to make it clear that this API doesn't return a full "Component" (view and all), but just its registered `Schema`

jacobsimionato: Could we use the terms ComponentApi and ComponentImplementation?

Even if you split up the data and view catalogs, I'd highly recommend having a first-class object to represent a component in each of them, even if it only has one property for now, because there's a good chance that we might want to add additional properties in the future. For example, we might want to add a concept of components being able to export functions, or components being marked as "hidden" to the LLM, or logging configuration etc. Having an object to represent a Component makes the design closer to other platforms and gives us this flexibility.

swift/core/Sources/A2UICore/Models/Node.swift
jacobsimionato: No worries. Sounds like this is a special case for Swift which makes sense.

swift/core/Sources/A2UICore/Catalogs/ComponentCatalog.swift
pinieb: Sure, that makes sense to me!

swift/core/Sources/A2UICore/Catalogs/ComponentCatalog.swift
pinieb: I've updated this to match the Catalog API surface in the blueprint :)

swift/core/Sources/A2UICore/Catalogs/ComponentCatalog.swift
pinieb: I've updated this to return `(any FunctionImplementation)?`

swift/core/Sources/A2UICore/Models/Node.swift
pinieb: I've updated the naming to `allChildNodes` and now have the scanning logic you mentioned here.

swift/core/Sources/A2UICore/Models/DataBinding.swift
pinieb: I've opened https://github.com/a2ui-project/a2ui/issues/2084 to track this decision. I'll leave the code as-is for the moment, but once we figure out which approach we want to go with, I'll make a PR for that change!

PR #2019, opened 2026-07-16
remove freestanding catalogs requirement in v1.0

wrenj:
For v1.0 - Catalogs may now use `$ref` to reference other catalogs. This removes the requirement that catalogs be freestanding / have no external references except to core libraries.


jacobsimionato:
I worry this makes the catalog ecosystem too closely tied to JSON schema. If we allow arbitrary $refs and $defs, then we are either on the hook to build infra to look up those catalogs at runtime from the network, or else developers need to manually provide files for each of them, which seems to undo the benefits of having one catalog that neatly encapsulates others. Let's discuss! I have a doc with some ideas which I'd like to get your thoughts on.

jacobsimionato:
Allowing references between catalogs by resolvable URL is also a security concern: https://chat.google.com/room/AAQACTY_CBc/QJaGfHvEBgc/QJaGfHvEBgc?cls=10

PR #2020, opened 2026-07-16
Allow mixing of catalogs, and use catalogId to resolve name conflicts

wrenj:
## Summary of Changes

This pull request updates the A2UI protocol to support the simultaneous use of multiple catalogs within a single UI surface. By requiring an explicit `catalogId` on individual components and function calls, the change eliminates potential name collisions and removes the need for surface-level catalog definitions. These modifications improve the scalability of the protocol, allowing for more complex, multi-catalog UI implementations while maintaining strict component resolution.

### Highlights

* **Catalog Mixing**: Enabled the mixing of multiple catalogs within a single surface, allowing for more flexible UI composition.
* **Explicit Catalog Resolution**: Introduced a mandatory `catalogId` property for all components and function calls to ensure unambiguous resolution across mixed catalogs.
* **API Simplification**: Removed the `catalogId` property from the `createSurface` message, shifting the responsibility of catalog identification to individual components and function calls.
* **Component Constraints**: Added a `rootOnly` boolean property to component definitions to restrict specific components to top-level placement.


specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -428,10 +428,20 @@ Each object in the `components` array of an `updateComponents` message defines a
 
 - `id` (`ComponentId`, required): A unique string that identifies this specific component instance. This is used for parent-child references.
 - `component` (string, required): Specifies the component's type (e.g., `"Text"`).
+- `catalogId` (string, optional): A string that uniquely identifies the catalog for this component, overriding the surface's default catalog. Useful when combining components from multiple catalogs.
+- `rootOnly` (boolean, optional, default `false`): If true, this component must be the top-level element and cannot be placed inside a `ChildList`. Top-level root components (e.g., a `Canvas`) will generally have this set to `true`.
```
jacobsimionato: To me, this seems to be leaking the details of specific catalogs into the core specification, so I'd prefer to not add this. Ultimately, `rootOnly` is something that a catalog can describe, and the most valuable place to enforce it is on the agent (allows for retries) or on the orchestrator (defend against malformed content). Agent developers and orchestrators can implement this validation in code on their side if they want as an extension to A2UI.

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -428,10 +428,19 @@ Each object in the `components` array of an `updateComponents` message defines a
...
 
 This structure is designed to be both flexible and strictly validated.
 
+#### Mixable catalogs and component collision resolution
+
+Clients can support components from multiple catalogs simultaneously (mixable catalogs). When a client declares `supportedCatalogIds`, it means components from any of those catalogs can be used. In the event of a component name collision (e.g., two catalogs define a `Button` component), the renderer uses the following resolution order:
+
+1. The component's `catalogId`, if provided.
+2. The `createSurface` message's `catalogId`, if provided.
+3. If neither resolves the collision unambiguously, the renderer returns an error (ambiguous catalog component).
```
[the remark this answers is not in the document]
jacobsimionato: I think we should *never* do automatic resolution, because it means that an agent needs to know the exact set of catalogs that the client supports in order to know whether an explicit catalogId is required. When the client's set of catalogs expands, this can create new collisions that break existing agents and cached content unexpectedly.

E.g.
1. Client supports standard catalog with "heading" component
2. Agent is developed which uses "heading" with no catalog ID specified.
3. User uses agent, and content is cached at orchestrator level
4. New client version released which now supports "advanced_catalog" with colliding "heading" component
5. Existing cached content can no-longer be rendered because the "heading" component reference is now ambiguous
6. New requests to the existing agent now also return ambiguous content, because the agent developer did not know that the client added support for an additional "heading" component.

As time goes on, I expect collisions to be frequent, because as soon as we run out of options to expand material catalog in a backwards-compatible way, we may want to introduce a new version of it "material_2" catlaog, which will certainly want to have colliding component names.

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -484,7 +494,7 @@ To ensure catalog schemas can be translated reliably into alternative, LLM-frien
    - All helper properties (such as common properties factored out of catalog items) MUST be inlined directly inside the properties block of each supporting component schema rather than referenced from a shared helper.
 3. **Restricted `$ref` Targets:**
    - Local `$ref` targets are restricted to referencing the catalog's top-level components or functions (e.g., `#/components/Text`, `#/functions/required`).
-   - External `$ref` targets MUST reference the standard types inside `common_types.json` (`https://a2ui.org/specification/v1_0/common_types.json#/$defs/...`), limited to the following allowed schemas:
+   - External `$ref` targets may reference other catalogs. When referencing `common_types.json` (`https://a2ui.org/specification/v1_0/common_types.json#/$defs/...`), targets are limited to the following allowed schemas:
```
jacobsimionato: Allow arbitrary references between catalogs makes it too hard for the SDKs to serialize and deserialize catalogs in a reliable and predictable way.

Now that we are allowing mixing and matching of catalogs, can't we just let the SDK ingest multiple catalog files and work from there?

specification/v1_0/docs/a2ui_protocol.md
wrenj: It ended up here because its in common_types, since thats where we define Component and ChildList. I'm fine removing these details from the a2ui_protocol.md spec if its too wordy here.

given the change (we mix all the components in catalogs together) we need some mechanism to say don't mix these catalogs/components. eg, gemini consumer has action cards, they add material3 catalog but don't want action cards nested in there. 

why first class support? 
- its managing interactions between catalogs
- there is no way to describe this in json schema (it won't look up component ids)
- it should be generally useful for removing theme/createSurface properties, since we can cleanly say here is a top level component for storing whatever properties

specification/v1_0/docs/a2ui_protocol.md
wrenj: i removed it from this PR (i had accidently created this branch off another one instead of main). that was a separate request from mandar since he had some use cases in mind. we can discuss on that PR https://github.com/a2ui-project/a2ui/pull/2019. I'm don't have a strong opinion here (I haven't seen any catalogs need to $ref other catalogs)

specification/v1_0/docs/a2ui_protocol.md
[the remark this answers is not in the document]
wrenj: sg, updated to require catalogId on component, and remove it from createSurface

specification/v1_0/docs/evolution_guide.md
```diff
@@ -39,7 +41,10 @@ Version 1.0 differs from 0.9 in the following ways:
...
 - Added `callFunction` message structure (`CallFunctionMessage`) to support server-initiated function execution. Removed `callableFrom` and `returnType` properties from the wire payload, relying on runtime catalog verification.
-- Updated the `createSurface` message (`CreateSurfaceMessage`) to rename the `theme` field to `surfaceProperties`, and allowed passing initial `components` and `dataModel` directly inside the payload.
+- Updated the `createSurface` message (`CreateSurfaceMessage`) to rename the `theme` field to `surfaceProperties`, allowed passing initial `components` and `dataModel` directly inside the payload, and removed `catalogId`.
+- Added a required `catalogId` to the base component definition (`ComponentCommon`) to cleanly isolate component namespaces.
+- Added a required `catalogId` to the `FunctionCall` definition so functions can be unambiguously resolved across catalogs.
+- Added a `rootOnly` boolean (default: false) to `ComponentCommon` to indicate if a component must be the top-level element. This allows restricting top-level components (like a Canvas) from being nested inside a `ChildList`.
```
gspencergoog:
```suggestion
- Added a `rootOnly` boolean (default: false) to `ComponentCommon` to indicate if a component must be the top-level element. This allows restricting top-level components (like a `Canvas`) from being nested inside a `ChildList`.
```

gspencergoog: I'm not sure I understand how `rootOnly` works.  Is it just saying "This component must have the component id '`root`'"?  If that's what it means, I find that a little weird, and it seems redundant. Can you explain what the motivation for it is?

specification/v1_0/docs/a2ui_protocol.md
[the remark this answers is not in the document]
jacobsimionato: In the long term, I definitely prefer what you have now, e.g. each Component has it's own `catalogId` baked in. It's a simple mental model that way - no tiered resolution logic. So I'm happy with this as-is!

My main fear is that requiring catalog ID on each component causes an immediate regression in inference performance for any people doing Gen UI with the JSON format who migrate to v1.0 protocol. Though, realistically, most people are not doing request-time Gen UI, and those that are can either stick with v0.9, or migrate to v1.0 + Express format. WDYT @gspencergoog ?

In my comment above, I actually was actually arguing that it's okay to keep catalogId in `createSurface`, but that if there is no catalogId in either createSurface or on the specific component, then we should error out rather than automatically picking from a catalog, which the original text implied.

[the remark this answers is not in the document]
gspencergoog: Why can't `catalogId` be optional on `createSurface`, where if it is specified it becomes the default for everything that doesn't specify it. Then those doing inference with the JSON format can optimize if they have only one catalog by specifying the "default" catalog id on the `createSurface`, and individual catalog ids on things that aren't the default.  It even seems sensible as an optimization for the transport protocol, since then you can reduce the payload size (although I guess in practice gzip will do that for you).

And yeah, if you don't specify one on `createSurface` _or_ on the component, then it's an error because it's not defined by the time you need it.

specification/v1_0/docs/evolution_guide.md
wrenj: I took this out, we can discuss in another PR to get the namespace change in at least

wrenj: took rootONly out of this PR

jacobsimionato:
While you were out, I tried to get the less controversial subset of your changes submitted in https://github.com/a2ui-project/a2ui/pull/2079. I didn't add you as a reviewer because you were out and I know that you wanted to go in this direction. I think what I have is a bit more complete with tests etc, so I'd like to submit that if you're okay with it.

Here is a quick summary of the similarities and differences from Gemini:

## What is common

* Both add an optional `catalogId` property to `ComponentCommon` and `FunctionCall` in `common_types.json`.
* Both make `catalogId` optional on `createSurface`, changing the required array from `["surfaceId", "catalogId"]` to `["surfaceId"]`.
* Both update capability schemas to clarify that `supportedCatalogIds` can be mixed within a single surface.
* Both update `a2ui_protocol.md` and `evolution_guide.md` to state that a component's explicit `catalogId` overrides the surface default `catalogId`.

## How they differ

* PR 2020 directly adds `catalogId` to `FunctionCall`, which fails JSON Schema validation under strict `unevaluatedProperties: false` rules. PR 2079 solves this by introducing `FunctionCommon` in `common_types.json` and updating all catalog functions in `catalog.json` to inherit it via `allOf`.
* PR 2079 explicitly excludes system functions like `@index` from inheriting `FunctionCommon`, preventing them from accepting a `catalogId`. PR 2020 does not restrict this.
* In `updateComponents`, `updateDataModel`, and `deleteSurface`, PR 2020 deletes the entire sentence requiring `createSurface` to be sent first. PR 2079 preserves the lifecycle requirement by changing the sentence to require only that `createSurface` was sent for the `surfaceId`.
* PR 2020 briefly describes a two-step resolution order for components. PR 2079 defines a three-step resolution order for both components and function calls, adding an explicit callout that there is no fallback to capabilities if neither an explicit nor a surface default `catalogId` is present.
* PR 2079 requires that all catalogs mixed within a surface conform to the same A2UI specification version. PR 2020 does not mention version rules.
* PR 2020 modifies legacy client and server files (`client_capabilities.json`, `server_to_client.json`). PR 2079 modifies updated renderer and agent files (`renderer_capabilities.json`, `agent_to_renderer.json`) and replaces lingering `clientCapabilities` references in `a2ui_protocol.md`.
* PR 2079 updates `schema_helper.py` in the Python SDK to traverse `allOf` subschemas when parsing catalog function arguments. PR 2020 contains no SDK changes.
* PR 2079 adds test cases across four test files to verify optional surface defaults, valid component and function overrides, and rejecting `catalogId` on `@index`. PR 2020 includes no tests.

wrenj:
> While you were out, I tried to get the less controversial subset of your changes submitted in #2079. I didn't add you as a reviewer because you were out and I know that you wanted to go in this direction. I think what I have is a bit more complete with tests etc, so I'd like to submit that if you're okay with it.
> 
> Here is a quick summary of the similarities and differences from Gemini:
> 
> ## What is common
> * Both add an optional `catalogId` property to `ComponentCommon` and `FunctionCall` in `common_types.json`.
> * Both make `catalogId` optional on `createSurface`, changing the required array from `["surfaceId", "catalogId"]` to `["surfaceId"]`.
> * Both update capability schemas to clarify that `supportedCatalogIds` can be mixed within a single surface.
> * Both update `a2ui_protocol.md` and `evolution_guide.md` to state that a component's explicit `catalogId` overrides the surface default `catalogId`.
> 
> ## How they differ
> * PR 2020 directly adds `catalogId` to `FunctionCall`, which fails JSON Schema validation under strict `unevaluatedProperties: false` rules. PR 2079 solves this by introducing `FunctionCommon` in `common_types.json` and updating all catalog functions in `catalog.json` to inherit it via `allOf`.
> * PR 2079 explicitly excludes system functions like `@index` from inheriting `FunctionCommon`, preventing them from accepting a `catalogId`. PR 2020 does not restrict this.
> * In `updateComponents`, `updateDataModel`, and `deleteSurface`, PR 2020 deletes the entire sentence requiring `createSurface` to be sent first. PR 2079 preserves the lifecycle requirement by changing the sentence to require only that `createSurface` was sent for the `surfaceId`.
> * PR 2020 briefly describes a two-step resolution order for components. PR 2079 defines a three-step resolution order for both components and function calls, adding an explicit callout that there is no fallback to capabilities if neither an explicit nor a surface default `catalogId` is present.
> * PR 2079 requires that all catalogs mixed within a surface conform to the same A2UI specification version. PR 2020 does not mention version rules.
> * PR 2020 modifies legacy client and server files (`client_capabilities.json`, `server_to_client.json`). PR 2079 modifies updated renderer and agent files (`renderer_capabilities.json`, `agent_to_renderer.json`) and replaces lingering `clientCapabilities` references in `a2ui_protocol.md`.
> * PR 2079 updates `schema_helper.py` in the Python SDK to traverse `allOf` subschemas when parsing catalog function arguments. PR 2020 contains no SDK changes.
> * PR 2079 adds test cases across four test files to verify optional surface defaults, valid component and function overrides, and rejecting `catalogId` on `@index`. PR 2020 includes no tests.

great thanks for adding the complete PR, ill close this

PR #2038, opened 2026-07-17
chore(ci): fix docs build_and_deploy skip check on PRs

gspencergoog:
Refactor docs.yml workflow to trigger on all pull requests and use dorny/paths-filter to conditionally execute the build_and_deploy job. This ensures the required status check gets reported as 'Skipped' on pull requests without documentation changes, rather than blocking the PR merge in a pending state.


josemontespg (approved):
Nice, thanks for the quick fix!

PR #2052, opened 2026-07-20
feat(swift): implement SurfaceViewModel core engine with dynamic value resolution

pinieb:
Ports the SurfaceViewModel — the central runtime engine that manages
component buffers, data model state, and resolved node tree building.

Changes:
- Add SurfaceViewModel: ObservableObject with thread-safe
  NSRecursiveLock-based access, @published rootNode, component buffer
  (raw JSONValue), data model with path-based updates, and theme state
- Implement classifySchema by inspecting Schema.jsonValue as raw JSON
  to identify A2UI property types (DynamicBoolean, DynamicString,
  DynamicNumber, DynamicValue, Action, ChildList) via $ref, oneOf,
  anyOf, and allOf traversal
- Implement resolveNode for recursive Node tree building with
  per-property classification and resolution
- Implement evaluateDynamicValue for resolving literals, DataBinding
  path lookups, and FunctionCall invocations
- Implement type-specific DataBinding resolvers (Bool, String, Double,
  JSONValue) with get/set closures and auto-rebuild
- Implement resolveAction for server events and client function calls
  with deferred context resolution
- Implement resolveChildList for static arrays and dynamic templates
  with data model iteration
- Add 16 tests with concrete TestSurfaceCatalog, TestActionHandler, and
  TestConcatFunction covering component updates, data model updates,
  dynamic value resolution, action resolution, child lists, and themes

Verification:
- swift test passes (133 tests in 11 suites)


swift/core/Sources/A2UICore/Errors/ClientServerError.swift
```diff
@@ -0,0 +1,43 @@
...
+
+  public init(from decoder: Decoder) throws {
+    let container = try decoder.singleValueContainer()
+    let discriminator = try container.decode(Discriminator.self)
+    if discriminator.code == ValidationFailedError.errorCode {
+      self = .validationFailed(try container.decode(ValidationFailedError.self))
+    } else {
+      self = .generic(try container.decode(GenericError.self))
+    }
+  }
```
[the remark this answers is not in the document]
pinieb: I disagree. This sacrifices modularity and code cleanliness for some dubious future using a custom decoder.

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
```diff
@@ -0,0 +1,632 @@
...
+import OrderedJSON
+
+/// Manages the active component buffer, two-way data model state, and
+/// active theme.
+///
+/// `SurfaceViewModel` is the central runtime engine that receives
+/// server-to-client messages, validates component definitions, resolves
+/// dynamic values (data bindings and function calls), and builds a
+/// tree of ``Node`` objects ready for rendering.
+public final class SurfaceViewModel: @unchecked Sendable, ObservableObject {
```
jacobsimionato, line 30: This is quite different to the architecture we have for the other libraries, which use `MessageProcessor`. I can't remember what we had in the design, but what was the rationale for making it so different?

Design: https://github.com/a2ui-project/a2ui/blob/main/blueprints/modules/a2ui_core.blueprint.md#4-the-processing-layer-messageprocessor

Some issues I see with this are:
- This is specific to a surface, but A2UI messages can address a range of different surfaces. So I assume the application developer would need to manually route messages between surfaces.
- This is quite a large monolithic class. We are planning to build conformance tests for the core libraries which will apply to specific classes, e.g. MessageProcessor, SurfaceModel etc. A different structure will be hard to rig up to these.
- There is no clear programmatic API for directly updating the system state. With SurfaceModel, ComponentModel etc, there is a way that developers can ergonomically get specific parts of the data and access/update it programmatically to do custom stuff.

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
```diff
@@ -0,0 +1,632 @@
...
+  private func classifySchema(_ schemaJSON: JSONValue) -> PropertyType {
+    // Check for $ref to A2UI common types
+    if let ref = schemaJSON["$ref"]?.stringValue {
+      if ref.contains("DynamicBoolean") { return .dynamicBoolean }
+      if ref.contains("DynamicString") { return .dynamicString }
+      if ref.contains("DynamicNumber") { return .dynamicNumber }
+      if ref.contains("DynamicValue") { return .dynamicValue }
+      if ref.contains("Action") { return .action }
+      if ref.contains("ChildList") { return .childList }
+    }
```
andrewkolos: I generally like to avoid `.contains` calls whenever possible.

In this specific case, `DynamicStringList` would be catch on `DynamicString`. This might not be a concern yet since it looks like a lot of types are intentionally not modeled yet.

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
```diff
@@ -0,0 +1,632 @@
...
+
+    return .standard
+  }
+
+  // MARK: - Node Resolution
+
+  /// Resolves a component by ID, using the component ID as both
+  /// definition and instance ID.
+  private func resolveNode(id: String, basePath: String? = nil) -> Node? {
```
andrewkolos: nit/edge case: I don't see any cycle/depth guard logic in here, so a malformed payload with cyclic references could cause the app to hang. I'm completely fine with putting this aside for now.

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
```diff
@@ -0,0 +1,632 @@
...
+
+  /// Retrieves a thread-safe copy of the active theme.
+  public func getActiveTheme() -> (any SurfaceTheme)? {
+    lock.withLock { activeTheme }
+  }
+
+  // MARK: - Tree Rebuilding
+
+  /// Rebuilds the node tree and publishes the new root.
+  private func rebuildTree() {
```
jacobsimionato: Let's leave this for now, but in the shared design from @andrewkolos , we're leaning towards putting the node resolution stuff in a separate class, mostly for separation of concerns. The model classes mostly represent data, and the node resolution follows references to build the tree representation.

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
jacobsimionato: Hey I actually totally misunderstood this class. I see, this is equivalent to SurfaceModel and SurfaceComponentsModel and part of MessageProcessor in the shared design. Please ignore the comment above - I have some other smaller comments.

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
```diff
@@ -0,0 +1,632 @@
...
+    self.surfaceID = surfaceID
+    self.catalog = catalog
+    self.actionHandler = actionHandler
+  }
+
+  // MARK: - Public API
+
+  /// Updates the component buffer with new component declarations after
+  /// validating them.
+  public func updateComponents(_ components: [[String: JSONValue]]) {
```
jacobsimionato: This class absorbs the functionality of SurfaceComponentsModel and ComponentModel in typescript, which handle representing the individual components. Is there a reason not to follow that approach?

That approach makes each model class simpler in itself, and it centralized the parsing and validation of the JSON content in MessageProcessor, which is kind of nice, because MessageProcessor is then equipped to output validation errors neatly etc. While the logic is not so complex now, it will get more complex as we need to support multiple protocol versions in a single codebase, so I'd recommend separating these out.

This is fine with me as a starting point though if you prefer to submit and then iterate, seeing as I know you're juggling a lot of PRs!

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
andrewkolos: Agreed. Perhaps we could reference our prior art for inspiration--`NodeGraph`[^1] in our Python code on main, and `NodeResolver` in my Dart[^2] and TS[^3] prototypes (we'll converge on a single name and implementation details later).

This approach might also buy us more performant tree rebuilds since these maintain persistent nodes rather than rebuilding the entire tree, for example, every time the user sends a keystroke to a TextField. Specifically, each node has its own reactive `props`, and each bound property subscribes to the specific data path it reads, and a write recomputes only that path.

This branches also have lots tests that could be ported to help increase our confidence in this code as well.

[^1]: https://github.com/a2ui-project/a2ui/blob/58f0d01937287fa7d703eb1c65565ce105fc0701/agent_sdks/python/a2ui_core/src/a2ui/core/state/node_graph.py#L28
[^2]: https://github.com/andrewkolos/genui/blob/c44d14ec366de689d3fd586749a9d55e46bdb131/packages/a2ui_core/lib/src/nodes/node_resolver.dart#L63
[^3]: https://github.com/andrewkolos/A2UI/blob/9a83ab2986518e38bd938a4219226a54cc64a089/renderers/web_core/src/v0_9/nodes/node-resolver.ts#L69

andrewkolos (commented):
More comments coming in a bit :)

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
```diff
@@ -0,0 +1,632 @@
...
+    definitionID: String,
+    instanceID: String,
+    basePath: String?
+  ) -> Node? {
+    guard let componentJSON = componentBuffer[definitionID],
+      let componentDict = componentJSON.dictionaryValue,
+      let type = componentDict["component"]?.stringValue
+    else {
+      return nil
+    }
```
andrewkolos: It's not uncommon for 1) an `updateComponents` message to describe a component that reference other components as its children and 2) the client hasn't heard a thing about those children yet (that often arrives in a follow-up `UpdateComponents` message that comes right after the first).

In this scenario, wouldn't the lookup return `nil` and every caller just drops its job of rendering this subtree  because `resolveChildList` skips nil children.

In this scenario, our prior art implements progressive rendering, putting placeholders in place of children that have yet to arrive. I don't remember off the top of my head how much better this tends to look compared to the more classical approach of just having components pop in as they arrive, so your mileage may vary.

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
pinieb: Good call, I've updated this logic to look for an exact match on the last segment of the fully specified type name.

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
pinieb: Great catch. One of my early prototypes had cycle detection, but it looks like it got lost along the way. I've added a `visited` set of instance IDs

jacobsimionato (approved):

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
pinieb: I've opened https://github.com/a2ui-project/a2ui/issues/2089 to track this. I think it will be easier to figure out the best/most consistent approach once I have a sample app and Basic Catalog to play with :)

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
pinieb: I've separated out the logic for SurfaceComponentsModel and ComponentModel into their own files

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
pinieb: https://github.com/a2ui-project/a2ui/issues/2090 to track this

PR #2058, opened 2026-07-21
feat(swift): Add A2UICore message processing with SurfaceGroupModel and MessageErrorMapper

pinieb:
Introduces the Swift A2UICore message processing pipeline, architected
to align with the canonical `web_core` TypeScript implementation.

### `SurfaceGroupModel`
Owns surface storage, lifecycle, and lookup. Tracks a per-surface
`sendDataModel` flag and aggregates flagged surfaces' data models via
`getClientDataModel()`. Thread-safe with an internal lock and
main-thread-published `surfacesMap`.

### `MessageErrorMapper`
A `Sendable` value type that converts `DecodingError` into
spec-compliant `ClientServerError` values (`.validationFailed` with
JSON pointer path, or `.generic`). Only decode failures are routed
through the action handler; semantic errors (surface/catalog not
found) throw directly, matching `web_core`'s pattern.

### `MessageProcessor`
Thin orchestrator that delegates surface storage to
`SurfaceGroupModel` and error conversion to `MessageErrorMapper`.
Handles `createSurface`, `updateComponents`, `updateDataModel`, and
`deleteSurface` messages. The `sendDataModel` flag on
`createSurface` sets a group-level flag instead of synthesizing a
synthetic `ResolvedAction`.

### `MessageParser`
Parses `ServerToClientMessage` from JSON strings or `Data`.
Includes `extractSurfaceID(fromLine:)` for best-effort surface ID
extraction when decoding fails, enabling proper error attribution.

## Tests
- `MessageParserTests` (5 tests)
- `MessageProcessorTests` (14 tests)
- `SurfaceGroupModelTests` (5 tests)
- `MessageErrorMapperTests` (4 tests)
- Shared test helpers extracted into `TestHelpers.swift`
- All 178 tests pass across 16 suites


pinieb:
/gemini please review again. That first review was before the latest commit

swift/core/Sources/A2UICore/Extensions/JSONValue+Path.swift
```diff
@@ -0,0 +1,286 @@
...
+    }
+  }
+
+  /// Returns the underlying integer value if this is an `.integer` or
+  /// a whole-number `.number` case.
+  public var intValue: Int? {
+    switch self {
+    case .integer(let value): return value
+    case .number(let value):
+      if value >= Double(Int.min) && value <= Double(Int.max) && value == value.rounded() {
```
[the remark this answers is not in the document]
pinieb: I'm going to move to using the failable `Int(exactly)` initializer instead of managing these boundaries directly.

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -0,0 +1,157 @@
...
+
+  // MARK: - Deprecated Forwarding Methods
+
+  /// The dictionary of active surfaces, published to the UI.
+  ///
+  /// - Deprecated: Access via `processor.group.surfacesMap`.
+  @available(*, deprecated, message: "Use group.surfacesMap")
+  public var surfaces: [String: SurfaceViewModel] {
+    group.surfacesMap
+  }
```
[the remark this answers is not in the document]
pinieb: I've deleted this property altogether.

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
jacobsimionato, line 1: This PR LGTM overall - thanks for coming closer to the shared design! I'll take another look hit approve on this once we can remove some diff bases!

swift/core/Sources/A2UICore/Models/SurfaceGroupModel.swift
```diff
@@ -0,0 +1,110 @@
...
+
+  // MARK: - Surface Lookup
+
+  /// Retrieves a surface by its ID.
+  public func surface(id: String) -> SurfaceViewModel? {
+    lock.withLock { surfaces[id] }
+  }
+
+  /// Returns a snapshot of all active surfaces.
+  public func allSurfaces() -> [String: SurfaceViewModel] {
```
jacobsimionato: Could it make sense to just expose an immutable map like this as a readonly property, and avoid exposing surface and allSurfaces functions?

swift/core/Sources/A2UICore/Models/SurfaceGroupModel.swift
```diff
@@ -0,0 +1,110 @@
...
+    lock.withLock { surfaces }
+  }
+
+  // MARK: - send Data Model
+
+  /// Marks a surface as requesting data-model reporting.
+  public func setSendDataModel(surfaceID: String, enabled: Bool) {
```
jacobsimionato: Why not expose this on SurfaceViewModel? The message processor can do the work to find the correct SurfaceViewModel and set this property, rather than us needing to plumb everything via this class.

swift/core/Sources/A2UICore/Models/SurfaceGroupModel.swift
```diff
@@ -0,0 +1,110 @@
...
+        sendDataModelSurfaces.remove(surfaceID)
+      }
+    }
+  }
+
+  /// Aggregates the data models of all surfaces that have
+  /// `sendDataModel` enabled.
+  ///
+  /// Returns `nil` if no surfaces have the flag set.
+  public func getClientDataModel() -> JSONValue? {
```
jacobsimionato: I think this is exposed on MessageProcessor in the shared design. Can we just expose it there?

swift/core/Sources/A2UICore/Processing/MessageParser.swift
```diff
@@ -0,0 +1,63 @@
...
+    try decoder.decode(ServerToClientMessage.self, from: jsonData)
+  }
+
+  /// Best-effort extraction of a `surfaceId` from a raw JSON line.
+  ///
+  /// This is used as a fallback when `parse(jsonString:)` fails, so
+  /// the caller can still attribute the error to the correct
+  /// surface. Returns `nil` if the JSON is malformed or contains no
+  /// `surfaceId` field.
+  public func extractSurfaceID(fromLine line: String) -> String? {
```
jacobsimionato: Can this be private and used from within parse(), to return a better exception? Just want to reduce our API surface to maintain and make life easy for developers.

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -12,138 +12,149 @@
...
     self.catalogs = catalogs
     self.actionHandler = actionHandler
+    self.surfaceGroupModel = SurfaceGroupModel()
   }
 
-  // MARK: - Surface Management
+  // MARK: - Surface Lookup & Management
 
   /// Returns the surface model for the given ID, if it exists.
   public func getSurface(_ id: String) -> SurfaceViewModel? {
```
jacobsimionato: SurfaceGroupModel already exists to expose the individual surface, so let's just expose that here, rather than the specific surfaces.

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -12,138 +12,149 @@
...
-    guard let catalog = catalogs.first(where: { $0.id == catalogID }) else {
-      return nil
-    }
+    guard let catalog = catalogs[catalogID] else { return nil }
     return createSurface(surfaceID: surfaceID, catalog: catalog)
   }
 
+  /// Creates a surface directly with a given catalog instance.
+  @discardableResult
+  public func createSurface(
```
jacobsimionato: How about making most of the below private, to minimize the public API surface that we need to keep stable, and ensure that there is one canonical way of doing things (e.g. `processor.surfaceGroup.createSurface(new SurfaceModel)`) etc.?

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -231,67 +236,89 @@ public final class MessageProcessor: @unchecked Sendable {
   }
 
   /// Updates a path in a surface's data model.
-  ///
-  /// - Parameters:
-  ///   - surfaceID: The surface to update.
-  ///   - path: The JSON Pointer path.
-  ///   - value: The value to set, or `nil` to remove.
   public func updateDataModel(
```
jacobsimionato: This should be private too. Developers can get the data model object for a specific surface if they want to do stuff like this.

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
jacobsimionato: There is no way to update the theme of a surface in A2UI, because theme is only exposed on CreateSurface, so I don't think we need this or getActiveTheme. These is just an immutable property of a SurfaceViewModel.

jacobsimionato: I think this should be private, right? The tree should automatically rebuild when data changes.

jacobsimionato, line 86: BTW is it necessary to support updates triggered from background queues here and throughout the SDK? It makes the logic throughout the SDK harder to reason about, because writes are only actioned after an async gap.

jacobsimionato: Can this be private?

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -12,138 +12,152 @@
...
-    catalogID: String
-  ) -> SurfaceViewModel? {
-    guard let catalog = catalogs.first(where: { $0.id == catalogID }) else {
-      return nil
-    }
-    return createSurface(surfaceID: surfaceID, catalog: catalog)
-  }
+  /// Throws on any failure (decoding error, missing surface, missing catalog).
+  /// Thrown parsing errors are also routed to `ActionHandling` via `MessageErrorMapper`.
+  public func process(line: String) throws {
```
jacobsimionato, line 77: Nit: In the shared design, we don't have this convenience method here, so wondering if it makes sense to avoid it. It does seem handy though.

jacobsimionato (approved):
Looks great overall - just nits. Approving so you can merge once you address the comments as you see fit!

swift/core/Sources/A2UICore/Models/SurfaceViewModel.swift
pinieb: If we put the tree resolution on the main queue, then there's a risk that it could cause UI hangs and jank, particularly in cases where the generated UI is complex or where the tree is being rebuilt a lot.

For the moment, I'd like to leave the computation on background queues, but I can take a pass in a later PR to make the queue hopping a little easier to reason about.

pinieb: I've opened https://github.com/a2ui-project/a2ui/issues/2124 to track this

pinieb: Good catch!

pinieb: Unmarked functions like this default to `internal` in Swift (only visible to the target they're defined in) so they wouldn't have made it to consumers, but you're right that `private` is more appropriate here.

PR #2065, opened 2026-07-21
Export wrap_as_json_array from a2ui.schema.validator for backward compatibility

nan-yu:
Export wrap_as_json_array from a2ui.schema.validator for backward compatibility


ditman (approved):
This came up during an internal import into google. Thanks for the quick fix @nan-yu!

PR #2070, opened 2026-07-22
spec(v1_0): restore `svgPath` and `DataBinding` support for `Icon.name`

gspencergoog:
This change resolves the unintended loss of data-binding support for the Icon component in v1.0.

- In the v1.0 basic catalog schema, restored the custom SVG path key name to svgPath (porting the fix from PR #1347), reverting the accidental rename to path.

- Restored DataBinding to the oneOf definition of Icon.name so that icon names can continue to be dynamically data-bound.

- Removed custom SVG svgPath -> path rename instructions from evolution_guide.md.

### DIFF
--- a/specification/v1_0/catalogs/basic/catalog.json
+++ b/specification/v1_0/catalogs/basic/catalog.json
@@ -162,12 +162,15 @@
                 {
                   "type": "object",
                   "properties": {
-                    "path": {
+                    "svgPath": {
                       "type": "string"
                     }
                   },
-                  "required": ["path"],
+                  "required": ["svgPath"],
                   "additionalProperties": false
+                },
+                {
+                  "$ref": "https://a2ui.org/specification/v1_0/common_types.json#/$defs/DataBinding"
                 }
               ]
             },
--- a/specification/v1_0/catalogs/basic/catalog.json
+++ b/specification/v1_0/catalogs/basic/catalog.json
@@ -163,7 +163,7 @@
                   "type": "object",
                   "properties": {
                     "svgPath": {
-                      "type": "string"
+                      "$ref": "https://a2ui.org/specification/v1_0/common_types.json#/$defs/DynamicString"
                     }
                   },
                   "required": ["svgPath"],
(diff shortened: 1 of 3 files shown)


specification/v1_0/catalogs/basic/catalog.json
```diff
@@ -162,12 +162,15 @@
...
                       "type": "string"
                     }
                   },
-                  "required": ["path"],
+                  "required": ["svgPath"],
                   "additionalProperties": false
+                },
+                {
+                  "$ref": "https://a2ui.org/specification/v1_0/common_types.json#/$defs/DataBinding"
                 }
```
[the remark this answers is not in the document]
gspencergoog: Added tests.

specification/v1_0/catalogs/basic/catalog.json
```diff
@@ -162,12 +162,15 @@
...
-                      "type": "string"
+                    "svgPath": {
+                      "$ref": "https://a2ui.org/specification/v1_0/common_types.json#/$defs/DynamicString"
                     }
                   },
-                  "required": ["path"],
+                  "required": ["svgPath"],
                   "additionalProperties": false
+                },
+                {
```
jacobsimionato, line 172: Does the express format compiler etc work with this oneOf construct?

It seems kind of bespoke to do a oneof between enum, data binding, or an object that contains an svgPath which itself is a dynamic string.

I think this is okay for now, but I worry this level of schema customization is not sustainable, hence the proposal for go/a2ui-catalogs-without-json-schema

jacobsimionato (approved):

specification/v1_0/catalogs/basic/catalog.json
gspencergoog: I don't think there's anything preventing it, but probably some nuance is lost in translation into the prompt.

The complicating factor here is the interpolation of the catalog schema into the prompt in a form that matches what the Express prompt describes how to generate.  In this case, it would bring in the description of the field, and indicate that it is "dynamic", which the LLM understands because dynamic values are explained in the Express prompt.

I'm not sure if your proposal would be any easier to translate into the prompt, though.  This change practically means that the `Icon.name` value can be a dynamic value, and it's probably easier for the LLM to understand (in few words in the prompt) if all aspects of the parameter can be dynamic.  If we left the `svgPath` as only allowing a string, for instance, I think that would trip it up because it would be told that the `name` was dynamic and then `svgPath` turns out to not exactly be dynamic.

Maybe a better design for this would be to allow `oneOf` `name` or `svgPath` but not both, and then individually the `name` is just a dynamic enum value, and the `svgPath` is just a dynamic string, so we don't need to think about the nested `oneOf`'s.

Issue #1745, opened 2026-06-24
[spec] children are unkeyed, making it impossible to move children correctly


yjbanov:
## Problem

Consider this example from https://a2ui.org/specification/v0.9.1-a2ui/#collection-scopes-relative-paths:

Data model:

```
{
  "company": "Acme Corp",
  "employees": [
    {"name": "Alice", "role": "Engineer"},
    {"name": "Bob", "role": "Designer"}
  ]
}
```

Component definition:

```
{
  "id": "employee_list",
  "component": "List",
  "children": {
    "path": "/employees",
    "componentId": "employee_card_template"
  }
},
{
  "id": "employee_card_template",
  "component": "Column",
  "children": ["name_text", "company_text"]
},
{
  "id": "name_text",
  "component": "Text",
  "text": { "path": "name" }
},
{
  "id": "company_text",
  "component": "Text",
  "text": { "path": "/company" }
}
```

A list of employees is renderer using the "employee_card_template" template component. On the first render everything is correct. However, consider a data update that swaps Alice and Bob:

```
{
  "company": "Acme Corp",
  "employees": [
    {"name": "Bob", "role": "Designer"}
    {"name": "Alice", "role": "Engineer"},
  ]
}
```

Because `List` children are unkeyed by anything, the data for "Bob" will flow into the card that used to render "Alice" and vice versa. This may produce correct visual result in basic (e.g. read-only) cases. However, this breaks in the following situations:

* If a screen reader focus is focused on Bob, instead of the focus moving with the widget to its new location, the data underneath the screen reader focus will unexpectedly be replaced with Alice. There will also be no screen reader notification to the user about the order change.
* If the user made a selection (e.g. on the web), that selection will be lost.
* If employee cards are editable the editing state may get corrupted, or simply confuse the user.
* If employee card widgets are stateful, their state will not move to a new location in the list, but instead will receive data for a wrong employee, which could violate some business rules.

## Proposal

The data optionally supplies an identifier:

```
{
  "company": "Acme Corp",
  "employees": [
    {"id": "123", "name": "Bob", "role": "Designer"}
    {"id": "234", "name": "Alice", "role": "Engineer"},
  ]
}
```

The list provides the property name that should be used as the child key, and the runtime applies it to every child widget:

```
{
  "id": "employee_list",
  "component": "List",
  "children": {
    "path": "/employees",
    "componentId": "employee_card_template",
    "key": "id" // NEW!
  }
}
```

[...]

Issue #1528, opened 2026-06-03
[Proposal] Resolve child components based on schema and not on hardcoded 'child'/'children' keywords. Also improve REF: handling


josemontespg:
# Make component binders schema-driven for child component resolution (remove hardcoded 'child'/'children' keywords)

## Motivation

Currently, both the Angular binder (`ComponentBinder`) and the Generic Binder (`GenericBinder`) rely on hardcoded property names (`child`, `trigger`, `content`, `children`) to identify which properties represent child component references.
This introduces a few limitations:

1. **Lack of Flexibility**: Custom component definitions cannot use alternative property names for child components (e.g., `header`, `footer`, `avatar`, `leftAction`).
2. **Boilerplate and Code Duplication**: Component implementations (such as `TabsComponent` in Angular) are forced to manually normalize child properties nested inside objects or arrays because the binder only resolves top-level properties matching the hardcoded names.
3. **Inconsistency**: Dynamic child lists template are resolved to `{id, basePath}` objects but static child lists (which are just arrays of IDs) are returned as raw string arrays, requiring renderers to write custom fallback code.

---

## Proposed Design (Updated)

We propose making both binders fully schema-driven for child component resolution using a combination of Zod runtime metadata, Zod branded types for compile-time safety, recursive binder resolution, and clean JSON Schema reference generation:

### 1. Tag and Brand Core Schemas via a Custom Helper Function

We define custom decoupled metadata properties on Zod schema definitions (`_def` objects) to control runtime binding and schema reference generation separately:
- `a2uiType`: Declares the binder's runtime mapping behavior (e.g. `CHILD`, `STRUCTURAL`, `ACTION`, `DYNAMIC`).
- `refPath`: Declares the JSON Schema `$ref` pointer path in the generated client catalog.

Instead of globally mutating Zod's prototype class—which creates side-effects, risks library name collisions, hinders tree-shaking, and introduces temporal coupling—we define a clean, side-effect-free **Custom Helper Function** inside `renderers/web_core/src/v0_9/schema/common-types.ts` to attach this metadata to the schemas in a type-safe way:

```typescript
export interface A2uiTypeDef {
  a2uiType?: 'Dynamic' | 'ComponentId' | 'ChildList' | 'Action';
  refPath?: string;
}

/**
 * Attaches A2UI metadata to a Zod schema's internal definition safely.
 */
export function withA2uiMetadata<T extends z.ZodTypeAny>(
  schema: T,
  metadata: A2uiTypeDef
[...]

Issue #1282, opened 2026-04-27
Implement Node Layer Architecture across Core and Renderers


jacobsimionato:
This is a proposal based on a comment that @yjbanov made on the original design for web_core. After thinking about it a bit more and seeing how the libraries evolved, I think he was right!

### Description and Rationale
We are introducing a "Node Layer" architecture into the A2UI core libraries and refactoring the existing rendering frameworks (React, Angular, Lit) to consume it. 

Currently, frameworks are responsible for complex tasks: traversing the flat `SurfaceModel` adjacency list, resolving data bindings using `GenericBinder`, handling template expansions (e.g., `ChildList`), and managing subscription lifecycles. This causes duplicated logic across frameworks, risks memory leaks due to forgotten binder disposals, and makes adding new framework adapters difficult.

The Node Layer centralizes this "business logic" into the core library. It transforms the flat component map into a living, reactive view hierarchy. Renderers will simply receive fully-resolved, type-safe `Node` instances and focus solely on mapping them to native UI primitives (pixels).

### Detailed Architecture & Design
The Node Layer represents a shift from frameworks directly managing components and data to having the `SurfaceModel` manage the entire view hierarchy lifecycle.

#### 1. Core Data Structures: The `Node`
The new architecture introduces a `Node<TProps>` interface. A `Node` represents a living, fully resolved component instance in the view hierarchy. By utilizing generics, the `Node` interface provides type-safe property access to the rendering frameworks.
- **`instanceId`**: A stable, unique identifier for this instance in the rendered tree. For templated nodes, this incorporates the data path (e.g., `'item-card-[/users/0]'`) to ensure React/Flutter keys remain stable.
- **`props`**: A reactive `Signal<TProps>`. Dynamic values are primitive strings/numbers/booleans, Actions are callable `() => void` closures, and Structural props (e.g., `child` or `children`) contain actual child `Node` references (not string IDs).
- **Lifecycle**: `onDestroyed` event and a `dispose()` method.

#### 2. Internal Node Engine inside SurfaceModel
The management of the Node tree is an implementation detail inside `SurfaceModel`. Because `SurfaceModel` encapsulates both the `DataModel` and the `SurfaceComponentsModel`, it is perfectly positioned to handle this.
- **Public API**: `SurfaceModel` exposes a single, reactive `rootNode: Signal<Node | undefined>`. Frameworks simply subscribe to this signal.
[...]

PR #2077, opened 2026-07-23
Node layer for web_core + React

andrewkolos:
Implements #1282 for TypeScript,

Adds an experimental layer to `@a2ui/web_core` that resolves a surface's components and data into a live tree, plus a React surface that renders from that tree.

Today, [`A2uiSurface`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/react/src/v0_9/A2uiSurface.tsx) renders by recursing over component ids: each component wrapper runs its own [`GenericBinder`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/web_core/src/v0_9/rendering/generic-binder.ts#L176) against raw definitions, child references stay strings until a wrapper renders them, and any renderer must reimplement the same resolution. The new [`NodeResolver`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/web_core/src/v0_9/nodes/node-resolver.ts#L84) moves that work into `web_core`: it turns a surface's flat component map into a live tree of [`ComponentNode`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/web_core/src/v0_9/nodes/component-node.ts#L57)s, exposed through a reactive [`rootNode`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/web_core/src/v0_9/nodes/node-resolver.ts#L89) signal. In a resolved node's props:

- child-reference properties hold live `ComponentNode` objects (template [`ChildList`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/web_core/src/v0_9/schema/common-types.ts#L132)s spawn one node per array item);
- every dynamic value is a [`ResolvedBinding`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/web_core/src/v0_9/nodes/resolved-binding.ts#L30): a snapshot of the current value. A path-bound payload yields the `WritableBinding` subtype carrying `set` (`isWritable()` narrows), while literal- and function-call-valued properties stay read-only, so an unchecked write is a type error instead of a silent no-op;
- actions are ready-to-call closures, with context resolved at dispatch time.

A node's props signal emits only when that node's own resolved properties change; child-internal changes stay on the child's signal. Every unresolved node carries a `state`: missing components are `pending` placeholders upgraded in place on arrival, a component whose type has no catalog entry is an `unknown-type` node the surface renders as an error (with `UNKNOWN_COMPONENT_TYPE` dispatched), and cyclic references render as placeholders with a dispatched [`CYCLIC_REFERENCE`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/web_core/src/v0_9/nodes/node-resolver.ts#L360) error. Node identity is parent-scoped, so one component id mounted in two positions yields two nodes and dropping one never tears down the other.

[`A2uiNodeSurface`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/react/src/v0_9/A2uiNodeSurface.tsx) (in `@a2ui/react`) is a drop-in sibling of `A2uiSurface` that renders this tree. Each component subscribes only to its own node's props, so a data change re-renders exactly the affected component. Existing view code runs unchanged: a generated per-implementation [`view`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/react/src/v0_9/adapter.tsx#L37) converts node props back to the value + `set<Prop>` and string-id shapes current views expect, and child references the catalog schema does not declare fall back to the existing [`DeferredChild`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/react/src/v0_9/A2uiSurface.tsx) recursion, so third-party catalogs render as-is.

## Behavioral changes

- **Client capabilities output changes for the basic catalog.** Per-usage `.describe()` calls on `ComponentIdSchema`/`ChildListSchema` were dropping the `REF:` pointer, so the basic catalog's eight child-reference properties (across Card, Row, Column, List, Tabs, Modal, and Button) were emitted as inline string schemas. They now emit `$ref: common_types.json#/$defs/ComponentId` (or `ChildList`) as intended. This fix stands on its own, independent of the node layer, and addresses the `REF:`-handling part of #1528 (the issue's larger ask, schema-driven child resolution in the existing binders, is untouched here).
- `GenericBinder` caches action closures across rebuilds, so resending an identical action-bearing component no longer emits new props (or re-renders) on either surface; a changed payload still produces a fresh closure.
- Invoking a catalog function that has no implementation now throws an `A2uiExpressionError` naming the function as schema-only. On the rendering path this surfaces as the same `EXPRESSION_ERROR` client error as before, with a clear message where a `TypeError`'s used to be.
- New exports, all experimental: the node layer (`ComponentNode`, `NodeResolver`, `ResolvedBinding`, and support types) from `@a2ui/web_core/v0_9`, and `A2uiNodeSurface` with its view types from `@a2ui/react/v0_9`. The two `index.ts` diffs are the full list.
- The React shell sample renders through `A2uiNodeSurface` when `?nodes` is in the URL; default rendering is unchanged.
- No breaking API changes: the new type parameters default to current behavior, existing implementations without a `view` render through the fallback, and `A2uiSurface` is untouched.

Known limitation: a template-spawned node's [`instanceId`](https://github.com/a2ui-project/a2ui/blob/0d46ea47/renderers/web_core/src/v0_9/nodes/component-node.ts#L70) names a list position, not a data item, so it is not stable across array insertions or reorders. Data-derived child keys depend on #1745.

If this PR is adopted, `A2uiSurface` can take these internals under its existing name and `A2uiNodeSurface` is deleted; the conversion in `node-view.tsx` exists only so current views render unchanged while the two are compared. Later, once catalog components are authored against resolved values directly (the portable web components proposal, #931, could be the place to do this), that conversion and `GenericBinder`'s synthesized `set<Prop>` shape lose their last consumer.


renderers/web_core/src/v0_9/nodes/component-node.ts
```diff
@@ -0,0 +1,185 @@
...
+    this.dataPath = dataPath;
+    this.props = signal(initialProps);
+  }
+
+  get disposed(): boolean {
+    return this._disposed;
+  }
+
+  get isPlaceholder(): boolean {
+    return this.type === PLACEHOLDER_TYPE;
```
jacobsimionato, line 94: What does this mean again? Could you add a doc comment about it?

renderers/web_core/src/v0_9/nodes/component-node.ts
```diff
@@ -0,0 +1,185 @@
...
+    this.cleanups.push(cleanup);
+  }
+
+  /**
+   * Replaces the resolved props, emitting only if a shallow comparison shows
+   * a change. Callers keep unchanged values reference-identical (child nodes
+   * come from the resolver's cache; untouched arrays keep their identity), so
+   * shallow comparison is exact rather than heuristic.
+   */
+  setProps(next: TProps): void {
```
jacobsimionato, line 142: I think we don't want application code to be able to call this. Can we provide a readonly public interface for Node, so that the NodeResolver has write access, but the rest of the application code can only read values and update the *data model* via the bindings in the node, rather than the node props directly?

renderers/web_core/src/v0_9/nodes/component-node.ts
```diff
@@ -0,0 +1,185 @@
...
+    };
+    const props = peekValue(this.props);
+    for (const [key, value] of Object.entries(props)) {
+      serialized[key] = serializeValue(value);
+    }
+    return serialized;
+  }
+}
+
+function serializeValue(value: unknown): unknown {
```
jacobsimionato, line 174: Does this need to be public?

renderers/web_core/src/v0_9/nodes/node-resolver.ts
```diff
@@ -0,0 +1,585 @@
...
+  const trimmed = dataPath.replace(/\/+$/, '') || ROOT_DATA_PATH;
+  return `${componentId}-[${trimmed}]`;
+}
+
+/**
+ * Returns `prev` whenever `next` is structurally identical to it, so
+ * unchanged props keep reference identity across rebuilds. Child
+ * `ComponentNode`s and action closures compare by identity.
+ */
+function stabilize(prev: unknown, next: unknown): unknown {
```
jacobsimionato, line 651: Can this be private?

renderers/web_core/src/v0_9/nodes/node-resolver.ts
```diff
@@ -0,0 +1,585 @@
...
+          allSame = false;
+        }
+      }
+      return allSame ? prev : out;
+    }
+  }
+  return next;
+}
+
+function isPlainObject(value: unknown): value is Record<string, unknown> {
```
jacobsimionato, line 691: Can this be private too?

renderers/web_core/src/v0_9/nodes/ref-fields.ts
```diff
@@ -0,0 +1,160 @@
...
+ */
+const SINGLE_REF_MARKER = 'A2UI_COMPONENT_REF:single';
+const LIST_REF_MARKER = 'A2UI_COMPONENT_REF:list';
+
+/**
+ * Declares a property holding the id of a single child component.
+ * Use in place of a bare string id so the node layer can resolve the child.
+ */
+export function componentReference(): z.ZodTypeAny {
+  return ComponentIdSchema.describe(`${SINGLE_REF_MARKER}|${ComponentIdSchema.description ?? ''}`);
```
jacobsimionato: We already do a hack like this in our schemas here: https://github.com/a2ui-project/a2ui/blob/d4723f29254520e1214d5004cb555d83eaafb828/renderers/web_core/src/v0_9/schema/common-types.ts#L96

We needed it so that when we convert an in-memory Zod catalog to a wire-format catalog, we can replace the subschemas for DynamicString etc with references to common_types.json.

Seems like you needed basically the same thing here - to be able to identify specific special subschemas to understand the meaning of different strings etc.

Could you use the same magic references, e.g. centralize some definition in common-types?

BTW I wrote a proposal at go/a2ui-catalogs-without-json-schema to design our own schema format that would make this less hacky. I'm not a fan of this description hack, but I don't know a way around it without abandoning JSON schema!

renderers/web_core/src/v0_9/nodes/component-node.ts
```diff
@@ -0,0 +1,185 @@
...
+/** The `type` of a node whose component definition has not arrived yet. */
+export const PLACEHOLDER_TYPE = 'Placeholder';
+
+/** Resolved node properties, keyed by the component's schema property names. */
+export type NodeProps = Record<string, unknown>;
+
+/**
+ * One resolved component instance in the rendered tree.
+ *
+ * A node's `props` hold fully resolved values: primitives for dynamic values,
```
jacobsimionato: Can we add some way to update the value of two-way bound variables from the node? E.g. for ChoicePicker, it should be possible to somehow update the bound "value" property of the node. I'm not sure the best way to do this. Some ideas:

A. `props['value']` returns a `BoundValue` object which exposes get and set. If we do this, then I think we should probably always return a BoundValue for properties that are DynamicValues etc in the schema, even if they happen to be populated with a static value at runtime. That way, component implementations can always access the value in the same way e.g. `props['value'].get()` and `props['value'].set('new value')`.

B. Expose setters as additional keys in props e.g. `props['setValue'](updatedValue)`. I think we basically did this in the generic binder.

C. Or expose some other parallel map of setters that is separate to props. E.g. `boundValues['value']('updated value')`.

Vercel did something between A and C I think: https://json-render.dev/docs/registry#component-props

renderers/react/src/v0_9/adapter.tsx
```diff
@@ -28,6 +28,13 @@ export interface ReactComponentImplementation extends ComponentApi {
     context: ComponentContext;
     buildChild: (id: string, basePath?: string) => React.ReactNode;
   }>;
+  /**
+   * The unwrapped view function `render` wraps. `A2uiNodeSurface` renders it
+   * directly with props resolved by the node layer, so the component works
+   * without `render`'s own binder. Absent on binderless implementations,
+   * which the node surface falls back to rendering through `render`.
+   */
+  view?: React.FC<ReactA2uiComponentProps<any>>;
```
jacobsimionato: Would it make sense for this to just accept a node and `buildChild` as the parameter? Then we can have extra sugar in `createComponentImplementation` that exposes props, in a typesafe way?

jacobsimionato (commented):
This is great! Neat way to prototype it!

renderers/react/src/v0_9/A2uiNodeSurface.tsx
```diff
@@ -0,0 +1,210 @@
...
+  }
+  if (value && typeof value === 'object') {
+    const result: Record<string, unknown> = {};
+    for (const [key, inner] of Object.entries(value)) {
+      result[key] = toViewValue(parent, inner, index);
+    }
+    return result;
+  }
+  return value;
+}
```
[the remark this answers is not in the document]
andrewkolos: Addressed in f77010fa5fc31c3282517614ae2dd77cb7ec4661

renderers/react/src/v0_9/A2uiNodeSurface.tsx
```diff
@@ -0,0 +1,210 @@
...
+  );
+  const subscribe = useCallback(
+    (onChange: () => void) => {
+      const resolver = new NodeResolver(surface, surface.catalog);
+      box.resolver = resolver;
+      onChange();
+      const stopEffect = effect(() => {
+        getValue(resolver.rootNode);
+        onChange();
+      });
```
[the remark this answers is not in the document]
andrewkolos: Addressed in 2f9f7d03dd52459ae5eba6da9a1af1d0092303c3

renderers/web_core/src/v0_9/nodes/node-resolver.ts
```diff
@@ -0,0 +1,585 @@
...
+    for (const [edgeKey, child] of record.childEdges) {
+      if (!newEdges.has(edgeKey)) {
+        this.disposeNode(child);
+        if (child.isPlaceholder) {
+          const stillWaiting = [...newEdges.values()].some(
+            other => other.isPlaceholder && other.componentId === child.componentId,
+          );
+          if (!stillWaiting) {
+            this.pendingParents.get(child.componentId)?.delete(record.node);
+          }
```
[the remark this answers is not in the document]
andrewkolos: Addressed in 75fafdc16dcbcbc2c6c48768d49ff7fe229ca198

renderers/web_core/src/v0_9/nodes/component-node.ts
andrewkolos: Done in 11938551ea7abbb83f7f32d6d4dd8d5cca0a37d8

renderers/web_core/src/v0_9/nodes/component-node.ts
andrewkolos: Done in 160a82b6. `ComponentNode` is now read-only with a protected constructor: identity fields, `props`, `onDestroyed`, `addCleanup`, `toJSON`. `setProps`, `dispose` are moved to `MutableComponentNode`, which only the resolver uses and the package does not export, so application code reads props and writes data only through a binding's `set`.

There is a caveat here in that `node.props` is still a `Signal`, so the low-level `setValue(node.props, ...)` remains a bypass for a determined client. We would need to add a read-only signal type to close that. I haven't explored this, but I'm assuming it would be a breaking change.

renderers/web_core/src/v0_9/nodes/component-node.ts
andrewkolos: It's not exported

renderers/web_core/src/v0_9/nodes/node-resolver.ts
andrewkolos: It's not exported

renderers/web_core/src/v0_9/nodes/ref-fields.ts
andrewkolos: Done in 58c3036f. Deleted my `A2UI_COMPONENT_REF` marker and switched to the same magic references you linked--the node layer now finds child refs by the `REF:common_types.json#/...` descriptions, so components just declare them with `ComponentIdSchema` / `ChildListSchema`. I didn't do that originally because `.describe()` at a call site wipes the whole description, magic reference included. That's also why generated capabilities were losing their `$ref`s (#1528). Added `componentIdWithDescription` / `childListWithDescription` that append instead of replace, moved the basic catalog onto them, and added a test that the generated `$ref`s persist.

renderers/web_core/src/v0_9/nodes/component-node.ts
andrewkolos: Done in 3135360a. Every dynamic property now resolves to a `ResolvedBinding` in props, shaped `{ value, set? }`, which is your option A with two changes. First, `set` is absent when the payload supplied a literal or a function call: there's nothing to write to, and the old synthesized setters silently swallowed those writes, so now an unchecked write doesn't compile.

Second, `value` is a snapshot rather than a `get()` closure. A live `get()` reads the data model at whatever moment the component happens to look, so under React's concurrent rendering two components can disagree about one value within a single frame; that's the tearing problem `useSyncExternalStore` exists to prevent, and the node surface already uses it for `rootNode`. A snapshot can't disagree with the emission that delivered it, and the same property means serialized trees and headless reads always describe a state that actually existed. It also keeps `toJSON` useful: a snapshot serializes as the actual value, while a closure can't appear in JSON at all, so it would have to serialize as a stand-in string the way action closures come out as `'<Action>'`, and serialized trees would lose their two-way values. Writes go through the node's data scope, so a component inside a template row writes to that row's item without knowing it's templated.

You're right that option B is what the generic binder already does: it puts extra `setValue`-style keys in props. I didn't touch that here, so nothing views consume changes shape; bindings only exist in node props, which are new here. I didn't want option B for node props because the `set<Name>` keys can only be typed with mapped types, which no other target language has. Moving the binder itself onto bindings would break `A2uiSurface`'s view contract, so that should probably wait for a major version bump.

Reading the json-render docs you linked, I agree that approach lies between A and C. They put the current value in `props` as a plain value and the state path it was bound to in a separate `bindings` map, and their `useBoundProp` hook combines the two into a `[value, setValue]` pair, same shape as `useState`, with the setter writing to that path. When the prop was a literal there's no path, and the setter it hands back does nothing. That's fine in an all-React library, because a React input displays whatever its state variable holds--if the setter does nothing, the variable never changes, the input visibly freezes, and the developer notices immediately. It doesn't carry over to us for two reasons. Flutter disables its selection controls by passing `onChanged: null` (`TextField` does it with `enabled: false`, decided the same way), so a view has to be able to tell "no write destination" apart from "has one", and a no-op function looks identical to a real one; an absent `set` is the signal a view can map onto that (and headless consumers render nothing that could visibly freeze). And the `[value, setValue]` hook idiom is React-specific; Dart and Swift have nothing like it, while a bundled value-plus-set object works as-is in every language I'm aware of.

renderers/react/src/v0_9/adapter.tsx
andrewkolos: Yes, I made this change in cf097093. `view` now takes `{node, buildChild}`, and both factories generate it. The generated wrapper subscribes to the node's props, converts them to the `ReactA2uiComponentProps` shape authors already write against, and resolves child ids through its own index. The author's function stays typed per component through `ResolveA2uiProps`, so the typesafe part of your suggestion is covered. Component code is unchanged, and the node never reaches it. `A2uiNodeSurface` is left as a dispatcher, plus a fallback that renders from raw definitions when an implementation has no `view`.

One can still write a `view` by hand without the factories, and that one takes the node directly. That's fine now that nodes are read-only. I exported `useSignalValue` in 6e4f06e0 so that's actually doable from outside the package.

Down the road I want a mapped type where dynamic props come through as `ResolvedBinding<T>`, so views consume bindings directly and the wrapper stops converting. I didn't add it here because nothing would use it yet. The components that replace the current catalog are the natural first users. Happy to go deeper on any of this.

renderers/web_core/src/v0_9/nodes/component-node.ts
jacobsimionato: This sounds good. I think that this approach of placeholders was used in the web rendering implementations, but is not really defined in a blueprint anywhere. It would be great if as part of your work here, you formalized this concept of placeholder in the blueprint. This then seems like a great starting point for adding APIs to define placeholder rendering behavior.

renderers/web_core/src/v0_9/nodes/component-node.ts
jacobsimionato: Ahhhh I forget that about TS!

renderers/web_core/src/v0_9/nodes/component-node.ts
jacobsimionato: Thank you for this extremely thoughtful analysis. I learned a lot, and I agree with your proposals here!

My only remaining concern is that the `ResolvedBinding` objects create another layer of indirection that might confused developers who are trying to build components which use dynamic values, but only as one-way bindings e.g. they only need to read and never write. In this case, it would be way easier for them if we can just expose primitive values directly in props rather than binding objects that they then need to unwrap.

I imagine we can add more convenience APIs later to improve on that, e.g. a read-only parallel `values` dictionary. I imagine the true solution is to update the catalog format so that catalog API authors indicate which bindings are one-way and which are two-way ahead of time, so that the SDK can automatically treat them different and only expose what is necessary for each.

renderers/web_core/src/v0_9/nodes/component-node.ts
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * Emission contract: `props` emits when this node's own resolved properties
+ * change, including when a child *reference* is replaced (a placeholder
+ * upgrade, a deletion, a list change). It does not emit when a child's
+ * internal properties change; subscribe to the child's `props` for that.
+ *
+ * The resolver creates, updates, and disposes nodes; application code reads
+ * them.
+ */
+export class ComponentNode<TProps extends NodeProps = NodeProps> {
```
jacobsimionato, line 42: How about making this an interface rather than a class, to avoid inheritance and just have a clean central implementation in MutableComponentNode etc?

renderers/react/src/v0_9/A2uiNodeSurface.tsx
```diff
@@ -0,0 +1,144 @@
...
+    }
+    return <View node={node} buildChild={buildChild} />;
+  },
+);
+NodeView.displayName = 'NodeView';
+
+/**
+ * Drop-in alternative to `A2uiSurface` rendering from a `NodeResolver`.
+ */
+export const A2uiNodeSurface: React.FC<{
```
jacobsimionato, line 100: My understanding is that this works with the existing definition of the basic catalog etc, because you managed to adapt the node API right back to the API that that existing components were written against in `node-view.tsx`. This is super neat!

If the above is correct, I think you can just replace the existing surface with this implementation because you've achieved the holy grail of making the architectural change we want, while preserving full compatibility with existing usage.

jacobsimionato (approved):
This is looking very good to me. I'd love at least one of @ditman or @josemontesp to approve before merging!

jacobsimionato:
Adding @ditman and @josemontesp to see what they think!

renderers/web_core/src/v0_9/schema/common-types.ts
```diff
@@ -97,6 +97,25 @@ export const ComponentIdSchema = z
...
 export type ComponentId = z.infer<typeof ComponentIdSchema>;
 
+/**
+ * Describes a component-id property without losing its `REF:` pointer.
+ * `.describe()` replaces the whole description, so calling it directly on
+ * {@link ComponentIdSchema} silently drops the pointer that the capabilities
+ * generator turns into a wire `$ref` and the node layer reads to classify
+ * child-reference properties.
+ */
+export function componentIdWithDescription(description: string): typeof ComponentIdSchema {
```
josemontespg, line 107: We might want other modifications to ComponentIdSchema in the future. To keep this function signature more flexible, I suggest a signature like:

`componentId({description?: 'myDescription'});`

This way we can easily add more modifications by adding optional properties to the function parameter, without having to modify all the call sites.

Same for the other function below.

renderers/web_core/src/v0_9/nodes/resolved-binding.ts
```diff
@@ -0,0 +1,74 @@
...
+ * `{"path": ...}` payload.
+ */
+export class ResolvedBinding<T> {
+  constructor(
+    readonly value: T,
+    readonly set?: (value: T) => void,
+  ) {}
+
+  /** Whether writes have a destination (the payload bound a data path). */
+  get writable(): boolean {
```
josemontespg, line 36: Would be nice to support type narrowing for this. 

I imagine a call site would like to do something like this:

```
if (isWritable(binding)) {
  binding.set('foo'); // notice no need for `!`
}
```

One way to do this is to use inheritance
```
WritableBinding extends Binding {
	constructor(readonly value, readonly set) {
       super(value);
    }
}
```

Then we can use type narrowing for safe access to `set`

renderers/web_core/src/v0_9/nodes/resolved-binding.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * A resolved two-way value in a node's props: a snapshot of the current
+ * value, plus a write capability present only when the payload bound a data
+ * path. `set` is absent for literal and function-call values, so a write
+ * without checking writability is a type error rather than a silent no-op.
+ *
+ * The snapshot is pinned at emission: a new binding arrives through the
+ * node's props whenever the underlying value changes.
+ *
+ * Named `ResolvedBinding` because `DataBinding` is the wire model of the
```
josemontespg, line 26: Is it relevant to adapters that this is a _Resolved_ binding, as opposed to just a binding?

How about naming this "Binding" ?

renderers/web_core/src/v0_9/nodes/resolved-binding.ts
josemontespg: Even better, if the props the components get are already typed with WritableBinding or plain Binding based on the schema, so that call sites don't even need to check isWritable

renderers/web_core/src/v0_9/nodes/node-resolver.ts
```diff
@@ -0,0 +1,705 @@
...
+    this.rootRecord = this.records.get(node);
+    setValue(this.rootNode, node);
+  }
+
+  private onComponentCreated(component: ComponentModel): void {
+    if (this._disposed) {
+      return;
+    }
+    if (component.id === ROOT_COMPONENT_ID && !this.rootRecord) {
+      this.buildRoot();
```
josemontespg, line 161: I see this is called in two places. Could this cause two root nodes being created in some scenario?

renderers/web_core/src/v0_9/nodes/node-resolver.ts
```diff
@@ -0,0 +1,705 @@
...
+    if (!api) {
+      this.surface.dispatchError({
+        code: 'UNKNOWN_COMPONENT_TYPE',
+        message: `Component '${componentId}' has type '${model.type}', which is not in catalog '${this.catalog.id}'.`,
+      });
+      return this.registerNode(
+        new MutableComponentNode(
+          instanceIdFor(componentId, dataPath),
+          componentId,
+          PLACEHOLDER_TYPE,
```
josemontespg, line 253: Should we differentiate between components that have not arrived (placeholders) and components that are not present in the catalog? This is using placeholder now, but maybe we want to use a type that shows there was an error.

renderers/web_core/src/v0_9/nodes/ref-fields.ts
```diff
@@ -0,0 +1,153 @@
...
+ * generator resolves into wire `$ref`s, reused here as the machine-readable
+ * classification source. Catalogs declare single-child properties with
+ * `ComponentIdSchema` (or `componentIdWithDescription`, which keeps the
+ * pointer when adding prose).
+ */
+const COMPONENT_ID_REF = 'REF:common_types.json#/$defs/ComponentId';
+const CHILD_LIST_REF = 'REF:common_types.json#/$defs/ChildList';
+
+/** Which properties of a component's schema reference child components. */
+export interface RefFields {
```
josemontespg, line 31: Nit: Is this intended to be used as a oneof? If so, maybe a type union would be more idiomatic

```
export RefFields = SingleChild | Children | NestedChildren;
```

renderers/web_core/src/v0_9/nodes/ref-fields.ts
```diff
@@ -0,0 +1,153 @@
...
+  /** Properties holding a single child component id. */
+  readonly single: ReadonlySet<string>;
```
josemontespg, line 33: nit: Could this be a string? Or maybe we could combine single and list into one Set

renderers/react/src/v0_9/A2uiNodeSurface.tsx
```diff
@@ -0,0 +1,144 @@
...
+      },
+      [surface, node],
+    );
+
+    if (node.isPlaceholder) {
+      return <div style={{color: 'gray', padding: '4px'}}>[Loading {node.componentId}...]</div>;
+    }
+    const impl = surface.catalog.components.get(node.type);
+    if (!impl) {
+      return <div style={{color: 'red'}}>Unknown component: {node.type}</div>;
```
josemontespg, line 86: I think this case is already handled by the node-resolver.ts, and is setting it to PLACEHOLDER, making this branch of code unreachable. (See my other comment there about using another type for the schema not found case)

renderers/react/src/v0_9/A2uiNodeSurface.tsx
```diff
@@ -0,0 +1,144 @@
...
+    }
+    const View = impl.view;
+    if (!View) {
+      return <RenderFallback surface={surface} node={node} impl={impl} buildChild={buildChild} />;
```
josemontespg, line 90: Could we add the `impl` to the interface of the node? This way renderers won't need to find the impl from the catalog.

Maybe we could also add `surface` to the interface of node. It would be nice if everything a renderer needs to paint a node is only the node instance.

josemontespg (commented):
Very cool PR 👏 Left some comments

renderers/web_core/src/v0_9/schema/common-types.ts
andrewkolos: Done in 2a9f6dc7

renderers/web_core/src/v0_9/nodes/resolved-binding.ts
andrewkolos: Much cleaner--thanks

renderers/web_core/src/v0_9/nodes/resolved-binding.ts
andrewkolos: `Binding` feels ambiguous with the `DataBinding` type we already export

renderers/web_core/src/v0_9/nodes/ref-fields.ts
andrewkolos: I don't think this suggestion would work as-is because a component can have both single child (multiple of them, too) and list props, but we can make the classification at the property level. a4f97a65

renderers/web_core/src/v0_9/nodes/ref-fields.ts
andrewkolos: [`Modal`](https://github.com/andrewkolos/A2UI/blob/b3e86e04c5959d46d00d2e1a9617972872a4d3ed/renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts#L316-L331) has multiple non-list children. I think this concern might be addressed in your other comment on ref-fields.ts

renderers/web_core/src/v0_9/nodes/node-resolver.ts
andrewkolos: Addressed in 89f7ea4b. Nodes carry `state` (`resolved` | `pending` | `unknown-type` | `cyclic`). Thanks to your suggestion I also found out that a pending node whose definition arrived with an unrecognized type would render a placeholder forever.

renderers/react/src/v0_9/A2uiNodeSurface.tsx
andrewkolos: Yes, that branch was unreachable. Addressed in 89f7ea4b, which adds the node `state` field discussed in another comment. The error now renders through an explicit `state === 'unknown-type'` branch.

renderers/react/src/v0_9/A2uiNodeSurface.tsx
andrewkolos: 5488aa64 added `node.impl`, which now gives a view everything it needs to render.

renderers/web_core/src/v0_9/nodes/node-resolver.ts
andrewkolos: There was a guard to avoid this issue, but I've since moved it into `buildRoot` to its more visible.

renderers/react/src/v0_9/A2uiNodeSurface.tsx
josemontespg: Nice, thanks!

renderers/web_core/src/v0_9/nodes/node-resolver.ts
josemontespg: Thanks!

renderers/web_core/src/v0_9/nodes/node-resolver.ts
josemontespg: Nice!

renderers/react/src/v0_9/A2uiNodeSurface.tsx
josemontespg: Nice, thanks for fixing!

renderers/web_core/src/v0_9/nodes/ref-fields.ts
josemontespg: Ah you are right, makes sense.

renderers/web_core/src/v0_9/nodes/ref-fields.ts
josemontespg: Makes sense, thanks for clarifying.

renderers/web_core/src/v0_9/nodes/resolved-binding.ts
josemontespg: Oh I see. Naming is hard 😅 ResolvedBinding SGTM.

renderers/web_core/src/v0_9/nodes/resolved-binding.ts
josemontespg: Awesome!

renderers/web_core/src/v0_9/schema/common-types.ts
josemontespg: Nice, thanks!

renderers/web_core/src/v0_9/nodes/node-resolver.ts
josemontespg, line 1: Would it makes sense to implement the validation feature here? Attaching (valid/errors) to ComponentNode.props

josemontespg (approved):
Looks good to me! Some minor comments left

Nit: Should we update a changelog?

renderers/web_core/src/v0_9/nodes/node-resolver.ts
josemontespg: Could be added in a follow up PR if you want

PR #2032, opened 2026-07-17
docs: add A2UI Ink community renderer

kokoro-ele:
## Summary
- add `kokoro-ele/a2ui-ink` to the Community Renderers table
- document its terminal/CLI platform and A2UI v0.9 support
- link the Apache-2.0 source repository and published `@evanyu/a2ui-ink` npm package

The renderer covers all 18 Basic Catalog components using keyboard-first Ink controls and reuses `@a2ui/web_core` for protocol processing, state, and data binding.

## Test plan
- [x] `yarn prettier --check docs/public/ecosystem/renderers.md`

Related to #1998

PR #2082, opened 2026-07-23
docs: add generative-mui (Material UI renderer) to community renderers

yessGlory17:
Adds **generative-mui** to the Community Renderers list.

**generative-mui** (`@yessglory/generative-mui-react`) is a **Material UI** renderer for A2UI — currently the only one in the list that targets MUI. The agent emits A2UI (a flat, id-referenced adjacency list + a JSON-Pointer-bound data model) and the library turns it into real Material UI components **inside the host app's existing `<ThemeProvider>`**, so the surface has no theme of its own and re-skins top-to-bottom when the host theme changes.

- **Spec:** A2UI v0.9.1 (v0.9 family). v0.8 not supported.
- **Coverage:** all 18 Basic Catalog components mapped 1:1 to MUI, plus an opt-in Extended Catalog (charts via `@mui/x-charts`, Table, Alert, Chip, Stepper, …).
- **Architecture:** framework-neutral core (`@yessglory/generative-mui-core` — Zod schemas, JSONL parser, JSON-Pointer, deterministic subscribable `SurfaceStore`, provider-agnostic agent tool schemas) + a React/MUI adapter, with a lint-enforced one-way `core ← react` dependency so parsing/store/tooling run on server/edge/RSC.
- **Two-way binding & validation:** inputs bound to `{"path":"/field"}`; `Button.checks` disable and surface helper text.
- **Streaming resilience:** incremental skeletons, append-only string deltas, cycle/depth guards, per-node error boundaries.
- **Security:** unregistered component types never execute; `sx`/`style`/`className` from the agent are stripped; ReDoS length cap on local `regex`.

**Links**
- GitHub: https://github.com/yessGlory17/generative-mui
- npm: https://www.npmjs.com/package/@yessglory/generative-mui-react (core: https://www.npmjs.com/package/@yessglory/generative-mui-core)
- License: MIT

I'll also post a short intro in Show and tell as suggested in the submission guide.


yessGlory17:
Thanks for the review — all three applied in the latest commit: dropped the temporal wording ("is a Material UI renderer"), switched the dependency notation to `react → core`, and fixed the term to "JSON Pointer".

yessGlory17:
@googlebot I signed it!

yessGlory17:
@googlebot I signed it!

yessGlory17:
Hi @ditman — this is a docs-only addition to the community renderers list, same file and format as the recent Ink (#2032) and BoteAI (#1674) entries you merged. The CLA check is now green and the gemini-code-assist suggestions have been applied. Whenever you have a moment, a review / workflow approval would be much appreciated. Thanks! cc @yjbanov

ditman:
@yessGlory17 have you considered deploying a demo app demonstrating your catalog? We have a demo app that you may re-skin and use that doesn't use any inference:

* https://github.com/a2ui-project/a2ui/tree/main/renderers/react/a2ui_explorer

ditman:
Note that there's a broken lint:

```
Checking formatting...
[warn] docs/public/ecosystem/renderers.md
```

You can run `./scripts/fix_format.sh` and it should apply the correct formatting everywhere. I'm sure the linter is complaining because the table doesn't have the right amount of whitespace or something in the doc (it's hard to see in github's diff tool)

yessGlory17:
Thanks @ditman! Fixed the formatting — ran the repo's Prettier config over `docs/public/ecosystem/renderers.md` and pushed; my new row was the longest in the first column, so Prettier re-padded the whole table. `prettier --check` is green now.

Good call on a demo. The repo already ships runnable examples that need no inference (they work in a built-in demo mode) — a Vite SPA, a Next.js app, and a `@mui/x-chat` integration that renders the assistant's replies as MUI. I'll look at re-skinning `a2ui_explorer` and putting up a hosted demo so people can try the catalog without cloning. Thanks for the pointer!

ditman (approved):
Let's land this, thanks for your work @yessGlory17, looking forward to your demo!

yessGlory17:
Thank you so much @ditman

PR #2085, opened 2026-07-23
fix(web_core): guard DataModel against prototype pollution and proper…

suyangw-g:
fix(web_core): guard DataModel against prototype pollution and property leakage

- Prevent prototype pollution by validating path segments against forbidden keys ('__proto__', 'constructor', 'prototype') in DataModel.parsePath().
- Use Object.prototype.hasOwnProperty.call() during property traversal in set() and get() to prevent inherited Object.prototype properties from leaking or being misidentified as existing state.
- Update JSDoc method documentation across DataModel to detail @param, @returns, and @throws conditions for forbidden path segments.
- Add comprehensive security and edge-case unit tests in data-model.test.ts.

### DIFF
--- a/lychee.toml
+++ b/lychee.toml
@@ -43,8 +43,9 @@ exclude = [
   "a2ui\\.org/a2a-extension",
   # Internal Google "go/" short links — only reachable by Googlers, by design.
   "^https?://go/",
-  # Valid Google vanity domain (Opal) that rejects/blocks automated checkers.
+  # Valid Google vanity domains / services that reject or block automated checkers.
   "opal\\.google",
+  "gemini\\.google\\.com",
   # Known-dead external references (verified 404, no published replacement).
   # Excluded to keep CI green; drop the referencing content if gone for good.
   "github\\.com/orgs/community/discussions/187776",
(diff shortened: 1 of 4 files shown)


andrewkolos (changes requested):
The two high priority bot comments are legitimate regressions. However, the simplest fix might just be to tighten up `isNumeric` since all array-indexing logic goes through it:

```ts
function isNumeric(value:string): boolean {
  return /^(0|[1-9]\d*)$/.test(value);
}
```

I'm verifying now and will push a commit if this turns out to be the case.

edit: forgot a pipe character

andrewkolos:
> The two high priority bot comments are legitimate regressions. However, the simplest fix might just be to tighten up `isNumeric` since all array-indexing logic goes through it:
> 
> ```ts
> function isNumeric(value:string): boolean {
>   return /^(0|[1-9]\d*)$/.test(value);
> }
> ```
> 
> I'm verifying now and will push a commit if this turns out to be the case.
> 
> edit: forgot a pipe character

for your consideration: 

https://github.com/a2ui-project/a2ui/commit/e879a0b7ab83750fd301e7f5cdd023df4d8802b9

all tests should pass.

andrewkolos (approved):
LGTM upon CI passing. I made a note to myself to come back and double check whether the vulnerability could actually be hit (or at least be latent), but I wanted to get this reviewed and code-correct before a big meeting block.

PR #2088, opened 2026-07-24
Add shared renderer fallback mechanism: web_core contract + React implementation (#2013)

### LINKED ISSUE #2013: [BUG]: React renderer shows developer `[Loading {id}...]` placeholders to end users

## Describe the Bug

The React renderer renders hard-coded developer placeholders directly into the surface tree, and there is no way to customize or disable them.

In `renderers/react/src/v0_9/A2uiSurface.tsx`, `DeferredChild` returns literal debug text when a component is not yet available:

```tsx
const componentModel = surface.componentsModel.get(id);

if (!componentModel) {
  return <div style={{ color: 'gray', padding: '4px' }}>[Loading {id}...]</div>;
}

...
```

Because `A2uiSurface` renders `<DeferredChild id="root" basePath="/" />`, this placeholder sits on the only rendering path and cannot be bypassed. During streaming, when a referenced component has not been created yet, end users briefly see text such as `[Loading root...]`, and `Unknown component: <type>` (red) when a catalog implementation is missing. These are debug-style placeholders (inline `color`/`padding`) that are not appropriate for production UIs.

## Steps to Reproduce

1. Render a surface with `A2uiSurface`.
2. Stream component updates incrementally (e.g., send `createSurface`, then `updateComponents` in separate messages, or add a parent component before its children).
3. Observe the surface between updates.
4. See `[Loading root...]` (and `[Loading <childId>...]` for not-yet-created children) rendered to the user.

## Expected Behavior

While a component is not yet available, the renderer should render nothing by default, or let the consumer provide a fallback. Debug text should not reach end users.

## Screenshots

Image

## Environment Details

- **OS**: macOS
- **Browser/Platform**: Chrome
- **SDK/Package Name & Version**: @a2ui/react v0.10.1, @a2ui/web_core v0.10.3
- **Protocol Version**: v0.9
- **Agent Framework & LLM Model**: ADK

phiftyfifty:
# Description

The React renderer shows developer debug placeholders (`[Loading {id}...]` divs and red `Unknown component:` text) to end users. This PR adds the shared fallback mechanism designed and agreed on in #2013: a small types-only contract in web_core, plus all three web renderers (React, Angular, Lit) implementing it. Fixes #2013.

Per the discussion with @ditman, this PR grew renderer by renderer so review could start early. All four packages are in now: web_core, React, Angular, and Lit, each following the shared contract with its own idiomatic delivery mechanism. Docs and the framework adapter blueprint are included too.

the design comment sketched a single info interface with two states, and the landed version is a three arm discriminated union that adds an `error` state and an optional `parentComponentType` to address feedback on the issue. 

What is in so far:

* **web_core**: `src/v0_9/rendering/fallback.ts`, a types-only discriminated union with three states (`loading`, `unknownComponent`, `error`), an info payload carrying `componentId`, optional `componentType` and `parentComponentType`, and `reason` for the error state. Exported from the v0_9 barrel, additive and non-breaking (verified: existing build and all 271 web_core tests pass unchanged, and the react package typechecks against the new types with zero edits to its own declarations).
* **React**: renders nothing by default in both fallback states, keeps developer signal via `console.warn` (same message text as the Lit renderer), and adds an optional `fallbacks` prop on `A2uiSurface` (`ReactNode` or `(info) => ReactNode` per state) delivered through a context so nested `DeferredChild` children receive it. The `unknownComponent` path also dispatches `surface.dispatchError({code: 'COMPONENT_NOT_FOUND', message, componentId, componentType})` once per distinct unknown type, deferred to a microtask so an `onError` listener can safely set state.
* **Angular**: same defaults, nothing renders and the existing console warnings still fire. Consumers pass optional `loadingTemplate` and `unknownComponentTemplate` `TemplateRef` inputs on `SurfaceComponent`, rendered through `ngTemplateOutlet` with component context. An optional `A2UI_FALLBACK_TEMPLATES` injection token, mirroring `A2UI_RENDERER_CONFIG`, carries the templates to nested component hosts. The `unknownComponent` path dispatches the same `COMPONENT_NOT_FOUND` as React.
* **Lit**: the visible `Loading surface...` default is removed while the `<slot name="loading">` is preserved for consumer projection. A nested child referenced before it streams no longer throws; it renders the fallback or nothing and re-renders when the component arrives via a targeted subscription that is cleaned up on disconnect and surface change. Consumer fallbacks are delivered through `@lit/context` so nested elements receive them, and the `unknownComponent` and error paths dispatch through the same shared channel, deferred out of render.

To be explicit about the behavior change: removing the visible React debug divs is deliberate and is the fix this issue asks for. It is a rendering default change, not an API break; every existing prop and export keeps working, and consumers who want visible fallbacks get them via the new `fallbacks` prop. I know a v1.0 restructure is in flight; this diff is small and additive on the v0_9 surface, and I am happy to rebase or adapt it if the restructure moves these files.

Tests: each renderer adds a matrix asserting render-nothing defaults for both states and consumer-provided fallbacks including nested delivery, plus dispatch dedup and, where relevant, subscription cleanup and cross-surface isolation. The six legacy placeholder-text assertions across the renderers are updated. React suite 462 passing, Angular suite 269 passing, Lit suite 109 passing, web_core suite 271 passing.

Docs: each renderer README gains a copy-paste custom-fallback example, and `blueprints/modules/a2ui_framework_adapter.blueprint.md` documents the shared fallback contract so other renderers can adopt it.

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].

For this PR:

(The diff is web only, so `e2e_test.sh`'s Flutter path is not exercised; the four web packages build and their test suites pass locally.)

## Demo

Recorded against the React gallery app with a demo surface containing a child that never arrives (loading) and an unregistered component type (unknownComponent). First half shows the new silent defaults, second half the same surface with a consumer-provided `fallbacks` prop. The gallery wiring and demo example are local harness only, not part of this PR.

fallback demo

| Default (renders nothing, warns once + dispatchError) | Custom `fallbacks` prop |
|---|---|
| default | custom |

### DIFF
--- a/blueprints/modules/a2ui_framework_adapter.blueprint.md
+++ b/blueprints/modules/a2ui_framework_adapter.blueprint.md
@@ -74,6 +74,54 @@ interface ComponentImplementation extends ComponentApi {
 
 The entrypoint widget/view for a specific framework. It is instantiated with a `SurfaceModel` (from the Core SDK). It listens to the model for lifecycle events and dynamically builds the UI tree, initiating the recursive rendering loop at the component with ID `root`.
 
+### Fallback States
+
+When the Surface's render loop cannot render a node, it surfaces one of two consumer-facing fallback states, named identically across every renderer: `loading` (the component is referenced but has not yet streamed in) and `unknownComponent` (the component's `type` has no implementation in the surface's catalog). A third state, `error` (a runtime exception thrown while rendering), is dispatch-only and is not a consumer fallback key.
+
+The payload passed to a fallback is `A2uiFallbackInfo`, a three-arm discriminated union keyed on `state`. The `componentType` field lives only on the `unknownComponent` arm; the `error` arm carries `reason: unknown` (the thrown value); every arm carries `componentId` and an optional `parentComponentType` hint that is absent at the surface root.
+
+```typescript
+export type A2uiFallbackInfo =
+  | {
+      /** The component is referenced but has not yet been streamed to the client. */
+      state: 'loading';
+      /** The id of the component that could not be rendered. */
+      componentId: string;
+      /** The `type` of the component that instantiated this one. Absent at the surface root. */
+      parentComponentType?: string;
+    }
+  | {
+      /** The component's `type` has no implementation in the surface's catalog. */
+      state: 'unknownComponent';
+      /** The id of the component that could not be rendered. */
+      componentId: string;
+      /** The component's declared type. */
+      componentType: string;
+      /** The `type` of the component that instantiated this one. Absent at the surface root. */
+      parentComponentType?: string;
+    }
+  | {
+      /** A runtime exception occurred while rendering the component. */
+      state: 'error';
+      /** The id of the component that could not be rendered. */
+      componentId: string;
+      /** The thrown value. If it is an `Error`, the stack is available on it. */
+      reason: unknown;
+      /** The `type` of the component that instantiated this one. Absent at the surface root. */
+      parentComponentType?: string;
+    };
+
+export type A2uiFallbackState = A2uiFallbackInfo['state'];
+```
+
+Renderers render nothing by default in both consumer-facing states and keep a `console.warn` for developer experience. Each renderer delivers consumer fallbacks in its own idiomatic way:
+
+- React: a context-delivered `fallbacks` prop on `A2uiSurface`, so nested pending children pick it up through context.
+- Angular: `loadingTemplate` and `unknownComponentTemplate` `TemplateRef` inputs on the surface, propagated to nested hosts through the `A2UI_FALLBACK_TEMPLATES` token.
+- Lit: a preserved `loading` slot for the root case plus a `fallbacks` property provided over `@lit/context` for nested children.
+
+On the `unknownComponent` path, and the Lit root `error` path, the renderer calls `surface.dispatchError` with a structured code such as `COMPONENT_NOT_FOUND` once per type, so the agent learns about hallucinated components.
+
 ---
 
 ## 3. Component Implementation Strategies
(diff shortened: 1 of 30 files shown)


matthewvilaysack:
Added the Angular renderer. Same idea as React underneath. `loadingTemplate` and `unknownComponentTemplate` are `TemplateRef` inputs on `SurfaceComponent`, and there is an optional `A2UI_FALLBACK_TEMPLATES` token so the templates reach nested hosts too, mirroring how `A2UI_RENDERER_CONFIG` already works. Defaults do not change. Nothing renders and the existing console warnings still fire. `unknownComponent` also dispatches `COMPONENT_NOT_FOUND` with the same shape and message as React. Added Karma specs for all of it, suite is green. Lit is next on this same PR.

matthewvilaysack:
Thanks for the review. Moved the warn and dispatch into a useEffect so render stays pure, and made the warned-types Set lazy so it is not reallocated on each render. Both are in the latest push.

On memoizing the fallbacks prop, I left that to the consumer on purpose. It is documented to be hoisted or memoized like any context value or render prop, and I would rather keep that explicit than hide it behind an internal memo.

matthewvilaysack:
Added the Lit renderer, so all four packages are now in: web_core, React, Angular, and Lit. Lit keeps the loading slot but drops the visible default text, guards the nested pending path that used to throw and re-renders when the child arrives, and delivers consumer fallbacks through @lit/context so nested elements receive them. Same COMPONENT_NOT_FOUND dispatch as the other renderers, deferred out of render. The cross renderer test matrix is complete now. Ready for review whenever you have a chance.

matthewvilaysack:
@ditman this is the implementation of what we discussed on #2013, all four packages are in and the branch is cleaned up to a single commit with a demo in the description. Ready for your pass whenever you get a chance.

nan-yu:
@ditman Can you take a look at this PR when you get a chance? Thanks!

PR #2093, opened 2026-07-24
feat(swift): add SwiftUI rendering layer with Surface, CatalogView, and DataBinding bridge

pinieb:
Ports the SwiftUI rendering layer that connects A2UICore's resolved
component tree to SwiftUI views.

Changes:
- Add CatalogView protocol for catalog-defined views constructed from
  resolved Node instances
- Add A2UIThemeKey environment key for propagating SurfaceTheme through
  the SwiftUI view hierarchy
- Add DataBinding+SwiftUI extension providing a single swiftUIBinding
  property that bridges A2UI's DataBinding to SwiftUI's Binding
- Add Surface view — the root SwiftUI view for a single A2UI surface
  that observes SurfaceViewModel and renders via the catalog's
  CatalogView, with theme propagated through the environment
- Remove placeholder A2UISwiftUI.swift stub
- Add 9 tests covering Surface initialization and equality,
  DataBinding swiftUIBinding get/set, theme environment key defaults
  and mutation, and CatalogView initialization

Verification:
- swift test passes (160 tests in 16 suites)


swift/swiftui/Sources/A2UISwiftUI/Surface.swift
```diff
@@ -0,0 +1,43 @@
...
+/// theme is propagated through the environment.
+public struct Surface<Catalog: CatalogView>: View {
+  @ObservedObject public var viewModel: SurfaceViewModel
+
+  private let catalogType: Catalog.Type
+  public let surfaceID: String
+
+  public init(viewModel: SurfaceViewModel, catalogType: Catalog.Type) {
+    self.viewModel = viewModel
+    self.catalogType = catalogType
```
jacobsimionato: What is CatalogType here?

pinieb: It was meant to be a way to give the CatalogView's type to the Surface since Swift isn't always nice about letting you do something like `Surface<CatalogType>`. `CatalogType` is basically a collection of `ComponentImplementation` types that plays nicely with SwiftUI.

Looking at the tests, maybe the specialized generic syntax will work though, so I might be able to streamline this initializer's API surface a little bit.

jacobsimionato: It would be great to avoid this if possible! Note that in https://github.com/a2ui-project/a2ui/pull/2079, we are allowing multiple catalogs to be used together in the same surface in v1.0, so we'll need to update the code to support that eventually. This parameter might make that more difficult.

jacobsimionato (approved):

swift/swiftui/Sources/A2UISwiftUI/Surface.swift
pinieb: Surface is now de-genericized!

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -68,25 +73,175 @@ public final class MessageProcessor: @unchecked Sendable, ObservableObject {
...
+  public struct CapabilitiesOptions: Sendable {
+    /// If true, full definitions of all catalogs will be included as inline catalogs.
+    public var includeInlineCatalogs: Bool
+
+    /// The protocol version to generate capabilities for (defaults to "v0.9.1").
+    public var version: String
+
+    public init(
+      includeInlineCatalogs: Bool = false,
+      version: String = "v0.9.1"
```
jacobsimionato: Maybe we should not have a default here, because otherwise it makes it a breaking change when we change the default. It'd be nice if this change didn't break users.

swift/swiftui/Sources/A2UISwiftUI/ComponentRegistry.swift
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/// A closure that constructs a SwiftUI view from a resolved engine node.
+public typealias ComponentViewBuilder = (Node) -> AnyView
+
+/// A registry mapping component keys to SwiftUI view builders.
+///
+/// Supports catalog-qualified lookups when a node specifies a catalog ID,
+/// with automatic fallback to an unqualified component type match.
+public final class ComponentRegistry: @unchecked Sendable {
```
jacobsimionato: Can we call this class `CatalogImplementation`, to better match the concepts used in our other rendering libraries?

In our other renderers, we just one parameterized Catalog implementation e.g.:
- Catalog<ComponentApi> is a catalog API
- Catalog<ComponentImplementation> is a catalog implementation.

I remember you mentioning that this might not be so easy in Swift, which is okay. But I still think that the API / Implementation naming would be a better match, and also that we should have a `ComponentImplementation` object, because this is the object that teams like Maps would implement in order to ship A2UI components that people can add to their catalogs.

swift/swiftui/Sources/A2UISwiftUI/ComponentRegistry.swift
```diff
@@ -0,0 +1,103 @@
...
+  ) {
+    register(
+      catalogID: catalog.id,
+      type: type,
+      builder: builder
+    )
+  }
+
+  /// Resolves and invokes the registered view builder for the given node.
+  public func render(node: Node) -> AnyView? {
```
jacobsimionato: Maybe this logic can be in `Surface.swift`, to have a more clear separation of concerns:
- Registry / Catalog implementation: Defines a set of components that can be rendered
- Surface: Does the actual rendering

swift/swiftui/Sources/A2UISwiftUI/ComponentRegistry.swift
```diff
@@ -0,0 +1,103 @@
...
+    builder: @escaping ComponentViewBuilder
+  ) {
+    let key = ComponentKey(
+      catalogID: catalogID,
+      type: type
+    )
+    builders[key] = builder
+  }
+
+  /// Registers a view builder for a component type within a specific catalog.
```
jacobsimionato: Is it possible for us to have a top-level `ComponentImplementation` object which combines the type name, schema, and builder function?

The component registry seems to just be a plain old swift object, so perhaps it's possible to now structure the objects more like in the other SDKs? Does the fact that we're switching to the `AnyView` approach give us more flexibility in the design of this than we previously had?

swift/swiftui/Sources/A2UISwiftUI/ComponentRegistry.swift
pinieb: It's less of a Swift language issue and more that it's very likely clients will need to show the same catalog on different UI frameworks in the same app, so we need to have separate API and Implementation types since there could be a one-to-many mapping from API to Implementation

PR #2094, opened 2026-07-24
fix(samples): remove workspace source override from gemini_enterprise samples

suyangw-g:
Remove `[tool.uv.sources]` specifying `a2ui-agent-sdk = { workspace = true }` from the Gemini Enterprise v0.8 sample packages.

- Prevents build and installation failures when deploying or running `agent_engine` and `cloud_run` samples as standalone packages outside the root workspace context.
- Allows `uv` and `pip` to resolve `a2ui-agent-sdk` directly from PyPI as declared in project dependencies.


ditman (approved):
/cc @nan-yu FYI!

PR #2098, opened 2026-07-25
Update script to remove label 'status: waiting-for-author-response'.

polina-c:
Renames `status: waiting-for-user-response` to `status: waiting-for-author-response`, and teaches the triage automation to clear it once the author replies.

### Behavior

* Renames the label across `scripts/triage.mjs`, the `a2ui-issue-triage` skill (criteria, dashboard, `suggest_triage.py`) and `docs/contributing/triage.md`.
* `scripts/triage.mjs` now removes the waiting label when the item's author has replied, and applies `status: needs-triage` in the same run rather than leaving it for the next one.
* The waiting-label check moved to the top of `flagReason`, so a parked item is off the triage queue entirely — previously it only suppressed the issue rules, not the PR or unanswered-comment ones.
* The run summary reports how many waiting labels were cleared.
* `docs/contributing/triage.md` notes which of the two labels is automated and which is applied by hand.

### Tests

* Unit tests for `authorHasResponded`, including the opening-post regression above.
* Reconciliation tests for the label being kept while awaiting the author, cleared once they reply, cleared-and-flagged in a single run, and backed off when a concurrent run already reconciled both labels.
* Split the `comment()` test factory into `comment()` (raw API payload) and `event()` (the normalized event `fetchContributions` produces), so each test states which layer it exercises.
* Fixes 6 pre-existing failures in `scripts/triage.test.mjs`. They went unnoticed because the suite never ran in CI — see below.

### CI

* `web_ci.yml` gains a `ci-scripts` job, gated on `scripts/**`, that runs `node --test scripts/*.test.mjs`. Nothing executed `triage.test.mjs` before, which is how the failures above survived on `main`.

TODO after merge: rename the label in GitHub.


scripts/triage.test.mjs
```diff
@@ -56,22 +57,30 @@ const pr = (overrides = {}) =>
...
   user: {login: 'maintainer', type: 'User'},
   ...overrides,
 });
 
+// The normalized event `fetchContributions` produces, which `flagReason` and
+// friends consume directly. Takes the same overrides as `comment`.
+const event = (overrides = {}) => {
+  const {created_at: createdAt, author_association: association, user} = comment(overrides);
+  return {createdAt, association, user};
+};
```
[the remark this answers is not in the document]
polina-c: thanks

fixed

scripts/triage.mjs
```diff
@@ -74,40 +82,68 @@ const labelNames = item =>
...
+  );
+}
 
-  return latest;
+/**
+ * True when the author has spoken since opening the item — the newest human
+ * contribution is theirs. Used to clear WAITING_LABEL. The opening post does not
+ * count: an item with no replies is exactly the one still waiting on its author.
+ *
+ * A maintainer who replies after the author hides that response from this check;
```
jacobsimionato: Checking my understanding: I guess this logic runs soon after any comment is posted, so if the author responds, it's very unlikely that someone else will post before this logic runs. Is that right?

jacobsimionato (approved):

scripts/triage.mjs
polina-c: Yes.

But after some thinking updated script and doc to clear label if author answered at least once after label was added.

Thank you for your question.

PR #2103, opened 2026-07-26
ci(triage): collapse concurrent triage runs into one

polina-c:
Adds a `concurrency` group to the triage workflow so a burst of issue events
collapses into a single reconciliation run.

## Why

The workflow triggers on `issues`, `issue_comment`, `pull_request_target`,
`pull_request_review` and `pull_request_review_comment`, and every run
reconciles the whole repository.

During a triage session today — adding `status: in-discussion` to 20 issues and
commenting on 42 more — that produced roughly 62 runs. Each walks every open
item and fetches its comments, which exhausted the GitHub App installation's
hourly API limit; the last five runs failed with:

```
RequestError [HttpError]: API rate limit exceeded for installation.
```

Any routine bulk triage can reproduce this.

## Why cancelling is safe

Runs are interchangeable: `scripts/triage.mjs` reads live state, reconciles
`status: needs-triage` across all open items, and never depends on what a
previous run did. Cancelling one part-way leaves items for the surviving run,
which reaches the same result. The daily cron backstops it either way.

## Alternative considered

Narrowing the trigger list would also cut the volume, but at the cost of the
prompt flagging the workflow exists for. This keeps the responsiveness and
removes only the redundant work.


nan-yu (approved):

josemontespg (approved):
Makes sense, thanks for the improvement!

gspencergoog:
Thanks!

PR #2120, opened 2026-07-28
feat(v1.0): add rootOnly component metadata to enforce root-level rendering

wrenj:
## What changed
Introduces an optional `rootOnly` boolean metadata property to component definitions in the A2UI v1.0 Catalog schema (`catalog_definition.json`). 
When `"rootOnly": true`, renderers must enforce that instances of the component appear only at the root of a surface and reject payloads that nest the component within another component.
## Why this is needed
A2UI v1.0 added support for mixing catalogs, so standalone components (`SuggestionBox`, `Canvas`, full-page iframe, `ToastMessage`) are now mixable with composable widget catalogs (`Button`, `Table`, `TextBox`). A mechanism is needed for catalog components to explicitly opt out of nesting. Since this involves interaction between catalogs using the ChildList in common_types, this can't be solved by individual catalogs themselves and we need a framework wide mechanism. Additionally, since A2UI represents component trees as a flat adjacency list on the wire, standard JSON Schema validation cannot detect when a standalone component is invalidly nested inside another component. 
## Why this approach
- **Zero wire overhead:** Defining `rootOnly` as static catalog metadata (parallel to `callableFrom` and `returnType` for functions) avoids adding validation properties to wire payloads in `common_types.json`.
- **Targeted scope:** A top-level `rootOnly` boolean cleanly solves the most common structural issue when mixing catalogs. More complex parent-child constraints (eg, MenuItem is the only child of Menu) remain the responsibility of individual component schemas.
## Testing
- Updated `specification/v1_0/json/catalog_definition.json`, `a2ui_protocol.md`, and `evolution_guide.md`.
- Added schema validation test cases in `catalog_root_only_checks.json` and updated `run_tests.py`.

### DIFF
--- a/specification/v1_0/docs/a2ui_protocol.md
+++ b/specification/v1_0/docs/a2ui_protocol.md
@@ -519,7 +519,7 @@ To ensure catalog schemas can be translated reliably into alternative, LLM-frien
      - Mandatory metadata fields outside the strict JSON validation properties to advertise interface details:
        - **`returnType`**: Must be a string enum indicating the return type (`string`, `number`, `boolean`, `array`, `object`, `any`, or `void`).
        - **`callableFrom`**: Must be a string enum indicating the execution boundary (`rendererOnly`, `agentOnly`, or `rendererOrAgent`). If omitted, it defaults to `rendererOnly`.
-7. **Component Structural Metadata:**
+7. **rootOnly components:**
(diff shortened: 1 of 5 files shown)


jacobsimionato:
Can we close this one now that we have the parent and children allowlists?

PR #2122, opened 2026-07-28
Simplify triage templates by removing instructions for agent.

polina-c:
Simplifies the triage templates and unblocks the Docs CI check.

## Changes

**`docs/contributing/triage-templates.md`** — removed the "How to use these replies"
section. The five numbered steps described how an agent should pick, apply, and fall back
from these replies, which is guidance that belongs with the triage skill rather than in the
template list itself. What remains is what the file is for: the canonical replies, each with
its case and the triage fields that go with it.

**`lychee.toml`** — excluded `https://www.diva-portal.org` from the Markdown link check.
The host intermittently stops responding altogether — connections hang well past the 20s
lychee timeout — which failed `Docs / build_and_deploy` on this PR even though the citation
lives in `eval/DESIGN.md` and is untouched here. The paper is still published there, so the
reference stays and only the check is skipped. This follows the existing precedent for
`ojs.aaai.org` in the same file.

## Testing

- `./scripts/fix_format.sh --check` passes.
- `lychee --config lychee.toml --no-progress "**/*.md"` no longer reports the timeout, and
  the exclusion introduces no new errors.

### DIFF
--- a/docs/contributing/triage-templates.md
+++ b/docs/contributing/triage-templates.md
@@ -1,26 +1,6 @@
 # A2UI Standard Triage Replies
 
-Canonical replies the team has agreed on for recurring triage cases. Check this list
-**before** drafting a new response.
-
-## How to use these replies
-
-1. **Match first.** If the issue or PR fits one of the cases below, use that reply.
-2. **Verbatim by default.** Post the reply as written when the case matches cleanly. These
-   texts encode team policy and agreed tone, so do not rewrite them for style — the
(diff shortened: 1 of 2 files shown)


ditman:
Please, run formatting and remember to add a description to the PR!

polina-c:
> Please, run formatting and remember to add a description to the PR!

done

nan-yu (approved):

PR #2123, opened 2026-07-28
feat: add specification for A2UI web app iframe component (v0.9)

sugoi-yuzuru:
## Description

This PR introduces the new URL-based web application iframe component (`web-app-frame-url`) for the A2UI protocol, adding a secure and isolated method to embed external web applications via URL. It also includes a new "Pong" community sample application to demonstrate its usage.

https://screencast.googleplex.com/cast/NDU2Njg0MjA3MTg0MjgxNnxjNzRjZjIyZS0zOA

### 🌟 Key Features & Additions

- **Web App Frame URL Component (`web-app-frame-url`)**:
  - **Component Implementation**: Added `web-app-frame-url.ts` to the Angular client to handle rendering external web apps inside a sandboxed iframe.
  - **Specification & Documentation**: Authored `web-frame-component_spec.md` to define the component's API, two-way data binding handshake behavior, and strict security considerations.
  - **Catalog Updates**: Registered the new component and its configuration in `mcp_app_catalog.json`.
  - **Sandboxing**: Introduced `sandbox-url.html` and updated `sandbox.ts` to facilitate secure `postMessage` cross-origin communications while strictly defining `<meta>` tag requirements and `Content-Security-Policy` limits for URL-based frames.

- **Pong Web Sample**:
  - **New Web App**: Created the `samples/community/web/pong/` sample app.
  - **A2UI Bridge**: Includes `pong_web_frame_bridge.js`, which handles the client-side `postMessage` handshake, state synchronization (player/cpu scores), and function calls with the A2UI host.
  - **Server Implementation**: Built a standalone Python HTTP server (`pong_server.py`) to serve the application and inject the bridge script seamlessly.

- **Python ADK Integration**:
  - Added necessary tool abstractions in `mcp_app_proxy/tools.py` and `agent.py` to allow the proxy agent to interact with and render the new Pong game component.

### 📝 Notes for Reviewers
- Please review the CSP rules and Origin verification logic detailed in `web-frame-component_spec.md` and `sandbox-url.html` to ensure the cross-origin communication remains completely secure.
- The Pong server is an independent application—ensure `uv` is configured properly if testing the sample locally, as per the `uv.lock` and `pyproject.toml` configurations.

## Pre-launch Checklist

One time:

- [ ] I signed the [CLA].

For this PR:

- [ ] If my branch is on a fork, I have verified that [scripts/e2e_test.sh](https://github.com/a2ui-project/a2ui/blob/main/scripts/e2e_test.sh) passes.


samples/community/web/pong/pong_server.py
```diff
@@ -0,0 +1,49 @@
...
+class Handler(http.server.SimpleHTTPRequestHandler):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, directory=DIRECTORY, **kwargs)
+
+    def do_GET(self):
+        parsed_path = urllib.parse.urlparse(self.path)
+        if parsed_path.path == '/pong_app_web_frame.html':
+            self.send_response(200)
+            self.send_header('Content-type', 'text/html')
+            self.send_header('Access-Control-Allow-Origin', '*')
```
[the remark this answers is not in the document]
sugoi-yuzuru: Updated.

samples/community/web/pong/pong_web_frame_bridge.js
```diff
@@ -0,0 +1,124 @@
+/*
+ * WebFrame Communication Layer
+ */
+let localPlayerScore = 0;
+let localCpuScore = 0;
+
+let functionCallId = 0;
+
+function dispatchAction(action, data = {}) {
+  window.parent.postMessage({type: 'a2ui_action', action, data}, '*');
```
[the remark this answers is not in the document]
sugoi-yuzuru: Fixed.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  private iframe = viewChild.required<ElementRef<HTMLIFrameElement>>('iframe');
+  private messageHandler: ((event: MessageEvent) => void) | null = null;
+  private dataSubscriptions: any[] = [];
+  private resizeTimeout: any = null;
+  private lastWidth?: number;
+  private lastHeight?: number;
+  private lastBoundRootValues: Record<string, string> = {};
+  private isProcessingAppWrite = false;
+  private expectedOrigin = '*'; // In production this should be validated
```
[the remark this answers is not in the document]
sugoi-yuzuru: fixed.

samples/community/agent/adk/mcp_app_proxy/catalogs/0.9/mcp_app_catalog.json
```diff
@@ -87,6 +87,50 @@
       ],
       "unevaluatedProperties": false
     },
+    "WebAppFrameUrl": {
```
jgindin, line 90: I suggest adding `description` to as many elements as possible.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
```diff
@@ -0,0 +1,405 @@
...
+      flex: 1;
+      max-width: 100%;
+      max-height: 100%;
+      border: none;
+      background-color: white; /* Ensure content is readable */
+    }
+  `,
+  template: ` <iframe #iframe [src]="iframeSrc()" [title]="'WebAppFrame'"></iframe> `,
+})
+export class WebAppFrameUrl extends CatalogComponent<any> implements OnDestroy, OnInit {
```
jgindin: There's some usages of `any`. Can they be tightened with actual types? If not, maybe use `unknown`?

go/tsjs-style#any

samples/community/agent/adk/mcp_app_proxy/catalogs/0.9/mcp_app_catalog.json
sugoi-yuzuru: added.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
```diff
@@ -0,0 +1,457 @@
...
+
+      const iframeEl = this.iframe().nativeElement;
+      if (!iframeEl || event.source !== iframeEl.contentWindow) {
+        return;
+      }
+
+      const data = event.data;
+      if (!data) return;
+
+      if (data.type === 'a2ui_sandbox_proxy_ready') {
```
jgindin: Should we define an enum for all the `data.type` values and the values sent back as strings in the `type` property?

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
sugoi-yuzuru: Thanks!
Updated to minimize `any`.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
```diff
@@ -0,0 +1,457 @@
...
+      const widthDiff =
+        targetWidth !== undefined && this.lastWidth !== undefined
+          ? Math.abs(targetWidth - this.lastWidth)
+          : 100;
+      const heightDiff =
+        targetHeight !== undefined && this.lastHeight !== undefined
+          ? Math.abs(targetHeight - this.lastHeight)
+          : 100;
+
+      if (targetWidth !== undefined && widthDiff >= 5) {
```
jgindin, line 196: I_think_ this means that if I'm particularly crafty, I can constantly resize the window, but the iframe would never resize.

I think we need to set the `lastHeight` / `lastWidth` outside of the `if` statements.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
```diff
@@ -0,0 +1,457 @@
...
+      }
+    }, 100);
+  }
+
+  private setupSandbox() {
+    if (this.messageHandler) {
+      window.removeEventListener('message', this.messageHandler);
+    }
+
+    this.messageHandler = async (event: MessageEvent) => {
```
jgindin, line 371: This method is .... large :-).

I suggest breaking it up, where this method checks the `data.type` and calls `handleFoo` methods. E.g., `handleProxyReady`, `handleFrameReady`, etc...

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
```diff
@@ -0,0 +1,457 @@
...
+            (this.props()['data']?.raw as any)?.paths ?? this.props()['data']?.value()?.paths ?? {};
+
+          if (dataPaths[data.key]) {
+            const dataPath = dataPaths[data.key];
+            const targetPath = data.subpath
+              ? `${dataPath}${data.subpath.startsWith('/') ? '' : '/'}${data.subpath}`
+              : dataPath;
+
+            const currentValue = surface.dataModel.get(targetPath);
+            if (JSON.stringify(currentValue) !== JSON.stringify(data.value)) {
```
jgindin: I _think_ `JSON.stringify` can be unstable such that even if `currentValue` and `data.value` are the same, the result of `JSON.stringify` on each could be different.

This is why folks have created 3rd party libraries to create a stable JSON string.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
```diff
@@ -0,0 +1,520 @@
+# A2UI WebApp Iframe Component Specification (v0.9)
```
jgindin, line 1: Can we reformat this file so it can be easily read in an IDE? (Right now, there are some lines that are extremely long.)

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
```diff
@@ -0,0 +1,520 @@
...
+## A Specification for Sandboxed, Rich Interactive Components in the Agent-to-UI Protocol
+
+Jul 27, 2026  
+Status: In progress
+
+# Abstract
+
+This specification document defines the A2UI Iframe Component (v0.9) for the secure, sandboxed rendering of rich interactive web applications and model-generated HTML content. This document serves two primary purposes:
+
+1. **Platform Implementation Blueprint:** It provides client-side platform developers with a strict, standard set of instructions to implement compliant `WebAppFrameUrl` and `WebAppFrameSrcdoc` components in any native programming language (e.g., TypeScript/Web, Kotlin/Android, Swift/iOS, or Dart/Flutter) while maintaining identical security and sandboxing guarantees.
```
jgindin: How does this work on, e.g., Android? Unless you're embedding a browser view, there's no such thing as frames like this what there is present in the web.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
```diff
@@ -0,0 +1,520 @@
...
+**Embedded app action**  
+The embedded app consumes this updated configuration and dynamically adjusts its internal rendering as needed.
+
+## 3.5. WebAppFrame Protocol vs. MCP App Bridge Protocol Equivalence
+
+The A2UI `WebAppFrame` contract and the MCP App Bridge (`@modelcontextprotocol/ext-apps/app-bridge`) share identical runtime capabilities and semantics, but differ in their underlying message envelope.
+
+While WebAppFrame uses custom top-level message string tags (`a2ui_*`) with flat keys optimized for simple scripts, the standard MCP App Protocol uses standard JSON-RPC 2.0 framing (`jsonrpc: "2.0"`).
+
+Because of this difference, if a developer wishes to use the formalized bridge approach for WebAppFrame, they should use the dedicated `@a2ui/web-bridge` SDK rather than the `@modelcontextprotocol/ext-apps/app-bridge` SDK, as the latter strictly enforces JSON-RPC 2.0 formatting.
```
jgindin: I'm not clear on why this is the case. If I'm using the A2UI approach, why would I think to use the MCP App SDK?

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
```diff
@@ -0,0 +1,520 @@
...
+
+# 4. Component catalog definition
+
+The two web frame components, _WebAppFrameUrl_ and _WebAppFrameSrcdoc_, shall be registered as distinct options in the A2UI v0.9 Component Catalog.
```
jgindin: Why do we need 2 components in the catalog? Isn't the fact that we use two iframes an implementation detail that could be encapsulated by one A2UI component?

samples/community/web/pong/pong_web_frame_bridge.js
```diff
@@ -0,0 +1,154 @@
...
+    document.body.style.width = `${containerDimensions.width}px`;
+  }
+  if (typeof containerDimensions.height === 'number') {
+    document.documentElement.style.height = `${containerDimensions.height}px`;
+    document.body.style.height = `${containerDimensions.height}px`;
+  }
+  if (typeof resize === 'function') resize();
+}
+
+window.addEventListener('message', event => {
```
jgindin: I'd suggest that here, too, this method be broken into smaller methods. This method can check the `data.type` and call an appropriate `handleFoo` method.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
sugoi-yuzuru: Developers might prefer to develop MCP Apps instead of the A2UI Iframe app. In that case, there are similar protocols but in a different bridge implementation. 

This paragraph is here to clarify that the bridge must be different from MCP Apps' approach.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
sugoi-yuzuru: While the specification is written primarily using web terminology (iframes, srcdoc, postMessage), it serves as a blueprint for configuring native WebView and WKWebView components to behave securely and communicate identically. 

And you are correct, this is assuming that you are serving a web-application so the native app will display this by a browser widget.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
sugoi-yuzuru: That’s a great question, and it highlights a subtle but important distinction in the architecture.

You are correct that the double-iframe sandbox pattern (an outer proxy iframe wrapping an inner sandbox iframe) is strictly an implementation detail that is completely hidden from the A2UI agent.

However, the reason there are two distinct components in the catalog (WebAppFrameUrl and WebAppFrameSrcdoc) has nothing to do with the double-iframe implementation. In fact, both components use the double-iframe setup under the hood.

They exist as two separate components in the catalog for two primary reasons:

1. Different JSON Schemas for the Agent
The A2UI Component Catalog dictates the JSON structure the LLM agent must generate. These two components take fundamentally different inputs:

WebAppFrameUrl requires a url property (a string pointing to a remote server).
WebAppFrameSrcdoc requires an htmlContent property (a potentially massive string of raw HTML/JS/CSS generated dynamically by the model).
2. Divergent Security Threat Models (CSP Enforcement)
As detailed in Section 5 of the specification, because the source of the content differs, the client renderer must apply completely different security pipelines before rendering the iframe:

For WebAppFrameUrl: The client is loading an external URL via the src attribute. It cannot manipulate the HTML of the remote site, so it must rely on the remote server's HTTP headers to provide the Content Security Policy (CSP). The client's primary security duty here is checking the URL against a strict Domain Allowlist.
For WebAppFrameSrcdoc: The client receives a raw HTML string generated by the LLM. Because LLM output can be unpredictable or maliciously prompted, the client must treat it as untrusted. The client must intercept this HTML, strip out any model-generated CSP tags, and forcibly inject a strict, network-free CSP <meta> tag into the <head> before rendering it as srcdoc.
Summary
While the double-iframe sandbox is indeed an encapsulated implementation detail, the agent does need to know whether it is passing a URL or raw HTML. Because they have different schemas and require the client to apply completely different security parsing rules, they are modeled as two separate components in the protocol catalog.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
sugoi-yuzuru: Great point! We definitely want type safety here, but because we're receiving these messages across the window.postMessage boundary, a traditional TypeScript enum could give us a false sense of security. Since the iframe might be written in vanilla JS (or by a third party), casting event.data to an enum wouldn't prevent malformed payloads from causing runtime errors.

Instead, I refactored this using a Zod Discriminated Union. We define the literal string types within the Zod schema and run IncomingWebFrameMessageSchema.safeParse(event.data). This gives us the exact same developer experience (TypeScript infers the type, so we get autocomplete and strict type-checking in the if/else blocks) but with actual runtime validation to gracefully ignore bad data.

I've also extracted these schemas into a new shared web-frame-messages.ts file so we can reuse them across other frame components without duplicating the type definitions.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
sugoi-yuzuru: done

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
sugoi-yuzuru: done

samples/community/web/pong/pong_web_frame_bridge.js
sugoi-yuzuru: done.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
sugoi-yuzuru: SG. I am now using `fast-json-stable-stringify` which is supported in Google.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-app-frame-url.ts
sugoi-yuzuru: The bug you are suggesting actually will happen IF we move the assignment out of the `if` statement.

samples/community/agent/adk/mcp_app_proxy/catalogs/0.9/mcp_app_catalog.json
```diff
@@ -87,6 +87,61 @@
...
+              "additionalProperties": {"type": "object"}
+            },
+            "allowedFunctions": {
+              "type": "object",
+              "description": "A map of authorized host client functions to JSON Schemas defining their expected arguments.",
+              "additionalProperties": {"type": "object"}
+            },
+            "mutableData": {
+              "type": "object",
+              "description": "A map of authorized data model keys to JSON Schemas defining their allowed values.",
```
jacobsimionato: This is cool!

samples/community/agent/adk/mcp_app_proxy/catalogs/0.9/mcp_app_catalog.json
```diff
@@ -87,6 +87,61 @@
...
     },
+    "WebAppFrameUrl": {
+      "type": "object",
+      "allOf": [
+        {
+          "$ref": "common_types.json#/$defs/ComponentCommon"
+        },
+        {
+          "type": "object",
+          "properties": {
```
jacobsimionato, line 98: Could it make sense to also have a way to pass arbitrary static data to the embedded app as a property, rather than needing to pass it via the data model?

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
```diff
@@ -0,0 +1,677 @@
...
+
+### C. Local client-side function execution (`a2ui_function_call`)
+
+Dispatched when the embedded app wants to invoke a registered local v0.9 function.
+
+**Message schema**
+
+```json
+{
+  "type": "a2ui_function_call",
```
jacobsimionato, line 240: Are these the same or similar as the API for MCP apps to interact with A2UI? Could we unify them somehow?

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/catalog.ts
```diff
@@ -127,6 +138,7 @@ export const DEMO_CATALOG = new Catalog(
     {name: 'McpApp', component: McpApp, schema: McpAppSchema},
     {name: 'PongScoreBoard', component: PongScoreBoard, schema: PongScoreBoardSchema},
     {name: 'PongLayout', component: PongLayout, schema: PongLayoutSchema},
+    {name: 'WebAppFrameUrl', component: WebAppFrameUrl, schema: WebAppFrameUrlSchema},
```
jacobsimionato, line 141: @josemontesp is there some way in Angular that we the application developer could pass some configuration data to the Component when installing it in a catalog?

I think it could be great for a developer to be able to set some static domain allowlist for these apps here, to statically protect the client against loading apps at arbitrary domains.

jacobsimionato (approved):

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
sugoi-yuzuru: They are very similar. However, I have made a conscious decision to keep them distinct. I have a table in the document https://screenshot-v2.corp.google.com/FCmW9CWE3CgfkeK showing how the mapping works. 

While there are some opportunities to mirror them, I believe partial mirroring will make things further convoluted and hard to pin-point which version of an app is implemented.  

Furthermore, mirroring the MCP Apps protocols could put us in a situation where we will need to keep updating our implementation to match resulting in additional workload to maintain the synchrony. 

By keeping them completely distinct, I believe it gives us more consistency for the A2UI Iframe (WebAppFrame) protocol and a more room to make independent feature evolution paths.

samples/community/agent/adk/mcp_app_proxy/catalogs/0.9/mcp_app_catalog.json
sugoi-yuzuru: Thanks!!
I noticed that the description was a bit too brief, I have added more details to clarify the intent. 

Basically, it is an allow-list of data fields that the embedded app is allowed to request mutation. Other data are considered read-only for embedded app (though mutable by the Agent/Host).

samples/community/agent/adk/mcp_app_proxy/catalogs/0.9/mcp_app_catalog.json
sugoi-yuzuru: Love the idea. 

I have introduced a new property `config` to allow static data assignment to embedded app. 
Please check [this commit](https://github.com/a2ui-project/a2ui/pull/2123/changes/93639e6d1f2579c11b55a15f0864aef774a3b3ba).

1. I have updated the spec document to discuss about it. In addition to the spec update, I added a section to discuss the difference between the 3 flavors of data: `config`, `data` without `mutableData`, and `data` with `mutableData`. 
2. I have updated the component catalog and implementation. 
3. I have updated the WebAppFrameUrl example Pong game to leverage this `config` to define the "matching-score" of the game. Previously, it was hard-coded to announce a winner who scored 3 points first. Now, the Agent will return a `config` to set the exact score to call the winner (5).

sugoi-yuzuru: Let me know if you think `config` is not the right design here.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/catalog.ts
sugoi-yuzuru: Oooo 

This would be cool. 

Maintaining an allowlist on the client-side catalog would be a nice protection against Agent hallucination and compromised A2UI payloads coming from Agent/MCP servers.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
jgindin: Makes sense. I'll read over the docs again to see if I missed this or if there's something that could be added.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
```diff
@@ -0,0 +1,700 @@
...
+    "unevaluatedProperties": false
+  }
+}
+```
+
+## 4.2. WebAppFrameSrcdoc schema definition
+
+Used to load standalone, sandboxed, model-generated HTML/JS layouts.
+
+```json
```
jgindin, line 437: Just a thought -- take it or leave it...

`WebAppFrameSrcdoc` and `WebAppFrameUrl` are quite similar. If we expect them to remain this way, maybe extract the common properties into a new `WebApp` schema, and then include that in both?

The `WebApp` schema would have all the properties that are common between the two.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
```diff
@@ -0,0 +1,700 @@
...
+{
+  "WebAppFrameSrcdoc": {
+    "type": "object",
+    "description": "Renders rich, model-generated HTML/JS bundles securely in a sandboxed safe content frame.",
```
jgindin, line 441: Actually, not *just* model-generated, but any time you have some pre-defined HTML/JS and want to just provide it inline.

Not sure how common that would be, so maybe it's not worth calling out explicity?

jgindin (approved):

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
sugoi-yuzuru: I think this is a reasonable POV. 
But I will leave it until I have an implementation of the 2 versions completed. 

We already had some spec updates through this review cycle and prematurely refactoring them might make us less agile to polish implementation or identify new ideas.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/web-frame-component_spec.md
sugoi-yuzuru: Yea. We can revise this later. 
It is similar to A2UI, not all A2UI are LLM inspired xP 
Many are simply templated, but this is a bit too detailed. 

Model-generated gives a better idea of the extent of what it can display, I think :)

sugoi-yuzuru:
Hi all, 

If there are any more comments, please LMK. I will address them as a follow up! 
Will merge this soon.

samples/community/client/angular/projects/mcp_calculator/src/a2ui-catalog/catalog.ts
josemontespg: In Angular we could use injection tokens that are provided at the app root level, with an optional type-safe configuration. The component can then inject that configuration directly into the component class.

PR #2125, opened 2026-07-29
Update lint rules.

polina-c:
The rules should be superset of Google3 rules.
Otherwise Google3 roll will fail linter checks.

Blocked CL: cl/952393108

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -48,5 +48,6 @@ linter:
     - unreachable_from_main
 
     # readability
+    - omit_obvious_local_variable_types
     - specify_nonobvious_property_types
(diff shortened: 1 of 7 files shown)


gspencergoog (commented):
Well, this is too bad.  I see that this change only removes non-obvious types from the code, which means that people (and agents) reading the code will sometimes misunderstand the types (by definition they are non-obvious, afterall).

Does the roll process run `dart fix` on the rolled code?  Because I think that would strip the types here, and take care of at least some differences in analyzer settings automatically.

polina-c:
I agree. Thank you for good suggestion.

PR #2155, opened 2026-08-03
feat(spec): add allowedParents/allowedChildren and Surface component to v1.0

wrenj:
## Summary
This PR adds support for set-membership composition hierarchy constraints (`allowedParents` and `allowedChildren`) to A2UI v1.0 component catalogs, and defines `Surface` as a reserved component type representing the root of a surface.

### Schema definitions (`specification/v1_0/json/`)
* **`catalog_definition.json`**: Added optional `allowedParents` and `allowedChildren` arrays to component schema definitions (`#/$defs/ComponentDefinition`). If omitted, all parent or child component types are permitted.
* **`common_types.json`**: Defined `Surface` (`#/$defs/Surface`) as the reserved component type representing an A2UI surface (`child: ["root"]`, `allowedParents: []`).
* **`renderer_to_agent.json`**: Added `UNALLOWED_PARENT` and `UNALLOWED_CHILD` error codes to `Validation Failed Error` for reporting parent-child hierarchy validation failures.

### Documentation (`specification/v1_0/docs/`)
* **`a2ui_protocol.md`**:
  * Clarified that `createSurface` creates the immutable component `Surface` with `child='root'`, and that `Surface` cannot be modified using `updateComponents`.
  * Added the **Composition validation rules** section under **UI composition**, including catalog examples for top-level root restrictions and container-restricted components.
* **`evolution_guide.md`**: Added v1.0 migration entries documenting `allowedParents`, `allowedChildren`, `Surface`, and the new renderer error codes.

### Tests (`specification/v1_0/test/`)
* Added `cases/composition_constraints.json` to test valid and invalid `allowedParents` and `allowedChildren` catalog configurations against `catalog_definition.json`.
* Added test cases to `cases/renderer_messages.json` to verify `UNALLOWED_PARENT` and `UNALLOWED_CHILD` error message payloads.
* Updated `testing_catalog.json` to include the `Surface` component type.

## Compatibility
These changes are catalog-level schema metadata and do not alter the wire format of component instances in `createSurface` or `updateComponents`.


specification/v1_0/json/renderer_to_agent.json
```diff
@@ -97,8 +97,9 @@
           "title": "Generic Error",
           "properties": {
             "code": {
+              "type": "string",
               "not": {
-                "const": "VALIDATION_FAILED"
+                "enum": ["VALIDATION_FAILED", "UNALLOWED_PARENT", "UNALLOWED_CHILD"]
```
gspencergoog, line 102: Do we need both?  I think a single error would be fine, since either way the result is clear.  (e.g. `INVALID_COMPOSITION`).

gspencergoog:
After reading this, I think I'm starting to warm up to an `allowRoot` explicit value.  Mainly because I don't like having a synthetic component (`Surface`).  I think that might be more confusing than helpful to an LLM, and it'll do some weird thing where it adds a `Surface` as a child to something.  It also complicates the implementation that now has to explicitly handle this synthetic type, and we have to deal with possible collisions with custom `Surface` components.

`allowedChildren` should only be allowed to be  empty for leaf components (like `Text`) that don't have children. `allowedParents` can be the empty list only when `allowRoot` is `always`.

```json
"ComponentDefinition": {
  "type": "object",
  "allOf": [
    { "$ref": "https://json-schema.org/draft/2020-12/schema" },
    {
      "type": "object",
      "properties": {
        "allowRoot": {
          "type": "string",
          "enum": ["any", "always", "never"],
          "default": "any",
          "description": "Root mounting policy: 'any' (can be root or child), 'always' (must be root), or 'never' (must be child)."
        },
        "allowedParents": {
          "type": "array",
          "items": { "type": "string" },
          "uniqueItems": true,
          "description": "Parent component types that can contain this component. Can only empty [] when allowRoot is 'always'.  Omitted means any parent type allowed."
        },
        "allowedChildren": {
          "type": "array",
          "items": { "type": "string" },
          "uniqueItems": true,
          "description": "Child component types allowed inside this container. Empty [] means leaf node (no children allowed). Omitted means any child type allowed."
        }
      }
    }
  ]
}
```

wrenj:
> After reading this, I think I'm starting to warm up to an `allowRoot` explicit value. Mainly because I don't like having a synthetic component (`Surface`). I think that might be more confusing than helpful to an LLM, and it'll do some weird thing where it adds a `Surface` as a child to something. It also complicates the implementation that now has to explicitly handle this synthetic type, and we have to deal with possible collisions with custom `Surface` components.
> 
> `allowedChildren` should only be allowed to be empty for leaf components (like `Text`) that don't have children. `allowedParents` can be the empty list only when `allowRoot` is `always`.
> 
> ```json
> "ComponentDefinition": {
>   "type": "object",
>   "allOf": [
>     { "$ref": "https://json-schema.org/draft/2020-12/schema" },
>     {
>       "type": "object",
>       "properties": {
>         "allowRoot": {
>           "type": "string",
>           "enum": ["any", "always", "never"],
>           "default": "any",
>           "description": "Root mounting policy: 'any' (can be root or child), 'always' (must be root), or 'never' (must be child)."
>         },
>         "allowedParents": {
>           "type": "array",
>           "items": { "type": "string" },
>           "uniqueItems": true,
>           "description": "Parent component types that can contain this component. Can only empty [] when allowRoot is 'always'.  Omitted means any parent type allowed."
>         },
>         "allowedChildren": {
>           "type": "array",
>           "items": { "type": "string" },
>           "uniqueItems": true,
>           "description": "Child component types allowed inside this container. Empty [] means leaf node (no children allowed). Omitted means any child type allowed."
>         }
>       }
>     }
>   ]
> }
> ```

It's either that or we just reserve the Surface / Root keyword in component names and drop the weird synthetic container (I was thinking this could be useful later if clients wanted to custom the Surface, but for now yeah it just adds complexity). 
```
"components": {
      "type": "object",
      "description": "Definitions for UI components supported by this catalog.",
      "propertyNames": {
        "not": {
          "const": "Surface"
        }
      },
      

allowedParents = ['Root', 'Surface', 'Whatever name we want to reserve'] 
```

My issue with allowRoot + allowedParents is they have some weird interplay
allowRoot = "always"
allowedParents = ["Menu"]
What's that mean?

jacobsimionato:
> > After reading this, I think I'm starting to warm up to an `allowRoot` explicit value. Mainly because I don't like having a synthetic component (`Surface`). I think that might be more confusing than helpful to an LLM, and it'll do some weird thing where it adds a `Surface` as a child to something. It also complicates the implementation that now has to explicitly handle this synthetic type, and we have to deal with possible collisions with custom `Surface` components.
> > `allowedChildren` should only be allowed to be empty for leaf components (like `Text`) that don't have children. `allowedParents` can be the empty list only when `allowRoot` is `always`.
> > ```json
> > "ComponentDefinition": {
> >   "type": "object",
> >   "allOf": [
> >     { "$ref": "https://json-schema.org/draft/2020-12/schema" },
> >     {
> >       "type": "object",
> >       "properties": {
> >         "allowRoot": {
> >           "type": "string",
> >           "enum": ["any", "always", "never"],
> >           "default": "any",
> >           "description": "Root mounting policy: 'any' (can be root or child), 'always' (must be root), or 'never' (must be child)."
> >         },
> >         "allowedParents": {
> >           "type": "array",
> >           "items": { "type": "string" },
> >           "uniqueItems": true,
> >           "description": "Parent component types that can contain this component. Can only empty [] when allowRoot is 'always'.  Omitted means any parent type allowed."
> >         },
> >         "allowedChildren": {
> >           "type": "array",
> >           "items": { "type": "string" },
> >           "uniqueItems": true,
> >           "description": "Child component types allowed inside this container. Empty [] means leaf node (no children allowed). Omitted means any child type allowed."
> >         }
> >       }
> >     }
> >   ]
> > }
> > ```
> 
> It's either that or we just reserve the Surface / Root keyword in component names and drop the weird synthetic container (I was thinking this could be useful later if clients wanted to custom the Surface, but for now yeah it just adds complexity).
> 
> ```
> "components": {
>       "type": "object",
>       "description": "Definitions for UI components supported by this catalog.",
>       "propertyNames": {
>         "not": {
>           "const": "Surface"
>         }
>       },
>       
> 
> allowedParents = ['Root', 'Surface', 'Whatever name we want to reserve'] 
> ```
> 
> My issue with allowRoot + allowedParents is they have some weird interplay allowRoot = "always" allowedParents = ["Menu"] What's that mean?

I quite like this PR as-is.

Re the LLM getting tripped up on the synthetic "Surface" etc, I guess that's just a prompt engineering challenge for us right? If we have inference performance issues, we can always explain it to the LLM in a slightly different way, or even omit this information completely.

wrenj:
> > > After reading this, I think I'm starting to warm up to an `allowRoot` explicit value. Mainly because I don't like having a synthetic component (`Surface`). I think that might be more confusing than helpful to an LLM, and it'll do some weird thing where it adds a `Surface` as a child to something. It also complicates the implementation that now has to explicitly handle this synthetic type, and we have to deal with possible collisions with custom `Surface` components.
> > > `allowedChildren` should only be allowed to be empty for leaf components (like `Text`) that don't have children. `allowedParents` can be the empty list only when `allowRoot` is `always`.
> > > ```json
> > > "ComponentDefinition": {
> > >   "type": "object",
> > >   "allOf": [
> > >     { "$ref": "https://json-schema.org/draft/2020-12/schema" },
> > >     {
> > >       "type": "object",
> > >       "properties": {
> > >         "allowRoot": {
> > >           "type": "string",
> > >           "enum": ["any", "always", "never"],
> > >           "default": "any",
> > >           "description": "Root mounting policy: 'any' (can be root or child), 'always' (must be root), or 'never' (must be child)."
> > >         },
> > >         "allowedParents": {
> > >           "type": "array",
> > >           "items": { "type": "string" },
> > >           "uniqueItems": true,
> > >           "description": "Parent component types that can contain this component. Can only empty [] when allowRoot is 'always'.  Omitted means any parent type allowed."
> > >         },
> > >         "allowedChildren": {
> > >           "type": "array",
> > >           "items": { "type": "string" },
> > >           "uniqueItems": true,
> > >           "description": "Child component types allowed inside this container. Empty [] means leaf node (no children allowed). Omitted means any child type allowed."
> > >         }
> > >       }
> > >     }
> > >   ]
> > > }
> > > ```
> > 
> > 
> > It's either that or we just reserve the Surface / Root keyword in component names and drop the weird synthetic container (I was thinking this could be useful later if clients wanted to custom the Surface, but for now yeah it just adds complexity).
> > ```
> > "components": {
> >       "type": "object",
> >       "description": "Definitions for UI components supported by this catalog.",
> >       "propertyNames": {
> >         "not": {
> >           "const": "Surface"
> >         }
> >       },
> >       
> > 
> > allowedParents = ['Root', 'Surface', 'Whatever name we want to reserve'] 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > My issue with allowRoot + allowedParents is they have some weird interplay allowRoot = "always" allowedParents = ["Menu"] What's that mean?
> 
> I quite like this PR as-is.
> 
> Re the LLM getting tripped up on the synthetic "Surface" etc, I guess that's just a prompt engineering challenge for us right? If we have inference performance issues, we can always explain it to the LLM in a slightly different way, or even omit this information completely.

Yeah I think just not telling the LLM about it would work. Which is the same as not having a synthetic "Surface" component and just reserving the "Surface" name. I'm fine either way, it probably comes down to what abstraction model works best in the SDK. Conceptually it seems useful (GE can edit surface.html and surface.ts to put all their surface specific logic in there, in v2 we can add properties and themes and other stuff there), but will defer to you all.

fwiw I added json schema to validate agents don't use "Surface" so we catch it early

```
"ComponentsList": {
      "type": "array",
      "description": "A list containing UI components for the surface.",
      "minItems": 1,
      "items": {
        "allOf": [
          {
            "$ref": "catalog.json#/$defs/anyComponent"
          },
          {
            "type": "object",
            "properties": {
              "component": {
                "not": {
                  "const": "Surface"
                }
              }
            }
          }
        ]
      }
    },
```

gspencergoog (approved):
Okay, I'm convinced.  It would be better if we didn't have to make a special reserved component, but I don't see a better way. Your point about the ambiguity in the `allowRoot` suggestion is a good one.

PR #2157, opened 2026-08-03
feat(spec): add requiresUserActivation to catalog schema and proposal

gspencergoog:
## Summary
Introduces the specification and catalog metadata for user-activation-restricted functions in A2UI (`requiresUserActivation: true`), enforcing Action Context execution boundaries across renderers to prevent automated function auto-invocation during layout rendering, dynamic property interpolation, passive input events, or direct server agent calls.

## Changes
- **Proposal Document (`specification/proposals/user_initiated_functions.md`)**: Defines the Action Context Enforcement model with Activation Event Classification, replacing `userInteractionLevel` component metadata and token objects to achieve full catalog agnosticism. Details asynchronous task boundary rules (scope persistence across `await` turns/Promises/Futures) and platform implementation designs for Web (`web_core`), Flutter (`a2ui_core`), Android (`Jetpack Compose`), and iOS (`SwiftUI`).
- **Catalog Definition Schema (`specification/v1_0/json/catalog_definition.json`)**: Adds `requiresUserActivation: boolean` property (default `false`) to catalog function definition schemas, with a schema conditional guard (`if/then`) ensuring `callableFrom: "agentOnly"` cannot be combined with `requiresUserActivation: true` (permitting `"rendererOnly"` or `"rendererOrAgent"`).
- **Basic Catalog (`specification/v1_0/catalogs/basic/catalog.json`)**: Annotates `openUrl` with `"requiresUserActivation": true`.
- **Documentation (`a2ui_protocol.md` & `basic_catalog_implementation_guide.md`)**: Updates baseline catalog specification and implementation guidance to mandate Action Context security enforcement for `openUrl`.

## Impact & Risks
- **Zero Breaking Changes to Component Schemas**: Fully catalog-agnostic; catalog component definitions require no metadata changes (`userInteractionLevel` is eliminated).
- **Security Enforcement**: Renderers implementing this specification block uninitiated calls to `requiresUserActivation` functions (such as embedded string interpolation `${openUrl(...)}`, layout rendering, passive `onBlurAction`/`onChangeAction` triggers, or direct server agent function invocation messages).

## Testing
- Schema validation of `catalog_definition.json` and `catalog.json`.
- Review of proposal specifications and platform design guides in `user_initiated_functions.md`.


specification/proposals/user_initiated_functions.md
```diff
@@ -0,0 +1,368 @@
...
+
+#### `web_core` Action Binder Implementation:
+```typescript
+// web_core: src/v0_9/rendering/generic-binder.ts
+export function bindAction(dataContext: DataContext, actionCall: ActionDefinition) {
+  return (eventOrOptions?: Event | { event?: Event }) => {
+    // Extract DOM Event from framework event argument or window fallback
+    const domEvent = eventOrOptions instanceof Event
+      ? eventOrOptions
+      : eventOrOptions?.event ?? (typeof window !== 'undefined' ? window.event : undefined);
```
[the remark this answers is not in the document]
gspencergoog: Updated bindAction code example to support React SyntheticEvent instances by checking for nativeEvent.

specification/proposals/user_initiated_functions.md
```diff
@@ -0,0 +1,368 @@
...
+
+  if (fn.requiresUserGesture && (!options?.gestureToken || !options.gestureToken.isValid)) {
+    throw new A2uiSecurityError(
+      `Execution blocked: Function '${name}' requires an active user gesture.`,
+      name,
+    );
+  }
+
+  // Consume token to prevent replay attacks
+  options?.gestureToken?.consume();
```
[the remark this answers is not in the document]
gspencergoog: Updated catalog invoker snippet to only consume gesture token when fn.requiresUserGesture is true.

specification/proposals/user_initiated_functions.md
```diff
@@ -0,0 +1,368 @@
...
+public struct ActionContext {
+    public let gestureToken: UserGestureToken?
+    public init(gestureToken: UserGestureToken? = nil) {
+        self.gestureToken = gestureToken
+    }
+}
+
+// TaskLocal Concurrency Support:
+public enum A2UIGestureScope {
+    @TaskLocal public static var currentToken: UserGestureToken?
```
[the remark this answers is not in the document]
gspencergoog: Updated @TaskLocal property declaration to public static let in Swift code snippet.

specification/proposals/user_initiated_functions.md
```diff
@@ -0,0 +1,372 @@
...
+
+## 3. Proposed Solution
+
+This proposal introduces an optional catalog attribute, `requiresUserGesture: boolean`, to function metadata definitions.
+
+When a function definition includes `requiresUserGesture: true` (such as `openUrl`), the A2UI framework guarantees that the function only executes in response to a physical user interaction (such as a button click, form submission, or touch gesture). Uninitiated calls during layout rendering, dynamic property interpolation, or reactive model updates are blocked and rejected by the renderer engine.
+
+### Protocol Schema Metadata (`requiresUserGesture`)
+
+In catalog function definitions, `requiresUserGesture` specifies whether the function requires an active user gesture context at invocation time:
```
wrenj: requireUserAction?

I think we call them actions right? gestures is a new term?

specification/proposals/user_initiated_functions.md
```diff
@@ -0,0 +1,372 @@
...
+
+```json
+{
+  "functions": {
+    "openUrl": {
+      "type": "object",
+      "description": "Opens the specified URL in a browser or handler. Requires an active user gesture.",
+      "returnType": "void",
+      "requiresUserGesture": true,
```
wrenj: do we have usecases where people want to call functions on load? other option is to just client function can only be called from user actions

specification/proposals/user_initiated_functions.md
gspencergoog: Well, but that's why I used Gesture.  Actions are already types in A2UI, and it doesn't require an `Action`, it requires a user gesture before a function can be invoked.  Also, this would apply even if the function were invoked in any context, even outside of an `Action`. For example, a text field validation function invokes inside of a user gesture (pressing a key), so it *would* be executed there (added a section on remaining risks to the proposal).

specification/proposals/user_initiated_functions.md
gspencergoog: Yes, client functions can be used to populate UI element values. For example, you can use `formatString` to format Text widget contents on load.

wrenj:
should we ensure that any Action event from server requiresUserGesure? eg make sure the LLM doesn't screw up and just emit action { event { "submit_time_off_form" } } and it submits without the user clicking a button,

gspencergoog:
> should we ensure that any Action event from server requiresUserGesure? eg make sure the LLM doesn't screw up and just emit action { event { "submit_time_off_form" } } and it submits without the user clicking a button,

@ditman Suggested that we add the ability to set a "providesUserInteraction" property (I called it `userInteractionLevel`) on catalog definitions, then you could mark buttons and text fields, etc. as having various levels of user interactivity, so now you can restrict these functions further so that they could only be triggered if they were being executed in the context of one of these.  I added that to this proposal, since they're complimentary.

specification/proposals/user_initiated_functions.md
```diff
@@ -0,0 +1,496 @@
...
+}
+
+// In Component onPressed handler:
+ElevatedButton(
+  onPressed: props.disabled ? null : () {
+    final token = UserGestureToken(
+      sourceComponentId: widget.id,
+      interactionLevel: 'action',
+    );
+    actionDispatcher.dispatch(props.action, gestureToken: token);
```
jacobsimionato: Could we automatically inject the token here too, rather than making the user instantiate it manually?

specification/proposals/user_initiated_functions.md
```diff
@@ -0,0 +1,496 @@
...
+    override val key: CoroutineContext.Key<*> = Key
+}
+
+// Jetpack Compose Button Component
+@Composable
+fun A2UIButton(componentId: String, label: String, enabled: Boolean = true, onAction: suspend () -> Unit) {
+    val coroutineScope = rememberCoroutineScope()
+    Button(
+        onClick = {
+            if (!enabled) return@Button
```
jacobsimionato, line 360: It'd be nice if this could be handled inside the SDK too

specification/proposals/user_initiated_functions.md
```diff
@@ -0,0 +1,496 @@
...
+2. **Restrict Token Event Types**: Framework binders must strictly limit `UserGestureToken` generation to activation events (`click`, `touchend`, `submit`, explicit key submission) and exclude passive events like mouse hover (`mouseenter`, `pointermove`).
+
+---
+
+## 3. Proposed Solution
+
+This proposal introduces two complementary catalog attributes:
+
+1. `requiresUserGesture: boolean` on catalog function definitions (e.g., `openUrl`).
+2. `userInteractionLevel: "action" | "input" | "none"` on catalog component definitions (defaulting to `"none"`).
```
jacobsimionato: I think this only works for granular components which have a singular purpose.

But some components might not fit into one category. `TextButton` might have a `DynamicString` as the input for its label, but also contain an action that can trigger a function. So should it be marked as "input" or "action"? It's fine for the button click to trigger the `createBicycle()` function, but not okay for the input dynamic value to do this.

Already, at the SDK level, we generally have separate APIs for resolving dynamic values and triggering actions. Can we just reuse these instead of introducing a new classification system at the component level? So, we just add `requiresUserGesture` to functions in the catalog definition, and nothing about any of the other APIs changes.

Then, perhaps we can lean into @wrenj 's wording e.g. `requireAction` or `actionOnly`.

gspencergoog: This is a good point.

gspencergoog:
Jacob had a good point about multi-use components, so I reworked this design so that:

1) `requiresUserGesture` is renamed to `requiresUserActivation`.  Because `requiresUserAction` is a bit dissonant with the fact that there is an `onBlurAction` and other passive actions. With this proposal, `onBlurAction` blocks functions marked `requiresUserAction: true`, so that might be confusing.
2) Only `requiresUserActivation` is needed: I eliminated the `userInteractionLevel`, and just make it so that it prevents any function marked as `requiresUserActivation` from running unless it runs as a result of a click or submit.

This makes it so that we don't need all of the token stuff, and it's not invasive to the catalog definitions.  You don't have as much control over the cases where a function can be executed, but if we find we need that, we can always add that in later.

PTAL.

specification/proposals/user_initiated_functions.md
gspencergoog: I eliminated tokens.

jacobsimionato (approved):
Thanks for iterating!!

PR #2166, opened 2026-08-04
spec(v1.0): compose ComponentCommon in envelope schema and simplify catalog definitions

wrenj:
Refactors v1.0 component schema validation by composing ComponentCommon at the protocol envelope level in agent_to_renderer.json, rather than requiring every component definition in every catalog to wrap itself in an allOf with ComponentCommon.

This simplifies catalog authoring and schema complexity across all v1.0 catalogs while guaranteeing baseline identity (id) and routing (catalogId) validation regardless of catalog implementation.


specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -142,7 +142,7 @@ The [`catalogs/basic/catalog.json`] schema contains the definitions for all spec
 
 **Swappable Catalogs & Validation:**
 
-The [`agent_to_renderer.json`] envelope schema is designed to be catalog-agnostic. It references components using a placeholder filename: `catalog.json` (specifically `$ref: "catalog.json#/$defs/anyComponent"`).
+The [`agent_to_renderer.json`] envelope schema is designed to be catalog-agnostic. Within [`ComponentsList`], it validates base properties against `common_types.json#/$defs/ComponentCommon` and references components using a placeholder filename: `catalog.json` (specifically `$ref: "catalog.json#/$defs/anyComponent"`).
```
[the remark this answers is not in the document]
gspencergoog: This is a good suggestion.

gspencergoog (approved):

specification/v1_0/json/agent_to_renderer.json
```diff
@@ -57,6 +57,9 @@
       "minItems": 1,
       "items": {
         "allOf": [
+          {
+            "$ref": "common_types.json#/$defs/ComponentCommon"
+          },
           {
             "$ref": "catalog.json#/$defs/anyComponent"
```
jacobsimionato: Here, we reference `anyComponent` from catalog.json (the currently selected catalog). `anyComponent` is a `oneOf` for different component types, each of which themselves are already force to conform to `ComponentCommon`. So adding `ComponentCommon` above is redundant, I believe.

See https://github.com/a2ui-project/a2ui/blob/79931d870fa9110fbac7c54f94ea3845a653b8db/specification/v1_0/catalogs/basic/catalog.json#L10-L21

And the requirement documented here for any catalog:
https://github.com/a2ui-project/a2ui/blob/79931d870fa9110fbac7c54f94ea3845a653b8db/specification/v1_0/docs/a2ui_protocol.md?plain=1#L564-L569

However, perhaps we should remove both of the above, and instead just include something in https://github.com/a2ui-project/a2ui/blob/main/specification/v1_0/json/catalog_definition.json that forces the constant "component" property to be set, but keep your change here which blends in `ComponentCommon` in the shared layer which is way neater.

I assume the overall motivation here is catalog-agnostic wire protocol conformance? To have a formal schema for that, we'd need a version of this that does not refer to the catalog schema as the magic `catalog.json`.

wrenj: I'm not seeing component catalogs include ComponentCommon, eg

https://www.gstatic.com/vertexaisearch/a2ui/v0_8/gemini_enterprise_custom_catalog.json

I think a better structure would be a Component wrapper that has id, catalogid, and component field which points to something from the catalog. Maybe thats what you were saying.

jacobsimionato: Yeah yeah, I think we agree! The link you showed is a v0.8 catalog, and that was before we added more guidance around exactly what a catalog should look like.

I prefer your structure, which requires less cruft in catalog.json. So to execute this, I'm suggesting that you update the v1.0 protocol docs that I linked to *stop* saying that each component needs to explictly fold in `ComponentCommon`, and update the basic catalog for v1.0 to not do this.

wrenj: @jacobsimionato @gspencergoog take a look now, I updated based on what I think the concensus is.

We would need to update the inference libraries I think to say you need to output Component, which is an allOf basecomponent and anyComponent

specification/v1_0/json/catalog_definition.json
```diff
@@ -187,7 +187,7 @@
               "items": {
                 "type": "string"
               },
-              "description": "The list of parent component type names that can contain this component type. If omitted, all parent component types are allowed. To restrict a component so it can appear only as the top-level component (id='root') of a surface, set \"allowedParents\": [\"Surface\"]. To allow a component as either the top-level component of a surface or a child of a specific container, specify both (e.g., \"allowedParents\": [\"Surface\", \"CanvasContainer\"]).",
+              "description": "The list of parent component type names that can contain this component type. If omitted, all parent component types are allowed. To restrict a component so it can appear only as the top-level component (id='root') of a surface, set \"allowedParents\": [\"Surface\"]. To allow a component as either the top-level component of a surface of a child of a specific container, specify both (e.g., \"allowedParents\": [\"Surface\", \"CanvasContainer\"]).",
```
jacobsimionato: I think revert this change?

specification/v1_0/test/cases/initial_state_validation.json
```diff
@@ -178,6 +178,23 @@
           ]
         }
       }
+    },
+    {
+      "description": "Invalid: component missing id required by ComponentCommon",
```
jacobsimionato, line 294: Nice test

specification/v1_0/json/common_types.json
```diff
@@ -35,9 +35,6 @@
         "catalogId": {
           "type": "string",
           "description": "The catalog ID for this component, overriding any surface-level default catalogId."
-        },
```
jacobsimionato, line 57: Why is this being removed? My understanding is that this is intentionally in ComponentCommon?

wrenj: I can revert this change if you prefer

Custom components don't need aria labels since they are defined in the HTML itself (eg, the time off form, or suggestion chips, or a canvas component), so I made it opt in. But either works for me if we want ot make sure accessibility gets attention. let me know, thanks

wrenj: from gemini

No, accessibility does not belong on ComponentCommon. 

While the intent to elevate accessibility is good, placing it on 

ComponentCommon
 conflates protocol envelope mechanics with component capability contracts.

Why ComponentCommon is the Wrong Place
ComponentCommon is for Universal Wire Envelopes, Not Component Props

In A2UI (specifically standardized in v1.0), 

ComponentCommon
 defines baseline properties that every single component instance must support across all catalogs (id and optional catalogId).
If a property is in ComponentCommon, agent_to_renderer.json validates and exposes it on every component unconditionally via allOf.
Many Components Cannot Meaningfully Support Top-Level Accessibility

Layout/Structural elements (Row, Column, Spacer, Divider) are presentation containers; assigning an accessible label/description at the container level is often a no-op or actively harms screen reader navigation.
Composite/Embedded widgets (e.g., custom HTML forms, embedded iframe widgets, Vega-Lite visualizers) manage their own internal DOM accessibility tree and will ignore top-level attributes.
LLM Prompt Efficiency and Hallucination Risk

Catalog schemas are fed into LLM prompts. If accessibility is stamped onto every component:
It wastes prompt tokens.
It encourages models to hallucinate accessibility attributes on components whose renderers do nothing with them (creating a false sense of compliance).
Consistency with Other Cross-Cutting Capabilities

A2UI handles other reusable capabilities (like validation 

Checkable
 or interactions 

Action
) as standardized reusable definitions in common_types.json that individual catalog components opt into via $ref. Accessibility should follow the exact same architectural pattern.

specification/v1_0/catalogs/basic/catalog.json
```diff
@@ -9,506 +9,431 @@
...
+        },
+        "accessibility": {
+          "$ref": "https://a2ui.org/specification/v1_0/common_types.json#/$defs/AccessibilityAttributes"
+        },
+        "weight": {
+          "type": "number",
+          "description": "The relative weight of this component within a Row or Column. This is similar to the CSS 'flex-grow' property. Note: this may ONLY be set when the component is a direct descendant of a Row or Column."
         }
-      ],
-      "unevaluatedProperties": false
```
gspencergoog, line 40: Why are we removing this?  It was added to prevent LLMs adding properties that weren't listed.

specification/v1_0/json/common_types.json
gspencergoog: I think having accessibility defined here makes sense. It provides a nudge for the LLM, but it's not required, so it doesn't have to take up extra tokens.

wrenj: sg, reverted that part of this PR

jacobsimionato (approved):

PR #2191, opened 2026-08-07
feat(angular): support universal web components in v0.9 renderer

josemontespg:
## Overview

Adds support in `@a2ui/angular` for rendering universal custom element Web Components, enabling the Angular renderer to share the same underlying Basic Catalog components provided by `@a2ui/web_core`.

## Details

- **`toWebComponent` Adapter**: Implements an adapter in `@a2ui/angular` that binds Angular context, event dispatching, and reactive state inputs to standard HTML custom element web components.
- **Component Host & Surface Integration**: Updates `ComponentHostComponent` and `SurfaceComponent` to dynamically render custom element tags alongside native Angular components while preserving schema validation and data binding.
- **Deduplication**: Replaces the redundant framework-specific basic catalog components in `@a2ui/angular` with the universal Web Component implementations from `@a2ui/web_core`.
- **Testing & Verification**: Updates Angular unit and explorer integration test suites, verifying full compatibility with all catalog examples.


josemontespg (commented):
Thanks for the feedback! I've addressed all the points by adding the `EnvironmentInjector` import, using the provided injector in `createComponent` to preserve dependency injection contexts, and explicitly copying the original component class to the returned implementation to prevent silent rendering failures.

PR #2199, opened 2026-08-07
fix(eval): make reporter.py parse on Python 3.10 and 3.11

### LINKED ISSUE #2198: [BUG]: eval declares Python >=3.10 but reporter.py only parses on 3.12+

## Describe the Bug

`eval/pyproject.toml` sets `requires-python = ">=3.10"`, but
`eval/iterative_format_optimizer/skills/inference-format-optimizer/scripts/utils/reporter.py`
does not parse below 3.12.

Line 363 puts backslashes inside the expression part of an f-string:

```python
report.append(f"- **Prompt**:\n  > {prompt_str.replace('\n', '\n  > ')}")
```

That became legal in 3.12 with PEP 701. On 3.10 and 3.11 it is a SyntaxError, so the module
never imports and three test modules fail at collection:

```
ERROR iterative_format_optimizer/skills/inference-format-optimizer/tests/test_optimize_format.py
ERROR iterative_format_optimizer/skills/inference-format-optimizer/tests/test_sync_history.py
ERROR iterative_format_optimizer/skills/inference-format-optimizer/tests/test_utils.py
E   SyntaxError: f-string expression part cannot include a backslash
```

`run_evals.yml` pins Python to 3.12, so CI cannot catch this. It only surfaces for someone
setting up `eval/` on a version the package claims to support, which is how I ran into it.

## Steps to Reproduce

1. `cd eval`
2. `uv sync --python 3.11`
3. `uv run python -m pytest`
4. Three collection errors with the SyntaxError above

Same result on 3.10. On 3.12 the suite runs and 127 tests pass.

## Expected Behavior

Either the package installs and tests cleanly on the versions `requires-python` advertises,
or the declared floor matches what the code actually needs.

## Environment Details

- **OS**: macOS 26.5.2 (arm64)
- **Browser/Platform**: Python 3.10 and 3.11 fail, 3.12 passes
- **SDK/Package Name & Version**: a2ui-eval 0.1.0 at 72fd6b3b
- **Protocol Version**: n/a
- **Agent Framework & LLM Model**: n/a, this reproduces without any model calls

## Additional Context

There are two reasonable fixes and I would rather not assume which you want:

1. Hoist the replacement into a local variable so the f-string has no backslash in the
   expression. Two lines in `reporter.py`, keeps the advertised 3.10 floor.
2. Raise `requires-python` to `>=3.12` so the metadata matches CI.

#1785 went in the direction of widening support so downstream projects on 3.10-3.13 could
adopt `a2ui-core`, which makes option 1 feel more consistent, but I may be missing context
on how much the optimizer scripts are meant to be portable.

I have the fix for option 1 ready and can open a PR as soon as you tell me which you prefer.

FelippeRoza:
# Description

`eval/pyproject.toml` declares `requires-python = ">=3.10"`, but `reporter.py` put backslashes inside an f-string expression, which only became valid in 3.12 with PEP 701. On 3.10 and 3.11 the module raised `SyntaxError` at import time, which took three test modules down at collection. `run_evals.yml` pins Python to 3.12, so CI never saw it.

This hoists the replacement into a local variable so the f-string contains no backslash. Output is unchanged.

```diff
-            report.append(f"- **Prompt**:\n  > {prompt_str.replace('\n', '\n  > ')}")
+            quoted_prompt = prompt_str.replace("\n", "\n  > ")
+            report.append(f"- **Prompt**:\n  > {quoted_prompt}")
```

Verified that the file parses on 3.10, 3.11 and 3.12, and that `uv run python -m pytest` in `eval/` goes from three collection errors on 3.11 to none. On 3.12 the suite is unchanged at 127 passed.

Two pre-existing failures in `TestArchiver` remain on macOS, unrelated to this change and present on `main` before it. I will file those separately.

Fixes #2198

As noted in the issue, the alternative is raising `requires-python` to `>=3.12` to match CI instead. If you would rather go that way, say so and I will close this.

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].
- [ ] I have built and run at least one [sample]. Not yet. This change is confined to the Python eval tooling and does not touch renderers or samples.

For this PR:

- [ ] I have updated the relevant CHANGELOG.md file. There is no `eval/CHANGELOG.md`; the CHANGELOG files in this repo cover the JS renderer packages.
- [ ] I updated/added relevant documentation. No documented behaviour changes.
- [x] My code changes (if any) have tests. No new test was added; the change is covered by the three existing test modules in `iterative_format_optimizer` that could not be collected at all on 3.10 or 3.11 before this fix.
- [ ] If my branch is on a fork, I have verified that [scripts/e2e_test.sh](https://github.com/a2ui-project/a2ui/blob/main/scripts/e2e_test.sh) passes. That script runs the Flutter restaurant_finder e2e tests and needs a Gemini API key; I do not have Flutter set up and this change does not touch that path. Happy to run it if you consider it necessary here.


Varun-S10:
Hi @FelippeRoza, there are merge conflicts on this branch. Please rebase on main when you get a chance so we can test and validate the changes. Thanks.

FelippeRoza:
Rebased onto main. Diff is unchanged at two lines, and `uv run python -m pytest` in `eval/` gives 127 passed locally.

One thing that may not be obvious from the checks list: every workflow here is sitting at `action_required`, so nothing beyond `cla/google` and `check-changes` has actually run. `Run A2UI Evals` and `Presubmit Lint` will need workflows approved before they can report.

polina-c (approved):

polina-c:
Thank you!

Issue #1298, opened 2026-04-28
The web renderers handle `Image` inconsistently with each other and with the spec


andrewkolos:
The three web renderers each implement the basic catalog `Image` differently, leading to visibly different rendered output for the same A2UI input. They also diverge from the spec on the default `fit` value.

| Renderer | Implementation | Default `fit` |
|----------|----------------|---------------|
| React | Bare `<img>` ([Image.tsx#L53](https://github.com/google/A2UI/blob/c55a37dc783146a607003ff3a3ecda92fbc36ac1/renderers/react/src/v0_9/catalog/basic/components/Image.tsx#L53)) | `'fill'` ([Image.tsx#L26](https://github.com/google/A2UI/blob/c55a37dc783146a607003ff3a3ecda92fbc36ac1/renderers/react/src/v0_9/catalog/basic/components/Image.tsx#L26)) — matches spec |
| Lit | `<img>` inside an `a2ui-image` custom element ([Image.ts#L25](https://github.com/google/A2UI/blob/c55a37dc783146a607003ff3a3ecda92fbc36ac1/renderers/lit/src/v0_9/catalogs/basic/components/Image.ts#L25), [Image.ts#L87-L93](https://github.com/google/A2UI/blob/c55a37dc783146a607003ff3a3ecda92fbc36ac1/renderers/lit/src/v0_9/catalogs/basic/components/Image.ts#L87-L93)) with `width: 100%; height: auto` in shadow CSS ([Image.ts#L38-L44](https://github.com/google/A2UI/blob/c55a37dc783146a607003ff3a3ecda92fbc36ac1/renderers/lit/src/v0_9/catalogs/basic/components/Image.ts#L38-L44)) | `'fill'` inline ([Image.ts#L84](https://github.com/google/A2UI/blob/c55a37dc783146a607003ff3a3ecda92fbc36ac1/renderers/lit/src/v0_9/catalogs/basic/components/Image.ts#L84)), but the wrapper's `height: auto` might make it visually irrelevant? |
[...]

Issue #1297, opened 2026-04-28
React (v0.9) basic catalog `Image` distorts when placed in a flex `Row`


andrewkolos:
Running the react shell with restaurant finder, the images rendered inside card layouts appear visibly stretched/squished--pixels are scaled non-uniformly to match the row's height.

## Repro steps

1. Run the restaurant-finder agent (`samples/agent/adk/restaurant_finder`).
2. Run the React shell (`samples/client/react/shell`).
3. Open the app at `http://localhost:5003/` and submit any prompt (or use `?mock=true` to skip the agent call).
4. Observe the restaurant photos in the result cards.

<img width="512" height="282" alt="Image" src="https://github.com/user-attachments/assets/f399fb5b-1d42-41b8-ba6d-cfd8af55386d" />

## Root cause

The basic catalog's `Row` defaults to `align-items: stretch`, so any flex child gets stretched to the row's height. The React `Image` renders a bare `<img>` whose default `objectFit` is `'fill'`, causing the image to scale non-uniformly along the stretched axis.

Related: https://github.com/google/A2UI/issues/1298

gspencergoog:
A2UI Triage: Classified as a P2 bug in the React renderer. We will investigate the image distortion when `Image` is used inside a flex `Row`. cc @ditman

Issue #666, opened 2026-02-23
Ergonomic API to implement custom catalogs on web


jacobsimionato:
Ergonomic API to implement custom catalogs on web see [Web renderers v0.9](https://docs.google.com/document/d/1dQoj-WyGxCysCdDU16TVtSPCeXyYNo1hdQ3-H6gSe-c/edit?tab=t.0) (internal Google document)

PR #2205, opened 2026-08-07
fix(lit): align basic catalog component behaviors with angular implementation

### LINKED ISSUE #1270: Support "portable A2UI web Components" which can be installed in a Catalog for *any* framework adapter
On the web, we have adapters for many different frameworks, that allow Angular developers to define components in an Angular-specific way, and React developers to define components in a React-specific way. This is great for the use case where a developer needs to implement a component for one specific framework, because they can do so in a familiar, ergonomic way.

But, there is also a use case where you want to define a component that can be used with any web framework via any framework adapter. This is especially true of shared components which some providers may vend. E.g. MySocialMedia company might implement a MySocialMediaPost Component library which can be used to display a specific post. Then different application developers might install the my_social_media_a2ui library from NPM and add that component to their catalog. In this case, we want to make it so that MySocialMedia does not need to implement 5 different versions of their component for different frameworks, but just implement it once in a low-dependencies way, and have it useable everywhere.

Also, developers of new A2UI framework adapters (e.g. for other frameworks like Svelte, Vue etc) should be able to implement support for this type of component cheaply, and give developers access to the ecosystem of Components.

Also see https://github.com/google/A2UI/issues/931 which describes this pattern specifically as a way to centralize the basic catalog implementation.

josemontespg:
## Overview

Aligns the `@a2ui/lit` Basic Catalog component implementations, styling contracts, and lifecycle behaviors with the `@a2ui/angular` reference implementation.

## Motivation & Architectural Rationale

This PR is a foundational step in our cross-framework rendering architecture:

1. **Prerequisite for Universal Components (`@a2ui/web_core`)**:
   In the subsequent PRs of this stack, the Basic Catalog implementations are extracted into unified, universal Web Components hosted in `@a2ui/web_core` and shared directly across `@a2ui/lit`, `@a2ui/angular`, and upcoming renderers (such as React). To ensure this migration produces zero behavioral drift or visual regressions, all renderers must first converge on a canonical DOM structure, CSS variable contract, and event model.

2. **Adopting the Canonical Reference Behaviors & Performance Enhancements**:
   - **Stable List Reconciliation (`Row`, `Column`, `List`)**: Uses Lit's `repeat` directive keyed by `child.basePath/child.id` to preserve DOM element identity and form input state across list reorders, insertions, and deletions.
   - **Comprehensive Form Validation & State Fallback (`TextField`)**: Renders all accumulated validation errors from `validationErrors` instead of only the first error, ensures full-width container styling (`width: 100%`, `box-sizing: border-box`), and provides a direct `dataContext.set` mutation fallback.
   - **Cross-Platform Layout Distribution (`Column`, `Row`)**: Adds `width: 100%` container styling, explicit gap fallbacks (`16px`), and `baseline` flex alignment mapping.
   - **Dynamic Tab Safety (`Tabs`)**: Implements `activeIndex` bounds clamping in `willUpdate()` so dynamic removals from `tabs` do not leave the component in an out-of-bounds state rendering `nothing`.
   - **Fast-Path Text Rendering (`Text`)**: Renders semantic HTML tags directly (`<h1>`–`<h5>`, `<em>`) for non-markdown variants (`h1`-`h5`, `caption`), wraps text with `:host { display: contents; }`, and adds flex weight styling.
   - **Media Sizing & Accessibility (`Image`, `Video`)**: Preserves intrinsic aspect ratios on `Image` and wraps `Video` in `.a2ui-video-container` with fallback accessibility messaging.
   - **Consistent Dialog Lifecycle & Styling (`Modal`)**: Adopts explicit `@state() isOpen` state tracking and standard overlay styling tokens.
   - **Light DOM Radio Scoping, Checkbox Cleanup & DOM Alignment (`ChoicePicker`)**: Aligns DOM structures and CSS classes with the Angular reference implementation (`.a2ui-choice-picker`, `.a2ui-option-label`, `.a2ui-option-text`, `.a2ui-chip`, `type="button"`), scopes radio input `name` attributes to surface and data context path to prevent cross-surface radio collisions in Light DOM, and omits `name` attributes on checkbox inputs.

## Changes Made

- **`Row`, `Column`, `List`**: Switched from `map()` to `repeat()` directive for key-tracked DOM node preservation; added `baseline` alignment and container gap variable fallbacks.
- **`Tabs`**: Added `activeIndex` bounds clamping in `willUpdate()` and updated tab bar class styling.
- **`Text`**: Added `NON_MARKDOWN_VARIANTS` fast-path rendering for header/caption variants, `:host { display: contents; }`, and `props.weight` flex styling.
- **`TextField`**: Renders all errors in `props.validationErrors`, adds input fallback to `dataContext.set`, and standardizes sizing.
- **`Video`**: Wrapped in `.a2ui-video-container` with fallback message for unsupported browsers.
- **`Image`**: Removed `width: 100%` default constraint to allow intrinsic sizing.
- **`Modal`**: Migrated to overlay hierarchy with explicit `@state() isOpen` tracking.
- **`ChoicePicker`**: Aligned DOM structures and CSS classes with the Angular reference implementation (`.a2ui-choice-picker`, `.a2ui-option-label`, `.a2ui-option-text`, `.a2ui-chip`, `type="button"`), scoped radio input `name` attributes to `a2ui-choice-${surfaceId}-${componentId}-${sanitizedPath}` to prevent collisions across surfaces in Light DOM, and omitted `name` attributes on checkboxes via `nothing`.

## Pre-launch Checklist

- [x] I signed the [CLA].

Related to #1270

### DIFF
--- a/renderers/lit/a2ui_explorer/tests/v0_9/29_movie-card.test.ts
+++ b/renderers/lit/a2ui_explorer/tests/v0_9/29_movie-card.test.ts
@@ -67,8 +67,8 @@ describe('Example: Movie Card', () => {
     button.click();
     await whenSettled(gallery);
 
-    const modal = querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLDialogElement;
-    expect(modal?.open).toBeTruthy();
+    const modal = querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLElement;
+    expect(modal).toBeTruthy();
 
     const video = querySelectorAllDeep(surface, 'video')[0] as HTMLVideoElement;
     expect(video).toBeTruthy();
--- a/renderers/lit/a2ui_explorer/tests/v0_9/36_modal.test.ts
+++ b/renderers/lit/a2ui_explorer/tests/v0_9/36_modal.test.ts
@@ -39,18 +39,14 @@ describe('Example: Modal', () => {
     expect(trigger).toBeTruthy();
 
     // Check modal is closed initially
-    expect(
-      (querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLDialogElement)?.open,
-    ).toBeFalsy();
+    expect(querySelectorAllDeep(surface, '.a2ui-modal-overlay').length).toBe(0);
 
     // Click trigger
     trigger.click();
     await whenSettled(gallery);
 
     // Check modal is open
-    expect(
-      (querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLDialogElement)?.open,
-    ).toBeTruthy();
+    expect(querySelectorAllDeep(surface, '.a2ui-modal-overlay').length).toBe(1);
 
     // Check content
     expect(getDeepTextContent(querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0])).toContain(
@@ -64,8 +60,6 @@ describe('Example: Modal', () => {
     await whenSettled(gallery);
 
     // Check modal is closed
-    expect(
-      (querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLDialogElement)?.open,
-    ).toBeFalsy();
+    expect(querySelectorAllDeep(surface, '.a2ui-modal-overlay').length).toBe(0);
   });
 });
--- a/renderers/lit/a2ui_explorer/tests/v0_9/36_modal.test.ts
+++ b/renderers/lit/a2ui_explorer/tests/v0_9/36_modal.test.ts
@@ -39,18 +39,14 @@ describe('Example: Modal', () => {
     expect(trigger).toBeTruthy();
 
     // Check modal is closed initially
-    expect(
-      (querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLDialogElement)?.open,
-    ).toBeFalsy();
+    expect(querySelectorAllDeep(surface, '.a2ui-modal-overlay').length).toBe(0);
 
     // Click trigger
     trigger.click();
     await whenSettled(gallery);
 
     // Check modal is open
-    expect(
-      (querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLDialogElement)?.open,
-    ).toBeTruthy();
+    expect(querySelectorAllDeep(surface, '.a2ui-modal-overlay').length).toBeGreaterThan(0);
 
     // Check content
     expect(getDeepTextContent(querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0])).toContain(
@@ -64,8 +60,6 @@ describe('Example: Modal', () => {
     await whenSettled(gallery);
 
     // Check modal is closed
-    expect(
-      (querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLDialogElement)?.open,
-    ).toBeFalsy();
+    expect(querySelectorAllDeep(surface, '.a2ui-modal-overlay').length).toBe(0);
   });
 });
--- a/renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts
+++ b/renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts
@@ -23,14 +23,24 @@ import {A2uiController} from '../../../a2ui-controller.js';
 @customElement('a2ui-audioplayer')
 export class A2uiAudioPlayerElement extends BasicCatalogA2uiLitElement<typeof AudioPlayerApi> {
   static override styles = css`
-    :host,
-    a2ui-audioplayer {
+    .a2ui-audio-player {
       display: flex;
       flex-direction: column;
       gap: var(--a2ui-spacing-xs, 0.25rem);
       background: var(--a2ui-audioplayer-background, transparent);
       border-radius: var(--a2ui-audioplayer-border-radius, 0);
       padding: var(--a2ui-audioplayer-padding, 0);
+      width: 100%;
+    }
+    .a2ui-audio-description {
+      font-size: var(--a2ui-font-size-s, 0.875rem);
+      color: var(
+        --a2ui-audioplayer-description-color,
+        var(--a2ui-text-caption-color, light-dark(#666, #aaa))
+      );
+    }
+    .a2ui-audio {
+      width: 100%;
     }
   `;
 
@@ -43,8 +53,14 @@ export class A2uiAudioPlayerElement extends BasicCatalogA2uiLitElement<typeof Au
     if (!props) return nothing;
 
     return html`
-      ${props.description ? html`<p>${props.description}</p>` : nothing}
-      <audio src=${props.url} controls></audio>
+      <div class="a2ui-audio-player">
+        ${props.description
+          ? html`<div class="a2ui-audio-description">${props.description}</div>`
+          : nothing}
+        <audio src=${props.url || nothing} controls class="a2ui-audio">
+          Your browser does not support the audio tag.
+        </audio>
+      </div>
     `;
   }
 }
--- a/renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts
+++ b/renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts
@@ -23,14 +23,21 @@ import {A2uiController} from '../../../a2ui-controller.js';
 @customElement('a2ui-audioplayer')
 export class A2uiAudioPlayerElement extends BasicCatalogA2uiLitElement<typeof AudioPlayerApi> {
   static override styles = css`
-    :host,
-    a2ui-audioplayer {
+    .a2ui-audio-player {
       display: flex;
       flex-direction: column;
       gap: var(--a2ui-spacing-xs, 0.25rem);
       background: var(--a2ui-audioplayer-background, transparent);
       border-radius: var(--a2ui-audioplayer-border-radius, 0);
       padding: var(--a2ui-audioplayer-padding, 0);
+      width: 100%;
+    }
+    .a2ui-audio-description {
+      font-size: var(--a2ui-font-size-s, 0.875rem);
+      color: var(--a2ui-text-caption-color, light-dark(#666, #aaa));
+    }
+    .a2ui-audio {
+      width: 100%;
     }
   `;
 
@@ -39,12 +46,18 @@ export class A2uiAudioPlayerElement extends BasicCatalogA2uiLitElement<typeof Au
   }
 
   override render() {
-    const props = this.controller.props;
+    const props = this.controller?.props;
     if (!props) return nothing;
 
     return html`
-      ${props.description ? html`<p>${props.description}</p>` : nothing}
-      <audio src=${props.url} controls></audio>
+      <div class="a2ui-audio-player">
+        ${props.description
+          ? html`<div class="a2ui-audio-description">${props.description}</div>`
+          : nothing}
+        <audio src=${props.url || nothing} controls class="a2ui-audio">
+          Your browser does not support the audio tag.
+        </audio>
+      </div>
     `;
   }
 }
--- a/renderers/lit/src/v0_9/catalogs/basic/components/Button.ts
+++ b/renderers/lit/src/v0_9/catalogs/basic/components/Button.ts
@@ -42,53 +42,46 @@ export class A2uiBasicButtonElement extends BasicCatalogA2uiLitElement<typeof Bu
    * - `--a2ui-button-margin`: The outer margin of the button. Defaults to `--a2ui-spacing-m`.
    */
   static override styles = css`
-    :host,
-    a2ui-basic-button {
-      display: inline-block;
-      margin: var(--a2ui-button-margin, var(--a2ui-spacing-m));
-    }
-    :where(:host, a2ui-basic-button) {
-      --_color-primary: var(--a2ui-color-primary, #17e);
-      --_button-border-radius: var(--a2ui-button-border-radius, var(--a2ui-spacing-s, 0.25rem));
-      --_button-padding: var(
+    .a2ui-button {
+      padding: var(
         --a2ui-button-padding,
         var(--a2ui-spacing-m, 0.5rem) var(--a2ui-spacing-l, 1rem)
       );
-      --_button-border: var(
+      border-radius: var(--a2ui-button-border-radius, var(--a2ui-spacing-s, 0.25rem));
+      border: var(
         --a2ui-button-border,
         var(--a2ui-border-width, 1px) solid var(--a2ui-color-border, #ccc)
       );
-    }
-    .a2ui-button {
-      --_a2ui-text-margin: 0;
-      --_a2ui-text-color: var(--a2ui-color-on-secondary, #333);
-      padding: var(--_button-padding);
+      cursor: pointer;
+      margin: var(--a2ui-button-margin, var(--a2ui-spacing-m, 0.5rem));
       background: var(--a2ui-button-background, var(--a2ui-color-surface, #fff));
       box-shadow: var(--a2ui-button-box-shadow, none);
       font-weight: var(--a2ui-button-font-weight, normal);
-      color: var(--_a2ui-text-color);
-      border: var(--_button-border);
-      border-radius: var(--_button-border-radius);
-      cursor: pointer;
-      display: inline-flex;
-      align-items: center;
-      justify-content: center;
-    }
-    .a2ui-button.a2ui-button-primary {
-      --_a2ui-text-color: var(--a2ui-color-on-primary, #fff);
-      background-color: var(--_color-primary);
+      --_a2ui-text-margin: 0;
+      --_a2ui-text-color: var(--a2ui-color-on-secondary, #333);
       color: var(--_a2ui-text-color);
     }
     .a2ui-button:hover {
       background-color: var(--a2ui-color-secondary-hover, #ddd);
     }
-    .a2ui-button.a2ui-button-primary:hover {
+    .a2ui-button.primary {
+      background: var(--a2ui-color-primary, #17e);
+      --_a2ui-text-color: var(--a2ui-color-on-primary, #fff);
+      color: var(--_a2ui-text-color);
+      border: none;
+    }
+    .a2ui-button.primary:hover {
       background-color: var(--a2ui-color-primary-hover, #fbd);
     }
-    .a2ui-button.a2ui-button-borderless {
+    .a2ui-button.borderless {
       background: none;
+      border: none;
       padding: 0;
-      color: var(--_color-primary);
+      color: var(--a2ui-color-primary, #17e);
+    }
+    .a2ui-button:disabled {
+      cursor: not-allowed;
+      opacity: 0.5;
     }
   `;
 
@@ -101,16 +94,22 @@ export class A2uiBasicButtonElement extends BasicCatalogA2uiLitElement<typeof Bu
     if (!props) return nothing;
 
     const isDisabled = props.isValid === false;
+    const variant = props.variant ?? 'default';
 
     const classes = {
       'a2ui-button': true,
-      ['a2ui-button-' + (props.variant || 'default')]: true,
+      [variant]: true,
+      ['a2ui-button-' + variant]: true,
     };
 
     return html`
       <button
+        type="button"
         class=${classMap(classes)}
-        @click=${() => !isDisabled && props.action && props.action()}
+        @click=${() => {
+          if (isDisabled) return;
+          props.action?.();
+        }}
         ?disabled=${isDisabled}
       >
         ${props.child ? html`${this.renderNode(props.child)}` : nothing}
--- a/renderers/lit/src/v0_9/catalogs/basic/components/Card.ts
+++ b/renderers/lit/src/v0_9/catalogs/basic/components/Card.ts
@@ -33,19 +33,17 @@ export class A2uiCardElement extends BasicCatalogA2uiLitElement<typeof CardApi> 
    * - `--a2ui-card-margin`: The outer margin of the card. Defaults to `--a2ui-spacing-m`.
    */
   static override styles = css`
-    :host,
-    a2ui-card {
-      display: block;
+    .a2ui-card {
+      padding: var(--a2ui-card-padding, var(--a2ui-spacing-m, 16px));
+      border-radius: var(--a2ui-card-border-radius, var(--a2ui-border-radius, 8px));
+      box-shadow: var(--a2ui-card-box-shadow, 0 2px 4px rgba(0, 0, 0, 0.1));
+      background: var(--a2ui-card-background, var(--a2ui-color-surface, #fff));
+      color: var(--a2ui-color-on-surface, #333);
       border: var(
         --a2ui-card-border,
         var(--a2ui-border-width, 1px) solid var(--a2ui-color-border, #ccc)
       );
-      border-radius: var(--a2ui-card-border-radius, var(--a2ui-border-radius, 8px));
-      padding: var(--a2ui-card-padding, var(--a2ui-spacing-m, 16px));
-      background: var(--a2ui-card-background, var(--a2ui-color-surface, #fff));
-      color: var(--a2ui-color-on-surface, #333);
-      box-shadow: var(--a2ui-card-box-shadow, 0 2px 4px rgba(0, 0, 0, 0.1));
-      margin: var(--a2ui-card-margin, var(--a2ui-spacing-m));
+      margin: var(--a2ui-card-margin, var(--a2ui-spacing-m, 16px));
     }
   `;
 
@@ -57,7 +55,9 @@ export class A2uiCardElement extends BasicCatalogA2uiLitElement<typeof CardApi> 
     const props = this.controller.props;
     if (!props) return nothing;
 
-    return html` ${props.child ? html`${this.renderNode(props.child)}` : nothing} `;
+    return html`
+      <div class="a2ui-card">${props.child ? this.renderNode(props.child) : nothing}</div>
+    `;
   }
 }
 
--- a/renderers/lit/src/v0_9/catalogs/basic/components/Card.ts
+++ b/renderers/lit/src/v0_9/catalogs/basic/components/Card.ts
@@ -22,30 +22,17 @@ import {A2uiController} from '../../../a2ui-controller.js';
 
 @customElement('a2ui-card')
 export class A2uiCardElement extends BasicCatalogA2uiLitElement<typeof CardApi> {
-  /**
-   * The styles of the card can be customized by redefining the following
-   * CSS variables:
-   *
-   * - `--a2ui-card-border`: The styling for the card border. Defaults to `--a2ui-border-width` width and `--a2ui-color-border` color.
-   * - `--a2ui-card-border-radius`: The border radius of the card. Defaults to `--a2ui-border-radius`.
-   * - `--a2ui-card-padding`: The padding of the card. Defaults to `--a2ui-spacing-m`.
-   * - `--a2ui-card-box-shadow`: The box shadow of the card. Defaults to `0 2px 4px rgba(0,0,0,0.1)`.
-   * - `--a2ui-card-margin`: The outer margin of the card. Defaults to `--a2ui-spacing-m`.
-   */
   static override styles = css`
-    :host,
-    a2ui-card {
-      display: block;
+    .a2ui-card {
+      padding: var(--a2ui-card-padding, var(--a2ui-spacing-m, 16px));
+      border-radius: var(--a2ui-card-border-radius, var(--a2ui-border-radius, 8px));
+      box-shadow: var(--a2ui-card-box-shadow, 0 2px 4px rgba(0, 0, 0, 0.1));
+      background: var(--a2ui-card-background, var(--a2ui-color-surface, #fff));
       border: var(
         --a2ui-card-border,
         var(--a2ui-border-width, 1px) solid var(--a2ui-color-border, #ccc)
       );
-      border-radius: var(--a2ui-card-border-radius, var(--a2ui-border-radius, 8px));
-      padding: var(--a2ui-card-padding, var(--a2ui-spacing-m, 16px));
-      background: var(--a2ui-card-background, var(--a2ui-color-surface, #fff));
-      color: var(--a2ui-color-on-surface, #333);
-      box-shadow: var(--a2ui-card-box-shadow, 0 2px 4px rgba(0, 0, 0, 0.1));
-      margin: var(--a2ui-card-margin, var(--a2ui-spacing-m));
+      margin: var(--a2ui-card-margin, var(--a2ui-spacing-m, 16px));
     }
   `;
 
@@ -54,10 +41,12 @@ export class A2uiCardElement extends BasicCatalogA2uiLitElement<typeof CardApi> 
   }
 
   override render() {
-    const props = this.controller.props;
+    const props = this.controller?.props;
     if (!props) return nothing;
 
-    return html` ${props.child ? html`${this.renderNode(props.child)}` : nothing} `;
+    return html`
+      <div class="a2ui-card">${props.child ? this.renderNode(props.child) : nothing}</div>
+    `;
   }
 }
 
--- a/renderers/lit/src/v0_9/catalogs/basic/components/CheckBox.ts
+++ b/renderers/lit/src/v0_9/catalogs/basic/components/CheckBox.ts
@@ -69,7 +69,8 @@ export class A2uiCheckBoxElement extends BasicCatalogA2uiLitElement<typeof Check
     input.invalid {
       outline: 1px solid var(--a2ui-checkbox-color-error, red);
     }
-    .error {
+    .error,
+    .a2ui-error-message {
       color: var(--a2ui-checkbox-color-error, red);
       font-size: var(--a2ui-font-size-xs, 0.75rem);
       margin-top: 4px;
@@ -97,10 +98,12 @@ export class A2uiCheckBoxElement extends BasicCatalogA2uiLitElement<typeof Check
             .checked=${props.value || false}
             @change=${(e: Event) => props.setValue?.((e.target as HTMLInputElement).checked)}
           />
-          ${props.label}
+          <span class="a2ui-check-box-text">${props.label}</span>
         </label>
         ${isInvalid && props.validationErrors?.length
-          ? html`<div class="error">${props.validationErrors[0]}</div>`
+          ? props.validationErrors.map(
+              (msg: string) => html`<div class="error a2ui-error-message">${msg}</div>`,
+            )
           : nothing}
       </div>
     `;
--- a/renderers/lit/src/v0_9/catalogs/basic/components/ChoicePicker.ts
+++ b/renderers/lit/src/v0_9/catalogs/basic/components/ChoicePicker.ts
@@ -37,19 +37,26 @@ export class A2uiChoicePickerElement extends BasicCatalogA2uiLitElement<typeof C
    */
   static override styles = css`
     :host,
-    a2ui-choicepicker {
+    .a2ui-choice-picker {
       display: flex;
       flex-direction: column;
+      width: 100%;
       gap: var(--a2ui-choicepicker-gap, var(--a2ui-spacing-xs, 0.25rem));
       padding: var(--a2ui-choicepicker-padding, 0);
     }
-    .options {
+    .options,
+    .a2ui-options-group {
       display: flex;
       flex-direction: column;
       gap: var(--a2ui-choicepicker-gap, var(--a2ui-spacing-xs, 0.25rem));
     }
-    label {
-      color: var(--a2ui-choicepicker-label-color, inherit);
+    label,
+    .a2ui-option-label {
+      display: flex;
+      align-items: center;
+      gap: var(--a2ui-choicepicker-gap, var(--a2ui-spacing-xs, 0.25rem));
+      cursor: pointer;
+      color: var(--a2ui-text-color-text, var(--a2ui-choicepicker-label-color, inherit));
       font-size: var(--a2ui-choicepicker-label-font-size, inherit);
     }
     :host,
@@ -63,41 +70,62 @@ export class A2uiChoicePickerElement extends BasicCatalogA2uiLitElement<typeof C
     .filter-input {
       background-color: var(--a2ui-color-input, #fff);
       color: var(--a2ui-color-on-input, #333);
-      border: var(--a2ui-textfield-border, var(--a2ui-border));
-      border-radius: var(--a2ui-textfield-border-radius, var(--a2ui-spacing-m));
+      border: var(
+        --a2ui-textfield-border,
+        var(--a2ui-border, 1px solid var(--a2ui-color-border, #ccc))
+      );
+      border-radius: var(--a2ui-textfield-border-radius, var(--a2ui-spacing-m, 4px));
       padding: var(
         --a2ui-choicepicker-filter-padding,
         var(--a2ui-spacing-xs, 4px) var(--a2ui-spacing-s, 8px)
       );
+      margin-bottom: var(--a2ui-choicepicker-filter-margin-bottom, var(--a2ui-spacing-s, 0.25rem));
       font-family: inherit;
     }
     .filter-input:focus {
       outline: none;
       border-color: var(--a2ui-textfield-color-border-focus, var(--a2ui-color-primary, #17e));
     }
-    .chips {
+    .a2ui-option-input {
+      width: var(--a2ui-choicepicker-checkbox-size, 1rem);
+      height: var(--a2ui-choicepicker-checkbox-size, 1rem);
+    }
+    .chips,
+    .a2ui-chips-group {
       display: flex;
       flex-direction: row;
       flex-wrap: wrap;
       gap: var(--a2ui-choicepicker-gap, var(--a2ui-spacing-xs, 0.25rem));
     }
-    .chip {
+    .chip,
+    .a2ui-chip {
       padding: var(
         --a2ui-choicepicker-chip-padding,
         var(--a2ui-spacing-s, 4px) var(--a2ui-spacing-m, 8px)
       );
       border-radius: var(--a2ui-choicepicker-chip-border-radius, 999px);
-      border: 1px solid var(--a2ui-color-border, #ccc);
-      background-color: var(--a2ui-color-surface, #fff);
+      border: var(--a2ui-choicepicker-chip-border, 1px solid var(--a2ui-color-border, #ccc));
+      background-color: var(--a2ui-choicepicker-chip-background, var(--a2ui-color-surface, #fff));
       color: var(--a2ui-color-on-surface, inherit);
       cursor: pointer;
-      font-size: var(--a2ui-font-size-xs, 0.75rem);
+      font-size: var(--a2ui-choicepicker-chip-font-size, var(--a2ui-font-size-xs, 0.75rem));
       font-family: inherit;
+      font-weight: var(--a2ui-choicepicker-chip-font-weight, normal);
+      transition: all 0.2s;
     }
-    .chip.selected {
-      background-color: var(--a2ui-color-primary, #007bff);
+    .chip.selected,
+    .chip.active,
+    .a2ui-chip.selected,
+    .a2ui-chip.active {
+      background-color: var(
+        --a2ui-choicepicker-chip-background-selected,
+        var(--a2ui-color-primary, #007bff)
+      );
       color: var(--a2ui-color-on-primary, #fff);
-      border-color: var(--a2ui-color-primary, #007bff);
+      border-color: var(
+        --a2ui-choicepicker-chip-background-selected,
+        var(--a2ui-color-primary, #007bff)
+      );
     }
   `;
 
@@ -108,23 +136,29 @@ export class A2uiChoicePickerElement extends BasicCatalogA2uiLitElement<typeof C
   }
 
   override render() {
-    const props = this.controller.props;
+    const props = this.controller?.props;
     if (!props) return nothing;
 
-    const selected = Array.isArray(props.value) ? props.value : [];
+    const rawVal = props.value;
+    const selected: string[] = Array.isArray(rawVal)
+      ? (rawVal as string[])
+      : typeof rawVal === 'string'
+        ? [rawVal]
+        : [];
     const isMulti = props.variant === 'multipleSelection';
     const isChips = props.displayStyle === 'chips';
 
     const toggle = (val: string) => {
-      if (!props.setValue) return;
+      const setter = props.setValue;
+      if (typeof setter !== 'function') return;
       if (isMulti) {
         if (selected.includes(val)) {
-          props.setValue(selected.filter((v: string) => v !== val));
+          setter(selected.filter((v: string) => v !== val));
         } else {
-          props.setValue([...selected, val]);
+          setter([...selected, val]);
         }
       } else {
-        props.setValue([val]);
+        setter([val]);
       }
     };
 
@@ -134,6 +168,16 @@ export class A2uiChoicePickerElement extends BasicCatalogA2uiLitElement<typeof C
         this.filter === '' ||
         String(opt.label).toLowerCase().includes(this.filter.toLowerCase()),
     );
+
+    // Radio group names are document-scoped while component ids are only
+    // surface-scoped, so the group name qualifies the component id with the
+    // surface id and the data context path (which disambiguates instances
+    // stamped from a list template within one surface).
+    const surfaceId = this.context?.dataContext?.surface?.id || 'default';
+    const componentId = this.context?.componentModel?.id || 'choice-picker';
+    const dataPath = this.context?.dataContext?.path || '/';
+    const sanitizedPath = dataPath.replace(/[^a-zA-Z0-9-]/g, '_');
+    const groupName = `a2ui-choice-${surfaceId}-${componentId}-${sanitizedPath}`;
 
     return html`
       ${props.label ? html`<label>${props.label}</label>` : nothing}
@@ -149,13 +193,24 @@ export class A2uiChoicePickerElement extends BasicCatalogA2uiLitElement<typeof C
             />
           `
         : nothing}
-      <div class=${classMap({options: true, chips: isChips})}>
+      <div
+        class=${classMap({
+          'a2ui-choice-picker': true,
+          options: true,
+          'a2ui-options-group': !isChips,
+          chips: isChips,
+          'a2ui-chips-group': isChips,
+        })}
+      >
         ${options.map((opt: any) =>
           isChips
             ? html`
                 <button
+                  type="button"
                   class=${classMap({
+                    'a2ui-chip': true,
                     chip: true,
+                    active: selected.includes(opt.value),
                     selected: selected.includes(opt.value),
                   })}
                   aria-pressed=${selected.includes(opt.value)}
@@ -165,13 +220,16 @@ export class A2uiChoicePickerElement extends BasicCatalogA2uiLitElement<typeof C
                 </button>
               `
             : html`
-                <label>
+                <label class="a2ui-option-label">
                   <input
                     type=${isMulti ? 'checkbox' : 'radio'}
+                    name=${isMulti ? nothing : groupName}
+                    value=${opt.value}
                     .checked=${selected.includes(opt.value)}
                     @change=${() => toggle(opt.value)}
+                    class="a2ui-option-input"
                   />
-                  ${opt.label}
+                  <span class="a2ui-option-text">${opt.label}</span>
                 </label>
               `,
         )}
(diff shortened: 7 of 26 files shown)


renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
```diff
@@ -24,41 +24,94 @@ import {LitComponentApi} from '../../../types.js';
...
-      <div
-        @click=${() => this.dialog?.showModal()}
-        class="a2ui-modal-trigger"
-        style="display: contents;"
-      >
-        ${props.trigger ? html`${this.renderNode(props.trigger)}` : nothing}
+      <div class="a2ui-modal-wrapper">
+        <div @click=${() => this.openModal()} class="a2ui-modal-trigger">
+          ${props.trigger ? this.renderNode(props.trigger) : nothing}
+        </div>
```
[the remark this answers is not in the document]
josemontespg: Implement this suggestion only if it aligns with the Angular implementation

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
```diff
@@ -24,41 +24,94 @@ import {LitComponentApi} from '../../../types.js';
...
+    .a2ui-modal-trigger {
+      cursor: pointer;
     }
-    dialog::backdrop {
+    .a2ui-modal-overlay {
+      position: fixed;
+      top: 0;
+      left: 0;
+      width: 100vw;
+      height: 100vh;
```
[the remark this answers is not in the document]
josemontespg: Implement this suggestion only if it aligns with the Angular implementation

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
```diff
@@ -24,41 +24,94 @@ import {LitComponentApi} from '../../../types.js';
...
+          ? html`
+              <div
+                class="a2ui-modal-overlay a2ui-modal"
+                @click=${(e: MouseEvent) => {
+                  if (e.target === e.currentTarget) {
+                    this.closeModal();
+                  }
+                }}
+              >
+                <div class="a2ui-modal-content" @click=${(e: Event) => e.stopPropagation()}>
```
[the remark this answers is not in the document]
josemontespg: Implement this suggestion only if it aligns with the Angular implementation

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
```diff
@@ -24,41 +24,94 @@ import {LitComponentApi} from '../../../types.js';
...
+                  <button type="button" class="a2ui-modal-close" @click=${() => this.closeModal()}>
+                    &times;
+                  </button>
```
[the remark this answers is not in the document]
josemontespg: Implement this suggestion only if it aligns with the Angular implementation

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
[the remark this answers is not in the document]
josemontespg: Checked Angular's `modal.component.ts` reference implementation: Angular does not currently set `role="dialog"` or `aria-modal="true"` on `.a2ui-modal-content`. To maintain strict 1:1 structural and behavioral alignment across both renderers before consolidating them into universal components in PR 3, keeping this aligned with Angular for now. We can add comprehensive ARIA/accessibility enhancements across all universal components in a dedicated follow-up PR.

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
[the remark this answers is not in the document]
josemontespg: Checked Angular's `modal.component.ts` reference implementation: Angular renders `<button class="a2ui-modal-close" (click)="closeModal()">&times;</button>` without an explicit `aria-label="Close"` or `type="button"`. To maintain exact cross-framework structural parity ahead of extracting universal web components in PR 3, keeping this consistent with Angular. We will plan a centralized accessibility pass on the universal components in `web_core`.

renderers/lit/a2ui_explorer/src/local-gallery.ts
```diff
@@ -239,15 +239,6 @@ export class LocalGallery extends LitElement {
 
   /**
    * Applies the user-selected primary color to `createSurface` messages.
-   *
-   * This is necessary for the explorer application to allow users to live-preview
-   * theme changes by injecting the selected color into the message stream.
-   * In a standard A2UI renderer deployment, this is not needed as the renderer
-   * simply processes messages as received from the agent, which is responsible
-   * for providing the correct theme.
```
josemontespg, line 247: Don't remove this jsdoc

josemontespg: Restored the JSDoc comment on `applyPrimaryColorToMessages`.

gspencergoog (approved):
Looks good to my Flutter developer eyes. :-)

renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts
```diff
@@ -23,14 +23,21 @@ import {A2uiController} from '../../../a2ui-controller.js';
...
       flex-direction: column;
       gap: var(--a2ui-spacing-xs, 0.25rem);
       background: var(--a2ui-audioplayer-background, transparent);
       border-radius: var(--a2ui-audioplayer-border-radius, 0);
       padding: var(--a2ui-audioplayer-padding, 0);
+      width: 100%;
+    }
+    .a2ui-audio-description {
+      font-size: var(--a2ui-font-size-s, 0.875rem);
+      color: var(--a2ui-text-caption-color, light-dark(#666, #aaa));
```
ditman: This makes it impossible for users to override the audio-description color from the text-caption-color, can the color be: `var(--a2ui-audioplayer-description-color, var(--a2ui-text-caption-color, light-dark(#666, #aaa))`, so the component can be overridden individually?

renderers/lit/src/v0_9/catalogs/basic/components/Card.ts
```diff
@@ -22,30 +22,17 @@ import {A2uiController} from '../../../a2ui-controller.js';
...
+      box-shadow: var(--a2ui-card-box-shadow, 0 2px 4px rgba(0, 0, 0, 0.1));
+      background: var(--a2ui-card-background, var(--a2ui-color-surface, #fff));
       border: var(
         --a2ui-card-border,
         var(--a2ui-border-width, 1px) solid var(--a2ui-color-border, #ccc)
       );
-      border-radius: var(--a2ui-card-border-radius, var(--a2ui-border-radius, 8px));
-      padding: var(--a2ui-card-padding, var(--a2ui-spacing-m, 16px));
-      background: var(--a2ui-card-background, var(--a2ui-color-surface, #fff));
-      color: var(--a2ui-color-on-surface, #333);
```
ditman, line 46: Why is color removed?

renderers/lit/a2ui_explorer/tests/v0_9/29_movie-card.test.ts
```diff
@@ -67,8 +67,8 @@ describe('Example: Movie Card', () => {
     button.click();
     await whenSettled(gallery);
 
-    const modal = querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLDialogElement;
-    expect(modal?.open).toBeTruthy();
+    const modal = querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLElement;
+    expect(modal).toBeTruthy();
```
ditman, line 71: Why did this expectation change, don't we have ".open" anymore?

renderers/lit/a2ui_explorer/tests/v0_9/36_modal.test.ts
```diff
@@ -39,18 +39,14 @@ describe('Example: Modal', () => {
...
 
     // Click trigger
     trigger.click();
     await whenSettled(gallery);
 
     // Check modal is open
-    expect(
-      (querySelectorAllDeep(surface, '.a2ui-modal-overlay')[0] as HTMLDialogElement)?.open,
-    ).toBeTruthy();
+    expect(querySelectorAllDeep(surface, '.a2ui-modal-overlay').length).toBeGreaterThan(0);
```
ditman:
```suggestion
    expect(querySelectorAllDeep(surface, '.a2ui-modal-overlay').length).toBe(1);
```

We expect a single element in the modal-overlay, no?

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
```diff
@@ -39,45 +39,57 @@ const ALIGN_MAP: Record<string, string> = {
...
 };
 
 @customElement('a2ui-basic-column')
 export class A2uiBasicColumnElement extends BasicCatalogA2uiLitElement<typeof ColumnApi> {
-  /**
-   * The styles of the column can be customized by redefining the following
-   * CSS variables:
-   *
-   * - `--a2ui-column-gap`: The gap between items in the column. Defaults to `--a2ui-spacing-m`.
-   */
```
ditman, line 51: Restore the JSDoc for the static styles property pls

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
```diff
@@ -39,45 +39,57 @@ const ALIGN_MAP: Record<string, string> = {
...
   static override styles = css`
     :host,
     a2ui-basic-column {
       display: flex;
       flex-direction: column;
-      gap: var(--a2ui-column-gap, var(--a2ui-spacing-m));
+      gap: var(--a2ui-column-gap, var(--a2ui-spacing-m, 16px));
```
ditman: I wouldn't add this 16px fallback; `--a2ui-spacing-m` should be defined by the default stylesheet, or a user override, and not defaulted here IMO

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
```diff
@@ -39,45 +39,57 @@ const ALIGN_MAP: Record<string, string> = {
...
   override updated(changedProperties: PropertyValues) {
     super.updated(changedProperties);
-    const props = this.controller.props;
+    const props = this.controller?.props;
     if (props) {
-      this.style.justifyContent = JUSTIFY_MAP[props.justify ?? ''] ?? 'flex-start';
-      this.style.alignItems = ALIGN_MAP[props.align ?? ''] ?? 'stretch';
+      this.style.display = 'flex';
+      this.style.flexDirection = 'column';
+      this.style.gap = 'var(--a2ui-column-gap, var(--a2ui-spacing-m, 16px))';
```
ditman: Why are we adding static styles here, they should go on the `static override styles` declaration. Also, we seem to be re-defining the gap again here?

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
```diff
@@ -39,45 +39,57 @@ const ALIGN_MAP: Record<string, string> = {
...
   override render() {
-    const props = this.controller.props;
+    const props = this.controller?.props;
     if (!props) return nothing;
 
     const children: ResolvedChildList = Array.isArray(props.children) ? props.children : [];
+    const getKey = (child: any) =>
+      typeof child === 'object' && child !== null
+        ? `${child.basePath ?? ''}/${child.id}`
+        : String(child);
```
ditman: Why is the `getKey` funtion defined inline? This could be a private member of the A2uiBasicColumnElement class, since it seems that it uses nothing from its closure, just the `child:any` incoming parameter. No need to recreate this function on every call to render.

renderers/lit/src/v0_9/catalogs/basic/components/DateTimeInput.ts
```diff
@@ -65,9 +65,12 @@ export class A2uiDateTimeInputElement extends BasicCatalogA2uiLitElement<typeof
     a2ui-datetimeinput {
       display: flex;
       flex-direction: column;
+      width: 100%;
```
ditman: I'm not a fan of all these `width: 100%` styles that we're adding (especially not being overridable by the user)

renderers/lit/src/v0_9/catalogs/basic/components/List.ts
```diff
@@ -14,50 +14,88 @@
...
   override render() {
-    const props = this.controller.props;
+    const props = this.controller?.props;
     if (!props) return nothing;
 
     const children: ResolvedChildList = Array.isArray(props.children) ? props.children : [];
-    return html`${map(children, child => html`${this.renderNode(child)}`)}`;
+    const listStyle = props.listStyle;
+    const direction = props.direction || 'vertical';
+    const getKey = (child: any) =>
```
ditman: Same comment about the `getKey` method as above.

renderers/lit/src/v0_9/catalogs/basic/components/List.ts
```diff
@@ -14,50 +14,88 @@
...
+      return html`
+        <ol class="a2ui-list ${direction}">
+          ${repeat(children, getKey, child => html`<li>${this.renderNode(child)}</li>`)}
+        </ol>
+      `;
+    }
+
+    if (listStyle === 'unordered') {
+      return html`
+        <ul class="a2ui-list ${direction}">
```
ditman, line 82: Using a2ui-list for both an ol/ul and a "list of arbitrary things" is a little bit weird. I think this should use slightly different classnames so you don't end up needing things like `list-style-position: inside`.

I'm not sure an ordered/unordered list should support horizontal mode, for example?

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
```diff
@@ -26,50 +26,107 @@ export class A2uiLitModal extends BasicCatalogA2uiLitElement<typeof ModalApi> {
    * The styles of the modal can be customized by redefining the following
    * CSS variables:
    *
-   * - `--a2ui-modal-backdrop-bg`: Controls the backdrop color of the dialog.
-   * - `--a2ui-modal-padding`: Padding inside the dialog content area. Defaults to `24px`.
-   * - `--a2ui-modal-border-radius`: Border radius of the dialog. Defaults to `8px`.
+   * - `--a2ui-modal-background`: Controls the background of the modal content.
```
ditman, line 29: modal-background is a very good toggle, thanks for adding this!

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
```diff
@@ -26,50 +26,107 @@ export class A2uiLitModal extends BasicCatalogA2uiLitElement<typeof ModalApi> {
...
+              </div>
+            `
+          : nothing}
       </div>
-      <dialog class="a2ui-modal a2ui-modal-overlay">
-        <form method="dialog" style="text-align: right;">
-          <button class="a2ui-modal-close">×</button>
-        </form>
-        ${props.content ? html`${this.renderNode(props.content)}` : nothing}
-      </dialog>
```
ditman, line 72: Shame, the implementation with dialog was fairly smart :P

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
```diff
@@ -32,52 +32,65 @@ const JUSTIFY_MAP: Record<string, string> = {
...
+  protected readonly api = RowApi;
+
   protected createController() {
     return new A2uiController(this, RowApi);
   }
 
   override updated(changedProperties: PropertyValues) {
     super.updated(changedProperties);
-    const props = this.controller.props;
+    const props = this.controller?.props;
```
ditman: In general, why is `controller` optional now?

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
```diff
@@ -32,52 +32,65 @@ const JUSTIFY_MAP: Record<string, string> = {
...
     if (props) {
-      this.style.justifyContent = JUSTIFY_MAP[props.justify ?? ''] ?? 'flex-start';
-      this.style.alignItems = ALIGN_MAP[props.align ?? ''] ?? 'stretch';
+      this.style.display = 'flex';
+      this.style.flexDirection = 'row';
```
ditman: Same comments as in the Column. Why is this setting static styles on this.style, instead of having them in the static `styles` property?

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
```diff
@@ -32,52 +32,65 @@ const JUSTIFY_MAP: Record<string, string> = {
...
+  baseline: 'baseline',
 };
 
 @customElement('a2ui-basic-row')
 export class A2uiBasicRowElement extends BasicCatalogA2uiLitElement<typeof RowApi> {
-  /**
-   * The styles of the row can be customized by redefining the following
-   * CSS variables:
-   *
-   * - `--a2ui-row-gap`: The gap between items in the row. Defaults to `--a2ui-spacing-m`.
```
ditman, line 50: Do not remove the JSDoc for the styles property

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
```diff
@@ -32,52 +32,65 @@ const JUSTIFY_MAP: Record<string, string> = {
...
-    const props = this.controller.props;
+    const props = this.controller?.props;
     if (props) {
-      this.style.justifyContent = JUSTIFY_MAP[props.justify ?? ''] ?? 'flex-start';
-      this.style.alignItems = ALIGN_MAP[props.align ?? ''] ?? 'stretch';
+      this.style.display = 'flex';
+      this.style.flexDirection = 'row';
+      this.style.gap = 'var(--a2ui-row-gap, var(--a2ui-spacing-m, 16px))';
+      if (props.justify) {
+        this.style.justifyContent = JUSTIFY_MAP[props.justify] || props.justify;
```
ditman: This allows setting invalid justifyContent values, via a non-empty `props.justify` value, which we probably don't want? (Same comment for Column)

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
```diff
@@ -32,52 +32,65 @@ const JUSTIFY_MAP: Record<string, string> = {
...
   override render() {
-    const props = this.controller.props;
+    const props = this.controller?.props;
     if (!props) return nothing;
 
     const children: ResolvedChildList = Array.isArray(props.children) ? props.children : [];
+    const getKey = (child: any) =>
+      typeof child === 'object' && child !== null
+        ? `${child.basePath ?? ''}/${child.id}`
+        : String(child);
```
ditman: `getKey` doesn't need to be an inline closure?

renderers/lit/src/v0_9/catalogs/basic/components/Tabs.ts
```diff
@@ -21,87 +21,87 @@ import {TabsApi} from '@a2ui/web_core/v0_9/basic_catalog';
...
   }
 
   @state() accessor activeIndex = 0;
 
+  override willUpdate(changedProperties: any) {
+    super.willUpdate(changedProperties);
+    const props = this.controller?.props;
+    if (props?.tabs && this.activeIndex >= props.tabs.length) {
+      this.activeIndex = Math.max(0, props.tabs.length - 1);
+    }
```
ditman, line 81: Is this to defend against the case that the number of tabs changes from the server? In this case, if activeIndex >= tabs.lengh, I think we could fall-back to 0. Resetting the interface to zero seems more natural to me than going to the last tab (which isn't going to be the one that the user wants either?)

renderers/lit/src/v0_9/catalogs/basic/components/Tabs.ts
```diff
@@ -21,87 +21,87 @@ import {TabsApi} from '@a2ui/web_core/v0_9/basic_catalog';
...
-   * - `--a2ui-tabs-header-background`: Default transparent.
-   * - `--a2ui-tabs-header-background-active`: Default `--a2ui-color-secondary`.
-   * - `--a2ui-tabs-header-color`: Default `--a2ui-color-on-surface`.
-   * - `--a2ui-tabs-header-color-active`: Default `--a2ui-color-on-secondary`.
-   * - `--a2ui-tabs-border`: Default `--a2ui-border-width` solid `--a2ui-color-border`.
-   * - `--a2ui-tabs-content-padding`: Default `0 var(--a2ui-spacing-m, 0.5rem)`.
-   */
+@customElement('a2ui-basic-tabs')
+export class A2uiBasicTabsElement extends BasicCatalogA2uiLitElement<typeof TabsApi> {
   static override styles = css`
```
ditman, line 37: It'd be nice if the styles static prop documents what `--a2ui-tabs-xxx` properties can be overridden by theme authors.

renderers/lit/src/v0_9/catalogs/basic/components/TextField.ts
```diff
@@ -79,11 +82,21 @@ export class A2uiBasicTextFieldElement extends BasicCatalogA2uiLitElement<typeof
...
     const isInvalid = props.isValid === false;
-    const onInput = (e: Event) => props.setValue?.((e.target as HTMLInputElement).value);
+    const onInput = (e: Event) => {
+      const val = (e.target as HTMLInputElement).value;
+      if (typeof props.setValue === 'function') {
+        props.setValue(val);
+      } else {
+        const rawVal = this.context?.componentModel?.properties['value'];
+        if (rawVal && typeof rawVal === 'object' && 'path' in rawVal) {
+          this.context.dataContext.set(rawVal.path, val);
```
ditman: We used to not set the value in the context before. Was this component broken earlier, or is this required because of the changes to the generic binder?

renderers/lit/src/v0_9/catalogs/basic/components/Video.ts
```diff
@@ -22,35 +22,36 @@ import {A2uiController} from '../../../a2ui-controller.js';
 
 @customElement('a2ui-video')
 export class A2uiVideoElement extends BasicCatalogA2uiLitElement<typeof VideoApi> {
-  /**
-   * The styles of the video can be customized by redefining the following
-   * CSS variables:
-   *
-   * - `--a2ui-video-border-radius`: Controls the rounded corners of the video. Defaults to `0`.
```
ditman, line 29: Please do not remove the overridable variable names :)

renderers/lit/src/v0_9/catalogs/basic/components/Video.ts
```diff
@@ -22,35 +22,36 @@ import {A2uiController} from '../../../a2ui-controller.js';
...
   }
 
   override render() {
-    const props = this.controller.props;
+    const props = this.controller?.props;
     if (!props) return nothing;
 
-    return html`<video src=${props.url} controls class="a2ui-video"></video>`;
+    return html`
+      <div class="a2ui-video-container">
```
ditman, line 55: I don't understand why we want all these "container" elements. Being lit, all these `render` methods already end up inside of the shadow-dom of the custom `a2ui-video` component. This feels super clunky!

renderers/lit/src/v0_9/tests/components/Text.test.ts
```diff
@@ -136,10 +136,10 @@ describe('Text Component', () => {
       e.context = context;
     });
 
-    const captionSpan = el.querySelector('span.a2ui-caption');
+    const captionSpan = el.querySelector('span.a2ui-text.caption, span.a2ui-caption');
```
ditman:
```suggestion
    const captionSpan = el.querySelector('span.a2ui-text.caption');
```

Let's be explicit in the selectors and the expectations; We don't need to test for .a2ui-caption anymore, do we?

renderers/lit/src/v0_9/tests/components/TextField.test.ts
```diff
@@ -163,9 +163,10 @@ describe('TextField Component', () => {
       e.context = context;
     });
 
-    const error = el.querySelector('.error');
-    assert.ok(error);
-    assert.strictEqual(error.textContent?.trim(), 'Email is invalid');
+    const errors = el.querySelectorAll('.error');
+    assert.strictEqual(errors.length, 2);
+    assert.strictEqual(errors[0].textContent?.trim(), 'Email is required');
+    assert.strictEqual(errors[1].textContent?.trim(), 'Email must contain @');
```
ditman, line 169: Good change!

ditman (commented):
A bunch of comments here, mainly:

1. Why do we need to add a container div as the root of the shadow DOM of every element? We already have a closed ShadowDOM, we don't need the container?
2. Why remove the JSDoc from the static "styles"? That should be useful for theme authors when customizing the elements without peeking into the code.

And then some more miscellaneous comments about the changes that are being introduced in some of the bits here!

renderers/lit/src/v0_9/tests/components/TextField.test.ts
josemontespg: Thanks!

renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts
josemontespg: Updated to `var(--a2ui-audioplayer-description-color, var(--a2ui-text-caption-color, light-dark(#666, #aaa)))`.

renderers/lit/src/v0_9/catalogs/basic/components/Card.ts
josemontespg: Restored `color: var(--a2ui-color-on-surface, #333);` and the styles JSDoc.

renderers/lit/a2ui_explorer/tests/v0_9/29_movie-card.test.ts
josemontespg: The component replaced native `<dialog>` with an overlay container that conditionally renders when open (matching Angular's modal), so the element no longer has a native `.open` property.

renderers/lit/a2ui_explorer/tests/v0_9/36_modal.test.ts
josemontespg: Applied suggestion.

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
josemontespg: Restored the JSDoc comment.

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
josemontespg: Removed the `16px` fallback so it uses `var(--a2ui-column-gap, var(--a2ui-spacing-m))`.

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
josemontespg: Removed `display`, `flexDirection`, and `gap` from `updated()`. Those are already declared in `static override styles`.

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
josemontespg: Extracted to a helper function

renderers/lit/src/v0_9/catalogs/basic/components/DateTimeInput.ts
josemontespg: The `width: 100%` default comes from the Angular basic catalog implementation where Google3 workflows rely on full-width inputs. Updated to `width: var(--a2ui-datetimeinput-width, 100%);` so consumers can override the width via CSS variables while preserving the default.

renderers/lit/src/v0_9/catalogs/basic/components/List.ts
josemontespg: Extracted to a helper function.

renderers/lit/src/v0_9/catalogs/basic/components/List.ts
josemontespg: The markup and classes directly mirror the Angular reference component (`renderers/angular/src/v0_9/catalog/basic/list.component.ts`), where the wire schema supports both `direction` ('vertical' | 'horizontal') and `listStyle` ('ordered' | 'unordered' | 'none'). I kept it identical here to avoid divergence ahead of extracting the universal components, but we can refine the class naming across all renderers.

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
josemontespg: Glad that helps!

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
josemontespg: `<dialog>` was convenient, but switching to the overlay container resolved backdrop click and layering issues across browsers while matching the Angular implementation.

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
josemontespg: Reverted the optional chaining back to `this.controller.props` across this component and all other basic catalog components.

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
josemontespg: Removed `display`, `flexDirection`, and `gap` from `updated()` here as well.

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
josemontespg: Restored the JSDoc comment.

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
josemontespg: Updated to `JUSTIFY_MAP[props.justify ?? ''] ?? 'flex-start'` (and `ALIGN_MAP` similarly) so only mapped values are applied.

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
josemontespg: Extracted to a static helper method on the class.

renderers/lit/src/v0_9/catalogs/basic/components/Tabs.ts
josemontespg: Good point. Changed the fallback to reset `activeIndex = 0`.

renderers/lit/src/v0_9/catalogs/basic/components/Tabs.ts
josemontespg: Restored and updated the JSDoc documenting all supported `--a2ui-tabs-*` custom properties.

renderers/lit/src/v0_9/catalogs/basic/components/TextField.ts
josemontespg: In standalone mocks without the generic binder attaching a synthetic `setValue` setter, `setValue` was undefined, so this was added as a defensive fallback. With the generic binder consistently generating setters in PR #2359, this fallback is no longer needed and has been simplified back to `props.setValue?.(val)`.

renderers/lit/src/v0_9/catalogs/basic/components/Video.ts
josemontespg: Restored the JSDoc comment for `--a2ui-video-border-radius`.

josemontespg:
Thanks @ditman for the review! I applied your suggestions and replied on the threads with details. Ready for another pass when you can 🙏

renderers/lit/src/v0_9/catalogs/basic/components/Video.ts
josemontespg: While an inner `div` is not strictly necessary to render the video, matching the Angular `video.component.ts` DOM structure minimizes DOM differences so we can land changes in Google3 more easily. We can revisit and simplify the DOM structure once we have a single source of truth for the basic catalog. Also, `BasicCatalogA2uiLitElement` renders in Light DOM (`createRenderRoot() { return this; }`), so the container div provides the styling boundary.

renderers/lit/src/v0_9/tests/components/Text.test.ts
josemontespg: Applied suggestion.

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
```diff
@@ -58,10 +60,18 @@ export class A2uiBasicRowElement extends BasicCatalogA2uiLitElement<typeof RowAp
     }
   `;
 
+  protected readonly api = RowApi;
+
   protected createController() {
     return new A2uiController(this, RowApi);
   }
 
+  private static getChildKey(child: any): string {
```
josemontespg: Instead of static, create a function outside the class

josemontespg: Updated. Moved `getChildKey` out of the class into a module-level function across `Row.ts`, `Column.ts`, and `List.ts`.

renderers/lit/a2ui_explorer/tests/v0_9/29_movie-card.test.ts
ditman: Ah yes, I noticed this a few files down the PR but forgot to remove this comment! thanks for the clarification!

renderers/lit/src/v0_9/catalogs/basic/components/DateTimeInput.ts
ditman: Thanks for adding the extra variable!

renderers/lit/src/v0_9/catalogs/basic/components/List.ts
ditman: Haha ok, then the angular implementation is a little bit weird. Good thing we can later fix it in a single place! Let's go!

renderers/lit/src/v0_9/catalogs/basic/components/Modal.ts
ditman: Yeah, it's just a shame that the native element was sssooo close to be enough but not really :'(

renderers/lit/src/v0_9/catalogs/basic/components/TextField.ts
ditman: Excellent, thanks for revisiting!

renderers/lit/src/v0_9/catalogs/basic/components/Video.ts
ditman: > `BasicCatalogA2uiLitElement` renders in Light DOM.

Ah yes, using the Light DOM is a breaking change at least when we adopt this on the Lit renderer, unless we can keep it "closed" over there :)

renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts
```diff
@@ -23,14 +23,24 @@ import {A2uiController} from '../../../a2ui-controller.js';
 @customElement('a2ui-audioplayer')
 export class A2uiAudioPlayerElement extends BasicCatalogA2uiLitElement<typeof AudioPlayerApi> {
   static override styles = css`
```
ditman, line 25: This is missing in the OG component, so we shouldn't block, but can you create a github issue to ensure that all the `styles` that can be overridden via CSS variables have some JSDoc? For example this one is missing things like `--a2ui-audioplayer-padding` or `--a2ui-auidoplayer-description-color`.

renderers/lit/src/v0_9/catalogs/basic/components/CheckBox.ts
```diff
@@ -97,10 +98,12 @@ export class A2uiCheckBoxElement extends BasicCatalogA2uiLitElement<typeof Check
             .checked=${props.value || false}
             @change=${(e: Event) => props.setValue?.((e.target as HTMLInputElement).checked)}
           />
-          ${props.label}
+          <span class="a2ui-check-box-text">${props.label}</span>
         </label>
         ${isInvalid && props.validationErrors?.length
-          ? html`<div class="error">${props.validationErrors[0]}</div>`
+          ? props.validationErrors.map(
+              (msg: string) => html`<div class="error a2ui-error-message">${msg}</div>`,
```
ditman, line 105: This is going to look a little bit when multiple errors are rendered and probably needs extra love, but let's go with it! It's better than before!

renderers/lit/src/v0_9/catalogs/basic/components/ChoicePicker.ts
```diff
@@ -37,19 +37,27 @@ export class A2uiChoicePickerElement extends BasicCatalogA2uiLitElement<typeof C
    */
   static override styles = css`
     :host,
-    a2ui-choicepicker {
+    a2ui-choicepicker,
```
ditman: I don't think you need the `a2ui-choicepicker` selector now that you added `.a2ui-choice-picker`, right?

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
```diff
@@ -39,25 +39,36 @@ const ALIGN_MAP: Record<string, string> = {
...
   end: 'flex-end',
   stretch: 'stretch',
+  baseline: 'baseline',
 };
 
+function getChildKey(child: any): string {
+  return typeof child === 'object' && child !== null
+    ? `${child.basePath ?? ''}/${child.id}`
+    : String(child);
+}
```
ditman, line 49: This definitely needs a better home since it's used in multiple parts; maybe we could have a `key` getter on the correct type `A2uiLitElement` or similar? (maybe on the Node layer?) This is can probably be added later though.

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
```diff
@@ -66,8 +77,16 @@ export class A2uiBasicColumnElement extends BasicCatalogA2uiLitElement<typeof Co
     super.updated(changedProperties);
     const props = this.controller.props;
     if (props) {
-      this.style.justifyContent = JUSTIFY_MAP[props.justify ?? ''] ?? 'flex-start';
-      this.style.alignItems = ALIGN_MAP[props.align ?? ''] ?? 'stretch';
+      if (props.justify) {
+        this.style.justifyContent = JUSTIFY_MAP[props.justify] ?? 'flex-start';
+      } else {
+        this.style.removeProperty('justify-content');
```
ditman, line 83: Why do we now need to "removeProperty" here? Do we have cases where the justify/align change on an element after it's been rendered?

renderers/lit/src/v0_9/catalogs/basic/components/Button.ts
```diff
@@ -101,16 +107,22 @@ export class A2uiBasicButtonElement extends BasicCatalogA2uiLitElement<typeof Bu
...
     const classes = {
       'a2ui-button': true,
-      ['a2ui-button-' + (props.variant || 'default')]: true,
+      [variant]: true,
+      ['a2ui-button-' + variant]: true,
     };
 
     return html`
       <button
+        type=${variant === 'primary' ? 'submit' : 'button'}
```
andrewkolos: nit: why set type to `submit`? This doesn't seem like documented behavior[^1] and I'm afraid this could cause weird behavior if a surface ever gets embedded in `<form>`. I have no evidence that this edge case would ever realistically happen, but I also can't think of any situation where setting `type` to `submit` is useful or expected

[^1]: (it's explicitly tested for, but--in current times--I'm afraid that doesn't inherently mean anything)

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
```diff
@@ -32,15 +32,23 @@ const JUSTIFY_MAP: Record<string, string> = {
   spaceAround: 'space-around',
   spaceEvenly: 'space-evenly',
   stretch: 'stretch',
+  baseline: 'baseline',
```
ditman: "baseline" is not a valid justify type (see: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/justify-content#values). (The Column component is *not* adding this, correctly)

renderers/lit/src/v0_9/catalogs/basic/components/Image.ts
```diff
@@ -75,20 +69,14 @@ export class A2uiImageElement extends BasicCatalogA2uiLitElement<typeof ImageApi
...
-    const classes = {
-      'a2ui-image': true,
-      [props.variant || '']: !!props.variant,
-    };
-
-    const styles = {
-      objectFit: props.fit || 'fill',
-    };
+    const variant = props.variant || 'default';
+    const fit = props.fit || 'cover';
```
andrewkolos, line 73: This defaults `fit` to `cover`, matching the [Angular implementation](https://github.com/a2ui-project/a2ui/blob/51b5da16a911eb1787a858b831a59af11c52ddaa/renderers/angular/src/v0_9/catalog/basic/image.component.ts#L79). The catalog schema declares [`.default('fill')`](https://github.com/a2ui-project/a2ui/blob/51b5da16a911eb1787a858b831a59af11c52ddaa/renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts#L69), and the spec declares [`"default": "fill"`](https://github.com/a2ui-project/a2ui/blob/51b5da16a911eb1787a858b831a59af11c52ddaa/specification/v0_9_1/catalogs/basic/catalog.json#L66).

This has been a piece of technical debt for a while (#1298) and it's finally come back to bite us. Since we have to break someone here, I'd rather us default to `fill`. That is, unless we decide `cover` is more intuitive (`fill` does have cases where it works quite poorly, see #1297). But that's probably a bigger, separate discussion to be had if anything.

ditman (commented):
Very minor comments now. Last thing; have you run the restaurant_finder demo with all of these changes to ensure it behaves *similar* to what it did before? Also the lit_explorer app?

renderers/lit/src/v0_9/catalogs/basic/components/Button.ts
josemontespg: Good catch, agreed. This was inherited directly from Angular's `button.component.ts` while aligning the two renderers in this PR.

Setting `type="submit"` provides no benefit since A2UI buttons dispatch client actions, and it risks unwanted form submissions if embedded in a `<form>`. I've updated it to always use `type="button"`.

renderers/lit/src/v0_9/catalogs/basic/components/Image.ts
josemontespg: I agree that this should be fixed, but I'd prefer to defer this to another PR. I'm aligning with the angular implementation, because it will produce less churn when rolling and enabling this in google3. After we do that, it will be easier to make these well scoped improvements to the basic catalog.

renderers/lit/src/v0_9/catalogs/basic/components/AudioPlayer.ts
josemontespg: Created issue #2461 to track adding complete JSDoc documentation for all overridable CSS variables across the basic catalog components.

renderers/lit/src/v0_9/catalogs/basic/components/CheckBox.ts
josemontespg: Thanks! Agreed, definitely something that can be refined whenever we give the error layout more love down the line.

renderers/lit/src/v0_9/catalogs/basic/components/ChoicePicker.ts
josemontespg: Good catch! Removed `a2ui-choicepicker` from the selector list.

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
josemontespg: Agreed. With the upcoming node-based architecture, this child key mapping will be much simpler (something like `child => child.id`). We'll consolidate this when refactoring the universal web component base class.

renderers/lit/src/v0_9/catalogs/basic/components/Column.ts
josemontespg: Because A2UI properties can technically be dynamic bindings from streaming messages or reactive data updates (even if dynamically clearing alignment is an uncommon edge case), an element may receive a `justify` or `align` value and later have it removed or cleared. Without `removeProperty`, the previously applied inline style would remain permanently on the element instead of reverting to the stylesheet default.

renderers/lit/src/v0_9/catalogs/basic/components/Row.ts
josemontespg: Good catch! Removed `baseline` from `JUSTIFY_MAP` to match CSS flexbox rules and align with `Column.ts`.

josemontespg:
Thanks for the review! I replied to the comments in threads.

> have you run the restaurant_finder demo with all of these changes to ensure it behaves _similar_ to what it did before? Also the lit_explorer app?

Yes, I ran both demos side by side comparing this PR with main. All is working as intended, with some minor rendering differences listed in the PR description.

renderers/lit/src/v0_9/catalogs/basic/components/DateTimeInput.ts
```diff
@@ -94,21 +112,59 @@ export class A2uiDateTimeInputElement extends BasicCatalogA2uiLitElement<typeof
...
+      } else {
+        props.setValue?.(date);
+      }
+    };
+
+    const handleTimeChange = (event: Event) => {
+      const time = (event.target as HTMLInputElement).value;
+      const date = rawValue.includes('T')
+        ? rawValue.split('T')[0]
+        : rawValue || new Date().toISOString().split('T')[0];
```
andrewkolos: if the current value is a time only (e.g. `09:30`) and we are in time-only mode (suppose the user is changing the value to `14:00`), then the saved value will be incorrect (`09:30T14:00:00` instead of `14:00`).

andrewkolos: nit: when in datetime mode and a time is chosen before the date is, we default to whatever the current date is in UTC. but unless the user is in one of the few countries that observe GMT year-round, we default to picking yesterday or tomorrow instead of today depending on what side of GMT/UTC they are on. we would have to do some string parsing to make it work, but perhaps we should build the default directly from `new Date()` (using `getFullYear`, `getMonth`, `getDate`).

josemontespg: Good catch! I have updated `handleTimeChange` to check `enableDate` and only emit the time string when in time-only mode. I also added `getLocalIsoDate()` to compute the local calendar date (`getFullYear`, `getMonth + 1`, `getDate`) instead of defaulting to UTC. In addition, I have added unit test coverage for time-only mode, datetime preservation, and local date fallbacks in the component tests PR up the stack.

renderers/lit/src/v0_9/catalogs/basic/components/ChoicePicker.ts
```diff
@@ -165,13 +212,16 @@ export class A2uiChoicePickerElement extends BasicCatalogA2uiLitElement<typeof C
                 </button>
               `
             : html`
-                <label>
+                <label class="a2ui-option-label">
                   <input
                     type=${isMulti ? 'checkbox' : 'radio'}
+                    name=${componentId}
+                    value=${opt.value}
```
andrewkolos, line 227: nothing guarantees `componentId` is globally unique, right? like two instances of a template could have inputs with the same `componentId` and thus `input.name`. then selecting one radio input in one of the two groups would incorrectly clear the other group

probably a pre-existing bug though

josemontespg: Yes, #2447
The fix here I think would be to have component ids that are unique, probably best handled in the node layer.

renderers/lit/src/v0_9/catalogs/basic/components/Button.ts
```diff
@@ -60,35 +60,39 @@ export class A2uiBasicButtonElement extends BasicCatalogA2uiLitElement<typeof Bu
...
+        --a2ui-button-padding,
+        var(--a2ui-spacing-m, 0.5rem) var(--a2ui-spacing-l, 1rem)
+      );
+      border-radius: var(--a2ui-button-border-radius, var(--a2ui-spacing-s, 0.25rem));
+      border: var(
+        --a2ui-button-border,
+        var(--a2ui-border-width, 1px) solid var(--a2ui-color-border, #ccc)
+      );
+      cursor: pointer;
+      margin: var(--a2ui-button-margin, var(--a2ui-spacing-m, 0.5rem));
```
andrewkolos, line 56: at first i got confused by `a2ui-basic-button` and `.a2ui-button` both being things until I noticed that former is not a CSS class

the upside (hardly) of my confusion is that I notice they both define `margin`. are Lit buttons going to have double the margin now? should we delete the `margin` specification from `a2ui-basic-button` to preserve parity?

ditman: This is a good catch; everything that targets the shadow-dom container of the element (like `:host` or `a2ui-basic-button`) may be removable now that we have the `.a2ui-button` class. (this has been done in other components btw, but I haven't exhaustively checked all instances)

renderers/lit/src/v0_9/catalogs/basic/components/ChoicePicker.ts
andrewkolos: Thanks for letting me know this is already filed; seems like there is a fix too (though I'm not sure it's complete--I'll see if I can look later). Regardless, thread resolved :)

renderers/lit/src/v0_9/catalogs/basic/components/Button.ts
```diff
@@ -60,35 +60,39 @@ export class A2uiBasicButtonElement extends BasicCatalogA2uiLitElement<typeof Bu
...
-      background-color: var(--_color-primary);
       color: var(--_a2ui-text-color);
+      border: none;
     }
-    .a2ui-button:hover {
-      background-color: var(--a2ui-color-secondary-hover, #ddd);
-    }
-    .a2ui-button.a2ui-button-primary:hover {
-      background-color: var(--a2ui-color-primary-hover, #fbd);
-    }
```
andrewkolos, line 87: nit: looks like these are lost in translation, unless we think this is going to break some image diffing test i dont know about

renderers/lit/src/v0_9/catalogs/basic/components/Icon.ts
```diff
@@ -43,30 +43,33 @@ export class A2uiIconElement extends BasicCatalogA2uiLitElement<typeof IconApi>
...
-      font-size: var(--_icon-size);
-      font-weight: normal;
+    .a2ui-icon {
+      display: inline-block;
+      width: var(--a2ui-icon-size, 24px);
+      height: var(--a2ui-icon-size, 24px);
+      font-size: var(--a2ui-icon-size, 24px);
       font-style: normal;
+      font-weight: normal;
+      font-family: var(--a2ui-icon-font-family, 'Material Icons', 'Material Symbols Outlined');
```
andrewkolos, line 53: FYI: this flips the default icon font from Material Symbols to Material icons. I'm assuming this is a product of preferring to inherit angular behavior for now (I'm assuming defaulting to Symbols is preferred long term); just making sure this is intentional

nit: if we do keep this change, [the doc comment on L42](https://github.com/a2ui-project/a2ui/blob/0aa51f47668e98ee976f623f432e6a538df12125/renderers/lit/src/v0_9/catalogs/basic/components/Icon.ts#L42) is stale

andrewkolos (approved):
LGTM with nits

renderers/lit/src/v0_9/catalogs/basic/components/Button.ts
josemontespg: Good catch from both of you! Following @ditman's suggestion, we've removed both `:host, a2ui-basic-button` and the unused `:where(:host, a2ui-basic-button)` container rules entirely now that styles are encapsulated on `.a2ui-button`.

This eliminates the duplicate margin and brings `Button.ts` into alignment with components like `Card`, `Slider`, and `Video` which don't target container elements.

renderers/lit/src/v0_9/catalogs/basic/components/Button.ts
josemontespg: Restored `.a2ui-button:hover` and `.a2ui-button.primary:hover` rules. Thanks for catching this!

renderers/lit/src/v0_9/catalogs/basic/components/Icon.ts
josemontespg: Yes, inheriting the Angular default (`'Material Icons', 'Material Symbols Outlined'`) for now to ensure consistency across renderers ahead of extracting the universal components in this stack. Updated the JSDoc on line 42 to reflect the new default.

ditman: Note that we should prefer material symbols over icons, the font is more modern and has cooler features like variants (and splitting); please create an issue to migrate from icons to symbols!

ditman (approved):
I have nothing more to say about this, let's go with it and see what comes out on the other side! :P

PR #2210, opened 2026-08-07
feat(spec): add bidirectional function calls (callRendererFunction & callAgentFunction) to A2UI v1.0

gspencergoog:
## Summary

Refines function call messaging in the unreleased A2UI v1.0 specification by replacing ambiguous `CallFunctionMessage` drafts with explicit, symmetric directional messages (`callRendererFunction` / `rendererFunctionResponse` and `callAgentFunction` / `agentFunctionResponse`), standardizing response payloads, and removing legacy action fields.

## Key Changes & Design Refinements

- **Directional Function Call & Response Specifications**:
  - `agent_to_renderer.json`: Introduced `callRendererFunction` (Agent ➔ Renderer call) and `agentFunctionResponse` (Agent ➔ Renderer response). Removed `surfaceId` from envelope and mandated `catalogId` inside `callFunction` using `allOf` schema composition. Removed legacy `ActionResponseMessage`.
  - `renderer_to_agent.json`: Introduced `callAgentFunction` (Renderer ➔ Agent call) and `rendererFunctionResponse` (Renderer ➔ Agent response). Removed legacy `wantResponse` and `actionId` fields from `action` events.
  - `common_types.json`: Added unified `FunctionResponse` definition (`functionCallId`, `value`, `error`) shared by both directional response messages. Removed legacy `wantResponse` and `responsePath` fields from `Action` events.
- **Protocol Documentation (`a2ui_protocol.md`)**:
  - Documented explicit response requirements for `callRendererFunction` (renderers MUST always respond with `rendererFunctionResponse` or `error`, even for `void` calls).
  - Added **Transport Interaction Patterns** section with sequence diagrams for HTTP POST request-response and WebSocket/gRPC streaming flows.
  - Added **Functions in A2UI Content Execution** section detailing polymorphic function usage in UI content, async pending states, and failure recovery policies.
  - Added **Function Location Resolution & Fallback Routing** section under `callAgentFunction` documenting implicit fallback routing for unregistered client functions.
  - Cleaned up obsolete references to `wantResponse`, `responsePath`, `actionId`, and `actionResponse`.
- **Evaluator & Test Suite**:
  - Added message validators for `callRendererFunction`, `callAgentFunction`, `agentFunctionResponse`, and `rendererFunctionResponse` in `validator.ts`.
  - Updated `call_function_message.json`, `call_agent_function_message.json`, and `function_response.json` test suites.
  - All 153 JSON schema validation tests pass across component, message, and protocol suites.

## Backward Compatibility & Scope

- **No Breaking Changes to Released Versions**: `CallFunctionMessage` was an unreleased v1.0 draft concept and did not exist in v0.9 / v0.9.1. This change refines unfinalized v1.0 draft schemas. Action events retain clean fire-and-forget semantics consistent with v0.9.

## Testing

1. Run the specification test suite using Python:
   ```bash
   cd specification/v1_0/test
   python3 run_tests.py
   ```
2. Verified all 153 JSON schema validation tests pass.


specification/v1_0/eval/src/validator.ts
```diff
@@ -97,6 +97,10 @@ export class Validator {
             validated = this.ajv.validate(`${schemaUri}#/$defs/UpdateDataModelMessage`, message);
           } else if (message.deleteSurface) {
             validated = this.ajv.validate(`${schemaUri}#/$defs/DeleteSurfaceMessage`, message);
+          } else if (message.callRendererFunction) {
+            validated = this.ajv.validate(`${schemaUri}#/$defs/CallRendererFunctionMessage`, message);
+          } else if (message.functionResponse) {
+            validated = this.ajv.validate(`${schemaUri}#/$defs/FunctionResponseMessage`, message);
```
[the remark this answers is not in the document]
gspencergoog: Updated Validator.run to validate actionResponse against ActionResponseMessage, and updated validateCustom to treat callRendererFunction, functionResponse, and actionResponse as valid v1.0 message types.

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -340,24 +340,26 @@ Agent responds with:
...
+  - `surfaceId` (string, optional): The surface ID where the call originated.
+  - `functionCallId` (string, required): A unique identifier for this invocation instance. The renderer MUST copy this ID verbatim into the subsequent `functionResponse` or `error` message.
+  - `callFunction` (object, required): The description of the function call.
+    - `call` (string, required): The registered name of the function to execute.
+    - `catalogId` (string, optional): Overrides the default catalog ID for this function call.
+    - `args` (object, optional): Arguments passed to the function, as defined by its schema in the catalog.
 
 **Security Boundaries and Verification:**
 
 Execution boundary verification (`agentOnly` vs `rendererOnly`) is enforced strictly at runtime by the renderer application:
```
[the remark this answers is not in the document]
gspencergoog: Updated protocol documentation to clarify runtime determination of function execution boundaries (callableFrom) from the active catalog definition.

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -340,24 +340,26 @@ Agent responds with:
...
 
 **Properties:**
 
-- `functionCallId` (string, required): A unique identifier for this invocation instance. The renderer MUST copy this ID verbatim into the subsequent `functionResponse` or `error` message.
-- `wantResponse` (boolean, optional, default `false`): Specifies whether the agent expects a response payload back from the renderer. If set to `true`, the renderer MUST reply with either a `functionResponse` or an `error` message.
-- `callFunction` (object, required): The description of the function call.
-  - `call` (string, required): The registered name of the function to execute.
-  - `args` (object, optional): Arguments passed to the function, as defined by its schema in the catalog.
+- `callRendererFunction` (object, required):
+  - `surfaceId` (string, optional): The surface ID where the call originated.
```
jacobsimionato: But if the agent is doing the call, how does it have an originating surface, and what should the renderer do with this information?

Is it just to make use of the default catalog ID for that surface? that could make sense, though it kind of seems overly complex to me vs just explicitly providing the catalog ID.

specification/v1_0/json/common_types.json
jacobsimionato: Can we remove this as part of this change? And maybe consult @wrenj ?

jacobsimionato: Update this to mention renderer *or* agent-side function

specification/v1_0/json/common_types.json
```diff
@@ -217,7 +217,7 @@
     },
     "FunctionCall": {
       "type": "object",
-      "description": "Invokes a named function on the renderer.",
+      "description": "Invokes a named function.",
```
jacobsimionato, line 231: How does the framework know whether a particular function call is intended to be client or agent side?

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -1251,25 +1255,73 @@ This message is sent when a user interacts with a component that has an agent ac
...
   "version": "v1.0",
   "functionResponse": {
-    "functionCallId": "ping-call-id-102",
-    "call": "pingAgent",
-    "value": true
+    "functionCallId": "verify_provider_99",
+    "error": {
+      "code": "PROVIDER_NOT_FOUND",
+      "message": "Provider ID PRV-102 was not found."
+    }
```
jacobsimionato, line 1389: I think it might be good to talk a little more about uses of functions within A2UI content in this doc, in addition to the agent-renderer protocol.

E.g. are agent functions and client functions polymorphic in the sense you can use either of them to provide dynamic values or button handlers, and nest them flexibly?

What if a function fails within a nested execution pipeline?

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -340,24 +340,26 @@ Agent responds with:
 }
 ```
 
-### `callFunction`
+### `callRendererFunction`
```
jacobsimionato, line 297: How does a server call a function on a client, seeing as servers typically can initiate requests to clients?

I think the answer is probably that the server would typically decide to call a client function in response to a client requests, and the client would be expected to make a further server request to deliver the result. E.g. client sends "Summarize my choices", server responds with `rendererFunctionCall(name: fetchScreenContext, id: abc)` client sends new request with `functionResponse('blah', id: abc)`, server sends `updateComponents` response.

jacobsimionato (commented):
Yay, it's happening! I believe in the review meeting there was feedback that it might be too early to add this. Personally, I like this change and I'm happy for it to happen now, seeing as this will allow us to remove `action.wantResponse`, and also open up a heap of new use cases.

Local functions was a very powerful feature for v0.9, and I think this opens to door to a lot of interesting patterns for v1.0.

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -1251,25 +1255,73 @@ This message is sent when a user interacts with a component that has an agent ac
 }
 ```
 
-### `functionResponse`
+### `callAgentFunction`
```
jacobsimionato, line 1317: Do agent functions need to be declared in catalogs?

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -340,24 +340,26 @@ Agent responds with:
...
 **Properties:**
 
-- `functionCallId` (string, required): A unique identifier for this invocation instance. The renderer MUST copy this ID verbatim into the subsequent `functionResponse` or `error` message.
-- `wantResponse` (boolean, optional, default `false`): Specifies whether the agent expects a response payload back from the renderer. If set to `true`, the renderer MUST reply with either a `functionResponse` or an `error` message.
-- `callFunction` (object, required): The description of the function call.
-  - `call` (string, required): The registered name of the function to execute.
-  - `args` (object, optional): Arguments passed to the function, as defined by its schema in the catalog.
+- `callRendererFunction` (object, required):
+  - `surfaceId` (string, optional): The surface ID where the call originated.
+  - `functionCallId` (string, required): A unique identifier for this invocation instance. The renderer MUST copy this ID verbatim into the subsequent `functionResponse` or `error` message.
```
jacobsimionato, line 304: If the renderer function returns void, does the renderer need to send a functionResponse back to the agent after it is called? Note that there is an additional server RPC required to do this, unlike renderer->agent functions where the server would be expected to return something (e.g. a HTTP 200) anyway.

If the function is declared with a return value, should we also have a `wantResponse` parameter (default true perhaps) which allows the agent to ignore the response and prevent the unnecessary RPC?

specification/v1_0/json/common_types.json
gspencergoog: Yes, we should be able to.  I'll add that change here.

specification/v1_0/json/common_types.json
gspencergoog: If the function isn't available on the renderer, then it is assumed to be an agent side function. If it turns out not to be, then the agent will send back an error, and then the renderer will send its own error back to the generating agent. Agents can also do their own vetting of the function calls on the way out to avoid the whole round trip.

This does allow a renderer to override an agent side function.  It think this is OK, but let me know if you think otherwise.

We could also have some kind of prefix or suffix (e.g. "agent_") for the function names too, but I thought just keep it simple for now.

specification/v1_0/docs/a2ui_protocol.md
gspencergoog: Okay, I'll add more detail.  The answer is that you can mix agent and renderer functions together: they all act the same.  If a function fails, it will generate an error, just like it does now, and the generating agent will see the error.

There are some odd cases when functions have side effects: they don't get rolled back if they already ran and succeeded, before another call fails.  It's best if functions don't have side effects, but they are useful sometimes (e.g. `openUrl` opens a web page. It won't get un-opened if a later function call in the same evaluation fails).

specification/v1_0/docs/a2ui_protocol.md
gspencergoog: The agent can't initiate the connection to the renderer, but if one already exists, it can send the request.  I see what you're saying though: if the renderer is using discrete connections to a server to make requests, there's no way for the agent to contact the renderer.  But typically these are HTTP SSE connections, and the server should be able to send an unrequested `rendererFunctionCall` to the renderer as part of the SSE stream as long as the stream has already been established.  A websocket connection would also allow two way connections.  gRPC might be harder, since the server isn't allowed to call the client there.

A solution to this would be some kind of heartbeat where the server can respond with the function call when it gets a heartbeat from the client.  Seems inefficient to be polling like that, though.

This feature was added at the request of a customer, but I feel like it is somewhat disconnected from A2UI: it seems like something that might be solved more robustly outside of the protocol.

specification/v1_0/docs/a2ui_protocol.md
gspencergoog: No.  If the function isn't available on the target surface, it is assumed to be an agent-side function.

specification/v1_0/docs/a2ui_protocol.md
gspencergoog: I think we should just always send a response.  The server likely wants to know it succeeded, even if there's no output, and I don't feel like the extra message is a lot of overhead. Adding an extra parameter to reason about likely increases the latency of the Agent's response a bit anyhow.  If it becomes an issue, we can add the optional parameter later without breaking anyone.

specification/v1_0/json/common_types.json
gspencergoog: Done.

specification/v1_0/docs/a2ui_protocol.md
gspencergoog: I think for some calls, it makes sense to have the surface ID as part of the call, to give it context to run the function in.  But I guess you could just make that a parameter of the function itself.

I think the idea of this was that it would allow using the default catalog on the surface, but let's just specify the catalog ID every time, and if they need the surface ID, they can supply that as an argument to the function.

specification/v1_0/json/common_types.json
gspencergoog: You can now get a response from the agent just by calling an agent function.  Is that sufficient @wrenj ?

specification/v1_0/docs/a2ui_protocol.md
jacobsimionato: Yeah... I can see how sometimes it might be helpful to pass a surfaceId. But having it as a built-in parameter seems weird to me seeing as the API to register local functions is completely surface agnostic.

specification/v1_0/json/common_types.json
jacobsimionato: My opinion is that we should push towards having quite explicit and robust identification of functions, because I think this is only going to get more complex. 

In the future I think we want A2UI components to be able to call:
- local functions baked into catalog itself (current functionality)
- component-specific functions e.g. functions exposed by MCP apps
- dynamic local functions exposed by local modules
- server functions baked into server binary
- functions provided by MCP servers that the agent dynamically discovers

So because many of these are dynamic, it'd be helpful to reduce the ambiguity around function resolution. Along the lines of your prefix approach, I do wonder about some path-based resolution like `agent/mcp_servers/weather_mcp_server/get_temp` or `renderer/rain_radar_mcp_app/set_location`. Initially, we could have `renderer/catalog/<catalog id>/<function name>` and `agent/<function name>`. I'd be tempted to add this soon, because MCP apps already can export functions, and it'd be great to have A2UI apps controlling them ASAP - @vpm238 is chatting to customers about this.

Though it also is perhaps too much of a leap. WDYT?

specification/v1_0/json/renderer_to_agent.json
```diff
@@ -58,23 +58,27 @@
       },
       "required": ["name", "surfaceId", "sourceComponentId", "timestamp", "context"]
     },
-    "functionResponse": {
+    "callAgentFunction": {
```
jacobsimionato, line 61: Random thought: If the agent calls nested server functions, we would send all the calls to the server in one "callAgentFunction" message, then execute the full pipeline on the server in one round trip.

I think the protocol should actually allow that already, because I believe the arguments in the FunctionCall can be dynamic values which are themselves function calls. We need some tricky logic on the client that sort of slices execution pipelines into server and client sections properly, but that seems surmountable.

Do you think this should already support this use case? Perhaps we should call it out in the protocol docs or something, that a callAgentFunction request can actually specify a pipeline of multiple functions? Perhaps it should even be puralized `callAgentFunctions` and `callClientFunctions` even? I dunno.

gspencergoog: Good point! Yes, I think the implementations should definitely recognize when they can do that. I'll update the `callAgentFunction` so it can take a chain.

specification/v1_0/json/common_types.json
gspencergoog: Yikes, that's a scary list. :-)

I think we should come up with a good naming scheme then.  Maybe something based on URIs, where the kind of call is the "protocol" section, like `renderer://catalog/functionName`, `mcp://weather/getTemp`, `agent://catalog/functionName`, `functionName` (no protocol defaults to renderer local).

specification/v1_0/json/renderer_to_agent.json
gspencergoog: To keep it simple to implement a renderer, it should be up to the implementation if they *want* to implement that: you could totally have a naive implementation that just sent each one separately.  Of course, if you use web_core or another of our core libraries, it should do that for you.

jacobsimionato: Agreed that it's fine to have naive implementations! I just want to make sure this is possible at the protocol level. I actually think that perhaps no changes are essential, but just wanted to flag it in case you think of some change we should make.

One change that might be nice: Make "error" include the function name, so that if the `callAgentFunction` request actually involves multiple functions, the client can later understand which specific one failed.

specification/v1_0/json/common_types.json
jacobsimionato, line 239: Perhaps call out that catalogId is only relevant to local function calls?

So for local functions:
- Use the catalogId here, or else the default catalog Id in create surface. If neither are present it's an error.

For server functions:
- The catalog Id specified on createSurface is irrelevant, and this call-level catalogId should always be null.

specification/v1_0/json/common_types.json
jacobsimionato: We were agreeing in chat, let's punt on this namespacing for now so we have more time to think about it.

How about we add an explicit `handlerLocation` (`local` vs `remote`) though? In the express compiler, we can always automatically populate this to save tokens. But at least this makes the handling logic less magical (e.g. automatically falling back to server resolution for unresolved local functions).

gspencergoog: I think I'd rather add a simplified naming scheme (e.g. "agent:functionName" for agent side functions, no other definitions for now) than add a bool that we'll have to remove later.

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -1265,32 +1319,84 @@ This message is sent when a user interacts with a component that has an agent ac
...
+### `callAgentFunction`
+
+This message is sent by the renderer to execute a function remotely on the agent (e.g. verifying a provider ID or checking inventory availability).
 
-This message is sent by the renderer to return the successful execution result of an agent-initiated function call. It is required only if the agent sent the `callFunction` request with `wantResponse: true`.
+**Function Location Resolution & Fallback Routing:**
+
+When the renderer evaluates a `FunctionCall` (from an action handler, validation check, or dynamic value), it determines the execution target using implicit fallback routing:
+
+1. **Local Lookup:** The renderer checks if a local renderer-side function with that name is registered in its local catalog/registry. If found, the renderer executes the function locally.
```
jacobsimionato, line 1335: Maybe mention that it only does this lookup if there is a catalogId specified in the function call, or a default catalog Id defined on the surface. If there is no catalog ID, then we assume it *must* be a server function?

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -1319,6 +1425,36 @@ This message is sent by the renderer to report runtime or execution errors to th
...
+1. **Async Evaluation:** The renderer dispatches the function call (e.g. emitting `callAgentFunction`) and enters an asynchronous evaluation state.
+2. **Pending UI State:** The renderer maintains a pending/loading state for the affected component binding (e.g., displaying a loading indicator or preserving existing component values) while awaiting `functionResponse`.
+3. **Value Resolution:** Upon receiving `functionResponse`, the renderer updates the local dynamic value or validation state with the returned `value`.
+
+### 3. Failure Propagation & Recovery Rules
+
+If a function call within a content pipeline fails (returns a `functionResponse` containing an `error` payload, times out, or triggers a transport error), the renderer applies the following recovery rules:
+
+- **Dynamic Value Binding Failure:** The property binding resolves to `null` (or a declared fallback value), and the renderer logs an evaluation error without crashing the surrounding component tree.
+- **Validation Rule Failure (`Checkable`):** The check rule evaluates as invalid, displaying the rule's specified error message to the user.
```
jacobsimionato, line 1455: So just to confirm: This means that the rendering SDKs have special handling for "Checkable" types, where it propagates the error to the component implementation, while for regular dynamic properties, it coerces the error to null and logs the error? I'm fine with that!

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -1319,6 +1425,36 @@ This message is sent by the renderer to report runtime or execution errors to th
...
+- **Action Pipeline Failure:** Halts execution of any subsequent steps in the action execution pipeline and dispatches a local error boundary event or toast notification. Any side effects of previously executed steps are preserved.
```
jacobsimionato, line 1456: I don't quite understand how this third point fits with the first two. My understanding is that this section is about what happens if an error occurs in any of the above scenarios. But this scenario seems to be more generic. Perhaps we just need to say that if an error occurs anywhere in a pipeline, the rest of the pipeline does not execute. I don't think we need to mention toasts etc.

jacobsimionato (approved):
Thank you for iterating and bearing with all my changes of mind!

wrenj:
1. When I bind a component property to an agent function, how does the initial load work? does the UI immediately do a callback to the agent? Callbacks are pretty heavyweight and latency is already an issue. Can we add an initialValue or something so the first UI doesn't need to make a bunch of server callbacks? I guess the SDK could populate it automatically.

2. How does this work on components that don't support these agent callbacks? eg, GC uses A2UI to load a custom widget and that's it; they don't want partners to start making all these callbacks. Is there some mechanism to disable it? component_schema.allowCallbacks or something?

gspencergoog:
> 1. When I bind a component property to an agent function, how does the initial load work? does the UI immediately do a callback to the agent? Callbacks are pretty heavyweight and latency is already an issue. Can we add an initialValue or something so the first UI doesn't need to make a bunch of server callbacks? I guess the SDK could populate it automatically.

Jacob and I were thinking that the renderer implementation would batch calls like this, so that they would be one RPC.  We were thinking more of cases like "foo(bar(baz()))" and batching nested calls, but I don't see why we couldn't do the same for startup.

An optional initial or placeholder value seems like a good idea, but I'm not quite sure how to fit that in to the function calls. This seems more like something that the component would provide (e.g. while a value is not yet set, there is a placeholder value or UI).

> 2. How does this work on components that don't support these agent callbacks? eg, GC uses A2UI to load a custom widget and that's it; they don't want partners to start making all these callbacks. Is there some mechanism to disable it? component_schema.allowCallbacks or something?

Hmm. I hadn't considered this.  At what granularity does it need to be disabled?  Component, catalog, surface, or session?  We should address this, but maybe in another PR.

Could this be a setting on the renderer/prompt generator instead of part of the protocol?

wrenj:
> > 1. When I bind a component property to an agent function, how does the initial load work? does the UI immediately do a callback to the agent? Callbacks are pretty heavyweight and latency is already an issue. Can we add an initialValue or something so the first UI doesn't need to make a bunch of server callbacks? I guess the SDK could populate it automatically.
> 
> Jacob and I were thinking that the renderer implementation would batch calls like this, so that they would be one RPC. We were thinking more of cases like "foo(bar(baz()))" and batching nested calls, but I don't see why we couldn't do the same for startup.
> 
> An optional initial or placeholder value seems like a good idea, but I'm not quite sure how to fit that in to the function calls. This seems more like something that the component would provide (e.g. while a value is not yet set, there is a placeholder value or UI).
> 
> > 2. How does this work on components that don't support these agent callbacks? eg, GC uses A2UI to load a custom widget and that's it; they don't want partners to start making all these callbacks. Is there some mechanism to disable it? component_schema.allowCallbacks or something?
> 
> Hmm. I hadn't considered this. At what granularity does it need to be disabled? Component, catalog, surface, or session? We should address this, but maybe in another PR.
> 
> Could this be a setting on the renderer/prompt generator instead of part of the protocol?

1.

Batches reduces RPCs but some mechanism to get that one initial rpc to 0 would be good, the agent SDK could process the agent generated a2ui, call any agent functions and get responses, and send them in the same payload as the createSurface, eg:

a2a datapart: 
[
createSurface { component { text { callAgentFunction ( "getText" }}}},
agentFunctionCall { "getText" { }}} ,
agentFunctionResponse { "getText" { result { "hello world" }}}
]

and we make some rule that renderers will use matching functioncall and responses that arrive in the same payload as the agentFunction gets created.

We'd need some rule around caching function call results too (if 10 components callAgentFunction getText, does that get called once or 10 times? What if another component does that call later in the session, do we reuse the value?)

2. 

SG to punt it. For now the renderer can throw an error if someone uses this, and put in the system instructions of the catalog agentFunctionCalls aren't supported.

gspencergoog:
> Batches reduces RPCs but some mechanism to get that one initial rpc to 0 would be good, the agent SDK could process the agent generated a2ui, call any agent functions and get responses, and send them in the same payload as the createSurface, eg:
> 
> a2a datapart: [ createSurface { component { text { callAgentFunction ( "getText" }}}}, agentFunctionCall { "getText" { }}} , agentFunctionResponse { "getText" { result { "hello world" }}} ]
> 
> and we make some rule that renderers will use matching functioncall and responses that arrive in the same payload as the agentFunction gets created.
> 
> We'd need some rule around caching function call results too (if 10 components callAgentFunction getText, does that get called once or 10 times? What if another component does that call later in the session, do we reuse the value?)

This is feeling pretty complicated.  I'd like to think about the right solution for this for a bit.

PR #2220, opened 2026-08-10
feat(spec): add ValidationResult schema and dynamic CheckRule validation to v1.0

gspencergoog:
## Summary

This PR updates the **v1.0 specification** (`catalog_definition.json`, `common_types.json`, `catalogs/basic/catalog.json`, `a2ui_protocol.md`, and `evolution_guide.md`) to support **Richer, Portable Validation** using dynamic `ValidationResult` objects.

## Key Changes

1. **`catalog_definition.json` (`specification/v1_0/json/`)**:
   - Added `$defs/ValidationResult` schema defining `valid` (`boolean`, required), `code` (`string`), `message` (`string`), and `severity` (`"error" | "warning" | "info"`, default `"error"`). Permits additional properties for custom validation metadata.
   - Added `"validationResult"` as an allowed `returnType` enum option in `FunctionDefinition`.
2. **`common_types.json` (`specification/v1_0/json/`)**:
   - Updated `$defs/CheckRule` so `condition` evaluates to a `ValidationResult` object (via `DataBinding` or `FunctionCall`), and made `message` an optional fallback string for backward compatibility.
3. **`catalogs/basic/catalog.json` (`specification/v1_0/catalogs/basic/`)**:
   - Updated `returnType` on standard validation check functions (`required`, `regex`, `length`, `numeric`, `email`) from `"boolean"` to `"validationResult"`.
4. **`a2ui_protocol.md` (`specification/v1_0/docs/`)**:
   - Added a dedicated "Component Validation & Check Rules" section documenting `ValidationResult` in `catalog_definition.json` and explaining extensibility.
5. **`evolution_guide.md` (`specification/v1_0/docs/`)**:
   - Updated evolution guide with `ValidationResult`, `"validationResult"` return type, and standard catalog check function updates.

## Testing

- Ran specification test suite (`python3 test/run_tests.py` / `yarn test`).
- All 156 specification test cases pass.


specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -988,26 +978,72 @@ A2UI v1.0 generalizes renderer-side logic into **Functions**. These can be used
 
 The renderer supports a set of named **Functions** (e.g., `required`, `regex`, `email`, `add`, `concat`) which are defined in the JSON schema (e.g. `catalogs/basic/catalog.json`) alongside the component definitions. The agent references these functions by name in `FunctionCall` objects. This avoids sending executable code.
 
-Input components (like `TextField`, `CheckBox`) can define a list of checks. Each failure produces a specific error message that can be displayed when the component is rendered. Note that for validation checks, the function must return a boolean.
+### Component Validation & Check Rules
+
+Input components (like `TextField`, `ChoicePicker`) and interactive elements (like `Button`) can define a list of `checks` (`CheckRule` objects).
```
jacobsimionato, line 983: Something I always wondered: Is the name "checks" simply a convention, like "child" or "children", which SDKs don't assume? e.g. can I define multiple different checks parameters for a single component? Or is it just one "checks" per component? Either way, I think we should be explicit here, so that SDK implementations don't assume one way or the other.

I think multiple checks per component may scale better to templates etc, e.g. if you want to have a CreateAccountForm component which has separate checks for username and email etc.

If we do allow for multiple checks per component, then could it make sense to not actually define a concept of "checks" and instead just define the standard "validationResult" object that validation functions emit and components understand. Then each component is free to add a DynamicValue parameter, conventionally called `checks` or `isValid`, which accomplishes pretty much this task without special-cased handling? If necessary, there could be a CombineValidationResults function that accepts multiple validation results and concatenates the results. I think the ability to specify multiple validation functions with merged results is one of the reasons we added the checks feature originally.

specification/v1_0/json/common_types.json
```diff
@@ -261,19 +250,57 @@
       ],
       "unevaluatedProperties": false
     },
+    "ValidationResult": {
```
jacobsimionato, line 253: Where is this $def expected to be referenced from?

I can see it's important to the schema somewhere in the spec, though it seems like it's more for consumption by SDK implementers than by transport-level schemas or catalogs, so I don't think it needs to be in common-types. Ultimately, some local function implementation emits a `ValidationResult`, and then some local Component implementation consumes the object - the SDK just pipes around an opaque payload.

Perhaps the schema could be a separate JSON file, or just inlined into `a2ui_protocol.md`?

Maybe we want local functions to be able to declare that they can return ValidationResult objects, so we might need a new enum element for returnType at https://github.com/a2ui-project/a2ui/blob/2226b0f881f06606468e2c0afebe945ff14efe24/specification/v1_0/json/catalog_definition.json#L158-L162

Or maybe they can just declare returning "object"?

specification/v1_0/docs/a2ui_protocol.md
gspencergoog: The name "checks" is actually only in the basic catalog.  The real constraint is that a parameter is defined with a `$ref` to `CheckRule` (I'll update the description here).  So, yes, if you had some kind of form with multiple generic fields, you could already define separate change functions for them using `CheckRule` properties.  The component gets to decide how to interpret the results.

Right now, this is kind of trying to be backward compatible with v0.9 by allowing a boolean or data path as well as a function, along with the (now optional) message.  Probably a better design is that a `CheckRule` parameter returns a `ValidationResult` always, can't return a `boolean`, and doesn't need a `message` parameter, since that's all in the result.

I suppose we could just eliminate all of this: any component property can define itself based on the output of a function, which can have any return value, so you could define a component that has a `checks` field, and just use the result of that field to drive your validation, and design your own result values.

I think this has some value in standardizing the validation checks, and providing guidance for folks who ask "but does it do form validation?", and lack the vision to see that they can just do it with functions.  I could be convinced that we don't need it in the protocol at all, and just provide some documentation guidance as to how you could set it up yourself.  What do you think?

specification/v1_0/json/common_types.json
gspencergoog: Yeah, you're right, it doesn't need to be in `common_types.json`.  I was loathe to add it to yet another JSON schema file, but yeah, it's just a template for validation functions to use as a basis their return type to standardize them.

specification/v1_0/docs/a2ui_protocol.md
jacobsimionato: I basically agree with all these thoughts and also feel on the fence! I'm willing to go with whatever your gut feeling is, knowing that we can evolve this in future protocol versions anyway.

I think at a minimum, we should include a validation mechanism in the basic catalog, that people can use for inspiration when designing other catalogs.

I feel quite convinced that it makes sense to standardize `ValidationResult`, or even just `Error` as an object, so that there is a standard catalog-independent interface that both functions and components can use. I kind of like the idea of it just being "error", seeing as we already have that concept in callClientFunction and callServerFunction etc. Perhaps we allow components to define a "void" dynamic value (sort of like `Stream<void>` in Dart) which is just a function result. If the parameter value is empty, validation passes. If it's an error, the SDK provides a reference to the error object, and the implementation can display the message and/or code. That way, we don't need to define a new `ValidationResult` or `Checks` or `CheckRule` - we just use existing function primitives.

gspencergoog: I think if we have anything, we should have a structured `ValidationResult` example that people can look at for most use cases.  I think it probably covers at least 90% of validation cases. It's extensible, so if you're in the 10% you can either roll your own entirely (not follow the `CheckRule` def), or add extra fields to `ValidationResult` (which doesn't have `additionalProperties: false`, so that's allowed).

I think either we define a framework here as a suggestion, or we leave it wide open and don't define anything.  I lean towards what's in this PR already because if we define nothing, we run the risk of people not realizing that they can do validation already.

So, let's keep `CheckRule`, since we already use that for A2UI validation checks (the validation that we do when we validate the A2UI itself, not the runtime checks.  This feels like explaining Inception!), and give `ValidationResult` in a separate schema file as an example of a return type for a validation check.

The remaining question is do we retain support for configuring a `bool` and a `message` in `CheckRule`, because we don't need it if ValidationResult has them.  A literal bool makes little sense, so I say we disallow bools.  And if we define `CheckRule` so that it is expected to return a ValidationResult (maybe just by convention so it's not in the `common_types.json`?), then we don't need to have `message` either.

I'll massage this PR some more and see what comes out. :-)

gspencergoog: Here's a pro/con analysis if you care to read one:

<details>
<summary>Pro/Con analysis of each of the three options</summary>

# Brainstorming: Pros & Cons of Including `ValidationResult` in the A2UI Specification

This document synthesizes the architectural discussion on PR [#2220](https://github.com/a2ui-project/a2ui/pull/2220) regarding whether A2UI should specify a standardized `ValidationResult` schema, reuse a generic `Error` pattern, or leave validation entirely unconstrained to catalog-specific primitives.

---

## Overview & Core Tension

In A2UI v1.0, validation is evaluated dynamically via renderer-side functions or data model bindings. The central debate is:

> **Should A2UI specify a canonical `ValidationResult` object schema, or should validation remain a catalog-specific convention using generic function primitives?**

---

## Architectural Options Matrix

| Feature / Criteria | Option 1: Standardized `ValidationResult` (Current PR) | Option 2: Generic `Error` / `Result` Pattern | Option 3: Pure Custom Primitives (No Spec) |
| :--- | :--- | :--- | :--- |
| **Protocol Footprint** | Moderate (`catalog_definition.json`) | Low (Reuses `Error` RPC shape) | Zero (Omitted from spec) |
| **Catalog Interoperability** | High (Common schema across catalogs) | Medium (Common error shape) | Low (Each catalog invents its own) |
| **UI Richness** | High (`valid`, `code`, `message`, `severity`) | Moderate (`code`, `message`) | Variable |
| **Developer / LLM Ergonomics** | High (Clear guidance & patterns) | Medium | Low (Must learn per catalog) |
| **Extensibility** | High (`additionalProperties: true`) | Medium | High |

---

## Detailed Option Analysis

```mermaid
sequenceDiagram
    autonumber
    participant Agent as Agent (LLM / System)
    participant Data as Data Model
    participant Component as Input Component (TextField)
    participant Func as Validation Function (validateEmail)

    Component->>Func: Evaluate CheckRule (condition)
    Func->>Data: Inspect /user/email
    alt Option 1: Standardized ValidationResult
        Func-->>Component: ValidationResult { valid: false, code: "INVALID_EMAIL", message: "Enter valid email", severity: "error" }
        Component->>Component: Render red border & error message
    else Option 2: Generic Error Payload
        Func-->>Component: Error { code: "INVALID_EMAIL", message: "Enter valid email" }
        Component->>Component: Interpret non-null Error as failure
    else Option 3: Catalog-Specific Custom Return
        Func-->>Component: { customStatus: 400, errText: "Enter valid email" }
        Component->>Component: Renderer requires catalog-specific handler
    end
```

---

### Option 1: Standardized `ValidationResult` Schema (Current PR State)

In this approach, `catalog_definition.json` defines a canonical `$defs/ValidationResult` schema and adds `"validationResult"` as a recognized `returnType` for catalog functions.

#### Structure:
```json
{
  "valid": false,
  "code": "EXPIRED_CARD",
  "message": "The card expiration date has passed.",
  "severity": "error"
}
```

#### Pros:
1. **Out-of-the-Box Coverage for 90%+ of Use Cases**: Input validation across forms, text fields, pickers, and buttons immediately works out of the box with a consistent contract.
2. **Prevents Ecosystem Fragmentation**: Avoids a world where Catalog A uses `isValid` + `errorMsg`, Catalog B uses `status` + `reason`, and Catalog C uses `errors[]`.
3. **Supports Severity Levels**: Differentiates between strict errors (blocking form submission), warnings (soft alerts), and informational hints.
4. **Extensible by Design**: Because `additionalProperties` is permitted, specialized catalogs can attach additional fields (e.g. `fieldPointer`, `suggestions`, `autoFixValue`) without breaking validation.
5. **Clear Guidance for LLMs & Developers**: Provides a straightforward answer to "How do I perform form validation in A2UI?".

#### Cons:
1. **Spec Surface Area**: Introduces a domain-specific validation structure into the core catalog definition spec.
2. **SDK Assumption**: Renderer SDKs and catalog authors need to implement handling for `ValidationResult` specifically.

---

### Option 2: Universal `Error` / `Result` Payload Pattern

In this approach, validation functions return either `null`/empty (meaning valid) or a standard A2UI `Error` object (matching the error shape used in `callFunction` / `actionResponse` RPCs).

#### Structure:
```json
// Valid: null or {}
// Invalid:
{
  "code": "INVALID_EMAIL",
  "message": "Please enter a valid email address."
}
```

#### Pros:
1. **Minimal Protocol Footprint**: Reuses the existing RPC `Error` payload shape without adding a new `$defs` schema.
2. **Conceptual Simplicity**: "No error = valid; Error present = invalid".

#### Cons:
1. **Lacks Severity Distinction**: Cannot distinguish between a hard `error`, a soft `warning`, or an `info` callout.
2. **Ambiguity Between Validation vs. Execution Failures**: Harder to distinguish between a user input validation check failure and an unexpected function execution crash.

---

### Option 3: Pure Custom Primitives (Zero Protocol Spec for Validation)

In this approach, the A2UI specification defines **neither** `CheckRule` nor `ValidationResult`. Validation is treated purely as generic function evaluation bound to component properties (e.g. `TextField.errorText`, `Button.disabled`).

#### Pros:
1. **100% Unopinionated Protocol**: Keeps A2UI protocol strictly limited to generic UI trees and RPC primitives.
2. **Maximum Flexibility**: Catalog authors can design complex multi-field validation graphs without any spec constraints.

#### Cons:
1. **High Risk of Fragmentation**: Every catalog author invents incompatible ways of performing validation.
2. **Poor LLM / Agent Prompting**: Agents generating A2UI payloads must learn catalog-specific validation quirks for every single catalog.
3. **No Cross-Catalog UI Behaviors**: Renderers cannot implement generic validation rules (e.g., auto-disabling submit buttons or focusing the first invalid field) across different catalogs.

---

## Summary Recommendation & Next Steps

### Recommendation
**Retain Option 1 (Standardized `ValidationResult` in `catalog_definition.json`)** as presented in the current PR. 

It strikes the ideal balance:
- It provides a **canonical, catalog-independent baseline** for form validation.
- It lives in **`catalog_definition.json`**, keeping `common_types.json` and transport envelopes clean.
- It is **extensible**, allowing advanced use cases to add custom fields without breaking standard renderers.

</details>

specification/v1_0/json/common_types.json
gspencergoog: It seems like a good place for this is in `catalog_definition.json`.

specification/v1_0/docs/a2ui_protocol.md
gspencergoog: One pro I liked out of that was that it provides a baseline for validation so that people can use validation functions from other catalogs.

specification/v1_0/json/common_types.json
```diff
@@ -263,17 +263,18 @@
     },
     "CheckRule": {
       "type": "object",
-      "description": "A single validation rule applied to an input component.",
+      "description": "A single validation check rule applied to an input component. The condition function or path evaluates to a ValidationResult object.",
       "properties": {
         "condition": {
-          "$ref": "#/$defs/DynamicBoolean"
+          "oneOf": [{"$ref": "#/$defs/DataBinding"}, {"$ref": "#/$defs/FunctionCall"}],
```
jacobsimionato, line 269: If this must evaluate to a ValidationResult, could it really be a direct data binding? I figure it must be a function, which can have args which are data bindings?

specification/v1_0/docs/a2ui_protocol.md
jacobsimionato: Thanks for the detailed analysis. This all SGTM. Agree that providing a baseline that enables interop of validation functions from different catalogs is good, especially because every use of ValidationResult is basically going to be the same.

jacobsimionato (approved):

specification/v1_0/json/common_types.json
gspencergoog: I mean, you can put a `ValidationResult` into a data model path.  This could be set by some renderer code, so it would be a useful use case.  I do see what you're concerned about, but I would just leave it to the receiving component or agent to provide an error if the value it receives isn't what it expected, just like it would in the current (`main` branch) implementation if the `DynamicBoolean`'s data path didn't contain a boolean as expected.

PR #2228, opened 2026-08-12
feat(v1_0): Add optional userMessage field to Action event definition and renderer_to_agent action message

### LINKED ISSUE #2014: [FEATURE]: Allow actions to have a human-readable string associated with them for display in conversation history
We want a way for chat conversations to represent data that the user has entered via A2UI.

## Scenario 

1. Let's say I have a conversation and the agent asks me to pick a date range, showing "Start vacation date ":<picker> "end vacation date"  <picker>, then a submit button.

2. I pick dates

3. I click submit, triggering an action with name "submit_dates" and the JSON of each date.

## Typical behavior

In the conversation history, most agent implementations will show a message like "User responded" etc as a placeholder for the action that has been submitted, to give visual feedback that the agent has been prompted. 

## Good behavior

A message like "Selected a vacation period from x/x/xxx to y/y/yyyy" shows up.

## Ideas

Perhaps the "action" schema could provide a property for providing a user-visible string? Perhaps that can reference functions, so that string format etc can be used to format a good string with values from the data model etc?

Though I wonder if the user visible string should actually be exactly the same as what the LLM sees? Perhaps that is the default way things work, then it's optional to add extra LLM-only context?

jacobsimionato:
### Summary & Rationale
Added an optional `userMessage` field (referenced to `DynamicString`) to v1.0 `Action` event definitions and `renderer_to_agent` action message payloads. This allows AI agents and renderers to specify or resolve human-readable descriptions of user actions for display in conversation history, action confirmation UI, or user feedback.

### Key Changes
- **v1.0 JSON Schemas**:
  - Updated `specification/v1_0/json/common_types.json` to add optional `userMessage` property (using `"#/$defs/DynamicString"`) to `Action.event`.
  - Updated `specification/v1_0/json/renderer_to_agent.json` to add optional `userMessage` property to `action` message payload.
- **v1.0 Specification Tests**:
  - Updated `specification/v1_0/test/cases/renderer_messages.json` with test case verifying `action.userMessage` validation.
  - Updated `specification/v1_0/test/cases/button_checks.json` with test case verifying `Action.event.userMessage` validation (literal string and dynamic path binding).

### Verification & Testing
- Executed v1.0 schema test suite: `python3 specification/v1_0/test/run_tests.py` (44 test cases passed cleanly).

Closes #2014


specification/v1_0/json/common_types.json
```diff
@@ -305,6 +305,10 @@
                   "type": "string",
                   "description": "The name of the action to be dispatched to the agent."
                 },
+                "displayName": {
+                  "type": "string",
```
wrenj: "$ref": "#/$defs/DynamicString"

specification/v1_0/json/common_types.json
```diff
@@ -305,6 +305,10 @@
                   "type": "string",
                   "description": "The name of the action to be dispatched to the agent."
                 },
+                "displayName": {
```
wrenj: this implies the name of action (eg, submit time off), not saying what happened (eg, submitted your time off)

consider alternatives:
displayText or userMessage

I think userMessage is clearest

specification/v1_0/json/common_types.json
jacobsimionato: Updated! Refactored `userMessage` to reference `#/$defs/DynamicString` so it supports both literal strings and dynamic path bindings.

specification/v1_0/json/common_types.json
jacobsimionato: Agreed, `userMessage` is much better! Renamed the property from `displayName` to `userMessage` across the v1.0 schemas and test cases.

jacobsimionato:
Thanks for the review @wrenj !

wrenj (approved):

gspencergoog (approved):

PR #2229, opened 2026-08-12
Allow object type in A2UI v1.0 DynamicValue schema

utkarshagrawalgoog:
# Description

Allow `object` type in the A2UI v1.0 `DynamicValue` schema while resolving validation conflicts.

This change:
1. Adds a restricted `object` definition to the union of allowed types in the A2UI v1.0 `DynamicValue` definition in `specification/v1_0/json/common_types.json`.
2. Restricts the generic object definition using `not` so it does not match objects containing `path` or `call` properties. This prevents a `oneOf` validation conflict with `DataBinding` (requires `path`) and `FunctionCall` (requires `call`).
3. Updates `specification/v1_0/test/testing_catalog.json` to define a `TestComponent` utilizing `DynamicValue`.
4. Adds integration test cases in `specification/v1_0/test/cases/dynamic_value_validation.json` verifying validation of:
   - Literal values (string, number, boolean, array)
   - Generic JSON objects (e.g. Vega specs)
   - Valid DataBindings and FunctionCalls
   - Invalid DataBindings and FunctionCalls (with extra properties)

This is required to support components like `VegaChart` whose `spec` property is defined as a `DynamicValue` in the composite catalog. Currently, the A2UI ADK validator rejects any `VegaChart` payload containing a raw JSON object for the `spec` field, causing tool validation errors during tests and runtime.

This change was originally proposed for `v0.9` in internal CL 955178940, but since `v0.9` is frozen, we are proposing this update to the `v1.0` specification.

Fixes: b/544958659


jacobsimionato:
AFAICT, this lets you specify specify a literal object value a component parameter. Previously, you could already plumb an object as the output of a function or data model reference, but not a literal value, which seems like an accident to me.

Could you add a test for this under `specification/v1_0/test/cases/`, showing an example of a valid (and possibly invalid) object being passed as a parameter?

utkarshagrawalgoog:
Added valid and invalid objects

jacobsimionato (approved):
Great tests! This looks good to me. Let's see what @gspencergoog thinks too!

gspencergoog (commented):
This LGTM. It makes me wonder if we should also prefix "call" and "path" to be `$call` `_call` , or `@call` or something to make collisions less likely.

gspencergoog (approved):

PR #2238, opened 2026-08-12
feat(spec): rename `callableFrom` to `allowedCallers`

gspencergoog:
## Summary

Refactors function boundary declarations in the A2UI v1.0 specification by replacing `callableFrom` with `allowedCallers` across catalog schemas, protocol documentation, and evaluation types.

## Changes

- Replaces `callableFrom` with `allowedCallers` (enum: `rendererOnly`, `agentOnly`, `rendererOrAgent`, default: `rendererOnly`) in `catalog_definition.json` to specify which roles are authorized to invoke a function.
- Enforces that catalog functions with `requiresUserActivation: true` restrict `allowedCallers` to `["rendererOnly"]` because direct agent invocations cannot supply user activation context.
- Updates TypeScript types in `specification/v1_0/eval/src/types.ts` to include `allowedCallers?: 'rendererOnly' | 'agentOnly' | 'rendererOrAgent'`.
- Updates `a2ui_protocol.md` and `evolution_guide.md` to reflect `allowedCallers` execution boundary enforcement for `callRendererFunction` messages and inline catalog definitions.

## Impact & Risks

- Breaking schema change for catalog authors migrating from early v1.0 drafts using `callableFrom` (no impact on v0.9 catalogs as caller restrictions were introduced in v1.0).

## Testing

- Ran schema validation tests via `uv run --with pytest pytest tests/schema/` in `agent_sdks/python/a2ui_agent`.
- Ran format check and lint scripts via `./scripts/fix_format.sh`.


jacobsimionato:
I like `allowedCallers` as a name!

But I think `AgentOnly` still makes sense, because that is a function that is implemented on the client, can be called by agents via `callClientFunction` etc, but can't be used as an action handler or to process input data for A2UI dynamic value parameters. It can *only* be called by the agent directly. I think that some of the functions @wrenj was interested in were like that, e.g. a `getClientContext()` function that the agent calls explicitly.

My personal preference would actually be to remove `allowedCallers` entirely because I'm not convinced it's a meaningful security or privacy barrier. But if we have it, we may as well support all 3 permutations of access?

I'm likely missing something - can you say more about the rationale?

gspencergoog:
> I like `allowedCallers` as a name!
> 
> But I think `AgentOnly` still makes sense, because that is a function that is implemented on the client, can be called by agents via `callClientFunction` etc, but can't be used as an action handler or to process input data for A2UI dynamic value parameters. It can _only_ be called by the agent directly. I think that some of the functions @wrenj was interested in were like that, e.g. a `getClientContext()` function that the agent calls explicitly.

Okay, I didn't see any point to something that was defined on the renderer but couldn't be called by it.  I guess I'm wrong, though. I'll add it back in.

> My personal preference would actually be to remove `allowedCallers` entirely because I'm not convinced it's a meaningful security or privacy barrier. But if we have it, we may as well support all 3 permutations of access?

Oh, I actually agree with you.  I feel like this is more of a hint to the LLM to not try to call functions it can't call than it is an actual security measure.  Actual security would need be handled by the function implementation in any case.

If we removed `allowedCallers`, then developers would have to modify their prompts to tell the LLM what not to use.

jacobsimionato (approved):
Thanks for iterating! LGTM!

PR #2251, opened 2026-08-13
feat(eval): classify a2ui_scorer failures and add failure_distribution metric (#2200)

### LINKED ISSUE #2200: [FEATURE]: Classify a2ui_scorer failures instead of collapsing them to a single 0.0
- [x] I have searched the existing issues to make sure this feature has not already been requested.

## Is your feature request related to a problem? Please describe.

`a2ui_scorer` in `eval/a2ui_eval/scorers.py` reduces every failure to the same thing:

```python
except Exception as e:
    return Score(value=0.0, answer=answer_text, explanation=str(e))
```

With `metrics=[accuracy()]`, the only thing that reaches a run summary is a pass rate. The per-sample `explanation` is in the log, but nothing aggregates it, so there is no way to ask "what kind of failures did this run produce?"

The committed baseline shows why that hurts. `eval/baselines/direct_json/run_meta.json` covers 51 samples, 5 of which fail:

| sample | schema_acc | quality_acc | code_tokens |
| --- | --- | --- | --- |
| productGalleryData | 1.0 | 0.0 | 802 |
| weatherForecast | 1.0 | 0.0 | 992 |
| standardFunctions | 1.0 | 0.0 | 446 |
| restaurantFinder | 0.0 | 0.0 | 1705 |
| rizzCharts | 0.0 | 0.0 | 16 |

`restaurantFinder` emitted 1705 code tokens and failed schema validation. `rizzCharts` emitted 16 and failed. Those are opposite problems. One is a rich payload that violates the schema, the other is essentially no answer at all. Both land as `schema_acc: 0.0`.

This matters more now that the iterative format optimizer (#2092) compares `direct`, `express`, `elemental` and `atom` against exactly these aggregates. When `schema_acc` moves between formats, there is no way to tell whether the format made models emit more malformed payloads or more empty responses. Those call for opposite fixes.

## Describe the Proposed Solution

Stop discarding the classification the SDK already produces.

`a2ui.core.exceptions` already defines a typed hierarchy: `A2uiParseError`, `A2uiValidationError`, `A2uiIntegrityError`, `A2uiRecursionError`, `A2uiCatalogError` and `A2uiCompileError`. On top of that, `A2uiValidationError` carries `A2uiErrorDetail(path, code, message)` with codes including `missing_field`, `type_mismatch`, `extra_field`, `invalid_value` and `invalid_pointer`. `validation/validator.py` builds all of it, and the scorer flattens the lot to `str(e)`.

Proposed:

1. Derive a coarse category from the exception type, and a finer one from `detail.code`.
2. Give the three branches that already exist before validation their own categories: no model output, solver-step compilation failure, and no A2UI payload found in the response. That last one is very likely what `rizzCharts` is.
3. Record the category on `Score.metadata` and add a small `@metric` that reports the distribution over a run.

Deliberately additive. `Score.value` stays 0.0/1.0 and `accuracy()` is untouched, so `schema_acc` keeps its current meaning and existing baselines stay comparable.

## Describe Alternatives Considered

- **Dict-valued score with per-key metrics.** Changes the shape of the metric output, which would break comparison against the committed baselines and `bin/report_evals.py`.
- **A second scorer emitting a categorical value with `frequency()`.** `frequency()` is built for exactly this, but a separate scorer would re-run parsing and validation only to label a failure that has already been computed once.
- **Leave it, and parse `explanation` strings downstream.** Fragile: those strings come from jsonschema and change whenever the schema does.

## Additional Context

No new dependencies are needed; everything required is already in the SDK and in Inspect. The pass/fail value does not change, so this should be safe to land ahead of further v1.0 eval work.

Happy to put up a PR with tests if you think this is worth having. I would want to agree the category names with you first, since they would end up in run metadata that other tooling might key off.

Varun-S10:
# Description

### Overview
This PR adds failure categorization and a distribution metric to `a2ui_scorer` so that evaluation runs provide clearer insights into why samples did not pass.

### What is Changing
1. **Failure Categorization (`classify_exception`)**:
   - Classifies errors into specific categories based on exception type and validation details (e.g. `validation_error:missing_field`, `validation_error:type_mismatch`, `validation_error:extra_field`, `integrity_error`, `parse_error`).
   - Categorizes pre-validation cases cleanly: `no_model_output`, `solver_step_compilation_failure`, and `no_a2ui_payload_found`.

2. **Score Metadata**:
   - Saves the category information in `Score.metadata` (`failure_category`, `coarse_category`, `error_type`, and error details where available).

3. **Additive Metric (`failure_distribution`)**:
   - Adds a `@metric failure_distribution` to `a2ui_scorer` to show counts and proportions of each failure reason across an evaluation run.

4. **Backwards Compatibility**:
   - `Score.value` continues to return `1.0` or `0.0`, and the `accuracy()` metric is unchanged so existing baselines and reporting tools stay fully compatible.

5. **Tests**:
   - Added unit tests in `eval/tests/test_scorers.py` covering all failure categories and the metric calculation.

Fixes #2200

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].

For this PR:


eval/a2ui_eval/scorers.py
```diff
@@ -15,25 +15,127 @@
...
+    A2uiParseError,
+    A2uiRecursionError,
+    A2uiValidationError,
+)
+from a2ui.parser.errors import A2uiCompilationError
+
+try:
+    from a2ui.inference_formats.experimental.express.errors import ExpressCompilerError
+except ImportError:
+    ExpressCompilerError = ()  # type: ignore[misc,assignment]
```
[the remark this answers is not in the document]
Varun-S10: Applied the suggested change.

Varun-S10:
Hi @ditman, Could you please take a look?

FelippeRoza:
Thanks for picking this up. Keeping `Score.value` and `accuracy()` unchanged is the part I was most concerned about, since it means the committed baselines under `eval/baselines/` stay comparable across the change.

One question on scope: the `pack_specs_hook.py` hunk looks unrelated to failure classification. Was it meant to be part of this PR?

I ask because I hit that same ANTLR failure setting `eval/` up locally, so I am sympathetic to wanting it fixed, but I am not sure the fallback is safe as written. If generation fails at a point where `specification/inference_formats/express/Express.g4` has changed, the build now succeeds against the previously generated parser and the only signal is a `print` to stdout. That would surface later as confusing Express parsing behaviour rather than as a build failure, which seems worse than the current hard error.

In my case the underlying cause was `antlr4-tools` failing to fetch `antlr4-4.13.2-complete.jar` even though the Maven URL was reachable from the same machine. Seeding the jar into `~/.m2/repository/org/antlr/antlr4/4.13.2/` was enough for a clean build. A JDK is also required to build `a2ui-agent-sdk` from source and I could not find that documented anywhere. Happy to file both separately if that would help.

If you would rather keep the fallback, would it be worth gating it on the generated files being newer than `Express.g4`, so a stale parser still fails the build?

ditman:
I'm not the right person to evaluate this, let me summon @nan-yu and/or @gspencergoog  here!

eval/a2ui_eval/scorers.py
```diff
@@ -15,25 +15,130 @@
...
+        if "recursion limit exceeded" in msg:
+            return "recursion_error", "recursion_error"
+
+        return "validation_error", "validation_error"
+
+    if isinstance(e, A2uiParseError):
+        msg = str(e).lower()
+        if "not found in response" in msg or "empty" in msg:
+            return "no_a2ui_payload_found", "no_a2ui_payload_found"
+        return "parse_error", "parse_error"
```
gspencergoog, line 130: `"not found in response" in msg` matches both `"A2UI tags '<a2ui-json>' and '</a2ui-json>' not found in response" `(when no tags exist) and `"A2UI close tag '</a2ui-json>' not found in response"` (when an opening tag is there, but is missing the close tag). Classifying unclosed blocks as `no_a2ui_payload_found` isn't quite right: unclosed/partial tags should probably be `parse_error`s.

Maybe do this:

```suggestion
    if isinstance(e, A2uiParseError):
        msg = str(e).lower()
        if "close tag" not in msg and ("not found in response" in msg or "empty" in msg):
            return "no_a2ui_payload_found", "no_a2ui_payload_found"
        return "parse_error", "parse_error"
```

eval/a2ui_eval/scorers.py
```diff
@@ -15,25 +15,130 @@
...
+            "missing root component" in msg
+            or "duplicate component id" in msg
+            or "references non-existent component" in msg
+            or "circular reference" in msg
+        ):
+            return "integrity_error", "integrity_error"
+        if "recursion limit exceeded" in msg:
+            return "recursion_error", "recursion_error"
+
+        return "validation_error", "validation_error"
```
gspencergoog, line 124: If an unreachable component error is wrapped by an `A2uiValidatorError`, it falls through to `validation_error`, misclassifying an integrity failure as a schema validation failure.

You should probably handle "not reachable from" here (raised by `analyze_topology` when orphan components are detected).

gspencergoog (commented):
Thanks for sending this PR, Varun.

PR #2255, opened 2026-08-13
feat(conformance): add v1.0 protocol test vectors and TypeScript test harness

gspencergoog:
# Title: feat(conformance): add v1.0 protocol test vectors and multi-language test runner filtering

## Summary
Establishes the Stage 1 (`Mise`) protocol conformance test vectors for A2UI v1.0 specification features, introduces the TypeScript conformance test runner harness in `renderers/web_core`, and implements `SUPPORTED_SPEC_VERSIONS` and `SKIP_TEST_NAMES` version filtering across TypeScript, Kotlin, PyTest schema, and Python agent conformance test runners.

## Changes
- **Bidirectional RPC Test Vectors**: Added [`conformance/core/rpc_functions.yaml`](file:///Users/gspencer/code/a2ui/v1_0_mise/conformance/core/rpc_functions.yaml) testing `callRendererFunction` (Agent-to-Renderer) execution, `callableFrom` authorization rules (`rendererOnly` vs `rendererOrAgent`), `requiresUserActivation` flags, and `callAgentFunction` (Renderer-to-Agent) response routing.
- **Multi-Catalog Resolution Test Vectors**: Added [`conformance/core/multi_catalog.yaml`](file:///Users/gspencer/code/a2ui/v1_0_mise/conformance/core/multi_catalog.yaml) testing component resolution across explicit and surface-default catalog IDs, unsupported catalog error handling, and protocol version mismatch checks.
- **Data Model Deletion Test Vectors**: Added [`conformance/core/data_deletion.yaml`](file:///Users/gspencer/code/a2ui/v1_0_mise/conformance/core/data_deletion.yaml) testing property deletion via `value: null`, parent object key preservation, and top-level key removal.
- **Built-in `@index` Function Test Vectors**: Added [`conformance/core/index_function.yaml`](file:///Users/gspencer/code/a2ui/v1_0_mise/conformance/core/index_function.yaml) testing `@index` and `@index(offset: 1)` in collection template iteration loops.
- **Validation Result Test Vectors**: Added [`conformance/core/validation_result.yaml`](file:///Users/gspencer/code/a2ui/v1_0_mise/conformance/core/validation_result.yaml) testing dynamic `ValidationResult` payload objects (`{ valid: false, code, message, severity }`) and primitive boolean fallbacks.
- **Composition Constraints Test Vectors**: Added [`conformance/core/composition_constraints.yaml`](file:///Users/gspencer/code/a2ui/v1_0_mise/conformance/core/composition_constraints.yaml) testing `allowedParents`, `allowedChildren`, and implicit `Surface` top-level container rules.
- **Data Model Pointer Test Vectors**: Added [`conformance/core/data_model_pointers.yaml`](file:///Users/gspencer/code/a2ui/v1_0_mise/conformance/core/data_model_pointers.yaml) testing RFC 6901 JSON Pointer escape resolution (`~0`, `~1`), array indexing, and `updateDataModel` path resolution.
- **v1.0 Agent SDK Inference Format Test Vectors**: Consolidated v1.0 Direct JSON `<a2ui-json>` envelope test case into [`conformance/agent/inference_format.yaml`](file:///Users/gspencer/code/a2ui/v1_0_mise/conformance/agent/inference_format.yaml).
- **Conformance Runner Filtering & Documentation**: Added top-level `SUPPORTED_SPEC_VERSIONS` and `SKIP_TEST_NAMES` configuration structures to TypeScript, Kotlin, PyTest schema, and Python agent runners, documenting their usage in [`conformance/README.md`](file:///Users/gspencer/code/a2ui/v1_0_mise/conformance/README.md).
- **TypeScript Test Runner Harness**: Created [`renderers/web_core/tests/conformance/conformance_test.mjs`](file:///Users/gspencer/code/a2ui/v1_0_mise/renderers/web_core/tests/conformance/conformance_test.mjs) with non-zero exit discovery checks, try-catch file load handling, and action dispatch validation.

## Impact & Risks
- Low risk; introduces test vectors and runner configuration without altering existing v0.8/v0.9 runtime behavior in `web_core` or language SDKs.
- Schema compliance strictly verified against `conformance_schema.json` and v1.0 canonical JSON schemas (`agent_to_renderer.json`, `renderer_to_agent.json`, `catalog_definition.json`, `common_types.json`).

## Testing
- **TypeScript Test Harness**: Run `node renderers/web_core/tests/conformance/conformance_test.mjs`. (Verified `210/210` test cases pass).
- **Kotlin Test Harness**: Run `./gradlew test` in `kotlin/agent_sdk_legacy`. (Verified `227/227` tests pass).
- **PyTest Schema Validation**: Run `uv run --with pyyaml --with jsonschema pytest conformance/tests/test_conformance_yaml.py`. (Verified `14 Passed, 1 Skipped`).
- **Python Agent SDK Harness**: Run `PYTHONPATH=agent_sdks/python/a2ui_agent/src:agent_sdks/python/a2ui_core/src uv run --with pyyaml --with jsonschema --with pydantic pytest agent_sdks/python/a2ui_agent/tests/conformance/test_conformance.py`. (Verified `184/184 Passed`).


renderers/web_core/tests/conformance/conformance_test.mjs
```diff
@@ -0,0 +1,172 @@
...
+  return yaml.load(content);
+}
+
+function runConformanceHarness() {
+  console.log('=====================================================');
+  console.log('A2UI Web Core TypeScript Conformance Test Harness');
+  console.log('=====================================================');
+
+  const files = [...findYamlFiles(CORE_DIR), ...findYamlFiles(AGENT_DIR)];
+  console.log(`Discovered ${files.length} conformance YAML test suite file(s).`);
```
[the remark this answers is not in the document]
gspencergoog: Added non-zero exit validation when 0 test files are discovered to prevent silent CI passes.

renderers/web_core/tests/conformance/conformance_test.mjs
```diff
@@ -0,0 +1,172 @@
...
+  const failures = [];
+
+  for (const filePath of files) {
+    const relativePath = path.relative(CONFORMANCE_ROOT, filePath);
+    const testCases = loadYamlFile(filePath);
+
+    if (!Array.isArray(testCases)) {
+      console.warn(`[SKIP] ${relativePath}: Content is not an array of test cases.`);
+      continue;
+    }
```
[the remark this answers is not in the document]
gspencergoog: Wrapped YAML file loading in a try-catch block to record malformed YAML files as test failures without interrupting remaining suites.

conformance/core/data_model_pointers.yaml
```diff
@@ -0,0 +1,82 @@
...
+            catalogId: "basic"
+            dataModel:
+              "a/b":
+                "c~d": "escaped_value"
+              "slash/key": 100
+  expect_data_model:
+    s1:
+      "a/b":
+        "c~d": "escaped_value"
+      "slash/key": 100
```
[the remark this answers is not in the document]
gspencergoog: Added an updateDataModel step using the escaped path /a~1b/c~0d to explicitly test RFC 6901 escape sequence resolution.

renderers/web_core/tests/conformance/conformance_test.mjs
```diff
@@ -0,0 +1,189 @@
...
+      'select_catalog test requires "expect_selected", "expect_catalog_schema", or "expect_error".',
+    );
+  }
+}
+
+function validateValidateTestCase(testCase) {
+  const {steps, payload, expect_error, expect_data_model} = testCase;
+  if (!steps && !payload && !expect_error && !expect_data_model) {
+    throw new Error('validate test case missing input payload/steps or assertions.');
+  }
```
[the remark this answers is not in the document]
gspencergoog: Updated validateValidateTestCase to require either steps or payload input.

renderers/web_core/tests/conformance/conformance_test.mjs
```diff
@@ -0,0 +1,189 @@
...
+    let testCases;
+    try {
+      testCases = loadYamlFile(filePath);
+    } catch (err) {
+      totalFailed++;
+      const failMessage = `  ✗ FAILED to load ${relativePath}: ${err.message}`;
+      console.error(failMessage);
+      failures.push({file: relativePath, name: 'YAML Parsing', error: err.message});
+      continue;
+    }
```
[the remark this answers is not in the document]
gspencergoog: Incremented totalTests inside the file loading catch block so summary metrics remain consistent.

conformance/core/composition_constraints.yaml
```diff
@@ -0,0 +1,89 @@
...
+    version: "1.0"
+  action: validate
+  steps:
+    - payload:
+        - version: "v1.0"
+          createSurface:
+            surfaceId: "s1"
+            catalogId: "basic"
+            components:
+              root:
```
nan-yu: This is not canonical A2UI messages. Why not?
```
- version: "v1.0"
          createSurface:
            surfaceId: "s1"
            catalogId: "basic"
            components:
              - id: root
                 component: Column
                 children: [t1]
              - id: t1
                 component: Text
                 text: "Hello World"
```

conformance/core/data_deletion.yaml
```diff
@@ -0,0 +1,95 @@
...
+              settings:
+                notifications:
+                  email: true
+                  push: false
+                  sms: true
+    - payload:
+        - version: "v1.0"
+          updateDataModel:
+            surfaceId: "s1"
+            path: "/settings/notifications/email"
```
nan-yu, line 66: I feel like one more test case is missing here. If we delete `/settings/notifications`, will all nested children be removed?

conformance/core/index_function.yaml
```diff
@@ -0,0 +1,92 @@
...
+                    item_index:
+                      component: Text
+                      text:
+                        call: "@index"
+                        args: {}
+                    item_text:
+                      component: Text
+                      text:
+                        path: ""
+
```
nan-yu, line 64: Missing `expected` output?

conformance/core/index_function.yaml
```diff
@@ -0,0 +1,92 @@
...
+              root:
+                component: Collection
+                items: "/players"
+                template:
+                  component: Text
+                  text:
+                    call: "@index"
+                    args:
+                      offset: 1
+
```
nan-yu, line 97: Missing expected?

conformance/core/validation_result.yaml
nan-yu, line 1: Should we rename this to be `client_functions.yaml`?

conformance/tests/test_conformance_yaml.py
```diff
@@ -51,3 +57,17 @@ def test_validate_conformance_yaml(yaml_path):
         jsonschema.validate(instance=yaml_data, schema=SCHEMA)
     except jsonschema.ValidationError as e:
         pytest.fail(f"{basename} failed schema validation: {e.message}")
+
+    if isinstance(yaml_data, list):
```
nan-yu: This only validates the conformance spec. Do we also need to check versions for the spec?

v1_0_implementation_plan.md
nan-yu, line 1: Do we want to check this in as well?

gspencergoog: This is just formatting changes.  I can move it to another PR if you want.

conformance/core/validation_result.yaml
gspencergoog: This is about validation functions (functions that return `ValidationResult`), not client functions in general.  But I guess it depends on what granularity you want to have in the conformance suite.  We could combine with with other function conformance vectors, but I think keeping it pretty granular is good.

conformance/core/index_function.yaml
gspencergoog: The only expectations you can assert in the conformance schema seem to be just passing without an exception, or  `expect_error`.  I'll add some expectations for the output to the conformance schema.

conformance/core/composition_constraints.yaml
gspencergoog: Good catch, yes, these should be lists. The tests didn't catch that because we haven't got an implementation to check against, so this didn't fail yet.

SInce we're still just in the first phase, the test harness only runs contract checks on the test runner files. It doesn't instantiate a `web_core` or `a2ui_core` engine or run protocol validations yet.

conformance/core/data_deletion.yaml
gspencergoog: Good point.  Added.

conformance/tests/test_conformance_yaml.py
gspencergoog: Good point.  Removed.

conformance/conformance_schema.json
```diff
@@ -178,7 +178,23 @@
     "ValidateTest": {
       "type": "object",
       "properties": {
-        "action": {"const": "validate"}
+        "action": {"const": "validate"},
```
gspencergoog, line 181: These are the expectations I added.  Is that a good set?

conformance/core/index_function.yaml
gspencergoog: Yes. Added.

v1_0_implementation_plan.md
gspencergoog: Actually, I found an error in the doc, and I added these changes to a different PR, and fixed that error too: https://github.com/a2ui-project/a2ui/pull/2261

conformance/conformance_schema.json
nan-yu: What about just the following?
-  expect_valid: boolean
- expect_error: object
- expect: object

I'm adding test cases for message_processor.yaml, which needs to associate dataModel, components to a surface, and also validate the surface lifecycle. Instead of introducing `expect_surfaces`, `expect_components`, and `expect_data_model`, and probably `expect_functions` separately, I'd like to keep the structure, for example,
```
  surfaces:
      surface1:
        exists: true
        sendDataModel: true
        components:
          - id: "root"
            component: "Column"
            children: ["btn1"]
          - id: "btn1"
            component: "Button"
            title: "Inline Button"
        dataModel:
          counter: 42
```

conformance/core/validation_result.yaml
nan-yu: Sounds good.

conformance/agent/inference_format.yaml
```diff
@@ -326,3 +324,41 @@
     - "### Catalog Schema:"
     - '"Text": {'
     - "---END A2UI JSON SCHEMA---"
+
+- name: test_agent_v1_0_direct_json_envelope_parsing
+  description: Verifies parsing Direct JSON v1.0 payload envelopes wrapped in <a2ui-json> sentinel tags.
+  catalog:
+    version: "1.0"
```
nan-yu: should we make it more explicit to be `protocol_version`?

gspencergoog: Yes, let's do that. I'll update it.

conformance/conformance_schema.json
gspencergoog: Okay, that makes sense, although it might make it validate more than necessary sometimes (we weren't validating the fact that there's a column before, for instance, just the indexed text items), since it has to include the structure.

Also, by definition this is a "contains" match, right, not "equals"?  Otherwise we'd have to duplicate the whole structure, and that feels fragile.

So an entire test case would like this?:

```yaml
- name: test_index_function_nested_path
  description: Verifies built-in @index function evaluation inside a deeply nested child property path of a template list.
  catalog:
    version: "1.0"
  action: validate
  steps:
    - payload:
        - version: "v1.0"
          createSurface:
            surfaceId: main
            catalogId: basic
            dataModel:
              users:
                - name: Alice
                  address:
                    city: Seattle
                - name: Bob
                  address:
                    city: Portland
            components:
              - id: root
                component: Collection
                items: "/users"
                template:
                  component: Column
                  children: [nested_text]
                  components:
                    - id: nested_text
                      component: Text
                      text:
                        call: "@index"
                        args:
                          offset: 1
  expect:
    surfaces:
      main:
        dataModel:
          users:
            - name: Alice
              address:
                city: Seattle
        components:
          - id: root
            component: Column
            components:
              - id: nested_text_0
                text: "1"
              - id: nested_text_1
                text: "2"
```

(I included a `dataModel` expectation here just for illustration: it wasn't included in this test).

instead of:

```yaml
  expect_components:
    main:
      nested_text_0:
        text: "1"
      nested_text_1:
        text: "2"
```

gspencergoog: Actually, in implementing this, having it match the nesting structure exactly feels pretty fragile.  Take a look at what I have implemented, which is more flattened, which matches the input structure more, since the input uses adjacency lists.

For example, I have this now:

```yaml
  expect:
    surfaces:
      main:
        dataModel:
          users:
            - name: Alice
              address:
                city: Seattle
            - name: Bob
              address:
                city: Portland
        components:
          root:
            component: Column
            children: ["nested_text_0", "nested_text_1"]
          nested_text_0:
            text: "1"
          nested_text_1:
            text: "2"
```

nan-yu: The example looks good. I think we can go with it now and update later.

> by definition this is a "contains" match, right, not "equals"? 

That's correct. It is not an `exact` match, but only a `contains` match.

nan-yu (approved):

agent_sdks/python/a2ui_agent/tests/conformance/test_conformance.py
```diff
@@ -48,6 +48,12 @@
     "CompileError": A2uiCompileError,
 }
 
+# Set of A2UI specification versions supported by this Python Agent SDK conformance harness.
+SUPPORTED_SPEC_VERSIONS = {"0.8", "0.9", "1.0"}
```
nan-yu: While working on other conformance test spec, I noticed inconsistency around the versions. Should we always add the leading `v` to match the protocol version in A2UI spec?

agent_sdks/python/a2ui_agent/tests/conformance/test_conformance.py
```diff
@@ -115,7 +121,7 @@ def load_tests(filename):
 
 
 def setup_catalog(catalog_config):
-    version = str(catalog_config["version"])
+    version = str(catalog_config["protocol_version"])
```
nan-yu: This is another inconsistency. The conformance test was set up using snake_case mostly, while A2UI spec uses camelCase.

To coordinate, I'd suggest using what we have for now, and convert to camelCase later in a separate PR. WDYT?

agent_sdks/python/a2ui_agent/tests/conformance/test_conformance.py
gspencergoog: Hmm. The v is kind of arbitrary.  I added it originally to make sure that it wasn't treated as a pure number, but now I question whether that is important. It makes it harder (slightly) to send it to APIs that expect semver versions, and it's easy to forget (like here).

Regardless, we should either remove the "v" entirely from the spec or we should be consistent.  I think it's probably too embedded in the spec to change easily now, so I'm going to make these have "v" in front of them.

agent_sdks/python/a2ui_agent/tests/conformance/test_conformance.py
gspencergoog: Yes, let's do that.  `camelCase` would be better.

PR #2256, opened 2026-08-13
feat(swift): Basic Catalog functions

pinieb:
# Description

This PR adds the 14 standard client-side functions defined in the A2UI v0.9.1 basic catalog schema to the Swift codebase.

The implementations align with the a2ui_core.blueprint.md module specifications and the basic_catalog_implementation_guide.md.

Included functions
and, or, not: Standard logical operations.
required, regex, length, numeric, email: Validation functions supporting text fields and conditional constraints.
formatNumber, formatCurrency: Uses standard Apple NumberFormatter logic mapped to the A2UI spec parameters.
formatDate: Parses numerical timestamps and ISO8601 strings and formats them using the provided TR35 format pattern.
formatString: Provides token parsing and recursive dynamic value resolution for string interpolation.
pluralize: Evaluates standard CLDR plural categories (zero, one, two, few, many, other). Includes an optional PluralResolver injection for host applications, falling back to an English-centric numeric heuristic since Foundation lacks dynamic plural categorization APIs.
openUrl: Implements the required security constraints by resolving relative URLs and restricting execution strictly to http: and https: schemes.
Technical details
All functions conform to the FunctionImplementation protocol, defining their FunctionAPI schema and an evaluate method.

Because Swift's evaluate protocol contract requires returning a synchronous JSONValue (unlike the web renderer which returns a reactive Signal), these functions act purely as static evaluators. Context reactivity for formatString and other dynamic function calls is expected to be managed by the binder or data context layers re-evaluating the tree when dependencies change.


jacobsimionato (approved):
LGTM! We will be creating conformance tests suites which should cover much of this, so we can hook these up later to help confirm the behavior matches other platforms and tweak as necessary.

PR #2284, opened 2026-08-15
refactor(python): migrate agent_sdks/python packages to python/

nan-yu:
## Summary
Migrates the Python SDK packages from `agent_sdks/python/` to the canonical `python/` directory and removes the legacy `agent_sdks/` structure.

## Changes
- **Directory Migration**:
  - Relocated `agent_sdks/python/a2ui_core` -> `python/a2ui_core`.
  - Relocated `agent_sdks/python/a2ui_agent` -> `python/a2ui_agent`.
  - Relocated `agent_sdks/python/release.sh` -> `python/release.sh`.
  - Relocated `agent_sdks/python/README.md` -> `python/README.md`.
  - Deleted obsolete `agent_sdks/` directory structure.
- **Repository Configuration & CI**:
  - Updated root `pyproject.toml`, `uv.lock`, and `.gitignore`.
  - Updated GitHub Actions workflows (`python_ci.yml`, `run_evals.yml`) to reference the new paths.
- **Documentation & Blueprints**:
  - Updated blueprint paths under `blueprints/codebases/python/`.
  - Updated documentation references across `docs/` and `conformance/`.
- **Test Deduplication & Helpers**:
  - Added centralized repository root, specification, and conformance path helpers (`find_repo_root`, `get_spec_dir`, `get_basic_catalog_path`, `get_basic_examples_dir`, `get_conformance_path`) in `a2ui.schema.utils` and `conformance_helpers.py`.
  - Refactored test suites to eliminate boilerplate directory traversal loops.

## Verification
- **Formatting**: `uv run pyink --check .` (Passed, 320 files checked).
- **Type Checking**: `uv run --all-packages mypy .` (Passed, 0 errors across 90 source files).
- **Test Suites**: `uv run pytest` (Passed all 783 tests across all workspaces).

TAG=agy
CONV=707c9b6b-b2da-4940-ba6e-67f0b4f1d563

PR #2285, opened 2026-08-15
feat(skills): add automated A2UI Python schema generation skill

nan-yu:
## Description

This pull request is stacked on top of #2284 (`v1_0_firing_1`).

- Adds the `a2ui-generate-python-schemas` skill with the dynamic Pydantic v2 code generator (`codegen_pydantic.py`).
- Supports automated schema generation, basic catalog bindings, and version adapters across any A2UI specification version (`v0.8`, `v0.9`, `v1.0`, etc.).
- Updates `python/a2ui_core/tests/test_generate_schemas.py` to verify syntax and generation validity across versions.

TAG=agy


nan-yu:
### Rationale for Custom `codegen_pydantic.py` over Generic Generators (`datamodel-code-generator` / `json-schema-to-pydantic`)

Generic third-party code generators (such as [datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator) or [json-schema-to-pydantic](https://github.com/richard-gyiko/json-schema-to-pydantic) struggle with A2UI’s complex schema graph, causing parser crashes and unmaintainable outputs:

1. **Complex Schema Graph & Recursion Limit Crashes**:
   - Generic generators frequently hit Python recursion limit crashes or produce invalid AST representations when encountering A2UI's recursive structures (e.g. `ChildList` ⇄ `Component`, nested `DataValueMapItem` trees, and recursive `DynamicValue` expressions).
   - Our generator handles recursion boundaries deterministically with forward references (`from __future__ import annotations`) and type aliases without blowing the call stack.

2. **Cross-File `$ref` & Circular Dependency Resolution**:
   - A2UI schemas reference types across multiple specification JSON files (e.g. `catalog.json` referencing `common_types.json`, capabilities schemas referencing catalog definitions). Generic tools either fail to resolve relative URI `$ref` paths correctly or generate Python modules with circular `import` cycles that fail at runtime.
   - `codegen_pydantic.py` resolves cross-schema references into a directed dependency graph, isolating shared types into `common_types.py` and preventing circular imports.

3. **Semantic Modular Partitioning per Version**:
   - Rather than dumping schemas into a single flat file or monolithic output, our generator cleanly partitions each specification version into structured submodules:
     - `constants.py` (version constants & literal types)
     - `common_types.py` (dynamic primitives, bindings, strict base models)
     - `server_to_client.py` / `agent_to_renderer.py` (outbound message envelopes)
     - `client_to_server.py` / `renderer_to_agent.py` (inbound action & error envelopes)
     - `client_capabilities.py` / `renderer_capabilities.py` (capability negotiation)
     - `__init__.py` (version export registry with explicit `__all__`)

4. **Catalog & Component Architecture Synthesis**:
   - Beyond raw Pydantic schemas, `codegen_pydantic.py` generates the executable catalog scaffolding required by `a2ui_core`:
     - Wraps each component definition into `ModelComponentApi(schema=...)` and synthesizes the `BASIC_COMPONENTS` registration map.
     - Synthesizes `FunctionApi` subclass signatures (`schema`, `return_type`, `name`) dynamically from catalog function specifications.
     - Generates the versioned `BasicCatalog` class bound to `_basic_catalog_id(SPEC_VERSION)` and catalog metadata.

5. **Protocol Versioning & Terminology Evolution**:
   - Handles terminology migration across specification versions (e.g. legacy `ServerToClient`/`ClientToServer` in v0.8/v0.9 vs. modern `AgentToRenderer`/`RendererToAgent` in v1.0+) while generating backwards-compatible aliases.
   - Automatically maintains the root `schema/__init__.py` multi-version enum (`A2uiProtocolVersion`) and multi-version envelope union types (`AgentToRendererMessage`, `RendererToAgentMessage`).

6. **Zero Dependencies & Clean Code Quality**:
   - Eliminates heavyweight, fragile third-party generator dependencies from our build and dev pipelines.
   - Emits clean, deterministic, PEP-484 compliant code that adheres to repository linting and formatting standards out of the box.

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
```diff
@@ -0,0 +1,2038 @@
...
+                if ref.endswith("Child"):
+                    return "Child"
+                if ref.endswith("Extensions"):
+                    return "Optional[Dict[str, Any]]"
+                if "common_types.json" in ref or ref.startswith("#/$defs/"):
+                    return ref.split("/")[-1]
+                elif ref.startswith("#/components/"):
+                    return f"{ref.split('/')[-1]}Component"
+                elif ref.startswith("#/"):
+                    return ref.split("/")[-1]
```
gspencergoog: Should we be more specific with these?  For example, should they check for `ref.endswith("/Component")` instead, to make sure it's not some other definition that happens to end with "Component" (e.g. to differentiate "#/$defs/DeletedComponent" from "#/$defs/Component")?

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
```diff
@@ -0,0 +1,2038 @@
...
+            if mapped not in mapped_items:
+                mapped_items.append(mapped)
+
+        return f"{class_name} = Union[{', '.join(mapped_items)}]\n"
+
+
+def generate_server_to_client(
+    s2c_data: Dict[str, Any],
+    dir_name: str = "v0_9",
+    spec_dot: str = "v0.9",
```
gspencergoog: Should we start defaulting to `v1.0` here?  And if not, why isn't this `v0.9.1`?

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
```diff
@@ -0,0 +1,2038 @@
...
+) -> Tuple[str, List[str]]:
+    """Generates server_to_client.py / agent_to_renderer.py content and message names."""
+    codegen = PydanticCodegen(dir_name, spec_dot)
+    codegen.allow_inline = False
+    is_modern_terminology = dir_name not in ("v0_8", "v0_9")
```
gspencergoog: Should this include `v1_0` now?

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
```diff
@@ -0,0 +1,2038 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="A2UI Dynamic Pydantic Codegen Engine",
+        epilog=(
+            "Examples:\n"
+            "  python codegen_pydantic.py --version v0.8   # Generate models for v0.8\n"
+            "  python codegen_pydantic.py --version v0.9   # Generate models for v0.9\n"
+            "  python codegen_pydantic.py --version v1.0   # Generate models for v1.0\n"
+            "  python codegen_pydantic.py --version v1.1   # Generate models for v1.1"
```
gspencergoog: Why do we need all four examples?

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
```diff
@@ -0,0 +1,2038 @@
...
+        ),
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    parser.add_argument(
+        "--version",
+        "-v",
+        required=True,
+        help=(
+            "Target protocol version to generate (e.g. 'v0.8', 'v0.9', 'v1.0', 'v1.1')"
```
gspencergoog: Why does this include "v1.1", when that doesn't exist yet?

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
nan-yu: Good point. Updated to force the match with a leading slash.

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
nan-yu: simplified with a single example.

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
nan-yu: We started to use modern terminologies from v1.0, so v1.0 should not be in the list. I also extracted it to be a separate helper method.

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
nan-yu: removed v1.1.

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
nan-yu: I refactored this method a bit to force the version to be a required parameter, otherwise, we have to update the default version every time when we have a new release.

gspencergoog (approved):

PR #2289, opened 2026-08-15
ci: scope format presubmit to changed files (#2232)

### LINKED ISSUE #2232: [FEATURE]: Scope presubmits to currently changed files rather than the whole repo
- [X] I have searched the existing issues to make sure this feature has not already been requested.

## Is your feature request related to a problem? Please describe.

Formatters and other automated checks run for the entire workspace on check in. In a recent Swift-only PR, I ran into this Dart formatting issue: https://github.com/a2ui-project/a2ui/actions/runs/31609384280/job/94168489841. This leaves contributors with the choice of bringing in unrelated formatting changes to their PRs to unblock themselves or to wait for a broader fix. As the monorepo grows, this will become more common.

## Describe the Proposed Solution

Automated presubmits should be pickier about where and when they run. For instance, we could scope the Dart formatter to run when files in the Dart codepaths have changed. Another option would be to only run presubmits on files in the active change. In that case, even if the Swift formatter ran, if the PR only contains TypeScript files, there would be no flags from the Swift formatter.

sanjayrohith:
Fixes #2232

## Problem

`scripts/fix_format.sh` runs Prettier, Pyink, `dart format`, `swift-format`, and ktfmt across the entire workspace on every PR. A contributor touching only Swift can be blocked by unrelated Dart formatting drift, as happened in [this run](https://github.com/a2ui-project/a2ui/actions/runs/31609384280/job/94168489841). Their choices are to pull unrelated formatting churn into their PR or to wait for someone else to fix it. As the monorepo grows this gets worse.

## Approach

The obvious fix — adding `paths:` filters to the `presubmit-lint` workflow — is a trap. `format-check` is a required check, and a required check that is filtered out never reports a conclusion, so PRs that skip it sit pending forever. So the job still always runs; it narrows its own work instead.

`fix_format.sh` gains a `--changed` mode that resolves the changed set from git and hands each formatter only its own files, skipping formatters whose languages the branch never touched. Default behaviour (whole repo) is unchanged, and post-submit on `main` still sweeps everything, so scoping can never let drift accumulate.

## Changes

**`scripts/fix_format.sh`**

- New flags: `--changed`, `--base <ref>`, `--plan`, `--help`. Existing invocations (`fix_format.sh`, `fix_format.sh --check`) behave exactly as before.
- Changed-set discovery unions committed, staged, unstaged, and untracked files, and drops deletions. It diffs against the **merge base**, not the branch tip, so commits that landed on `main` after the branch point are not dragged in.
- Per-language selection mirrors each formatter's existing whole-repo scope, so `--changed` can never widen what a formatter touches:
  - Dart is still limited to `samples/client/flutter` and `renderers/flutter`.
  - Swift is still limited to `Package.swift` and `swift/`.
  - Python filters out `/generated/` explicitly, because pyink's `extend-exclude` in `pyproject.toml` applies to directory walks but **not** to explicitly listed files. Without this, scoped runs would have formatted generated code that whole-repo runs skip.
  - ktfmt is a Gradle task rather than a file-level binary, so changed Kotlin sources are mapped to their owning Gradle module and only those modules run.
- `--plan` prints which formatters would run and over how many files, then exits. It needs no language toolchains.

**`.github/workflows/presubmit-lint.yml`**

- `format-check` computes its scope from `github.event.pull_request.base.sha` on PRs, and stays in whole-repo mode for pushes to `main`.
- The scope step drives toolchain installation off `--plan` output: Node, Python, and Dart setup are each skipped when that formatter has no work. A Swift-only PR now installs none of the three.
- `fetch-depth: 0` so the merge base is reachable.
- The new unit tests run in the `static-checks` job.

**`scripts/test_fix_format.py`** — 21 tests covering the scoping logic.

**`CONTRIBUTING.md`** — documents `--changed`, `--base`, and `--plan`.

## Verification

The issue's exact scenario, run against this repo with a single added Swift file:

```
$ ./scripts/fix_format.sh --plan --base upstream/main
--- formatting plan ---
mode: changed (base: upstream/main)
prettier: 0 file(s)
pyink: 0 file(s)
dart: 0 file(s)
swift: 1 file(s)
ktfmt: 0 module(s)
```

Feeding that through the workflow's gating logic:

```
prettier=false (SKIP setup)
pyink=false   (SKIP setup)
dart=false    (SKIP setup)
```

Dart formatting can no longer block a Swift-only PR, and the job stops installing three toolchains it will not use.

Scoped mode on this PR's own diff:

```
$ ./scripts/fix_format.sh --check --changed --base upstream/main
Running Prettier formatting for Node/Web assets...   All matched files use Prettier code style!
Running Pyink for Python files...                    1 file would be left unchanged.
Skipping Dart format: no matching files changed.
Skipping swift-format: no matching files changed.
Skipping ktfmt: no matching files changed.
Done.                                                # exit 0
```

Correct failure behaviour — a deliberately malformed TS file is caught, and only that file is checked:

```
[warn] renderers/web_core/src/__scope_probe.ts
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
# exit 1
```

Backwards compatibility — the default whole-repo path is unchanged and still passes:

```
$ ./scripts/fix_format.sh --check
Running Prettier formatting for Node/Web assets...  All matched files use Prettier code style!
Running Pyink for Python files...                   320 files would be left unchanged.
```

Unit tests:

```
$ python3 -m unittest scripts.test_fix_format
Ran 21 tests in 0.7s
OK
```

They cover: Swift-only / TypeScript-only / Dart-only isolation, `Package.swift` at the root, generated-Python exclusion, Dart outside the two formatted roots, Kotlin→Gradle-module mapping (including modules without ktfmt), untracked files, deleted files, unioning committed and working-tree changes, merge-base isolation from unrelated `main` commits, whole-repo mode, and the two error paths.

## Notes for reviewers

- **Prettier file selection is an extension allowlist** (`PRETTIER_EXTENSIONS`). Handing Prettier the raw changed set with `--ignore-unknown` was simpler, but then any changed file at all — a `.sh`, a `.png` — marks Prettier as needed and forces a Node install. The allowlist covers Prettier's built-in parsers; a Prettier *plugin* adding a new language would need a line here. The whole-repo post-submit run is the backstop if that is ever missed.
- **`mapfile` is deliberately avoided** so the script keeps running on the bash 3.2 that ships with macOS.
- **Out of scope, happy to follow up:** `workspace-lint` still runs `yarn build:all` + `lint:all` for every PR, and `static-checks` still checks license headers repo-wide. Both have the same "unrelated work blocks your PR" shape, but scoping them means reasoning about the Yarn workspace/wireit graph, which felt like a separate change rather than something to bundle here.

### DIFF
--- a/.github/workflows/presubmit-lint.yml
+++ b/.github/workflows/presubmit-lint.yml
@@ -83,10 +83,14 @@ jobs:
       # runs and narrows its own work below.
       - name: Determine formatting scope
         id: scope
+        # Passed through env rather than interpolated into the script body:
+        # `${{ }}` expansion inside `run:` is a template-injection vector.
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
         run: |
           set -euo pipefail
-          BASE_SHA="${{ github.event.pull_request.base.sha }}"
-          if [ "${{ github.event_name }}" = "pull_request" ] && [ -n "$BASE_SHA" ]; then
+          if [ "$EVENT_NAME" = "pull_request" ] && [ -n "${BASE_SHA:-}" ]; then
             SCOPE_ARGS="--changed --base $BASE_SHA"
           else
             # Post-submit on main sweeps the whole repo, so drift can never
@@ -119,7 +123,11 @@ jobs:
         with:
           sdk: "3.12.2"
       - name: Run enforce-formatting check
-        run: ./scripts/fix_format.sh --check ${{ steps.scope.outputs.args }}
+        env:
+          SCOPE_ARGS: ${{ steps.scope.outputs.args }}
+        # SCOPE_ARGS is intentionally unquoted: it is either empty or the two
+        # words "--changed --base <sha>", and must word-split.
+        run: ./scripts/fix_format.sh --check $SCOPE_ARGS
 
   workspace-lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/presubmit-lint.yml
+++ b/.github/workflows/presubmit-lint.yml
@@ -62,6 +62,9 @@ jobs:
       - name: Run fix_licenses unit tests
         run: python3 -m unittest scripts.test_fix_licenses
 
+      - name: Run fix_format scoping unit tests
+        run: python3 -m unittest scripts.test_fix_format
+
       - name: Check license headers
         run: ./scripts/fix_licenses.py --check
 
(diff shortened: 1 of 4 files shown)


sanjayrohith:
Both applied, thanks.

**Redundant `git diff` calls** — agreed. `git diff <merge_base>` compares the merge base against the working tree, so it already covers committed, staged, and unstaged changes; the `--cached` and bare `git diff` calls added nothing. I kept `git ls-files --others` alongside it, since untracked files don't appear in that diff. The existing tests (including the committed-plus-dirty case) still pass, and I added one more that pins the staged-file case explicitly — that's the behaviour that would have silently regressed if the reasoning had been wrong.

**`echo` → `printf`** — applied. For accuracy though: the practical impact is nil here, since `echo` only misreads a path named exactly `-n`/`-e`/`-E`, and no formatter's extension filter would select such a path anyway. Correct habit regardless, and free.

Also in this push, unrelated to the review: the `zizmor` `template-injection` failure at `presubmit-lint.yml:122`. `${{ steps.scope.outputs.args }}` was being interpolated into a `run:` body; it and the two `github.*` expressions now go through `env:` instead. Checked with zizmor 1.25.2 locally — reproduces on the previous revision, clean on this one.

polina-c:
I am rejecting your PR - our policy is to have all files formatted in a right way.

If formatting rules change upstream, someone should reformat all the files.

PR #2301, opened 2026-08-17
feat(python): regenerate and verify v0.9 pydantic models and basic catalog

nan-yu:
## Summary

Regenerates and verifies Pydantic v2 data models for the A2UI v0.9 protocol and basic catalog using the updated automated codegen skill, stacked on top of #2287.

## Changes
- Updated `.agents/skills/a2ui-generate-pydantic-models` to generate clean direct imports, explicit `__all__` re-exports, and `function_apis.py`.
- Regenerated Pydantic models in `python/a2ui_core/src/a2ui/core/schema/v0_9/` (`common_types.py`, `server_to_client.py`, `client_to_server.py`, `client_capabilities.py`, `constants.py`, `__init__.py`).
- Regenerated basic catalog schemas in `python/a2ui_core/src/a2ui/core/basic_catalog/v0_9/` (`components.py`, `function_apis.py`, `styles.py`, `__init__.py`).
- Updated unit test assertions in `test_codegen_pydantic.py` and `test_accessibility.py`.
- Verified formatting (`pyink`), static type checking (`mypy`), and full test suite (722 passed).

TAG=agy


.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
```diff
@@ -622,32 +623,131 @@ def generate_schema_init(
...
+        "A2uiClientMessage",
+        "ClientToServerMessage",
+        "RendererToAgentMessage",
+        "A2uiClientActionMessage",
+        "A2uiClientErrorMessage",
+        "A2uiClientAction",
+        "A2uiRendererAction",
+        "A2uiValidationError",
+        "A2uiGenericError",
+        "A2uiClientError",
```
gspencergoog: We've shifted terminology to "Renderer" instead of "Client".  Should we make those changes in this implementation? (not necessarily in this PR, although it looks like there's a switch for it here that we could use)

gspencergoog (approved):

.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
nan-yu: A2uiClientError is not defined in the spec explicitly, so I renamed to A2uiRenderError without an alias. Other old terminologies are defined as is with an alias, e.g. `A2uiRendererActionMessage = A2uiClientActionMessage`.

PR #2303, opened 2026-08-17
feat(python): generate Pydantic models and basic catalogs for A2UI spec v0.8 and v1.0

nan-yu:
## Description

This PR squashes and integrates the Pydantic data model and basic catalog generation for A2UI specifications **v0.8** and **v1.0** (squashing PR #2303 and PR #2308).

### Summary of Changes:
- **v0.8 Generation**:
  - `schema/v0_8/`: Constants, common types (baseline without accessibility/checks), client/server messages, client capabilities, and catalog definition.
  - `basic_catalog/v0_8/`: Component models and basic catalog registration.
- **v1.0 Generation**:
  - `schema/v1_0/`: Constants, common types (`FunctionCommon`, `IndexSystemFunction`, `Checkable`, `Surface`, `Extensions`), agent/renderer messages, capabilities, and catalog definition.
  - `basic_catalog/v1_0/`: Component models, function APIs, operator APIs (`@index`), and function implementations (`{"valid": bool}` validation).
- **Basic Catalog Architecture**:
  - Extracted shared utilities (`expression_parser.py`, `locale_config.py`) to `basic_catalog/`.
  - Maintained version-isolated `operator_apis.py` and `function_impls.py` in `v0_9/` and `v1_0/`.
- **Dynamic Codegen Generator**:
  - Dynamically compiles all `$defs` from `common_types.json` in topological order.
  - Supports `patternProperties` for extensible models (`Extensions`).
  - Auto-generates version-scoped imports in `components.py` and package `__init__.py`.

### Verification:
- `uv run pytest python/` (729 passed)
- `uv run --all-packages mypy .` (0 errors)
- `uv run pyink --check .` (clean)
- All skill tests and integration checks passed.

TAG=agy


.agents/skills/a2ui-generate-pydantic-models/scripts/codegen_pydantic.py
```diff
@@ -463,201 +474,200 @@ def validate_version_field(cls, v: Any, info: ValidationInfo) -> Any:
...
-        )
-
-    if "ValidationResult" in defs:
-        common_blocks.append(
-            codegen.compile_object_def("ValidationResult", defs["ValidationResult"])
-        )
+    if not defs:
+        # Baseline fallback for versions without common_types.json (e.g. v0.8)
+        common_blocks.extend([
+            """class DataBinding(StrictBaseModel):
```
gspencergoog: This seems fragile.  Right now, if we change the schema, then this needs to be regenerated/updated.  It would be ideal if we can codegen directly from the JSON without having to update this.

I did something similar in https://github.com/a2ui-project/a2ui/blob/v1_0_firing_ts/typescript/web_core/scripts/generate-zod-schemas.mjs for the web_core library.

It may be that it doesn't map well to Python, but we should see if there are some packages available that could help.

nan-yu: I removed some hard-coded code templates, and broke down the codegen_pydantic.py into separate modules. I also tried other python libraries ([datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator) or [json-schema-to-pydantic](https://github.com/richard-gyiko/json-schema-to-pydantic)). However, none of them works seamlessly with the custom script. Here is the rationale: https://github.com/a2ui-project/a2ui/pull/2285#issuecomment-5318591415.

gspencergoog (approved):

PR #2307, opened 2026-08-17
remove hardcoded url so mcp demo works in mcp inspector

wrenj:


sugoi-yuzuru (approved):
Can you add a comment that this should be changed for productionized version for better security posture?

wrenj:
> Can you add a comment that this should be changed for productionized version for better security posture?

Leaving as is since this is how MCP Apps does it 

https://github.com/modelcontextprotocol/ext-apps/blob/10195ad91851502134930e9b80ec2c04e277a720/src/message-transport.ts#L120

PR #2331, opened 2026-08-18
feat(swift): Core validation rules, reactivity, and type coercion

pinieb:
# Summary

This PR extends the `A2UICore` engine with support for validation check rules, improved type coercion according to the A2UI specification, synchronous state update notifications, and schema classification for numbers, integers, and dynamic string lists.

This is part 1 of 4 in the series implementing the Swift Basic Catalog.

# Changes

## Validation and checks
* Added `ResolvedCheck.swift` defining the `ResolvedCheck` model for evaluating boolean conditions and failure messages.
* Added `checks`, `validationErrors`, and `isValid` convenience properties to `Node.swift`.
* Added action guard validation in `SurfaceViewModel.swift` to prevent event dispatch and emit a `.validationFailed` error when node checks fail.

## Reactivity and state management
* Added `dataDidChange` (`CurrentValueSubject<JSONValue, Never>`) to `DataModel.swift` and `componentsDidChange` to `SurfaceComponentsModel.swift` to enable synchronous tree recomputation upon state mutations.
* Added support for root JSON pointer paths (`""` and `"/"`) in `DataModel.get` and `DataModel.set`.

## Schema classification and type coercion
* Added schema classification for `DynamicStringList`, `CheckRule`, `Checkable`, `ComponentId`, `number`, and `integer` types in `SurfaceViewModel.swift`.
* Added recursive property schema extraction across `allOf`, `anyOf`, and `oneOf` blocks.
* Added `coerceToString` in `SurfaceViewModel.swift` to convert numbers and booleans to strings according to the protocol specification (§Type conversion).
* Implemented `resolveDynamicStringList` for string array bindings.

## Cleanup and conformances
* Added `Equatable` conformances for resolved types in `Resolved+Conformances.swift` and `Node`.
* Removed redundant `MessageProcessor` dictionary initializer in `MessageProcessor.swift` and migrated test callers.

# Verification

Run the test suite with:
```bash
swift test
```

Includes unit tests covering:
* Validation errors reflection and action suppression on validation failure in `SurfaceViewModelTests.swift`.
* String coercion from numbers and booleans.
* Dynamic string list resolution and mutations.
* `Node` equatable conformances in `ModelTests.swift`.
* All 396 existing core and integration tests.


swift/core/Sources/A2UICore/Models/ResolvedCheck.swift
```diff
@@ -0,0 +1,55 @@
...
+
+  public static func == (lhs: ResolvedCheck, rhs: ResolvedCheck) -> Bool {
+    lhs.condition == rhs.condition && lhs.message == rhs.message
+  }
+}
+
+extension Node {
+  /// All resolved validation checks for this node.
+  public var checks: [ResolvedCheck] {
+    (properties["checks"] as? [ResolvedCheck]) ?? []
```
jacobsimionato: I believe the intention in the specification is not to special case "checks" based on property name, but instead based on whether the property is a CheckRule based on this schema - https://github.com/a2ui-project/a2ui/blob/7541f953050cd58b80f0bf5d85fe2d63192af305/specification/v0_9/json/common_types.json#L233

I'm sorry for the slow review though. I'm also okay if you submit this as-is and then fix it later.

swift/core/Sources/A2UICore/State/SurfaceComponentsModel.swift
```diff
@@ -24,35 +24,53 @@ import Foundation
 public final class SurfaceComponentsModel: @unchecked Sendable, ObservableObject {
```
jacobsimionato, line 24: I thought we were moving towards just having all the A2UI stuff run on the main queue, to avoid having to make the classes typesafe and accessible from any queue via locks. Is that the case? It'd be great to be clear about the threading model and the rationale for it! Probably it's written somewhere but I missed it!

jacobsimionato (approved):
Sorry for the delay!!

swift/core/Sources/A2UICore/State/SurfaceComponentsModel.swift
pinieb: I haven't done the full pass to update everything yet. I think there are still two classes (DataModel and SurfaceComponentsModel) that need to move over. I left a comment on https://github.com/a2ui-project/a2ui/issues/2124 to call these out specifically.

I'm planning to do a pass through the outstanding issues once the Basic Catalog lands :)

swift/core/Sources/A2UICore/Models/ResolvedCheck.swift
pinieb: Ah I missed this. The agent really likes adding these convenience methods even when I tell it not to do it. It seems to think that because the Basic Catalog (mostly) uses one key (for example, "children") that it should definitely 100% have a convenience method that it then treats as a structural rule 🙃

I've updated the code to not look for the "checks" key and instead rely on the schema type.

PR #2333, opened 2026-08-19
feat(swift): validate surface theme against catalog themeSchema in MessageProcessor

pinieb:
## Summary

Implements JSON Schema validation for the `theme` dictionary on `CreateSurfaceMessage` in `MessageProcessor`.

When an incoming `CreateSurfaceMessage` includes a `theme` object and the surface catalog defines a `themeSchema`, `MessageProcessor` now validates the theme instance against that schema. If validation fails, a `ValidationFailedError` is constructed with the specific JSON Pointer path and error description, preventing surface creation and dispatching the error to `actionHandler`. When `themeSchema` is `nil` or no theme is provided, validation passes unconditionally.

## Changes

* `MessageProcessor.swift`: Added `themeSchema` validation in `validateAndProcess(.createSurface)` and `createSurface`, including error path extraction.
* `MessageErrorMapper.swift`: Added direct mapping support for `ValidationFailedError` to `ClientServerError.validationFailed`.
* `MessageProcessorTests.swift`: Added unit tests for valid theme against schema, invalid theme against schema (throwing and non-throwing), and theme with nil schema.
* `MessageErrorMapperTests.swift`: Added test verifying `ValidationFailedError` mapping.

## Testing

* `swift-format lint`: Passed with 0 formatting or linting errors.
* `swift test`: 388 unit tests pass across 41 test suites.

### DIFF
--- a/swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
+++ b/swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
@@ -255,34 +255,29 @@ public final class MessageProcessor: ObservableObject {
     }
   }
 
-  // MARK: - Message Processing (Strongly-Typed)
+  // MARK: - Message Processing
 
-  /// Processes a single server-to-client message, throwing validation or missing surface errors.
+  /// Processes a single server-to-client message.
   ///
-  /// Errors thrown during validation are reported to `ActionHandling` via `MessageErrorMapper`
-  /// before being rethrown.
-  public func process(message: ServerToClientMessage) throws {
+  /// Any validation or lifecycle errors are mapped via `MessageErrorMapper`
+  /// and reported to `ActionHandling`.
+  public func process(message: ServerToClientMessage) {
     do {
       try validateAndProcess(message)
     } catch {
       let surfaceID = extractSurfaceID(from: error, fallback: message.surfaceID)
       let clientError = errorMapper.map(error, surfaceID: surfaceID)
       actionHandler?.handle(error: clientError, from: surfaceID)
-      throw error
     }
   }
 
-  /// Processes a single server-to-client message without throwing.
-  ///
-  /// Errors are routed directly to `ActionHandling`.
-  public func processMessage(_ message: ServerToClientMessage) {
-    _ = try? process(message: message)
-  }
-
   /// Processes an array of server-to-client messages.
-  public func processMessages(_ messages: [ServerToClientMessage]) {
+  ///
+  /// Any validation or lifecycle errors are mapped via `MessageErrorMapper`
+  /// and reported to `ActionHandling`.
+  public func process(messages: [ServerToClientMessage]) {
     for message in messages {
-      processMessage(message)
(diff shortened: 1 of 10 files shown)


swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
jacobsimionato: Thanks for adding this!

Gemini and I had one architectural suggestion regarding where the error handling and `actionHandler` reporting live:

### 1. Moving validation error handling to `process(message:)`
Currently, the error inspection, mapping via `MessageErrorMapper`, and reporting to `actionHandler?.handle(error:from:)` are scoped inside `process(line: String)`. 

If a caller uses the strongly-typed entry points (`process(message: ServerToClientMessage)` or `processMessages(_:)`), any `ValidationFailedError` or `GenericError` thrown during theme/component validation will bypass `actionHandler` reporting (and in `processMessage(_:)`, `try?` will silently swallow it).

Could we move the validation error handling into the canonical `process(message:)` (or wrap `validateAndProcess`) so that all processing paths consistently report client validation errors back to the server? For example:

```swift
public func process(message: ServerToClientMessage) throws {
  do {
    try validateAndProcess(message)
  } catch {
    let surfaceID = extractSurfaceID(from: error, fallback: message.surfaceID)
    let clientError = errorMapper.map(error, surfaceID: surfaceID)
    actionHandler?.handle(error: clientError, from: surfaceID)
    throw error
  }
}
```

### 2. Supporting `process(line: String)` cleanly alongside the blueprint
In the `a2ui_core` blueprint, `MessageProcessor` is designed around consuming strongly-typed message objects (`processMessages`), keeping string/JSONL decoding decoupled in parser classes (like `MessageParser`).

That said, having `process(line: String)` directly on `MessageProcessor` clearly solves a real, practical ergonomics use case for streaming and transport layers in Swift. To support this cleanly while keeping the core processing pipeline consistent, we could have `process(line: String)` handle only string-decoding errors (`MessageParseError`) and then delegate straight to `process(message:)`:

```swift
public func process(line: String) throws {
  let message: ServerToClientMessage
  do {
    message = try parser.parse(jsonString: line)
  } catch {
    let surfaceID = (error as? MessageParseError)?.surfaceID ?? "unknown"
    let clientError = errorMapper.map(error, surfaceID: surfaceID)
    actionHandler?.handle(error: clientError, from: surfaceID)
    throw error
  }

  // Delegates to canonical message processing (which handles validation error reporting)
  try process(message: message)
}
```

What do you think?

jacobsimionato (approved):
Approving so you can submit in your timezone

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
pinieb: I've been meaning to get rid of `process(line:)` for a while now. At best, I think it adds unnecessary complexity around exposing parsing errors to the client for them to figure out how to handle without adding value to that process.

It has been removed! 🎉

PR #2337, opened 2026-08-19
feat(python): a2a-sdk 0.3/1.x dual-compat Part helpers

WillBeebe:
Hooman here. 👋  

I've been using A2UI a lot, it's lovely. This PR is an attempt to bring A2UI to a wider audience. Thank you for your consideration.

// below is an llm summary

Today `a2ui-agent-sdk` pins `a2a-sdk>=0.3.0,<0.4.0` (https://github.com/a2ui-project/a2ui/blob/main/agent_sdks/python/a2ui_agent/pyproject.toml#L22). ADK 2.x already accepts a2a-sdk>=0.3.4,<2, and Agent Engine / A2A protocol 1.0 stacks ship a2a-sdk 1.x. That <0.4 pin (plus 0.3-only Part(root=DataPart) construction) is what blocks those consumers.

This PR:
• adds a2ui.a2a._compat so the same helpers work on a2a-sdk 0.3 (pydantic DataPart / TextPart) and 1.x (protobuf Part)
• widens the published pin to a2a-sdk>=0.3.0,<2 while keeping this repo’s CI on 0.3 until samples drop Part.root

That's Python-type dual-compat (same idea as ADK’s _compat). It is not a2a-sdk’s enable_v0_3_compat / supported_interfaces wire layer — servers still opt into that separately.

We've run the library tests here on a2a-sdk 0.3.25 + ADK 2.3.0 (269 passed). Outside this repo we've used the 1.x-native helpers heavily with A2UI 0.8 / 0.9 and Gemini Enterprise over A2A. Sample rewrites are a follow-up (ready).

Here's a decent LLM summary the goes over more of the details. Thank you for your consideration.

A2A SDK **1.x** is the native protocol-1.0 stack. Staying on 0.3 is staying on a compat generation. `_compat` is what lets `a2ui-agent-sdk` run on **either** without rewriting samples yet.

## Why 1.x vs staying on 0.3

A2UI is not an ADK or model feature. It rides on A2A: the agent card advertises the A2UI extension, the client activates it, the executor emits **data Parts** (`application/json+a2ui` / `application/a2ui+json`). If those Parts cannot be built or read on the SDK people actually install, generative UI is blocked.

`a2a-sdk` 1.x is that stack:

| | **0.3** | **1.x** |
|---|---|---|
| Spec | A2A 0.3 | A2A **1.0** (proto-first, ProtoJSON) |
| Types | Pydantic `Part(root=DataPart\|TextPart)` | Flat protobuf `Part` (`text` / `data` / `raw` / `url` oneof) |
| Card | single `url` | `supported_interfaces` (JSON-RPC + REST + gRPC) |
| Server | `A2AStarletteApplication` wrapper | `create_jsonrpc_routes` / `create_rest_routes` you mount into Starlette/FastAPI |
| Extension activation | `RequestContext.add_activated_extension` | `call_context.state["activated_extensions"]` |
| Extension params | dict constructor | protobuf `Struct` via `.update()` |

That is not cosmetic. ADK 2.x’s `[a2a]` extra already accepts `a2a-sdk>=0.3.4,<2` and **tells new integrations to target 1.x**. Agent Engine A2A stacks that ship ADK 2.x + `a2a-sdk` 1.x already run that combo. This package was still pinned `<0.4` and constructed 0.3-only types, so it **cannot import** next to them:

- `Part(root=DataPart(...))` / `Part(root=TextPart(...))`
- `RequestContext.add_activated_extension`
- `AgentExtension(params=dict)`

Without this change, anyone on that stack has to vendor the SDK. With it, the published package can sit in a 1.x environment.

1.x also gives real product surface that 0.3 does not: first-class gRPC, REST, stricter streaming (Task-first vs one Message), and one JSON-RPC mount that still speaks 0.3 on the wire via `enable_v0_3_compat=True`. That last piece is **a2a-sdk’s** protocol compat, not this PR.

## Two compat layers (easy to mix up)

**1. a2a-sdk `enable_v0_3_compat`** — *wire protocol*. A 1.x **server** still accepts 0.3 JSON-RPC/REST clients. Internal types are protobuf; the SDK translates on the boundary. JS `@a2a-js/sdk` `legacyCompat: {enabled: true}` is the client twin.

**2. `a2ui.a2a._compat` (this PR)** — *Python types*. A2UI library code used to do `Part(root=DataPart(...))`. That constructor does not exist on 1.x. `_compat` detects the installed generation (`StreamResponse` is 1.x-only) and builds/reads the right shape.

Same idea as ADK’s own `_compat`: one public API, two SDK generations.

What it actually covers:

- **Construct:** `make_text_part` / `make_data_part` / `make_agent_extension`
- **Read:** `is_data_part`, `data_part_dict`, `part_metadata`, `part_media_type`
- **Activate:** `add_activated_extension` — 0.3 method if present, else `call_context.state`
- **Bridge old callers:** `data_part_view` so `get_a2ui_datapart(...).metadata.get(...)` still works on 1.x

Public helpers (`create_a2ui_part`, `is_a2ui_part`, `get_a2ui_data`, `extract_user_action`, `try_activate_a2ui_extension`) all go through that. Callers stop touching `part.root`.

## Why this PR shape

This repo cannot jump to 1.x-only today: restaurant finder and other samples still use `Part.root`. So this PR:

1. Widens the **published** pin to `a2a-sdk>=0.3.0,<2`
2. Keeps workspace **CI on 0.3** via `constraint-dependencies` until samples migrate
3. Leaves sample rewrites out of scope

That is conservative and mergeable. Consumers who already run 1.x get a working SDK; this repo’s samples keep passing.

Extra helpers that are useful even on 0.3:

- `extract_user_action` — mime-tagged **and** untagged inbound `userAction` (clients often omit the A2UI mime on clicks)
- `A2uiPartConverter` accepts ADK 2.x converters that return a **list** of Parts
- `A2uiSubagentMap` no longer inspects `part.root`

Out of scope on purpose (and should stay out): Gemini Enterprise / Agent Engine last-mile, Cloud Run normalize, Python floor bump. Follow-up can migrate samples to `extract_user_action` and drop the 0.3 branch later.

## What

- Add `a2ui.a2a._compat` for Part / AgentExtension construction and reads.
- Route `create_a2ui_part`, `is_a2ui_part`, `get_a2ui_datapart`, parse helpers, and `try_activate_a2ui_extension` through it.
- New helpers: `get_a2ui_data`, `part_data_as_dict`, `extract_user_action`.
- `A2uiSubagentMap` no longer inspects `part.root`.
- `A2uiPartConverter` accepts ADK 2.x converters that return a list of Parts.
- Widen published pin to `a2a-sdk>=0.3.0,<2`. Workspace `constraint-dependencies` keeps CI on 0.3 until samples stop using `Part.root`.

## Out of scope

Restaurant finder / other samples still use `part.root`. Follow-up PR can switch them to `extract_user_action`.

## Test

```bash
uv run --package a2ui-agent-sdk pytest \
  agent_sdks/python/a2ui_agent/tests/a2a \
  agent_sdks/python/a2ui_agent/tests/adk \
  agent_sdks/python/a2ui_agent/tests/conformance -q
```

269 passed (a2a-sdk 0.3.25, google-adk 2.3.0).


agent_sdks/python/a2ui_agent/src/a2ui/a2a/_compat.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def part_metadata(part: Part) -> dict[str, Any]:
+    """Reads a Part's metadata as a plain dict."""
+    if IS_A2A_V1:
+        from google.protobuf.json_format import MessageToDict
+
+        if part.HasField("metadata"):
+            return MessageToDict(part.metadata)
+        return {}
```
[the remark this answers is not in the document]
WillBeebe: Are you suggesting theses three lines be MessageToDict(part.metadata)

WillBeebe:
CLA updated.

WillBeebe:
Follow-up stack (branched from this PR):

1. **[#2363](https://github.com/a2ui-project/a2ui/pull/2363)** — sibling CI job that actually runs `a2ui-agent-sdk` on **a2a-sdk 1.x**. Isolated venv (workspace `uv sync --all-packages` cannot resolve 1.x while sample packages still pin `<0.4`). Pytest is `tests/a2a` through `_compat`. Also drops `HasField` on protobuf map metadata.

2. **[#2364](https://github.com/a2ui-project/a2ui/pull/2364)** — first-party ADK samples (restaurant finder, custom-components-example) stop using `Part.root`. Community samples and the 0.3 workspace constraint stay for a later cut.

`pack_specs_hook.py` is **not** in this stack: it is build hygiene (bundled schema / skip ANTLR regen when assets are already present), not the 1.x Part-type gap.

WillBeebe:
Hi @polina-c

Curious what you think.

polina-c:
Would you mind to create an issue separately, that would explain problem that this PR is solving, and link the issue in the first line of description?

\+ @nan-yu , who is more familiar with Python codebase.

WillBeebe:
@polina-c yes no problem at all. I'll take care of that. :)

jacobsimionato:
Assigning to @nan-yu to take a look at this. In principle, it seems great, but also we are looking at decoupling the A2A integration from our SDKs, so this may not be aligned with our roadmap.

If that is the case, either we might want to accept this but release a separate agent SDK A2A integration package, or else encourage the community to build and maintain such an integration package.

agent_sdks/python/a2ui_agent/src/a2ui/a2a/_compat.py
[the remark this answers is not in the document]
nan-yu: My review agent suggested that `part.metadata` is a `ScalarMapContainer`, not a `google.protobuf.message.Message`. Passing it to `MessageToDict()` throws:
`AttributeError: 'google._upb._message.ScalarMapContainer' object has no attribute 'DESCRIPTOR'`.

## Fix

Use `dict(part.metadata)` for map fields, and add a `None` guard:
```python
def part_metadata(part: Part) -> dict[str, Any]:
    """Reads a Part's metadata as a plain dict."""
    if part is None:
        return {}
    if IS_A2A_V1:
        meta = getattr(part, "metadata", None)
        if meta is None:
            return {}
        if hasattr(meta, "DESCRIPTOR"):  # In case metadata is a Struct Message
            from google.protobuf.json_format import MessageToDict
            return MessageToDict(meta)
        return dict(meta)
    return getattr(getattr(part, "root", None), "metadata", None) or {}
```

## Reproducing Unit Test

```Python
def test_part_metadata_v1_raises_attribute_error_on_map_container(monkeypatch):
    """Demonstrates that MessageToDict fails on protobuf map containers."""
    monkeypatch.setattr(_compat, "IS_A2A_V1", True)

    # In protobuf Python, map fields are ScalarMapContainer objects (dict-like without DESCRIPTOR)
    class MockScalarMapContainer(dict):
        pass

    mock_proto_part = SimpleNamespace(
        WhichOneof=lambda field: "data" if field == "content" else None,
        metadata=MockScalarMapContainer({"mimeType": "application/a2ui+json"}),
        media_type="",
    )

    # Calling part_metadata (or is_a2ui_part) triggers MessageToDict(part.metadata)
    with pytest.raises(AttributeError, match="DESCRIPTOR"):
        _compat.part_metadata(mock_proto_part)


def test_part_metadata_v1_none_raises_attribute_error(monkeypatch):
    """Demonstrates that part_metadata(None) crashes on 1.x without a None guard."""
    monkeypatch.setattr(_compat, "IS_A2A_V1", True)

    with pytest.raises(AttributeError):
        _compat.part_metadata(None)
```

agent_sdks/python/a2ui_agent/src/a2ui/a2a/_compat.py
```diff
@@ -0,0 +1,182 @@
...
+    if part is None:
+        return False
+    if IS_A2A_V1:
+        return part.WhichOneof("content") == "text"
+    from a2a.types import TextPart  # type: ignore[attr-defined]
+
+    return isinstance(getattr(part, "root", None), TextPart)
+
+
+def part_text(part: Part) -> str:
```
nan-yu, line 90: Other methods check if Part is None, should we also check it here?

agent_sdks/python/a2ui_agent/src/a2ui/a2a/_compat.py
```diff
@@ -0,0 +1,182 @@
...
+    if IS_A2A_V1:
+        from google.protobuf.json_format import MessageToDict
+
+        data = MessageToDict(part.data)
+        return data if isinstance(data, dict) else None
+    data = part.root.data
+    return data if isinstance(data, dict) else None
+
+
+def part_media_type(part: Part) -> Optional[str]:
```
nan-yu, line 119: Checking if Part is None here?

agent_sdks/python/a2ui_agent/src/a2ui/a2a/_compat.py
```diff
@@ -0,0 +1,182 @@
...
+    if callable(add_activated):
+        add_activated(uri)
+        return
+    call_context = getattr(context, "call_context", None)
+    if call_context is None:
+        return
+    state = getattr(call_context, "state", None)
+    if state is None:
+        return
+    activated = state.setdefault("activated_extensions", set())
```
nan-yu, line 162: If `state["activated_extensions"]` was already initialized as a `list` by the server framework or middleware, `isinstance(activated, set)` evaluates to `False` and `state["activated_extensions"] = {uri}` overwrites the existing list, dropping previously activated extensions.

Fix: Handle both set and list:
```python
activated = state.get("activated_extensions")
if isinstance(activated, set):
    activated.add(uri)
elif isinstance(activated, list):
    if uri not in activated:
        activated.append(uri)
else:
    state["activated_extensions"] = {uri}
```

PR #2349, opened 2026-08-20
feat(web_core): add A2uiValidator and topology/integrity validation for Python parity

gspencergoog:
## Summary
Adds validation modules (`A2uiValidator`, integrity checking, recursion and path validation, and topology analysis) to TypeScript `@a2ui/web_core` to achieve parity with the Python validation implementation. Integrates validation routines into `MessageProcessor` and `SurfaceComponentsModel`, adding support for mixed catalogs, schema-driven reference extraction, and pre-mutation atomic updates.

## Changes
- **Validating Module (`typescript/web_core/src/validating/`)**:
  - Added `A2uiValidator` in `validator.ts` for validating message envelopes across protocol versions and component hierarchies (`validateProtocolEnvelope`, `validateComponents`).
  - Added `validateComponentIntegrity` in `integrity-checker.ts` to detect duplicate IDs, missing roots, and dangling child references across single or mixed catalogs (`CatalogOrRefMapInput`).
  - Added `analyzeChildRefSchema`, `isChildSchema`, `isChildListSchema`, and `isChildOrChildListSchema` to identify child and child-list references dynamically from Zod schemas and JSON Schema `$ref` definitions (`ComponentId`, `ChildList`).
  - Added `buildComponentRefMap` with `WeakMap` caching (`getOrCreateRefMap`) to map component types to single and list reference properties.
  - Added `getComponentReferences` generator to extract child IDs from scalar fields, arrays, dynamic template objects (`{componentId, path}`), and `{child}` structures.
  - Added `validateRecursionAndPaths` to enforce global structure depth (`MAX_GLOBAL_DEPTH = 50`), function-call depth (`MAX_FUNC_CALL_DEPTH = 5`), and RFC 6901 JSON pointer syntax (`RELAXED_PATH_PATTERN`).
  - Added `analyzeTopology` in `topology-analyzer.ts` to detect cycles, self-references, and unreachable orphan components.
  - Added `ValidationConfig` presets (`STRICT_VALIDATION` and `RELAXED_VALIDATION`).
- **State & Error Handling (`typescript/web_core/src/state/`, `errors.ts`)**:
  - Added `A2uiIntegrityError` (`INTEGRITY_ERROR`) and `A2uiRecursionError` (`RECURSION_ERROR`) extending `A2uiValidationError`.
  - Added `detectCycles`, `validateTopology`, `validateReferences`, `getChildReferences`, `getChildIds`, `setCatalog`, `getAll`, and `has` to `SurfaceComponentsModel`.
  - Added optional `catalog` association to `ComponentModel` and `SurfaceComponentsModel` to support multi-catalog surfaces.
- **Message Processing Integration (`typescript/web_core/src/processing/`)**:
  - Integrated `validateRecursionAndPaths` and `validateCandidateTopology` into `MessageProcessor.processMessages` to validate candidate component updates before mutating state.
  - Added per-component `catalogId` support in `MessageProcessor`, resolving schemas and reference properties against the target catalog.
  - Updated component updates to replace properties instead of shallow merging existing properties.
  - Updated `validateCompositionConstraints` to resolve child references using schema introspection.
  - Standardized version adapter class naming to PascalCase (`V0Point8Adapter`, `V0Point9Adapter`, `V1Point0Adapter`) in `adapters/factory.ts`.
- **Validation Functions & Consumers (`src/v1_0/functions/`, `samples/`, `tools/`)**:
  - Standardized naming of v1.0 validation function APIs and implementations (`RequiredV1Point0Api`, `RegexV1Point0Api`, `LengthV1Point0Api`, `NumericV1Point0Api`, `EmailV1Point0Api`).
  - Added per-component catalog resolution in `NodeResolver`.
  - Added explicit generic type parameters (`ReactComponentImplementation`) to `MessageProcessor` instances in `samples/client/react/shell/src/App.tsx` and `tools/composer/src/lib/v09Viewer.tsx`.
- Addresses issue 2297.

## Impact & Risks
- **Breaking Changes**:
  - Component updates in `updateComponents` now replace existing properties instead of shallow merging them. Updates that omit required properties defined in the catalog schema will fail validation.
  - Unknown component types are rejected when `allowUnknownElements` is `false` (`STRICT_VALIDATION` default).
  - Version adapter classes and v1.0 validation function constants were renamed (for example, `V1_0VersionAdapter` to `V1Point0Adapter`, `RequiredV1_0Api` to `RequiredV1Point0Api`).
- **Performance**:
  - Schema introspection results are cached in a `WeakMap` keyed by `Catalog`, avoiding repeated reflection overhead.
  - Pre-mutation candidate topology validation runs DFS graph traversal in $O(V + E)$ bounded by a maximum depth of 50.
- **Dependencies**: No new external dependencies added.

## Testing
- **Verification Commands**:
  - `yarn --cwd typescript/web_core test`
- **New Test Coverage**:
  - `typescript/web_core/src/validating/validator.test.ts`: Unit tests covering child reference extraction, duplicate ID detection, missing root detection, dangling references, path regex parsing, recursion limits, cycle detection, orphan detection, mixed catalog validation, and JSON Schema `$ref` introspection.
  - `typescript/web_core/src/processing/message-processor.test.ts`: Integration tests for mixed catalogs on single surfaces, multi-surface payloads, inline `createSurface.components`, relaxed vs strict validation modes, full property replacement on updates, composition constraints, and pre-mutation rollback on validation failures.
  - `typescript/web_core/src/state/surface-components-model.test.ts`: Unit tests for inlined topology analysis, cycle detection, self-reference detection, recursion depth limits, missing root handling, and dangling references.


typescript/web_core/src/validating/integrity-checker.ts
```diff
@@ -0,0 +1,239 @@
...
+    throw new A2uiIntegrityError(`Missing root component: No component has id='${rootId}'`);
+  }
+
+  if (allowDanglingReferences) {
+    return;
+  }
+
+  // 3. Check for dangling references
+  for (const comp of components) {
+    const compId = comp.id !== undefined && comp.id !== null ? String(comp.id) : 'Unknown';
```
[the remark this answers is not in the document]
gspencergoog: Added defensive check ensuring comp is a non-null object before accessing comp.id.

typescript/web_core/src/validating/integrity-checker.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  if (typeof item === 'object' && item !== null) {
+    if ('path' in item && typeof item.path === 'string') {
+      const path = item.path;
+      if (!RELAXED_PATH_PATTERN.test(path)) {
+        throw new A2uiValidationError(`Invalid path syntax: '${path}'`);
+      }
+    }
+
+    const isFuncV08 = 'functionCall' in item && typeof item.functionCall === 'object';
```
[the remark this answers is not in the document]
gspencergoog: Added non-null check for functionCall object.

typescript/web_core/src/validating/validator.ts
```diff
@@ -0,0 +1,161 @@
...
+    components: Array<Record<string, any>>,
+    refFieldsMap: ComponentRefMap = STANDARD_REF_MAP,
+    config: ValidationConfig = STRICT_VALIDATION,
+    options: ValidateComponentsOptions = {},
+  ): void {
+    validateComponentIntegrity(components, refFieldsMap, config);
+    analyzeTopology(components, refFieldsMap, config);
+    if (!options.skipRecursionCheck) {
+      validateRecursionAndPaths(components);
+    }
```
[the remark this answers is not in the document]
gspencergoog: Moved recursion check ahead of integrity and topology checks to fail fast on deep structures.

typescript/web_core/src/validating/integrity-checker.ts
```diff
@@ -0,0 +1,512 @@
...
+  return res.isChild || res.isChildList;
+}
+
+/**
+ * Builds a ComponentRefMap dynamically by inspecting component Zod schemas.
+ *
+ * @param catalogOrComponents Catalog instance, array of ComponentApi objects, or Map of ComponentApis.
+ * @returns ComponentRefMap containing single and list reference properties.
+ */
+export function buildComponentRefMap(
```
nan-yu, line 251: One comment I got from Jacob when working on the API design doc was this component ref map maintains duplicate reference pointer extraction algorithms because MessageProcessor along with SurfaceComponentsModel also maintains a similar component reference graph. 

The new proposal is to inline topology and cycle checks directly into SurfaceComponentsModel during message processing. More details are documented in https://docs.google.com/document/d/17BuULV88IlsUStOFthGMgfeW1XMtLUyQDm3oflgXi-k/edit?resourcekey=0-fltDWdFS4VEdDHSTzq--QA&tab=t.lcmh7rwipp8t#bookmark=id.y3mg7magugtk.

The Python version also needs to be updated.

gspencergoog: Okay I addressed this.  I made `SurfaceComponentsModel` the single source of truth for graph traversal and validation.

typescript/web_core/src/state/surface-components-model.ts
```diff
@@ -33,8 +64,29 @@ export class SurfaceComponentsModel {
...
+
+  /**
+   * Associates a catalog or reference map with this model for schema-driven child detection.
+   *
+   * @param catalogOrRefMap Catalog instance or ComponentRefMap.
+   */
+  setCatalog(catalogOrRefMap: Catalog<any, any> | ComponentRefMap): void {
+    this.catalogOrRefMap = catalogOrRefMap;
+    this.refMap =
+      catalogOrRefMap instanceof Catalog ? buildComponentRefMap(catalogOrRefMap) : catalogOrRefMap;
```
nan-yu: It builds the componentRefMap with a single catalog. With the mixed-catalogs support, a surface may include components from multiple catalogs.

We need to build the map with the corresponding catalog for each component separately.

typescript/web_core/src/state/surface-components-model.ts
```diff
@@ -16,13 +16,44 @@
...
+  ComponentRefMap,
+  getComponentReferences,
+  MAX_GLOBAL_DEPTH,
+} from '../validating/integrity-checker.js';
 
 /**
- * Manages the collection of components for a specific surface.
+ * Configuration options for surface component topology validation.
+ */
+export interface TopologyValidationConfig {
```
nan-yu: Can we reuse the existing ValidationConfig and add any missing properties there?

typescript/web_core/src/state/surface-components-model.ts
```diff
@@ -16,13 +16,44 @@
...
+  maxDepth?: number;
+}
+
+/**
+ * Manages the collection of components for a specific surface and performs
+ * inlined graph topology, cycle detection, and completeness validation.
  */
 export class SurfaceComponentsModel {
   private components: Map<string, ComponentModel> = new Map();
+  private catalogOrRefMap?: Catalog<any, any> | ComponentRefMap;
```
nan-yu: Because a surface may contain components from multiple catalogs, I added the Catalog reference to the ComponentModel in the python version and run schema validation on each component with the corresponding catalog.

typescript/web_core/src/state/surface-components-model.ts
```diff
@@ -16,13 +16,44 @@
...
+  private refMap?: ComponentRefMap;
```
nan-yu, line 40: Do we need both catalogOrRefMap and refMap?

typescript/web_core/src/validating/validator.ts
```diff
@@ -0,0 +1,175 @@
...
+   * @throws {A2uiValidationError} If envelope format or path syntax is invalid.
+   * @throws {A2uiIntegrityError} If component references or graph integrity are violated.
+   * @throws {A2uiRecursionError} If recursion limits or circular component links are detected.
+   *
+   * @example
+   * ```ts
+   * validator.validate(payloadMessage, catalog);
+   * ```
+   */
+  public validate(
```
nan-yu: I'm wondering if we really need a standalone, top-level entry point for all validations (schema checks, topology checks, etc.).

In the Python implementation, I removed the standalone `validate()` method and instead made `MessageProcessor.processMessages()` the single entry point. Individual sub-validations are executed internally as part of the processing pipeline rather than via a separate top-level pass.

What do you think?

gspencergoog: The only thing I worry about is that without a top level validation step, the individual validations will validate their portion, but not the entire thing: like the difference between a unit test and an integration test.  For example, we might validate the topology of each piece, but put it all together and it might still produce an invalid topology overall.

nan-yu: MessageProcessor will handle full component topology checks to ensure overall validity. Instead of having a separate validator, we can make MessageProcessor the primary entry point for top-level validation.

typescript/web_core/src/state/surface-components-model.ts
gspencergoog: No, removed `catalogOrRefMap`. Thanks!

typescript/web_core/src/state/surface-components-model.ts
gspencergoog: Yes.  Done in 5f827b1aa5ca033af8982b593e75a9c8376587c9

typescript/web_core/src/validating/validator.ts
gspencergoog: Okay, sure.  Done.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -537,19 +506,36 @@ export class MessageProcessor<T extends ComponentApi> {
...
+      if (typeof rawCatalogId === 'string' && rawCatalogId) {
+        const found = this.catalogs.find(c => c.id === rawCatalogId);
+        if (!found) {
+          throw new A2uiValidationError(
+            `Unknown catalog ID '${rawCatalogId}' for component '${id}'. Available catalogs: ${this.catalogs.map(c => c.id).join(', ')}`,
+          );
+        }
+        targetCatalog = found;
+      } else {
+        const existing = surface.componentsModel.get(id);
```
nan-yu: For a component in UpdateComponent message, if catalogId is not specified, should we use the default catalogId in the createSurface message, instead of the existing component.catalogId?

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -599,225 +599,7 @@ export class MessageProcessor<T extends ComponentApi> {
...
-      if (!allowOrphanComponents) {
-        for (const comp of components.values) {
-          if (!visited.has(comp.id)) {
-            throw new A2uiValidationError(
-              `Orphaned component '${comp.id}' is not reachable from root`,
-            );
-          }
-        }
-      }
+      surface.componentsModel.validateTopology(this.validationConfig);
```
nan-yu: It validates the topology after updating the state (componentsModel). If the validation failed, the state has been updated already. Should we validate before update?

typescript/web_core/src/state/surface-components-model.ts
```diff
@@ -92,6 +153,177 @@ export class SurfaceComponentsModel {
     }
   }
 
+  /**
+   * Extracts child component references for a given component identifier.
+   *
+   * @param componentId The ID of the component whose children should be resolved.
+   * @returns Array of tuples containing `[referencedId, propertyPath]`.
+   */
+  getChildReferences(componentId: string): Array<[referencedId: string, propertyPath: string]> {
```
nan-yu, line 162: This looks similar to `extractChildIds` in MessageProcessor. We may want to consolidate them into a single implementation.

Ideally, if we go with extractChildIds, we don't need the componentRef map.

typescript/web_core/src/validating/index.ts
nan-yu, line 1: Jacob also suggested to name this folder to be `validation`, so we will have
- processing
- validation
- resolution (rename `rendering` to `resolution`)

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Yes, I think that's right.  No `catalogId` is equivalent to specifying the default catalog ID.

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Yes.  Fixed.

typescript/web_core/src/validating/index.ts
gspencergoog: Okay, let's rename in another PR.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -399,6 +368,10 @@ export class MessageProcessor<T extends ComponentApi> {
   processMessages(messages: unknown): void {
     if (!messages) return;
 
+    if (this.validationConfig) {
+      validateRecursionAndPaths(messages);
+    }
+
     if (this.validationConfig?.targetVersion) {
       this.validateTargetVersion(messages);
```
nan-yu, line 376: Thanks for spotting this change in the Python PR. I remove target_version from the python code because we allow multiple catalogs in a surface, which may come from multiple protocol versions. It no longer makes sense to have a single target version.

typescript/web_core/src/state/component-model.ts
```diff
@@ -34,11 +35,13 @@ export class ComponentModel {
    * @param id The unique identifier for this component.
    * @param type The component type name.
    * @param initialProperties The initial properties for the component.
+   * @param catalog Optional Catalog associated with this component.
    */
   constructor(
     readonly id: string,
     readonly type: string,
     initialProperties: Record<string, any>,
+    readonly catalog?: Catalog<any, any>,
```
nan-yu, line 44: nit: is the catalog optional? A component should always have a catalog associated. If no catalogId is specified in the updateComponents message, it should be the surface's default catalog.

nan-yu (commented):
LGTM in with 2 comments:
- The targetVersion can be removed in a separate PR
- Should the catalog be optional in the ComponentModel?

typescript/web_core/src/state/component-model.ts
gspencergoog: Yes, it's optional.  If it isn't specified, then it uses the default catalog specified by the `createSurface` op.

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Okay.  Like you say below, let's remove that in another PR.

typescript/web_core/src/state/component-model.ts
nan-yu: ComponentModel needs the surface's default catalog at initialization time since it can't access it after the instance is created. Therefore, I feel like this parameter shouldn't be optional.

nan-yu (approved):

typescript/web_core/src/state/component-model.ts
gspencergoog: Ah, good catch. I was thinking about the wire protocol where `catalogId` is optional on `updateComponents`, but you're talking about the constructor parameter on `ComponentModel`. You're right: at the model level, the processor resolves the catalog upfront, so `ComponentModel` should hold a required `Catalog`.

PR #2350, opened 2026-08-20
feat(web_core): implement Stage 3 Sauce-TS bidirectional RPC and @index function

gspencergoog:
## Summary
Implements bidirectional RPC function invocation (`RpcHandler`) and the `@index` system function for A2UI protocol v1.0 in TypeScript `web_core`. This enables client renderer function execution requested by the server agent and outbound agent function calls from the client renderer.

## Changes
- **Bidirectional RPC Handler (`typescript/web_core/src/v1_0/rpc/rpc-handler.ts`)**:
  - Adds `RpcHandler` to execute remote renderer functions and dispatch outbound agent function calls.
  - Implements `handleCallRendererFunction` to execute renderer functions with catalog lookup, argument schema validation, signal unwrapping, and error responses.
  - Enforces `allowedCallers` execution boundaries (`rendererOnly`, `rendererOrAgent`, `agentOnly`) and `requiresUserActivation` checks.
  - Implements `callAgentFunction` to send outbound function calls via `outboundListener` with promise tracking and configurable timeouts.
  - Implements `handleAgentFunctionResponse` to resolve or reject pending outbound promises.
  - Adds `dispose` method to clear timers and reject pending promises with `RpcErrorCode.CANCELLED`.
  - Defines `RpcErrorCode` enum and `RpcError` error class.
- **System Functions (`typescript/web_core/src/v1_0/functions/system_functions.ts`)**:
  - Implements `IndexImplementation` for the `@index` function using `IndexApi`.
  - Computes iteration indices from `DataContext` paths or `getIndex()`, ignoring alphanumeric property names and applying optional offsets.
  - Exports `SYSTEM_FUNCTIONS` collection containing `IndexImplementation`.
- **Basic Catalog Organization (`typescript/web_core/src/v1_0/basic_catalog/`)**:
  - Moves v1.0 validation function APIs and implementations to `basic_catalog/functions/`.
  - Re-exports validation functions through `basic_catalog/index.ts`.
  - Updates `v1_0/index.ts` to export `rpc-handler.js` and `system_functions.js`.
- **Unit Tests (`typescript/web_core/src/v1_0/rpc/rpc.test.ts`)**:
  - Adds 18 unit tests in `rpc.test.ts` covering inbound RPC execution, boundary checks, user activation, and outbound calls.
  - Tests verify timeouts, `@index` path resolution, crypto fallback IDs, disposal, and error handling.

## Impact & Risks
- **Impact**: Adds protocol v1.0 bidirectional function execution and `@index` context resolution to TypeScript `web_core`.
- **Breaking Changes**: Validation functions are relocated to `basic_catalog/` and are no longer exported directly from the `v1_0/` barrel. Consumers importing validation functions from `a2ui/v1_0` must import them from `a2ui/v1_0/basic_catalog`.
- **Performance & Dependencies**: Adds no external dependencies. In-flight RPC calls use `Map` lookups and native timers with automatic cleanup on response, timeout, or disposal.

## Testing
1. Run `web_core` unit tests:
   ```bash
   yarn --cwd typescript/web_core test
   ```
2. Verify all 451 tests in `typescript/web_core` pass, including all 18 new test cases in `typescript/web_core/src/v1_0/rpc/rpc.test.ts`.


typescript/web_core/src/v1_0/rpc/rpc-handler.ts
```diff
@@ -0,0 +1,413 @@
...
+
+    // 3. Validate boundary execution constraint (allowedCallers / callableFrom)
+    const boundary =
+      funcImpl.allowedCallers ?? (funcImpl as any).callableFrom ?? 'rendererOnly';
+    if (boundary !== 'rendererOrAgent' && boundary !== 'agentOnly') {
+      return this.createResponseError(
+        functionCallId,
+        RpcErrorCode.INVALID_FUNCTION_CALL,
+        `Function '${call}' cannot be called by agent (callableFrom is ${boundary}).`,
+      );
```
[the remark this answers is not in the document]
gspencergoog: Updated boundary check and error message to strictly use allowedCallers.

typescript/web_core/src/v1_0/rpc/rpc-handler.ts
```diff
@@ -0,0 +1,413 @@
...
+      call = callOrOptions as FunctionCall;
+      effectiveTimeoutMs = timeoutMs ?? this.defaultTimeoutMs;
+    } else {
+      call = functionCallIdOrCall;
+      const opts = (callOrOptions as {functionCallId?: string; timeoutMs?: number}) ?? {};
+      functionCallId =
+        opts.functionCallId ??
+        (typeof globalThis.crypto?.randomUUID === 'function'
+          ? globalThis.crypto.randomUUID()
+          : `call-${Date.now()}-${Math.random().toString(36).slice(2, 9)}`);
```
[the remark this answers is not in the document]
gspencergoog: Updated fallback ID generator to produce consistent length padded string.

typescript/web_core/src/processing/adapters/v1_0.ts
```diff
@@ -80,6 +77,33 @@ export class V1Point0Adapter extends BaseVersionAdapter {
...
+      const crf = msgObj.callRendererFunction as Record<string, unknown>;
+      const cf = (crf?.callFunction || {}) as Record<string, unknown>;
+      ops.push({
+        type: 'callRendererFunction',
+        functionCallId: String(crf?.functionCallId || ''),
+        call: String(cf?.call || ''),
+        version: this.version,
+        catalogId: typeof cf?.catalogId === 'string' ? cf.catalogId : undefined,
+        args:
+          cf?.args && typeof cf.args === 'object' && !Array.isArray(cf.args)
```
nan-yu, line 90: Compared with the Python implementation, do we need a flag to indicate whether user_activation is done?

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -93,6 +115,10 @@ export interface MessageProcessorOptions {
   adapterRegistry?: VersionAdapterResolver;
   /** Validation configuration rules. */
   validationConfig?: ValidationConfig;
+  /** Listener receiving outbound renderer messages intended for the agent. Required for callAgentFunction. */
+  outboundListener?: OutboundMessageListener;
+  /** Default timeout in milliseconds for callAgentFunction requests (default: 30000ms). */
+  defaultTimeoutMs?: number;
```
nan-yu, line 128: Great idea with the timeout. I need to steal it and add to the Python implementation :)

typescript/web_core/src/rpc/rpc-handler.ts
nan-yu, line 1: I like the idea of extracting a separate RpcHandler for single responsibility concern. I'm thinking to refactor the Python implementation to adopt it. Should we move it outside of the `v1_0` package for future versions?

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -68,8 +82,16 @@ export type ProcessableMessagePayload =
...
 
-export type {RendererCapabilities, ValidationConfig};
-export {STRICT_VALIDATION, RELAXED_VALIDATION};
+export type {RendererCapabilities, ValidationConfig, OutboundMessageListener, CallOptions};
+export {STRICT_VALIDATION, RELAXED_VALIDATION, RpcError, RpcErrorCode};
+
+/**
+ * Contextual execution options for message processing.
+ */
+export interface ProcessMessagesOptions {
```
nan-yu: Nit: Should we rename to ExecutionContext, or ProcessContext? It feels like more of a context instead of a configuration option to me.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -68,8 +82,16 @@ export type ProcessableMessagePayload =
...
+  /** Whether execution occurs in an active user gesture context. */
+  isUserActivated?: boolean;
```
nan-yu, line 100: I'm wondering whether we need to keep naming consistency across multiple languages. This is used in the primary `processMessages` interface. It would be good to define it in the blueprint.

I like `isUserActivated` more than `user_activation_present` because it is shorter and simple. I'll update the Python code to match it.

typescript/web_core/src/v1_0/rpc/rpc-handler.ts
```diff
@@ -0,0 +1,556 @@
...
+  callAgentFunction<T = unknown>(
+    surfaceId: string,
+    call: FunctionCall,
+    options?: CallOptions,
+  ): Promise<T>;
+
+  /**
+   * Invokes a remote function on the server agent using positional parameters.
+   *
+   * @deprecated Prefer the options-bag overload: `callAgentFunction(surfaceId, call, options)`.
```
nan-yu: We can remove the deprecated function.

typescript/web_core/src/v1_0/rpc/rpc-handler.ts
```diff
@@ -0,0 +1,556 @@
...
+   * @param surfaceId The ID of the surface requesting execution.
+   * @param functionCallId The unique ID for this invocation instance.
```
nan-yu: The other method allows functionCallId to be provided rather than generated internally. I'm wondering which part controls the functionCallId?

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -205,6 +218,29 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
...
 
+  /**
+   * Invokes a remote function on the server agent using an options bag.
+   *
+   * @param surfaceId The ID of the surface requesting execution.
+   * @param call The function call details.
+   * @param options Optional invocation options (custom functionCallId, timeoutMs).
+   * @returns A promise resolving to the agent function return value.
+   */
+  callAgentFunction<TRes = unknown>(
```
nan-yu, line 236: This is much simpler than the Python implementation with `create_call_agent_function_message` followed by `register_pending_future`. I'll update the Python code to match.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -351,11 +387,12 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
...
 
   /**
-   * Processes a list of messages, a message wrapper, or raw operations.
+   * Processes a list of messages, a message wrapper, or raw operations synchronously.
    *
    * @param messages The messages or operations to process.
+   * @param options Contextual execution options.
    */
-  processMessages(messages: ProcessableMessagePayload): void {
+  processMessages(messages: ProcessableMessagePayload, options?: ProcessMessagesOptions): void {
```
nan-yu: Do we need to return `RendererFunctionResponseMessage` if `processMessages` process `callAgentFunction` synchronously?

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -387,8 +417,52 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
...
+  }
+
+  /**
+   * Asynchronously processes messages, executing RPC calls and returning all produced responses.
+   *
+   * @param messages The messages or operations to process.
+   * @param options Contextual execution options.
+   * @returns Array of rendererFunctionResponse messages produced during execution.
+   */
+  async processMessagesAsync(
```
nan-yu, line 450: Most of logic in `processMessagesAsync` and `processMessages` is identical. What about extracting the shared logic to a private method?

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -443,6 +518,90 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
...
+            },
+          },
+        };
+        this.rpc.handleCallRendererFunction(callMsg, dataContext, isUserActivated);
+        break;
+      }
+    }
+  }
+
+  private async processOperationAsync(
```
nan-yu, line 583: This is also almost identical to the synchronous method. We may want to dedup.

typescript/web_core/src/rpc/rpc-handler.ts
gspencergoog: Hmm, moving it is a good idea.  It should be fairly version agnostic, although it wouldn't be used by v0.9.

typescript/web_core/src/processing/adapters/v1_0.ts
gspencergoog: I think we should make the python one act more like TypeScript:

In TypeScript, it intentionally keeps the version adapters ([V1Point0Adapter](https://github.com/a2ui-project/a2ui/blob/v1_0_sauce_ts/typescript/web_core/src/processing/adapters/v1_0.ts#L22)) responsible for wire deserialization without a dependency on the runtime state.

The user activation state is passed into `processMessages` / `processMessagesAsync` in the execution context options:

```typescript
messageProcessor.processMessages(messages, { isUserActivated: true });
```

When we're handing the operations in `MessageProcessor`, it evaluates `options?.isUserActivated ?? op.isUserActivated ?? false` and passes that down to `RpcHandler.handleCallRendererFunction(...)` to enforce `funcImpl.requiresUserActivation` checks.

The `InternalCallRendererFunctionOp` still supports an optional `isUserActivated?:` boolean field if an operation is constructed directly (like in unit tests).

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Yeah, otherwise you can't clean up stale ones.

typescript/web_core/src/rpc/rpc-handler.ts
gspencergoog: I'll move `RpcHandler` out of `src/v1_0/rpc/` to `src/rpc/rpc-handler.ts` and re-export it from the root.

nan-yu: > although it wouldn't be used by v0.9

v0.9 adapter should not extract the RPC function call. We should be good.

typescript/web_core/src/rpc/rpc-handler.ts
```diff
@@ -0,0 +1,556 @@
...
+  /**
+   * Invokes a remote function on the server agent using an options bag.
+   *
+   * @param surfaceId The ID of the surface requesting execution.
+   * @param call The function call details.
+   * @param options Optional invocation options (custom functionCallId, timeoutMs).
+   * @returns A promise resolving to the agent function return value.
+   */
+  callAgentFunction<T = unknown>(
+    surfaceId: string,
```
nan-yu, line 231: Is callAgentFunction always associated with a surface?

typescript/web_core/src/processing/adapters/v1_0.ts
nan-yu: That makes sense. I'll update the Python code to match the TypeScript version.

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Sounds good. Let's standardize on `isUserActivated` (and `is_user_activated` in Python) across the blueprint and all SDKs.

typescript/web_core/src/v1_0/rpc/rpc-handler.ts
gspencergoog: The caller initiating the RPC controls the `functionCallId`:
- Inbound (`callRendererFunction`): The agent generates `functionCallId`, and the renderer echoes it back in `rendererFunctionResponse`.
- Outbound (`callAgentFunction`): The renderer generates `functionCallId` (using `randomUUID()`), tracks it in `pendingAgentCalls`, and the agent echoes it back in `agentFunctionResponse`.

Allowing `options.functionCallId` is only an escape hatch for deterministic testing or custom tracing; standard callers will let `RpcHandler` generate it automatically.

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Good catch. I will extract a `prepareOperations(messages: ProcessableMessagePayload): InternalOperation[]` helper for both sync and async methods.

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Good point. I'll extract an `applyStateOperation(op)` method to share across both sync and async dispatchers.

typescript/web_core/src/rpc/rpc-handler.ts
gspencergoog: Yes (it's in the spec that way).

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Do you mean `callRendererFunction`?

If you mean `callAgentFunction`: No response message is returned. `callAgentFunction` is outbound from the renderer to the agent; when the agent replies with `agentFunctionResponse`, `processMessages` simply resolves the pending `Promise` on the client and returns void.

If you were thinking of inbound `callRendererFunction`: Synchronous `processMessages` returns `void` because catalog functions in TypeScript can be asynchronous. Instead, the resulting `RendererFunctionResponseMessage` is emitted over `outboundListener` as soon as execution completes (or collected via `processMessagesAsync`).

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Sounds good. I'll update it.

gspencergoog:
Here are the changes I made:

- `RpcHandler` Relocation: Move `RpcHandler` from `src/v1_0/rpc/` to `src/rpc/` so it is treated as a core engine service reusable across present and future protocol versions.
- `OutboundListener` Response Streaming: Update `RpcHandler.handleCallRendererFunction` to emit `RendererFunctionResponseMessage` directly through `outboundListener` when present. This enables streaming transports to receive function execution responses even when using synchronous `processMessages`.
- API Streamlining: Remove the deprecated positional overload on `callAgentFunction` to keep a single, clean options-bag signature (`callAgentFunction(surfaceId, call, options?)`).
- Naming Parity: Rename `ProcessMessagesOptions` to ExecutionContext to align with the Python implementation and the protocol blueprint.

PTAL.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -443,6 +518,90 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
...
+          callRendererFunction: {
+            functionCallId: op.functionCallId,
+            callFunction: {
+              call: op.call,
+              catalogId: op.catalogId,
+              args: op.args,
+            },
+          },
+        };
+        this.rpc.handleCallRendererFunction(callMsg, dataContext, isUserActivated);
```
nan-yu: The async `handleCallRendererFunction` function is invoked in the synchronous `processMessages` function. We may want to block the callRendererFunction message from the synchronous call:

```typescript
case 'callRendererFunction':
  throw new A2uiProcessingError(
    "Operation 'callRendererFunction' is asynchronous and requires using processMessagesAsync()."
  );

```

typescript/web_core/src/processing/message-processor.ts
nan-yu: Sorry, wrong copy/paste. I meant `callRendererFunction`. I added another comment about whether we should allow synchronous `processMessages` to handle the async `callRendererFunction`.

typescript/web_core/src/v1_0/rpc/rpc-handler.ts
nan-yu: Thanks for the explanation. That makes a lot of sense.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -68,8 +82,21 @@ export type ProcessableMessagePayload =
...
+/**
+ * Contextual execution options for message processing.
+ */
+export interface ExecutionContext {
+  /** Whether execution occurs in an active user gesture context. */
+  isUserActivated?: boolean;
+}
+
+/**
+ * @deprecated Prefer `ExecutionContext`.
```
nan-yu: nit: ProcessMessagesOptions was introduced in this PR. We can remove it directly without deprecation.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -443,9 +498,76 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
       case 'updateDataModel':
         this.processUpdateDataModelOp(op);
         break;
+      case 'agentFunctionResponse':
+        this.rpc.handleAgentFunctionResponse({
+          version: 'v1.0',
```
nan-yu: v1.0 -> op.version?

nan-yu: `version` is not in `InternalOperation` yet, but I think we need to add it.

typescript/web_core/src/processing/message-processor.ts
nan-yu: Never mind. The comment was based on the old implementation. I see `applyStateOperation` skips `callRendererFunction, which addressed my concern.

typescript/web_core/src/processing/message-processor.ts
nan-yu: `applyStateOperation` skips `callRendererFunction, which addressed my concern.

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Well, but it doesn't have to be that way.  If they specify an `outboundListener` then they can listen to responses to these calls, even if they called the sync version.

Here are the different use cases:
- `processMessages`: Callers register an `outboundListener`. In this mode, incoming message streams can be passed directly to `processMessages`; UI state updates apply synchronously, while `callRendererFunction` triggers asynchronously in the background and streams the resulting `RendererFunctionResponseMessage` back out through the `outboundListener`. Blocking this with an exception would break mixed batches that contain both UI mutations and function calls, and lose the other parts of the batched input every time the exception throws.
- `processMessagesAsync`: When the caller needs to `await` all function executions and collect the responses directly in the return value (like in an HTTP endpoint or test harness), `processMessagesAsync` returns a `Promise`.

Because `outboundListener` does delivery for both, keeping `processMessages` so it handles the renderer function calls makes it so that they aren't forced to use async execution unless they specifically need to await responses.

typescript/web_core/src/rpc/rpc-handler.ts
```diff
@@ -0,0 +1,543 @@
...
+    }
+    return null;
+  }
+
+  private resolveFunctionImplementation(
+    catalogId: string | undefined,
+    call: string,
+    context?: DataContext,
+  ): {funcImpl: FunctionImplementation} | {error: string} {
+    let catalog: Catalog<any> | undefined;
```
nan-yu, line 340: Per `All catalog IDs specified at the component and function-call levels and at the surface-level must refer to catalogs which use the same A2UI specification version.` in https://github.com/a2ui-project/a2ui/blob/main/specification/v1_0/docs/a2ui_protocol.md#mixable-catalogs-and-component-resolution-logic, we also need to verify if the function catalog conforms to the same protocol version.

gspencergoog: Good catch. I removed the fallback to `this.catalogs[0]`.

I think the python version was defaulting to the `[0]` one too in [message_processor.py](https://github.com/a2ui-project/a2ui/blob/v1_0/python/a2ui_core/src/a2ui/core/processing/message_processor.py#L237)

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Yes. Added `version` to `InternalAgentFunctionResponseOp`, populated it in `V1Point0Adapter`, and updated `handleAgentFunctionResponse` to use `op.version ?? this.version` instead of hardcoding `'v1.0'`.

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Good point.  Removed.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -443,9 +493,76 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
...
+        callRendererFunction: {
+          functionCallId: op.functionCallId,
+          callFunction: {
+            call: op.call,
+            catalogId: op.catalogId,
+            args: op.args,
+          },
+        },
+      };
+      this.rpc.handleCallRendererFunction(callMsg, dataContext, isUserActivated);
```
nan-yu: Missed it. The synchronous `processOperation` method calls the async `rpc.handleCallRendererFunction` without an await sounds problematic.

gspencergoog: Well, I switched it to this to prevent unhandled exceptions:

```ts
      void this.rpc.handleCallRendererFunction(callMsg, dataContext, isUserActivated).catch(err => {
        console.error(`Unhandled error in callRendererFunction (${op.call}):`, err);
      });
```

But since this is a sync function, we can't `await` here.

gspencergoog: The responses will go to the `outboundListener` in the sync case, so they do get handled eventually, assuming the caller provided a listener.

typescript/web_core/src/rpc/rpc-handler.ts
nan-yu: I'll fix the Python version as well :)

nan-yu (approved):

PR #2359, opened 2026-08-20
feat(web_core): enhance generic binder schema inference

josemontespg:
# Description

Enhances `@a2ui/web_core` `GenericBinder` and `scrapeSchemaBehavior` to recognize schema `$defs` references and child-reference metadata during schema inference, types resolved child lists, and guards structural template parsing.

## Motivation

In `@a2ui/web_core`, `scrapeSchemaBehavior` previously identified A2UI primitives (`Action`, `DataBinding`, `ChildList`) solely through structural union checks (`ZodUnion` options). When component schemas were defined using descriptions referencing `$defs` (such as `#/$defs/Action`, `#/$defs/ChildList`, `#/$defs/DynamicString`, or `REF:common_types.json#/$defs/DataBinding`), or using custom metadata like `a2uiChildRef`, `scrapeSchemaBehavior` failed to match them and defaulted to `STATIC`.

Additionally, child lists inferred through `ResolveA2uiProp` resolved to `any`, and structural binding did not explicitly guard against arrays before checking template object fields.

## Key Changes

1. **Schema Description and Metadata Matching (`scrapeSchemaBehavior`)**:
   - Inspects `_def.description` across unwrapped `ZodOptional`, `ZodNullable`, and `ZodDefault` wrappers.
   - Detects action schemas via `ACTION_REF` (`REF:common_types.json#/$defs/Action`) or description text matching `Triggers a server-side event`.
   - Detects child lists via `_def.a2uiChildRef === 'child-list'` or `#/$defs/ChildList` in descriptions.
   - Detects dynamic bindings via `DATA_BINDING_REF` (`REF:common_types.json#/$defs/DataBinding`) or `#/$defs/Dynamic`, while excluding `ZodObject` and `ZodArray` so nested containers continue recursive traversal.

2. **Array Guard in Structural Template Binding (`GenericBinder`)**:
   - Adds an `!Array.isArray(value)` check in `bindStructuralTemplate` so static child ID arrays (`string[]`) pass through untouched and are not evaluated as `{ path, componentId }` templates.

3. **Type Improvements (`ResolveA2uiProp`)**:
   - Exports the `ResolvedChildRef` interface (`{ id: string; basePath: string }`).
   - Updates `ResolveA2uiProp<T>` to type `ChildList` properties as `(string | ResolvedChildRef)[]` instead of `any`.

4. **Internal Binding Modularization**:
   - Extracts `bindDynamicValue`, `bindAction`, and `bindStructuralTemplate` from `resolveAndBind` for clearer separation of binding logic.

5. **Build and Test Coverage**:
   - Adds unit tests in `generic-binder.test.ts` verifying description-based inference, nested object/array traversal, and static pass-through for unannotated schemas.
   - Updates `renderers/web_core/tsconfig.json` to exclude `node_modules` and `dist`.

## Pre-launch Checklist

- [x] I signed the [CLA].

Related to #1270

### DIFF
--- a/renderers/web_core/src/v0_9/rendering/generic-binder.ts
+++ b/renderers/web_core/src/v0_9/rendering/generic-binder.ts
@@ -16,7 +16,13 @@
 
 import {z} from 'zod';
 import {ComponentContext} from './component-context.js';
-import {Action, ChildList, DataBinding, FunctionCall} from '../schema/common-types.js';
+import {
+  Action,
+  ChildList,
+  DataBinding,
+  FunctionCall,
+  childRefKindOf,
+} from '../schema/common-types.js';
 
 // --- Schema Scraping ---
 
@@ -55,8 +61,15 @@ export function scrapeSchemaBehavior(schema: z.ZodTypeAny): BehaviorNode {
   return getFieldBehavior(schema);
 }
 
+// TODO(#2443): Export and reuse these schema reference constants from a central location across web_core.
+const ACTION_REF = 'REF:common_types.json#/$defs/Action';
+const DATA_BINDING_REF = 'REF:common_types.json#/$defs/DataBinding';
+const DYNAMIC_REF_PREFIX = '#/$defs/Dynamic';
+
 function getFieldBehavior(type: z.ZodTypeAny, propertyName?: string): BehaviorNode {
   let current = type;
+
+  let description = current._def?.description || '';
 
   // Unwrap optionals/nullables/defaults
   while (
@@ -64,11 +77,33 @@ function getFieldBehavior(type: z.ZodTypeAny, propertyName?: string): BehaviorNo
     current._def.typeName === 'ZodNullable' ||
     current._def.typeName === 'ZodDefault'
   ) {
+    if (!description && current._def.description) {
+      description = current._def.description;
+    }
     current = current._def.innerType;
+  }
+  if (!description && current._def.description) {
+    description = current._def.description;
   }
 
   if (propertyName === 'checks') {
     return {type: 'CHECKABLE'};
+  }
+
+  if (description.startsWith(ACTION_REF)) {
+    return {type: 'ACTION'};
+  }
+
+  if (childRefKindOf(current) === 'child-list' || description.includes('#/$defs/ChildList')) {
+    return {type: 'STRUCTURAL'};
+  }
+
+  if (
+    (description.startsWith(DATA_BINDING_REF) || description.includes(DYNAMIC_REF_PREFIX)) &&
+    current._def.typeName !== 'ZodObject' &&
+    current._def.typeName !== 'ZodArray'
+  ) {
+    return {type: 'DYNAMIC'};
   }
 
   // Structural matching for A2UI primitives using typeName to avoid dual-module instanceof issues
@@ -122,13 +157,21 @@ type DynamicTypes = DataBinding | FunctionCall;
 type IsDynamic<T> = DataBinding extends NonNullable<T> ? true : false;
 
 /**
+ * A resolved reference to a child component, containing its unique ID and bound data context path.
+ */
+export interface ResolvedChildRef {
+  id: string;
+  basePath: string;
+}
+
+/**
  * Maps raw Zod inferred types to their resolved runtime equivalents.
  * For example, an `Action` object becomes a callable `() => void` function.
  */
 export type ResolveA2uiProp<T> = [NonNullable<T>] extends [Action]
   ? (() => void) | Extract<T, undefined>
   : [NonNullable<T>] extends [ChildList]
-    ? any | Extract<T, undefined>
+    ? (string | ResolvedChildRef)[] | Extract<T, undefined>
     : Exclude<T, DynamicTypes> extends never
       ? any
       : Exclude<T, DynamicTypes>;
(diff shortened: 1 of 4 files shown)


renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -58,19 +58,50 @@ export function scrapeSchemaBehavior(schema: z.ZodTypeAny): BehaviorNode {
...
+  if (
+    propertyName === 'text' ||
+    propertyName === 'label' ||
+    propertyName === 'value' ||
+    propertyName === 'url' ||
+    description.includes('#/$defs/Dynamic') ||
+    description.includes('#/$defs/DataBinding')
+  ) {
+    return {type: 'DYNAMIC'};
+  }
```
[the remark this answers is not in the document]
josemontespg: Agent do this.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -328,8 +363,70 @@ export class GenericBinder<T> {
         return value; // The 'checks' property itself remains as the original rules array
       }
 
-      case 'STATIC':
+      case 'STATIC': {
+        if (value && typeof value === 'object') {
```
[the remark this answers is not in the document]
josemontespg: Agent do this.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -328,8 +363,70 @@ export class GenericBinder<T> {
...
+                const res: any = {};
+                for (const [k, v] of Object.entries(val)) res[k] = resolveDeepSync(v);
+                return res;
+              };
+              this.context.dispatchAction(resolveDeepSync(value));
+            };
+          }
+        }
         return value;
+      }
```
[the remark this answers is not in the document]
josemontespg: Agent do this.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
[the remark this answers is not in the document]
josemontespg: Fixed. Updated `getFieldBehavior` to ensure we do not short-circuit to `DYNAMIC` if the unwrapped schema is a `ZodObject` or `ZodArray`, allowing nested properties and arrays to be traversed and resolved recursively.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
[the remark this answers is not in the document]
josemontespg: Fixed. Added `!Array.isArray(value)` to ensure arrays are not erroneously checked as plain objects in the `STATIC` fallback block.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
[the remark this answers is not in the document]
josemontespg: Fixed. Extracted common binding logic into private helper methods (`bindDynamicValue`, `bindAction`, `bindStructuralTemplate`), eliminated code duplication between typed and static cases, and introduced the strongly-typed `ResolvedChildRef` interface.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -58,19 +58,52 @@ export function scrapeSchemaBehavior(schema: z.ZodTypeAny): BehaviorNode {
 function getFieldBehavior(type: z.ZodTypeAny, propertyName?: string): BehaviorNode {
   let current = type;
 
+  let description = current._def?.description || '';
```
ditman, line 72: I know we had there before, but it'd be nice if we extracted all these things that access deep zod properties to a zod-specific library. Anyway, that can be done later, this _generic_ binder is super tied to zod everywhere else :P

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -58,19 +58,52 @@ export function scrapeSchemaBehavior(schema: z.ZodTypeAny): BehaviorNode {
...
 
+  if (
+    propertyName === 'action' ||
+    description.includes('#/$defs/Action') ||
+    description.includes('Triggers a server-side event')
+  ) {
+    return {type: 'ACTION'};
+  }
+
+  if (propertyName === 'children' || description.includes('#/$defs/ChildList')) {
```
ditman: Is this "description.includes" or is this `description === #/$defs/ChildList`? (for all these accesses)

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -58,19 +58,52 @@ export function scrapeSchemaBehavior(schema: z.ZodTypeAny): BehaviorNode {
...
+    description = current._def.description;
+  }
 
   if (propertyName === 'checks') {
     return {type: 'CHECKABLE'};
   }
 
+  if (
+    propertyName === 'action' ||
+    description.includes('#/$defs/Action') ||
```
ditman: Should these special paths be extracted to a constant somewhere? it probably doesn't need to be exported.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -339,8 +401,20 @@ export class GenericBinder<T> {
         return value; // The 'checks' property itself remains as the original rules array
       }
 
-      case 'STATIC':
+      case 'STATIC': {
+        if (value && typeof value === 'object' && !Array.isArray(value)) {
+          if (('componentId' in value || 'id' in value) && 'path' in value) {
+            return this.bindStructuralTemplate(value, path, isSync);
+          }
```
ditman: Why do we need special casing here? Instead of modifying the resolveAndBind to determine that this STATIC bit needs the same handling as the STRUCTURAL above?

ditman (approved):
This LGTM in general, I left a few comments that are minor, mostly questions about the approach

renderers/web_core/src/v0_9/rendering/generic-binder.ts
andrewkolos: piggy-backing: I might be unclear on what this logic is meant for. Does this just handle the case where the client specifies `z.any()` (and the property doesn't have one of the magical names like `value`/`action`)? In that scenario, I would assume that the user would just want the data passed through as-is rather than risk being misinterpreted as a data binding or action.

Are these heuristics here to support ported angular components that don't declare a schema or something (which sounds like a problem with the Angular adapter, if anything)? I feel like I'm missing something and would need a good bit of time to rethink all this binding code into something more intuitive (and thus more likely to get consistent across other languages/platforms)

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -58,19 +58,52 @@ export function scrapeSchemaBehavior(schema: z.ZodTypeAny): BehaviorNode {
...
+      propertyName === 'label' ||
+      propertyName === 'value' ||
+      propertyName === 'url' ||
+      description.includes('#/$defs/Dynamic') ||
+      description.includes('#/$defs/DataBinding')) &&
+    current._def.typeName !== 'ZodObject' &&
+    current._def.typeName !== 'ZodArray'
+  ) {
+    return {type: 'DYNAMIC'};
+  }
```
andrewkolos: I am not sure any of these fallback heuristics are even necessary after #2077 merged ([code](https://github.com/a2ui-project/a2ui/pull/2077/changes#diff-eaeb0c3a10fbfb1aa6e28d7a77b1df733a33b02cf19fd338a1bc777eedce3b75)). In addition, I don't like that in

```ts
{
  value: {id: 'row', path: '/rows'},
  other: {id: 'row', path: '/rows'},
}
```

`value` and `other` get interpreted differently despite having the same payload.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -226,76 +267,97 @@ export class GenericBinder<T> {
...
+
+  private bindStructuralTemplate(
+    value: any,
+    path: string[],
+    isSync: boolean,
+  ): ResolvedChildRef[] | any {
+    if (value && typeof value === 'object' && !Array.isArray(value)) {
+      const templatePath = value.path;
+      const templateComponentId = value.componentId || value.id;
```
andrewkolos: What produces `id` instead of `componentId`?

andrewkolos (commented):
I like the schema-unwrapping improvements, but I have some concerns with the schema fallback logic that's based on property names, shapes, and free-text descriptions. See my other comments.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
andrewkolos: > Are these heuristics here to support ported angular components that don't declare a schema or something (which sounds like a problem with the Angular adapter, if anything)?

Re-reading the PR description, i see this is the case. But I still maintain this is probably an issue to be resolved in the Angular adapter, not in web_core.

In addition, reading through https://github.com/a2ui-project/a2ui/blob/b356d61b30652b33d738dcf42a4ccfa4eb099a86/renderers/angular/src/v0_9/core/component-binder.service.ts, parity wouldn't be achieved. For example, Angular has special logic for `trigger` and `content`.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
josemontespg: Agreed. Created issue #2443 to track extracting a shared Zod schema reflection and unwrapping utility across @a2ui/web_core.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
josemontespg: The wire descriptions include extra metadata (`REF:common_types.json#//ChildList|...`), so `===` doesn't match.

However, following PR #2393, we now check `childRefKindOf(current) === 'child-list'` metadata directly instead of parsing descriptions.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
josemontespg: Done. Extracted the reference strings into constants at the top of the file, with a `TODO(#2443)` to export them from a central location across `web_core`.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
josemontespg: Good point. Removed the fallback heuristics from `case 'STATIC'` entirely, leaving it as a simple `return value;`. If a property isn't typed as dynamic or structural in its schema, it passes through as-is.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
josemontespg: Agreed. This PR was drafted before 2077 was merged. Removed all property name checks (value, text, label, url, action, children). The binder now relies strictly on schema types and metadata.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
josemontespg: Nothing on the wire—this was accidentally conflated with ResolvedChildRef. Removed || value.id; it now strictly checks componentId.

josemontespg:
Thanks @ditman and @andrewkolos for the reviews! We've updated the PR accordingly: removed magic property names, strictly enforced `componentId` per the wire spec, removed runtime fallback heuristics from `case 'STATIC'`, and filed #2443 to track extracting shared schema reflection utilities. Detailed replies posted inline.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -55,22 +61,51 @@ export function scrapeSchemaBehavior(schema: z.ZodTypeAny): BehaviorNode {
...
   }
+  if (!description && current._def.description) {
+    description = current._def.description;
+  }
 
   if (propertyName === 'checks') {
     return {type: 'CHECKABLE'};
   }
 
+  if (description.startsWith(ACTION_REF) || description.includes('Triggers a server-side event')) {
```
andrewkolos: nit: The `description.includes('Triggers a server-side event')` part of the condition appears redundant to me (by the same logic in my earlier comments) and likely to become stale if/when the spec involves.

andrewkolos (approved):
LGTM with nit

renderers/web_core/src/v0_9/rendering/generic-binder.ts
josemontespg: Done. Removed `description.includes('Triggers a server-side event')`, keeping only the `description.startsWith(ACTION_REF)` check along with structural `ZodUnion` detection.

PR #2363, opened 2026-08-21
ci(python): run a2ui-agent-sdk tests on a2a-sdk 1.x

WillBeebe:
Stacked on #2337. **Only the last commit is new** (`ci(python): run a2ui-agent-sdk tests on a2a-sdk 1.x`). Review against `WillBeebe:feat/a2a-sdk-1x-part-types` if this PR still contains #2337.

## Why

#2337 widens the published `a2ui-agent-sdk` pin to `a2a-sdk>=0.3.0,<2`, but workspace CI stays on **0.3** (`constraint-dependencies` plus sample packages that still pin `<0.4`). The `IS_A2A_V1` branch of `_compat` is untested in CI.

A full clone of `python-ci` cannot `uv sync --all-packages` on 1.x: other workspace members (e.g. `file-upload-summarizer`) still require `a2a-sdk<0.4`.

## What

Sibling job **Python CI (a2a-sdk 1.x)**:

- Isolated venv (install from `/tmp` so workspace `constraint-dependencies` do not apply)
- `a2ui-core` + `a2ui-agent-sdk` + `a2a-sdk>=1.0.0,<2`
- `pytest agent_sdks/python/a2ui_agent/tests/a2a` — the helpers that go through `_compat`

Default `python-ci` is unchanged (0.3, full workspace).

Also: `part_metadata` no longer calls `HasField` on the protobuf map (raises `ValueError` on 1.x).

Locally: 10 passed on a2a-sdk **1.1.2** (`IS_A2A_V1=True`).

## Out of scope

- Sample `Part.root` rewrites (follow-up)
- `tests/adk` / `tests/conformance` still import 0.3 `DataPart` at test-construction time
- Lifting workspace `constraint-dependencies`

PR #2364, opened 2026-08-21
feat(python): stop using Part.root in first-party ADK samples

WillBeebe:
Stacked on #2363 (which is stacked on #2337). **Only the last commit is new.** Review against `WillBeebe:feat/a2a-sdk-1x-ci`.

## Why

#2337 left restaurant finder / other samples on `Part.root`. That is the 0.3-only construction the dual-compat helpers exist to replace.

## What

First-party ADK samples stop touching `part.root`:

- `samples/agent/adk/restaurant_finder`
- `samples/agent/adk/custom-components-example`

They now use `make_text_part` / `part_data_as_dict` (and `_compat.is_text_part` for logging).

Export `make_text_part` from `a2ui.a2a.parts`.

## Out of scope

- Community samples (`samples/community/**`) still use `Part.root`
- Workspace `constraint-dependencies` stays on a2a-sdk 0.3 (other packages still pin `<0.4`)
- AgentCard / Starlette 1.x server rewrite


WillBeebe:
Stacked on #2363 (#2337). Community samples and workspace `constraint-dependencies` stay 0.3.

jacobsimionato:
I replied to https://github.com/a2ui-project/a2ui/pull/2337. Let's get @nan-yu 's input and agree on direction first.

samples/agent/adk/custom-components-example/agent_executor.py
```diff
@@ -19,13 +19,12 @@
...
 from a2a.server.tasks import TaskUpdater
 from a2a.types import (
-    DataPart,
-    Part,
     Task,
     TaskState,
-    TextPart,
     UnsupportedOperationError,
 )
+from a2ui.a2a import _compat
```
nan-yu, line 26: In general, we should avoid importing private `_compat` in public sample code. We can either export `is_text_part`, `is_data_part`, and `part_text` from 
`a2ui.a2a.parts`
 alongside `make_text_part` and `part_data_as_dict`, or make `_compat` public.

PR #2374, opened 2026-08-24
fix(dart): implement atomic two-pass component validation in MessageProcessor

diegolopezrm:
## Summary

Ports the atomic two-pass component validation from #2332 (Swift) to the Dart implementation.

`MessageProcessor._processUpdateComponents` validated and applied components in a single pass. If an update payload contained a valid component followed by an invalid one, the valid component was already committed to `SurfaceComponentsModel` when the error was raised, leaving the surface in a half-updated state that no message describes and the client has no way to recover.

This mirrors the structure #2332 introduced in Swift:

1. **Validation pass** — every component in the batch is checked for a required `id` and, when the component does not already exist, a required `component` type. Any failure throws and the batch is abandoned without touching surface state.
2. **Mutation pass** — runs only once the whole batch has passed validation.

## Reproduction

Before this change, the following left `root` committed even though the batch was rejected:

```dart
processor.processMessages([
  CreateSurfaceMessage(surfaceId: 's1', catalogId: catalog.id),
]);

// throws A2uiValidationError for the second component...
processor.processMessages([
  UpdateComponentsMessage(surfaceId: 's1', components: [
    {'id': 'root', 'component': 'Text', 'text': 'valid'},
    {'component': 'Text', 'text': 'no id'},
  ]),
]);

// ...but 'root' was already in the surface.
surface.componentsModel.get('root'); // non-null before, null after
```

## Behavioural note

The "cannot create a component without a type" check moves from the mutation site into the validation pass. It keeps the same meaning — the type is required only when the component does not already exist, so updating an existing component without repeating its type still works — but it is now evaluated for the whole batch before anything is applied. A test covers that case explicitly so the relaxation is not lost.

## Tests

Five tests added to `dart/a2ui_core/test/processor_test.dart`:

- a fully valid batch still applies every component;
- a batch rejected for a missing `id` leaves the surface untouched;
- a batch rejected for a missing `component` type leaves the surface untouched;
- a rejected batch does not overwrite components applied by an earlier accepted batch;
- updating an existing component without repeating its type keeps working.

The three regression tests fail on `main` and pass with this change. The other two guard against over-restriction.

`dart test` in `dart/a2ui_core`: 74 passing (69 before). `dart analyze` reports the same two pre-existing `unnecessary_library_directive` infos as `main` and nothing new.


diegolopezrm:
@googlebot I signed it!

diegolopezrm:
@googlebot I signed it!

diegolopezrm:
@googlebot I signed it!

jacobsimionato (approved):

jacobsimionato:
Thank you for this nice improvement!!

PR #2379, opened 2026-08-24
feat(python): refactor MessageProcessor version adapters and organize conformance suites

nan-yu:
## Summary

This PR refactors `MessageProcessor` version adapters and reorganizes core conformance suites across A2UI protocol versions (v0.8, v0.9, v1.0).

### Key Changes
- **Version Adapters**: Introduced `BaseVersionAdapter` pattern (`v0_8.py`, `v0_9.py`, `v1_0.py`) using `get_adapter(version: A2uiProtocolVersion)` to translate version-specific message payloads into standard `InternalOperation` objects.
- **MessageProcessor Refactoring**: Updated `MessageProcessor` to process internal operations directly. Consolidated `get_renderer_capabilities` signature to accept `versions: List[ProtocolVersion]` and removed redundant `get_client_capabilities`.
- **Envelope & Validation**: Added strict validation enforcing that `surfaceId` must be a string and that envelopes containing multiple conflicting update actions raise a `ValidationError`.
- **Conformance Suites**: Reorganized monolithic `message_processor.yaml` into version-focused conformance files (`message_processor_v0_8.yaml`, `message_processor_v0_9.yaml`, `message_processor_v1_0.yaml`).

### Verification
- **Core Conformance Tests**: Passed 70/70 active conformance test cases.
- **Python Unit Tests**: Passed 807/807 tests with zero failures (`uv run pytest python/ -v`).

### DIFF
--- a/python/a2ui_core/src/a2ui/core/processing/adapters/__init__.py
+++ b/python/a2ui_core/src/a2ui/core/processing/adapters/__init__.py
@@ -12,23 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .message_processor import MessageProcessor
-from .operations import (
-    InternalCreateSurfaceOp,
-    InternalDeleteSurfaceOp,
-    InternalOperation,
-    InternalUpdateComponentsOp,
-    InternalUpdateDataModelOp,
-)
-from .adapters import VersionAdapter, VersionAdapterFactory
+from .base import VersionAdapter
+from .v0_8 import V0_8VersionAdapter
+from .v0_9 import V0_9VersionAdapter
+from .v1_0 import V1_0VersionAdapter
+from .factory import DEFAULT_PROTOCOL_VERSION, VersionAdapterFactory
 
 __all__ = [
-    "MessageProcessor",
-    "InternalOperation",
-    "InternalCreateSurfaceOp",
-    "InternalUpdateComponentsOp",
-    "InternalUpdateDataModelOp",
-    "InternalDeleteSurfaceOp",
+    "DEFAULT_PROTOCOL_VERSION",
     "VersionAdapter",
+    "V0_8VersionAdapter",
+    "V0_9VersionAdapter",
+    "V1_0VersionAdapter",
     "VersionAdapterFactory",
 ]
--- a/python/a2ui_core/src/a2ui/core/processing/adapters/__init__.py
+++ b/python/a2ui_core/src/a2ui/core/processing/adapters/__init__.py
@@ -13,16 +13,16 @@
 # limitations under the License.
 
 from .base import VersionAdapter
-from .v0_8 import V0_8VersionAdapter
-from .v0_9 import V0_9VersionAdapter
-from .v1_0 import V1_0VersionAdapter
+from .v0_8 import V0Point8Adapter
+from .v0_9 import V0Point9Adapter
+from .v1_0 import V1Point0Adapter
 from .factory import DEFAULT_PROTOCOL_VERSION, VersionAdapterFactory
 
 __all__ = [
     "DEFAULT_PROTOCOL_VERSION",
     "VersionAdapter",
-    "V0_8VersionAdapter",
-    "V0_9VersionAdapter",
-    "V1_0VersionAdapter",
+    "V0Point8Adapter",
+    "V0Point9Adapter",
+    "V1Point0Adapter",
     "VersionAdapterFactory",
 ]
--- a/python/a2ui_core/src/a2ui/core/processing/adapters/factory.py
+++ b/python/a2ui_core/src/a2ui/core/processing/adapters/factory.py
@@ -14,9 +14,9 @@
 
 from typing import Any, Dict, Optional
 from .base import VersionAdapter
-from .v0_8 import V0_8VersionAdapter
-from .v0_9 import V0_9VersionAdapter
-from .v1_0 import V1_0VersionAdapter
+from .v0_8 import V0Point8Adapter
+from .v0_9 import V0Point9Adapter
+from .v1_0 import V1Point0Adapter
 from ...exceptions import A2uiValidationError
 from ...schema import ProtocolVersion, AgentToRendererMessagePayload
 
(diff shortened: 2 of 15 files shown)


nan-yu:
All code review comments have been addressed:

1. **Payload Dict Validation**: Added checks in `BaseVersionAdapter.extract_operations` to verify that payload and action objects are valid dictionaries before operation extraction.
2. **v0.8 `dataModelUpdate` Contents Parsing & List Guard**: Added support for legacy v0.8 `contents` typed array parsing and guarded `raw_comps` type in `V0_8VersionAdapter`.
3. **Defensive Components List Check**: Added defensive type validation for `op.components` in `MessageProcessor._process_update_components_op`.
4. **Conformance Test Runner Assertions**: Added assertions for `theme`, `sendDataModel`, and `dataModel` state in `test_conformance.py`.

python/a2ui_core/src/a2ui/core/processing/adapters/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+from .base import VersionAdapter
+from .v0_8 import V0_8VersionAdapter
+from .v0_9 import V0_9VersionAdapter
+from .v1_0 import V1_0VersionAdapter
+from .factory import DEFAULT_PROTOCOL_VERSION, VersionAdapterFactory
+
+__all__ = [
+    "DEFAULT_PROTOCOL_VERSION",
+    "VersionAdapter",
+    "V0_8VersionAdapter",
```
gspencergoog: I struggled with these names (I know, you are just using the ones I had in `web_core`): I wish we didn't need the `_`, but we do, because when we get to version 11.0 someday, it would have the same name as version 1.10 otherwise (or any other three-digit version) (`V110VersionAdapter`).

And having both V and Version seems redundant.

I suppose we could be using `Version0Point8Adapter` or something?

python/a2ui_core/src/a2ui/core/validating/validator.py
```diff
@@ -82,7 +82,7 @@ class A2uiValidator:
 
     def validate_protocol_envelope(
         self,
-        messages: List[Dict[str, Any]],
+        messages: Any,
```
gspencergoog: Can we restrict the type a bit more here?

python/a2ui_core/src/a2ui/core/processing/message_processor.py
```diff
@@ -56,31 +62,39 @@ def process_messages(
         if self.strict_mode:
             self.validator.validate_protocol_envelope(message_list)
```
gspencergoog: When `messages` is passed as a single message dictionary (when it's     `AgentToRendererMessage` or  `Dict[str, Any]`), the `get` returns `[]` because `"messages"` is not a key in the dictionary, so it'll bypass the checks because it's an empty list.

python/a2ui_core/src/a2ui/core/processing/adapters/factory.py
```diff
@@ -0,0 +1,114 @@
...
+                raw_item: Any = item
+                if hasattr(raw_item, "model_dump"):
+                    raw_item = raw_item.model_dump(by_alias=True, exclude_none=True)
+                if isinstance(raw_item, dict):
+                    if "version" in raw_item and isinstance(raw_item["version"], str):
+                        ver_enum = cls._parse_version(raw_item["version"])
+                        if ver_enum:
+                            return cls.get_adapter(ver_enum)
+                    if any(
+                        k in item
```
gspencergoog: Should this be `raw_item`?

python/a2ui_core/src/a2ui/core/processing/adapters/factory.py
```diff
@@ -0,0 +1,114 @@
...
+            raw_payload = raw_payload.model_dump(by_alias=True, exclude_none=True)
+
+        if isinstance(raw_payload, list):
+            for item in raw_payload:
+                raw_item: Any = item
+                if hasattr(raw_item, "model_dump"):
+                    raw_item = raw_item.model_dump(by_alias=True, exclude_none=True)
+                if isinstance(raw_item, dict):
+                    if "version" in raw_item and isinstance(raw_item["version"], str):
+                        ver_enum = cls._parse_version(raw_item["version"])
```
gspencergoog: If it includes a version, but the version is bad (_parse_version doesn't parse it), then it will fall back to the default version.  It should probably throw an error if it finds a version it can't parse (vs. not having a version at all).

python/a2ui_core/src/a2ui/core/processing/adapters/__init__.py
nan-yu: Since we want to expand `Point`, what about `V0Point8Adapter`?
Or maybe even `V0P8Adapter` because we may end up having `V1Point1Point11Adapter` and `V1Point11Point1Adapter`, which seems still very verbose.

python/a2ui_core/src/a2ui/core/validating/validator.py
nan-yu: reverted the change. The method will be deleted in the next PR.

python/a2ui_core/src/a2ui/core/processing/message_processor.py
nan-yu: Thanks for catching this. I've updated process_messages to handle single message dictionary.

python/a2ui_core/src/a2ui/core/processing/adapters/factory.py
nan-yu: Thanks. Added exception for invalid versions.

python/a2ui_core/src/a2ui/core/processing/adapters/factory.py
nan-yu: updated

python/a2ui_core/src/a2ui/core/processing/adapters/__init__.py
nan-yu: I updated to `V0Point8Adapter` with the assumption that different patch versions should share the same minor version.

gspencergoog: Yeah, that works.

gspencergoog (approved):

PR #2380, opened 2026-08-25
feat(spec): add core A2UI templates specification, version 0.1 schema, and examples

jacobsimionato:
## Summary

This PR introduces the authoritative **A2UI Templates Specification**, formal JSON Schema, and multi-document YAML test fixtures under `specification/proposals/templates/`.

Templates provide a declarative, human-readable mechanism for authoring reusable UI subtrees in nested YAML. At generation time, an LLM emits compact, high-level template signatures (such as `UserProfile("u1", "Alice")`), which the backend expands into canonical Basic Catalog primitives in a single synchronous pass before emitting standard A2UI protocol messages (`createSurface`, `updateComponents`) to the client.

---

## Architectural Highlights

- **Zero Client Overhead**: Client renderers (@a2ui/react, @a2ui/lit, @a2ui/angular, @a2ui/flutter) require no custom widget code or dynamic plugins. They continue rendering 100% standard Basic Catalog primitives.
- **Context & Token Efficiency**: Agents output concise function-like signatures rather than deeply nested component trees.
- **Deterministic Synthetic ID Generation**: Implements canonical ID synthesis (`{parent}_{slot}_{index}_{type}`) guaranteeing zero ID collisions when multiple template instances are placed on a single surface.
- **Loop Expansion Boundaries**: Differentiates server-side synchronous unrolling (for literal parameter arrays) from client-side reactive loops bound to the client DataModel.
- **Two-Way DataModel Interoperability**: Preserves dynamic data binding paths (`{path: "/user/name"}`) and path prefix plumbing.
- **Strict Template Versioning (`0.1`)**: Enforces top-level `version: "0.1"` across all template definitions to support smooth schema evolution across future protocol releases.
- **Safety & Recursion Guards**: Formal call-stack tracking algorithm (`TemplateCycleError`) and `MAX_EXPANSION_DEPTH = 32` recursion limits.

---

## Files Added

- `specification/proposals/templates/README.md`: Complete, authoritative specification with pipeline diagrams, grammar rules, and multi-language conformance checklist.
- `specification/proposals/templates/schema/template_definition.json`: Canonical JSON Schema (Draft 2020-12) enforcing required `version: "0.1"`, `templateId`, `parameters`, and nested `layout`.
- `specification/proposals/templates/examples/*.yaml`: 11 multi-document YAML template fixtures and test suites (`basic_cards`, `nested_lists`, `slot_composition`, `user_profile`, `salary_card`, `team_roster`, etc.).

---

## Stack Context
- **PR 1 of 4**: Core Specification, Schema & Test Fixtures (This PR)
- **PR 2 of 4**: Python Agent SDK Template Engine Implementation
- **PR 3 of 4**: Community Templates Sample App & Interactive Studio
- **PR 4 of 4**: Future Proposals (Client-Expanded Templates & Python Fluent Builders)

### DIFF
--- a/specification/proposals/templates/README.md
+++ b/specification/proposals/templates/README.md
@@ -1,0 +1,582 @@
+# A2UI Templates Specification
+
+## Abstract
+
+This document defines the authoritative specification for the **A2UI Templates System**. 
+
+Templates provide a declarative, human-readable mechanism for authoring reusable UI subtrees in nested YAML or via native programmatic render functions. At generation time, an agent or Large Language Model (LLM) outputs compact, high-level template invocations (such as `UserProfile("u1", "Alice")` or `PayrollSummary("Eng", true)`). The backend SDK expands these invocations into standard, flat A2UI Basic Catalog components in a single synchronous pass before emitting standard A2UI protocol messages (`updateComponents`) to the client.
+
+This architecture achieves three fundamental objectives:
+1. **Human Readability**: Layouts are authored as clean, natural tree hierarchies in YAML or native language code, eliminating artificial IDs and flat array boilerplate.
+2. **Token Efficiency & Context Optimization**: LLMs emit concise function-like signatures rather than verbose, deeply nested UI component trees, saving prompt context and generation tokens.
+3. **Zero Client Overhead**: Client renderers (@a2ui/react, @a2ui/lit, @a2ui/angular, @a2ui/flutter) implement only standard Basic Catalog primitives (`Card`, `Column`, `Row`, `Text`, `Divider`, `Icon`, `Button`). No custom client widgets, dynamic code deployment, or renderer plugins are required.
+
+---
+
+## 1. Architectural Overview & Execution Pipeline
+
+The template engine operates entirely server-side within the A2UI Agent SDK:
+
+```
++-------------------------------------------------------------------------------+
+| 1. Authoring & Registration                                                   |
+|    - Static YAML files (with multi-doc '---' support & cross-references)      |
+|    - Dynamic resolvers (data-binding mode & programmatic AST mode)            |
++---------------------------------------+---------------------------------------+
+                                        |
+                                        v
++-------------------------------------------------------------------------------+
+| 2. Synthetic Catalog Generation                                               |
+|    - TemplateProcessor scans registered templates                             |
+|    - Synthesizes virtual A2UI Component Catalog JSON schema                   |
++---------------------------------------+---------------------------------------+
+                                        |
+                                        v
++-------------------------------------------------------------------------------+
+| 3. Model Prompt Generation & Inference                                        |
+|    - PromptGenerator injects template signatures into system prompt           |
+|    - LLM emits compact Express DSL / JSON:                                    |
+|      root = TeamRoster("Engineering", [TeamCard("Core", [...])])              |
++---------------------------------------+---------------------------------------+
+                                        |
+                                        v
++-------------------------------------------------------------------------------+
+| 4. Synchronous Template Expansion (TemplateProcessor)                         |
+|    - Resolves dynamic callbacks or executes programmatic render functions     |
+|    - Substitutes typed parameters & evaluates dot-notation paths              |
+|    - Unrolls server loops & plumbs component slot arguments                   |
+|    - Assigns deterministic synthetic IDs: {parent}_{slot}_{index}_{type}      |
++---------------------------------------+---------------------------------------+
+                                        |
+                                        v
++-------------------------------------------------------------------------------+
+| 5. Standard A2UI Protocol Emission                                            |
+|    - Emits standard createSurface and updateComponents messages               |
+|    - Payload contains 100% standard Basic Catalog primitives                  |
++---------------------------------------+---------------------------------------+
+                                        |
+                                        v
++-------------------------------------------------------------------------------+
+| 6. Client Renderer Execution                                                  |
+|    - Standard Basic Catalog renderers paint the flat component graph          |
+|    - Binds reactive two-way paths to client DataModel                         |
++-------------------------------------------------------------------------------+
+```
+
+---
+
+## 2. Template Taxonomy
+
+A2UI defines three template modalities:
+
+### A. Declarative Static Templates (YAML)
+Declarative static templates define immutable layout trees authored in YAML files. All parameter placeholders are populated directly from arguments provided by the model or caller.
+
+```yaml
+version: "0.1"
+templateId: UserProfile
+description: Standard identity card for team members.
+parameters:
+  userId:
+    type: string
+    description: Unique user identifier.
+  userName:
+    type: string
+    description: Full name of the user.
+  role:
+    type: string
+    description: Position or title within the team.
+    default: Team Member
+
+layout:
+  component: Card
+  child:
+    component: Column
+    children:
+      - component: Icon
+        name: person
+      - component: Text
+        text: ${userName}
+        variant: h3
+      - component: Text
+        text: ${role}
+        variant: caption
+```
+
+### B. Dynamic Templates: Data-Binding Mode (`resolver + layout`)
+Data-binding dynamic templates decouple public model parameters from private or live backend data. 
+- **Public Surface**: The model only sees and outputs high-level lookup parameters (e.g. `employeeId: "emp_101"`).
+- **Server Resolver**: At expansion time, a backend callback executes (e.g., querying an internal HR database or API) and produces a data dictionary.
+- **Layout Inflation**: The resolved dictionary is merged with the caller's arguments and applied to an underlying static YAML layout.
+
+```python
+# Server Registration
+dynamic_salary = DynamicTemplate(
+    template_id="EmployeeSalaryCard",
+    resolver=fetch_private_compensation,
+    layout=salary_yaml_layout,
+    description="Verified compensation card. Pass only employeeId.",
+)
+```
+
+**Security Rationale**: Confidential numbers (e.g., executive salaries, PII) never enter the model's prompt context window or inference transcript.
+
+### C. Dynamic Templates: Programmatic Render Mode (`render_fn`)
+Programmatic dynamic templates bypass static YAML layouts entirely. Instead, a developer writes a native function in the host language (Python, Dart, etc.) that accepts parameters and directly returns a UI component tree.
+
+```python
+def render_payroll_summary(department: str = "Engineering", includeBonus: bool = True) -> dict:
+    total_base = sum(e.salary for e in db.get_dept(department))
+    total_bonus = sum(e.bonus for e in db.get_dept(department))
+    
+    rows = []
+    for emp in db.get_dept(department):
+        cols = [
+            {"component": "Text", "text": emp.name, "variant": "body"},
+            {"component": "Text", "text": f"${emp.salary:,}", "variant": "body"}
+        ]
+        if includeBonus:
+            cols.append({"component": "Text", "text": f"${emp.bonus:,}", "variant": "body"})
+        rows.append({"component": "Row", "justify": "spaceBetween", "children": cols})
+
+    return {
+        "component": "Card",
+        "child": {
+            "component": "Column",
+            "children": [
+                {"component": "Text", "text": f"Payroll Summary: {department}", "variant": "h2"},
+                {"component": "Divider", "axis": "horizontal"},
+                *rows,
+                {"component": "Divider", "axis": "horizontal"},
+                {"component": "Text", "text": f"Total Budget: ${total_base + (total_bonus if includeBonus else 0):,}", "variant": "h3"}
+            ]
+        }
+    }
+
+payroll_tmpl = DynamicTemplate(
+    template_id="PayrollSummary",
+    render=render_payroll_summary,
+    description="Dynamic payroll calculation matrix.",
+)
+```
+
+**Advantages**:
+- Full host language power: arbitrary `for` loops, mathematical calculations, currency formatting, conditionals, and recursion.
+- Polymorphic return values: accepts raw dictionaries/maps, dataclasses, or typesafe fluent builder objects (`Card()`, `Column()`).
+
+---
+
+## 3. Loop Processing: Server Unrolling vs. Client Data Model
+
+A critical question in template design is: **When does the template engine unroll loops vs. when are collections handled by the client runtime?**
+
+### Comparison Matrix
+
+| Dimension | Server-Side Template Unrolling | Client-Side DataModel Loop |
+| :--- | :--- | :--- |
+| **Execution Point** | Backend SDK (`TemplateProcessor`) during template expansion. | Client Renderer runtime during paint/state updates. |
+| **Data Source** | Static YAML literals, LLM invocation arguments, or server resolver output. | Client `DataModel` tree (e.g. `/session/cart/items`). |
+| **Component Output** | Emits $N$ concrete Basic Catalog component nodes with synthesized unique IDs. | Emits a single repeater/collection component with a `DataBinding` path. |
+| **Client Requirement** | Zero. Standard Basic Catalog renderers render it like any other static UI. | Requires client-side repeater support or SDK session re-renders. |
+| **Dynamic Mutation** | Immutable once generated unless the agent emits a new `updateComponents` message. | Reactively re-renders when the client data model updates (e.g. via button events). |
+
+### Example 1: Server-Side Template Loop Unrolling
+A template author specifies a loop over a parameter array:
+
+#### Template Definition:
+```yaml
+version: "0.1"
+templateId: TeamGoalList
+parameters:
+  teamName: {type: string}
+  goals: {type: array}
+layout:
+  component: Card
+  child:
+    component: Column
+    children:
+      - component: Text
+        text: "Objectives: ${teamName}"
+        variant: h2
+      - component: Column
+        children:
+          loop:
+            param: goals
+            as: goal
+            item:
+              component: Row
+              justify: spaceBetween
+              children:
+                - component: Text
+                  text: ${goal.title}
+                - component: Text
+                  text: ${goal.priority}
+                  variant: caption
+```
+
+#### Model Invocation:
+```text
+root = TeamGoalList("Core Team", [
+  {title: "Ship Protocol", priority: "High"},
+  {title: "Write Tests", priority: "Medium"}
+])
+```
+
+#### Expanded Output Sent to Client:
+The engine unrolls the loop into concrete, synthetic child components:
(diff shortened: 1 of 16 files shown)


jacobsimionato:
Thank you for the detailed feedback! All comments have been addressed in the latest commit:

1. **LoopDefinition**: Added `anyOf: [{"required": ["template"]}, {"required": ["item"]}]` to ensure loops explicitly specify what component layout to instantiate. Also constrained `children` in `NestedTemplateNode` to prevent generic object matching.
2. **Required Parameters**: Added top-level `requiredParameters` array and supported `required` boolean in `ParameterDefinition` to declare required parameters.
3. **`headerAction` Type**: Changed `headerAction` from `type: action` to `type: child` in `section_card.yaml` to match its usage in container children.
4. **Disambiguated Template IDs**: Renamed template IDs in multi-doc bundles (`BasicUserProfile`, `BasicMetricCard`, `CompositeSectionCard`, `TwoColumnLayoutWithHeader`, `NestedTeamRoster`, `NestedTeamCard`) to ensure zero naming collisions across all example files.
5. **`headerChild` Sample Data**: Updated `two_column_layout.yaml` sampleData from `""` to a valid component node `{"component": "Text", "text": "Header Title"}`.
6. **`normalized_type` Clarification**: Explicitly clarified in `README.md` Section 7 that `normalized_type` refers to lowercase component/template names.

specification/proposals/templates/schema/template_definition.json
```diff
@@ -0,0 +1,255 @@
...
+          "$ref": "#/$defs/NestedTemplateNode",
+          "description": "Self-contained inline component layout to instantiate for each item."
+        },
+        "as": {
+          "type": "string",
+          "description": "Optional variable scope alias for accessing item fields in expressions."
+        }
+      },
+      "additionalProperties": false
+    },
```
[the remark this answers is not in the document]
jacobsimionato: Good catch! Updated `LoopDefinition` in `template_definition.json` to require `anyOf: [{"required": ["template"]}, {"required": ["item"]}]` so that every loop explicitly specifies its instantiation model. Also constrained `children` in `NestedTemplateNode` to ensure loops are prioritized and not matched as unconstrained objects.

specification/proposals/templates/schema/template_definition.json
```diff
@@ -0,0 +1,255 @@
...
+      "type": "string",
+      "description": "Human-readable description of what this template represents."
+    },
+    "parameters": {
+      "type": "object",
+      "description": "Mapping of parameter names to their explicit A2UI type definitions.",
+      "additionalProperties": {
+        "$ref": "#/$defs/ParameterDefinition"
+      }
+    },
```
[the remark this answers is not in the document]
jacobsimionato: Thanks for pointing this out! Added a top-level `requiredParameters: [...]` array to the schema, and also expanded `ParameterDefinition.required` to support a boolean so parameters can declare their own requiredness directly.

specification/proposals/templates/examples/section_card.yaml
```diff
@@ -0,0 +1,52 @@
...
+    description: Heading text displayed at the top of the section card.
+  description:
+    type: string
+    title: Section Description
+    description: Subordinate descriptive text beneath the title.
+    default: ""
+  headerAction:
+    type: action
+    title: Header Action Callback
+    description: Event action or child component triggered from the header.
```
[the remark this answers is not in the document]
jacobsimionato: Agreed! Updated `headerAction` to `type: child` with title `Header Action Component` to properly reflect that it represents a UI component rendered inside the Column container children.

specification/proposals/templates/examples/slot_composition.yaml
```diff
@@ -0,0 +1,80 @@
+version: "0.1"
+templateId: SectionCard
```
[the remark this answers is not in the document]
jacobsimionato: Good catch! Renamed the template IDs in the bundle examples to avoid any duplicate registration conflicts (`BasicUserProfile`, `BasicMetricCard`, `CompositeSectionCard`, `TwoColumnLayoutWithHeader`, `NestedTeamRoster`, `NestedTeamCard`). Now all template IDs across all example files are unique.

specification/proposals/templates/examples/two_column_layout.yaml
```diff
@@ -0,0 +1,38 @@
...
+      axis: horizontal
+    - component: Row
+      children:
+        - component: Column
+          children: ${leftChildren}
+        - component: Column
+          children: ${rightChildren}
+
+sampleData:
+  headerChild: ""
```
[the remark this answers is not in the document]
jacobsimionato: Fixed! Updated `headerChild` in `sampleData` to a valid Text component node `{"component": "Text", "text": "Header Title"}`.

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,579 @@
...
+
+```
+Let instance_id be the ID passed to expand_template(instance_id, template_id, params)
+
+1. Root Node:
+   id = instance_id
+
+2. Single Child Slot (e.g. card.child):
+   id = "{parent_id}_{slot_name}_{normalized_type}"
+   Example: "root_child_column"
```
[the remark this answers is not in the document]
jacobsimionato: Thank you! Added the explicit clarification directly in `README.md` Section 7 that `normalized_type` is the component or template type converted to lowercase (e.g., `Column` -> `column`, `UserProfile` -> `userprofile`).

specification/proposals/templates/schema/template_definition.json
gspencergoog, line 1: Would it make sense for this to be a YAML schema instead of a JSON schema?

I mean, given that the format is in YAML, that makes sense to me.

e.g. https://yaml-schema.net/

I'm not sure it actually matters much, and I think the end result is the same, so maybe it makes sense to keep all the schema in one form. It was just a thought.

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,582 @@
...
+3. **Zero Client Overhead**: Client renderers (@a2ui/react, @a2ui/lit, @a2ui/angular, @a2ui/flutter) implement only standard Basic Catalog primitives (`Card`, `Column`, `Row`, `Text`, `Divider`, `Icon`, `Button`). No custom client widgets, dynamic code deployment, or renderer plugins are required.
+
+---
+
+## 1. Architectural Overview & Execution Pipeline
+
+The template engine operates entirely server-side within the A2UI Agent SDK:
+
+```
++-------------------------------------------------------------------------------+
```
gspencergoog: This could easily be a Mermaid diagram and be a lot nicer.

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,582 @@
+# A2UI Templates Specification
```
gspencergoog: Nit: Should it maybe be "A2UI Template Specification" (singular)?

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,582 @@
...
+
+## Abstract
+
+This document defines the authoritative specification for the **A2UI Templates System**.
```
gspencergoog: Is it a System or a Specification?

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,582 @@
+# A2UI Templates Specification
+
+## Abstract
+
+This document defines the authoritative specification for the **A2UI Templates System**. 
+
+Templates provide a declarative, human-readable mechanism for authoring reusable UI subtrees in nested YAML or via native programmatic render functions. At generation time, an agent or Large Language Model (LLM) outputs compact, high-level template invocations (such as `UserProfile("u1", "Alice")` or `PayrollSummary("Eng", true)`). The backend SDK expands these invocations into standard, flat A2UI Basic Catalog components in a single synchronous pass before emitting standard A2UI protocol messages (`updateComponents`) to the client.
```
gspencergoog, line 7: The examples are all "programmatic render functions", so maybe we don't need the "nested YAML" case in the first sentence?  Besides, seems weird to next YAML inside of A2UI (or maybe that's not what that means)?

gspencergoog: I'm actually a little confused about this reading the doc: it seems like the "nested YAML" part is the template *definitions*, while the "programmatic render functions" is the *instantiations*.  This description makes it sound like they are interchangable.

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,582 @@
...
+|    - Standard Basic Catalog renderers paint the flat component graph          |
+|    - Binds reactive two-way paths to client DataModel                         |
++-------------------------------------------------------------------------------+
+```
+
+---
+
+## 2. Template Taxonomy
+
+A2UI defines three template modalities:
```
gspencergoog: Modalities feels like the wrong word.  To me that implies doing the same thing in different ways.  This seems like they are three different parts that serve different purposes.

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,582 @@
...
+     - `enum` $\rightarrow$ `{"type": "string", "enum": param.values}`
+     - `object` $\rightarrow$ `{"type": "object", "properties": ...}`
+     - `array` $\rightarrow$ `{"type": "array", "items": ...}`
+     - `child` $\rightarrow$ `{"$ref": "#/$defs/ComponentReference"}`
+     - `children` $\rightarrow$ `{"$ref": "#/$defs/ComponentReferenceList"}`
+   - It copies the `description` to provide context for the model.
+3. The synthetic catalog is passed to `PromptGenerator` (Express, Elemental, Atom, Direct JSON), which automatically writes the syntax rules and documentation into the system prompt.
+
+### Surface Isolation Boundary
+```
```
gspencergoog: This could be a Mermaid diagram too.

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,582 @@
...
+If a string field value exactly matches the parameter token regex `^\$\{([a-zA-Z0-9_]+(?:\.[a-zA-Z0-9_]+)*)\}$`:
+- The entire field is replaced by the native runtime value of the parameter.
+- **Integers remain integers**: `level: ${userLevel}` $\rightarrow$ `level: 4` (not `"4"`).
+- **Booleans remain booleans**: `checked: ${isActive}` $\rightarrow$ `checked: true` (not `"true"`).
+- **Objects and Arrays remain native structures**: `items: ${memberList}` $\rightarrow$ `items: [{...}, {...}]`.
+
+### Rule 2: Embedded String Interpolation
+If a parameter token appears alongside other characters (e.g. `"Hello, ${userName}!"` or `"${dept} - ${code}"`):
+- All tokens are replaced by their string representations.
+- Evaluates to a native string.
```
gspencergoog, line 500: Hmm.  But what about the reactive UI client side of things?  What if I want to use the `formatString` function to put a data model item into the string and have it be reactive if the data model item changes?  Can I escape the string interpolation so that is possible?

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,582 @@
...
+```
+
+The template engine replaces `${headerAction}` with the synthesized single child node, and concatenates the `${children}` component list into the body column.
+
+---
+
+## 9. Error Handling, Cycle Guards & Safety
+
+### Recursion & Circular Reference Detection
+Templates can invoke other templates. However, circular references (`A` invokes `B`, which invokes `A`) must be caught immediately to prevent stack overflow.
```
gspencergoog: What about nesting?  If I have a "Card" with a "Carousel" inside of it, containing other "Card"s, I would think that should be allowed, since they're different instances of the Card template.

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,582 @@
...
+
+- **Call Stack Tracking**: The `TemplateProcessor` maintains an active `_call_stack: Set[str]` throughout expansion.
+- **Cycle Guard**: Before expanding any template, if `template_id in _call_stack`, expansion terminates immediately with `TemplateCycleError: Circular template reference detected: A -> B -> A`.
+- **Maximum Depth Guard**: An absolute limit (`MAX_EXPANSION_DEPTH = 32`) enforces termination even in non-identical runaway recursion.
```
gspencergoog: We should standardize on the max depth defaults.  I think in other cases, we're using 50.

specification/proposals/templates/schema/template_definition.json
jacobsimionato: I read that people normally use JSON schema to represent schemas for YAML files, so I just stuck with that. I don't feel strongly though - we can change this at any time.

From search:
"The most widely used method to validate YAML files (such as GitHub Actions, Kubernetes manifests, or custom configs) is to write a standard JSON Schema but format it as a YAML file"

specification/proposals/templates/README.md
jacobsimionato: Updated the execution pipeline ASCII box diagram to a clean Mermaid flowchart diagram!

specification/proposals/templates/README.md
jacobsimionato: Updated the title to singular: `# A2UI Template Specification`.

specification/proposals/templates/README.md
jacobsimionato: Clarified phrasing to: "This document defines the authoritative specification for **A2UI Templates**."

specification/proposals/templates/README.md
jacobsimionato: Changed "modalities" to "authoring styles" (Static YAML Templates, Dynamic Data-Binding Templates, and Dynamic Programmatic Render Functions).

specification/proposals/templates/README.md
jacobsimionato: Replaced the isolation boundary ASCII art with a clean Mermaid flowchart diagram!

specification/proposals/templates/README.md
jacobsimionato: That is a great observation and highlights an inherent tension: in A2UI catalogs, the client runtime already uses `${/path}` for reactive data model interpolation inside `formatString(...)`. Using `${param}` for server-side template expansion means any client-side expressions in template layouts would either need escaping (`\${/path}`) or risk colliding.

We plan to transition template substitution to standard double curly braces `{{ param }}` (like Jinja/Mustache/Helm). That cleanly separates the two scopes without requiring escaping:
- `{{ param }}` is evaluated server-side at template expansion time.
- `${/path}` is untouched by the template engine and passes through for client-side reactive evaluation (e.g. `formatString("Hello ${/user/name}, dept: {{ dept }}")`).

For now, `\${...}` escapes any literal `${...}` so it passes through to the client intact.

specification/proposals/templates/README.md
jacobsimionato: Clarified this in Section 11! The cycle guard specifically prevents *definition-level* circular references (where a template's own layout unconditionally references itself in a cycle, causing infinite expansion).

In contrast, *instance nesting*—where a caller passes a `Card` instance into a slot of another `Card` instance (e.g. `Card(child=Carousel(items=[Card(...)]))`)—is fully supported and encouraged, as each instance has a unique synthetic ID and is bounded by caller input.

specification/proposals/templates/README.md
jacobsimionato: Updated `MAX_EXPANSION_DEPTH` default from 32 to 50 to standardize across A2UI!

jacobsimionato:

specification/proposals/templates/README.md
jacobsimionato: Re whether we need YAML at all, I think the main use case is to provide a smooth workflow from composer app to code, which people seem to like. E.g. you get the composer to create some YAML, then any SDK can use it. If we only allow the programmatic option, then the composer needs to output code, which is more of a pain, and ideally it should also be able to import code (to preview / modify existing layouts) which is probably too painful for us to pursue.

Let's discuss whether YAML is worthwhile tomorrow. I think @yjbanov maybe agrees with you that we should focus on programmatic.

Re nesting YAML inside A2UI. We're not doing that - the template process parses the YAML and then outputs pure JSON. The reason to use nested YAML here is because it's a better authoring experience for humans. Most of the time people author A2UI content and have a bad time, they should actually be authoring templates. Now we are designing templates, let's make it a human-friendly format (whether data or programmatic API).

specification/proposals/templates/README.md
jacobsimionato: Update: We have gone ahead and implemented this across the specification and all template examples! Double-curly Mustache syntax (`{{ param }}`) is now the canonical format for template parameter substitutions (with optional whitespace, e.g. `{{ param }}` or `{{param}}`).

This cleanly disambiguates compile-time template substitutions from client-side reactive expressions (`${/data/path}` and catalog functions like `formatString(...)`), ensuring client expressions pass through untouched without requiring any escaping.

specification/proposals/templates/schema/template_definition.json
```diff
@@ -0,0 +1,312 @@
...
+          "anyOf": [
+            {
+              "type": "object",
+              "required": ["loop"],
+              "properties": {
+                "loop": {"$ref": "#/$defs/LoopDefinition"}
+              },
+              "additionalProperties": false
+            },
+            {"type": "array"},
```
nan-yu, line 280: Should we limit this to be an array of `NestedTemplateNode`, rather than arbitrary array?

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,737 @@
...
+| **Expanded Component Output** | `catalogId` is **never** emitted on expanded component dicts (illegal property in v0.9).                                  | `catalogId` is automatically injected on any component whose catalog differs from surface default. |
+
+---
+
+## 6. Sub-Template Imports & Global Identity
+
+Templates decouple their local invocation name from their optional global identity:
+
+- **`name`** (Required): Component tag name used in layout trees and LLM prompts (`^[A-Za-z][A-Za-z0-9_]*$`). `templateId` is accepted as a backwards-compatible alias.
+- **`id`** (Optional): Globally unique URI or URN string (e.g. `https://company.org/templates/card/v1.json`) for package distribution and version pinning.
```
nan-yu, line 436: How do we enforce or validate the global uniqueness of template IDs? Is there a plan for a centralized template registry for discovery and resolution?

specification/proposals/templates/README.md
```diff
@@ -0,0 +1,737 @@
...
+- **`imports`** (Optional): List of global template IDs, or dictionary mapping local aliases to global template IDs (`imports: { VendorCard: "https://vendor.com/card.json" }`).
```
nan-yu, line 437: Do we allow remote network fetching at runtime, or must all imported template IDs be pre-loaded into a local registry ahead of time?

PR #2393, opened 2026-08-25
feat(react): render A2uiSurface through the node layer

andrewkolos:
Follow-up to #2077.

Flips `A2uiSurface` over to the node layer that PR added to web_core, and deletes `A2uiNodeSurface`. The surface builds a single `NodeResolver`, which turns the flat component map and the data model into a live tree of `ComponentNode`s, and each component subscribes to the props of its own node. A data write should then only re-render the component whose resolved props changed, rather than recursing from the root over component ids. `A2uiNodeSurface` only existed so the two paths could be compared, and it never shipped in a release, so deleting it outright seemed fine.

### Eager resolution

Eager resolution is the riskiest part of the change, though I doubt it matters much for the way A2UI actually gets used. The existing `A2uiSurface` resolves a child only when something renders it, so a closed `Modal`'s content and the inactive children of `Tabs` sit unresolved until they get revealed. Building the whole tree up front means those catalog functions all run when the resolver builds or reconciles the tree, and any errors they hit get reported at that point too, well before the user has touched the widget that would have shown them. `Modal` and `Tabs` are the only ones affected in the basic catalog, because they decide whether to render a child from their own React state. `Card` and `Button` decide whether to render the child from the property value, which the resolver reads anyway. Any component that decides from its own state behaves the same way, so this is a question about the node layer's contract, not about those two components. Making it lazy would mean the resolver has to know which subtrees are actually mounted, which I'd rather not build here, and whatever this PR settles on is what the other node-layer renderers inherit.

### Behavior

- A populated surface still shows its loading state on the first committed render. Its content appears after the resolver's initial update. Tests that assert immediately after mounting need to wait for that update.
- When a child arrives after its parent, replacing the placeholder causes the parent to render again. The old renderer kept that update local to the child. `adapter.test.tsx` has a test for the new expectation.
- Unknown component types and cyclic references are now reported through `onError` as `UNKNOWN_COMPONENT_TYPE` (its message reads `Unknown component type: <declared type>`) and `CYCLIC_REFERENCE`. Each is reported once per component and data path, and reported again if the condition is fixed and then reintroduced.
- `UNRESOLVED_CHILD_REFERENCE` is reported when `buildChild` cannot find the requested instance.

### API changes

The first three changes affect the node-layer API. The final change affects views that call `buildChild`.

- Repeated references to the same component receive distinct `instanceId`s, with `#n` for later occurrences (`a`, `a#2`) and the data scope appended for template instances (`a-[/items/0]`). The characters those forms are built from are escaped in the component id and data path, so an authored id cannot collide with them; ordinary ids keep their existing form.
- `ComponentNode.type` retains an unknown component's declared type, including in `toJSON`. Its `state`, rather than a sentinel type name, identifies it as a placeholder.
- `WritableBinding` now exposes the authored data path as `path`.
- `buildChild(id, basePath)` selects an instance created by the payload. It no longer creates an instance at a caller-chosen path, and a scoped template child must be requested with its `{id, basePath}` pair rather than resolving through a bare-id fallback. An unresolved notice lists the paths where matching instances exist.


renderers/react/tests/v0_9/node-surface.test.tsx
```diff
@@ -232,14 +331,142 @@ describe('A2uiNodeSurface', () => {
...
+    add(surface, 'root', 'UnmarkedParent', {child: 'kid'});
+    add(surface, 'kid', 'Text', {text: 'never reached'});
+
+    render(<A2uiSurface surface={surface} />);
+
+    expect(screen.getByText(/does not mark the referencing property/)).toBeDefined();
+    expect(screen.queryByText('never reached')).toBeNull();
+    const reported = errors.filter(e => e.code === 'UNRESOLVED_CHILD_REFERENCE');
+    expect(reported).toHaveLength(1);
+    expect(String(reported[0]?.message)).toContain('componentId()');
```
josemontespg, line 402: Would be nice to expect the entire error message, to make it easier for readers of this test to understand the expected behavior. `component()` on its own is a bit cryptic.

Same for the other tests that check on error messages to be displayed in the UI

renderers/react/tests/v0_9/node-surface.test.tsx
```diff
@@ -232,14 +331,142 @@ describe('A2uiNodeSurface', () => {
...
+    };
+    const catalog = new Catalog<ReactComponentImplementation>('scoper', [Scoper, TextImpl]);
+    const surface = new SurfaceModel<ReactComponentImplementation>('surf-scoper', catalog);
+    add(surface, 'root', 'Scoper', {child: 'kid'});
+    add(surface, 'kid', 'Text', {text: 'scoped child'});
+
+    render(<A2uiSurface surface={surface} />);
+
+    expect(screen.getByText(/instances exist at \//)).toBeDefined();
+    expect(screen.queryByText(/does not mark the referencing property/)).toBeNull();
```
josemontespg: I'm not sure what this error message means, without looking at the implementation. Can we improve this error message?

renderers/react/src/v0_9/A2uiSurface.tsx
```diff
@@ -14,119 +14,153 @@
...
+    const buildChild = useCallback<NodeBuildChild>(
+      (child, basePath) => {
+        if (isComponentNode(child)) {
+          return <NodeView key={child.instanceId} surface={surface} node={child} />;
+        }
+        // The resolver turns every child reference it can identify into a
+        // node, so a leftover id was never classified. Distinguish the two
+        // causes a catalog author can actually have.
+        const requested = basePath ?? node.dataPath;
+        const detail = surface.componentsModel.get(child)
```
josemontespg, line 81: Could we handle this case as an error type in web_core? This problem seems common to all renderers

renderers/react/src/v0_9/A2uiSurface.tsx
```diff
@@ -14,119 +14,153 @@
...
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-      [surface, id, basePath, componentModel],
-    );
-
-    const buildChild = useCallback(
-      (childId: string, specificPath?: string) => {
-        const path = specificPath || context.dataContext.path;
+    const buildChild = useCallback<NodeBuildChild>(
+      (child, basePath) => {
+        if (isComponentNode(child)) {
```
josemontespg, line 74: Would be nice to also have type narrowing for the other node cases, to help with exhaustiveness.

```typescript
if (isComponentNode(child)) {}
else if (isErrorNode(child)) {}
else if (isPlaceholderNode(child)) {}
...
else {
  checkExhaustive(child) // Fails at compile time when child is not narrowed to `never`
}
```

I'm not sure if this is possible with the current type signature of child. Maybe we will need an enum node type to make exhaustiveness checks work correctly.

renderers/react/src/v0_9/A2uiSurface.tsx
```diff
@@ -14,119 +14,153 @@
...
 
-    return <ComponentToRender context={context} buildChild={buildChild} />;
+    if (node.state === 'unknown-type') {
+      return <div style={{color: 'red'}}>Unknown component type: {node.type}</div>;
+    }
+    if (node.isPlaceholder) {
+      return <LoadingPlaceholder componentId={node.componentId} />;
+    }
+    const impl = node.impl;
+    if (!impl) {
```
josemontespg, line 106: Same here, I think this should also be an error type calculated in web_core, as this seems to be common for all renderers.

renderers/react/src/v0_9/A2uiSurface.tsx
```diff
@@ -14,119 +14,153 @@
...
+  const {context, rawBuildChild} = useNodeView(node, buildChild);
+  const Render = impl.render;
+  if (!context) {
+    return <LoadingPlaceholder componentId={node.componentId} />;
+  }
+  return <Render context={context} buildChild={rawBuildChild} />;
+};
 
-const ResolvedChild = memo(
+const NodeView = memo(
```
josemontespg, line 64: Can we extract this NodeView component into its own file? It would be nice if all the logic for rendering a node correctly could be abstracted away from the surface component. Then the surface would only need to get the root and instantiate the root node.

I think this pattern would port well to other frameworks too

renderers/react/CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 ## Unreleased
 
-- (v0_9) Add `A2uiNodeSurface`, a surface renderer driven by `NodeResolver` from `@a2ui/web_core`; component implementations gain an optional node-driven `view` (see `NodeViewProps` and `useSignalValue`) ([#2077](https://github.com/a2ui-project/a2ui/pull/2077)).
+- (v0_9) Component implementations may supply a `view` that renders from a resolved `ComponentNode` (see `NodeViewProps` and `useSignalValue`) ([#2077](https://github.com/a2ui-project/a2ui/pull/2077)).
```
josemontespg, line 3: Is the rewrite of this changelog entry intentional?

renderers/web_core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
...
+- (v0_9) The child-reference marker on `ComponentIdSchema` and `ChildListSchema` now lives in the schema's metadata, so `.describe()` and other schema-rebuilding methods no longer drop it. Hand-authored `REF:` descriptions are still recognized ([#2393](https://github.com/a2ui-project/a2ui/pull/2393)).
+
+- (v0_9) Add the node layer: `NodeResolver` resolves a surface's components and data into a live tree of read-only `ComponentNode`s, with dynamic properties resolved to `ResolvedBinding`/`WritableBinding` and distinct pending, unknown-type, and cyclic placeholder states. Sibling instance ids are always distinct, unresolvable and cyclic references are reported through `onError` once per component and data path while the condition persists, model events delivered late reconcile against current model state, and child-reference detection covers `ChildList` unions and plain arrays of component ids ([#2077](https://github.com/a2ui-project/a2ui/pull/2077), [#2393](https://github.com/a2ui-project/a2ui/pull/2393)).
```
josemontespg, line 5: Same here, is this rewrite intentional?

josemontespg (approved):
This looks good! Nice to see the Node architecture being used in React. Looking forward to the support in the other web renderers!

I left some comments, but I'm giving approval anyway because they can be addressed as follow up changes if you want. The PR in the current state is good 👏

andrewkolos:
There are some really good review comments here. That being said, I'm settling on just giving a pinky promise to follow up on these so I can merge this ASAP in interest of facilitating @josemontespg's work on universal web components.

PR #2395, opened 2026-08-25
fix(swift): safe failable initializers for basic functions without force try

pinieb:
## Summary

Replaces forced unwrapping (`try!`) with defensive schema compilation inside failable initializers across all 14 Basic Catalog functions.

## Changes

* **Failable initializers (`init?()`)**: Moves JSON Schema parsing into `public init?()` on each function class in `swift/core/Sources/BasicCatalog/Functions/`. If schema construction fails at runtime, an assertion failure is emitted and `nil` is returned instead of crashing the host process.
* **Protocol stability**: Leaves `FunctionImplementation` unchanged.
* **Test suites**: Updates unit tests in `swift/core/Tests/BasicCatalogTests/` to account for failable initialization.

## Testing

* `swift test --package-path swift/core` (all 415 tests pass)

### DIFF
--- a/swift/core/Sources/A2UICore/Models/Resolved+Conformances.swift
+++ b/swift/core/Sources/A2UICore/Models/Resolved+Conformances.swift
@@ -81,4 +81,3 @@ public struct ResolvedArray: Resolved, Equatable, Sendable {
     return true
   }
 }
-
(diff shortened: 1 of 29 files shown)


pinieb:
The bot's comment about force unwrapping being the standard for issues that could only be programmer error (i.e. hard-coded schema strings) is correct. I had initially favored that, but thought Swift had generally moved on from that with the introduction of macros like `#URL`. As it turns out, this case is very much still alive and we don't need to overcomplicate this.

PR #2398, opened 2026-08-25
refactor(python): update validation logic to support mixed catalogs

nan-yu:
- Update validation logic to support mixed catalogs
- Consolidates schema validation into PayloadValidator and surface topology validation into SurfaceComponentsModel, removing monolithic A2uiValidator.

Stacked on top of #2379.

TAG=agy


eval/a2ui_eval/scorers.py
```diff
@@ -93,8 +95,51 @@ async def score(
...
+                    or "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"
+                )
+                create_msg = {
+                    "version": ver_prefix,
+                    "createSurface": {
+                        "surfaceId": first_surface,
+                        "catalogId": cat_id,
+                    },
+                }
+                all_messages = [create_msg] + all_messages
```
nan-yu: We should not need to manipulate the generated a2ui messages. I'm going to check if we can revert these changes but still make the eval CI pass.

python/a2ui_core/src/a2ui/core/state/component_model.py
gspencergoog, line 115: When we set the properties here, do we want to overwrite them, or merge them?

python/a2ui_core/src/a2ui/core/state/component_model.py
```diff
@@ -13,8 +13,75 @@
...
+    key: str, val: Any = None, known_ids: Optional[Set[str]] = None
+) -> bool:
+    """Checks if property key or value structure matches component reference conventions or reference types."""
+    if isinstance(val, ComponentReference):
+        return True
+
+    if (
+        key in ("child", "children", "items", "components")
+        or key.endswith("Child")
+        or key.endswith("children")
```
gspencergoog: We can't do validation this way (with a fixed set of names) because it is catalog specific and tied to the basic catalog.  

The way to determine if a key is a child property is to see if the property is defined in the schema as a $ref to ChildList or ComponentId in `common_types.json`, because then the property in a custom catalog can be called "tabs" or "body" or whatever, as long as it uses a `$ref` of the right type.

python/a2ui_core/src/a2ui/core/state/surface_components_model.py
```diff
@@ -44,6 +58,65 @@ def remove_component(self, component_id: str) -> None:
...
+            prospective_components[comp_model.id] = comp_model
+
+        validate_composition_constraints(prospective_components)
+        try:
+            validate_component_integrity(
+                prospective_components, root_id=root_id, config=config
+            )
+            analyze_topology(prospective_components, root_id=root_id, config=config)
+        except Exception as e:
+            raise A2uiValidationError(str(e)) from e
```
gspencergoog: If the other calls in the try emit errors that are more specific (e.g. `A2uiIntegrityError`), then catching them and re-raising this error will lose all of the juicy details from the other errors.

python/a2ui_core/src/a2ui/core/validation/payload_validator.py
```diff
@@ -0,0 +1,573 @@
...
+class ValidationConfig(BaseModel):
+    """Configuration options for A2UI payload and component validation."""
+
+    model_config = ConfigDict(frozen=True)
+
+    allow_orphan_components: bool = False
+    allow_dangling_references: bool = False
+    allow_missing_root: bool = False
+    allow_unknown_elements: bool = False
+    target_version: Optional[str] = None
```
gspencergoog: I don't think this is ever checked.

python/a2ui_core/src/a2ui/core/validation/payload_validator.py
```diff
@@ -0,0 +1,573 @@
...
+from jsonschema import Draft202012Validator
+from ..exceptions import A2uiValidationError, A2uiErrorDetail, A2uiCatalogError
+from ..catalog.catalog import Catalog, TComponent, TFunction
+
+
+class A2uiValidatorError(A2uiValidationError):
+    """Exception raised when an A2UI Catalog payload validation fails."""
+
+
+class ValidationConfig(BaseModel):
```
gspencergoog, line 42: I think this is missing the `allowed_messages` field.

https://github.com/a2ui-project/a2ui/blob/main/agent_sdks/python/a2ui_agent/agent_development.md?plain=1#L82

python/a2ui_core/src/a2ui/core/processing/adapters/factory.py
```diff
@@ -39,8 +39,10 @@ def register_adapter(cls, adapter: VersionAdapter) -> None:
         cls._adapters[adapter.version] = adapter
 
     @classmethod
-    def get_adapter(cls, version: ProtocolVersion) -> VersionAdapter:
-        """Resolves the version adapter for the specified protocol version enum."""
+    def get_adapter(cls, version: Union[ProtocolVersion, str]) -> VersionAdapter:
+        """Resolves the version adapter for the specified protocol version enum or string."""
+        if isinstance(version, str):
+            version = cls._parse_version(version) or DEFAULT_PROTOCOL_VERSION
```
gspencergoog, line 45: Passing an unknown or malformed version here (e.g. "v99.0" or "invalid") silently falls back to DEFAULT_PROTOCOL_VERSION (v0.9) instead of raising an error.  We should probably check and make sure it matches a version regex.

python/a2ui_core/src/a2ui/core/state/component_model.py
nan-yu: I think we should overwrite them. Each component in the `updateComponents` message is a complete, self-contained component. If properties were merged, removed or obsolete properties would linger around. Especially when a component type changes (from Button to Text), merging would leave invalid properties.

gspencergoog: Yes, I agree.  There was some merging code in the web_core that I'll nuke.

python/a2ui_core/src/a2ui/core/state/component_model.py
nan-yu: I was hoping this is just a fallback because I have `if isinstance(val, ComponentReference):  return True` before this.

Here are the classes around ComponentReferences:
```python
class ComponentReference:
    """Base marker class for all A2UI component references."""
    
class SingleReference(str, ComponentReference):

class ListReference(ComponentReference):

ComponentId = SingleReference
Child = SingleReference

class TemplateChildList(StrictBaseModel, ListReference):
ChildList = Union[List[ComponentId], TemplateChildList]
```

However, it doesn't handle plain Python dictionaries well.

nan-yu: We'll need the component's schema to check if it is a child reference. That makes me think we need to build the  component reference map anyway. This also helps caching the references.

gspencergoog: Well, but it also restricts what custom component authors can call things.  For instance, they could have a "children" with kid's names on a form about family members, or a "hasChild" property that is a bool.  The `isinstance` check would fail, and the fallback would incorrectly categorize it.

I'd just leave out any key checks that are catalog specific and rely on types.

nan-yu: I ended up implementing https://github.com/a2ui-project/a2ui/blob/3cb086d73753f568d0b8881340f1ff905a74c312/python/a2ui_core/src/a2ui/core/catalog/reference_map.py for the component reference map. 

I still kept the heuristic property keys for v0.8 because it doesn't have references.

python/a2ui_core/src/a2ui/core/state/surface_components_model.py
nan-yu: Good point! I've removed the `try...except` re-wrapping block in `validate_components_update()`. 
Now `validate_component_integrity` and `analyze_topology` propagate their structured `A2uiValidationError` / `A2uiIntegrityError` instances directly with all `A2uiErrorDetail` diagnostic properties (paths, codes, and messages) intact.

python/a2ui_core/src/a2ui/core/validation/payload_validator.py
nan-yu: Yes, I was attempting to remove the target_version. I don't think we need it any more, so removed with this PR.

python/a2ui_core/src/a2ui/core/validation/payload_validator.py
nan-yu: Thanks. Added.

python/a2ui_core/src/a2ui/core/processing/adapters/factory.py
nan-yu: This was addressed by #2379.

gspencergoog (approved):

PR #2377, opened 2026-08-24
feat(swift): add Basic Catalog component schemas and registry

pinieb:
## Summary

Adds the standard Basic Catalog component schemas and registry to the Swift Core SDK, defining all 18 standard A2UI components and 14 built-in client functions across supported specification versions (v0.9, and v0.9.1).

> Note: This PR depends on the validation checks and property resolution changes in the base PR.

## Key changes

* **Component schemas (`BasicCatalogComponents.swift`)**:
  * Implemented JSON schemas for all 18 standard Basic Catalog components:
    * Layout and containers: `Row`, `Column`, `List`, `Card`, `Tabs`, `Modal`, `Divider`
    * Display: `Text`, `Image`, `Icon`, `Video`, `AudioPlayer`
    * Inputs and controls: `Button`, `TextField`, `CheckBox`, `ChoicePicker`, `Slider`, `DateTimeInput`
  * Added `BasicCatalogComponents.allComponents` to expose the full list of `ComponentAPI` definitions.

* **Function registry (`BasicFunctions.swift`)**:
  * Added `BasicFunctions.allFunctions` aggregating all 14 standard client-side functions: `and`, `email`, `formatCurrency`, `formatDate`, `formatNumber`, `formatString`, `length`, `not`, `numeric`, `openUrl`, `or`, `pluralize`, `regex`, and `required`.

* **Catalog factory and constants (`BasicCatalog.swift`)**:
  * Defined canonical catalog URIs for v0.9, and v0.9.1.
  * Added standard `themeSchema` for Basic Catalog surfaces.
  * Added `createCatalog` helper and preconfigured instances (`v09Catalog`, `v091Catalog`, `allCatalogs`).

## Verification

Ran the Swift test suite:

```bash
cd swift/core && swift test

PR #2400, opened 2026-08-25
feat(swift): Basic Catalog SwiftUI component implementations

pinieb:
## Summary

Adds SwiftUI view implementations for the 18 standard components in the A2UI Basic Catalog, along with the `BasicCatalogImplementation` registry for SwiftUI renderers.

This PR builds on top of #2377 by providing the concrete view layer that binds reactive component models to native SwiftUI views.

## Changes

* **Component views**: Added SwiftUI implementations under `swift/swiftui/Sources/BasicCatalog/`:
  * **Text & media**: `A2UIText`, `A2UIImage`, `A2UIIcon` (with SVG path parsing and SF Symbol fallbacks), `A2UIAudioPlayer`, and `A2UIVideo`.
  * **Containers & layout**: `A2UIRow`, `A2UIColumn`, `A2UIList`, `A2UICard`, `A2UITabs`, `A2UIModal`, and `A2UIDivider`.
  * **Controls & inputs**: `A2UIButton`, `A2UITextField`, `A2UICheckBox`, `A2UIChoicePicker`, `A2UISlider`, and `A2UIDateTimeInput`.
* **Catalog registry**: Added `BasicCatalogImplementation.allComponents` to export all 18 SwiftUI `ComponentImplementation` instances paired with their component APIs.
* **Theming**: Added `Theme+Extensions.swift` for applying catalog theme colors and styles to component hierarchies.
* **Data binding**: Added SwiftUI helpers in `DataBinding+SwiftUI.swift` to bind engine state to native SwiftUI input controls.
* **Test coverage**: Added unit and integration tests in `swift/swiftui/Tests/A2UISwiftUITests/` covering component rendering, state updates, SVG path parsing, and catalog resolution.

## Verification

Ran the test suite locally:

```bash
cd swift/swiftui
swift test
```


swift/core/Sources/BasicCatalog/BasicCatalog.swift
```diff
@@ -0,0 +1,74 @@
...
+          "agentDisplayName": {
+            "type": "string"
+          }
+        }
+      }
+      """
+  )
+
+  /// Creates a Catalog instance with all Basic Catalog components and functions for a specified catalog ID.
+  public static func createCatalog(
```
jacobsimionato: Can we call this `createBasicCatalog` seeing as it's kind of specific to the basic catalog? I know that all parameters can be overridden from defaults, but they all default to basic catalog settings. In general, we've tried to avoid having generic APIs and logic which depend on the basic catalog, so that developers think of the basic catalog as an example, rather than a default. We might even put the basic catalog in a separate repo some day.

swift/swiftui/Sources/BasicCatalog/A2UIRow.swift
```diff
@@ -0,0 +1,89 @@
...
+    HStack(alignment: verticalAlignment, spacing: spacing) {
+      if justify == "end" || justify == "center" || justify == "spaceAround"
+        || justify == "spaceEvenly"
+      {
+        Spacer(minLength: 0)
+      }
+
+      ForEach(Array(children.enumerated()), id: \.element.id) { index, childNode in
+        let isWeighted =
+          childNode.double(for: "weight") != nil || childNode.int(for: "weight") != nil
```
jacobsimionato, line 66: Yay, this works way better with the node architecture than it did before!

swift/swiftui/Sources/BasicCatalog/BasicCatalogImplementation.swift
```diff
@@ -0,0 +1,83 @@
...
+import BasicCatalog
+import SwiftUI
+
+/// Provides pre-configured SwiftUI component implementations for all 18 basic components.
+public enum BasicCatalogImplementation: Sendable {
+
+  /// All 18 concrete component implementations for SwiftUI.
+  public static var allComponents: [ComponentImplementation] {
+    [
+      ComponentImplementation(api: BasicCatalogComponents.text) { node in
```
jacobsimionato: While it's trivial, I think it would be best to put the `ComponentImplementation` alongside the view declaration in each file e.g. `A2uiButton.swift` etc.

That way, someone can import `A2uiButton` and import it into their own catalog without rewrapping it or extracting it from the basic catalog. It makes a nice separation of concerns - each component file contains everything needed for that component, and this file just aggregates the components.

swift/swiftui/Sources/BasicCatalog/A2UIButton.swift
```diff
@@ -0,0 +1,92 @@
...
+// limitations under the License.
+
+import A2UICore
+import A2UISwiftUI
+import OrderedJSON
+import SwiftUI
+
+/// SwiftUI component view for the A2UI Basic Catalog `Button` component.
+public struct A2UIButton: View {
+  @Environment(\.a2uiTheme) private var theme
```
jacobsimionato, line 34: This is cool!

jacobsimionato (approved):

swift/core/Sources/BasicCatalog/BasicCatalog.swift
pinieb: This isn't really intended to be a generic catalog builder, but `BasicCatalog.createBasicCatalog()` feels a little goofy at the call site. I think I'm more inclined to make it `private` since it's intended to be a helper.

I also need to update the naming to follow Swift's factory rules and call it `makeCatalog` instead of `create` 🙃

swift/swiftui/Sources/BasicCatalog/BasicCatalogImplementation.swift
pinieb: That seems reasonable to me, but with one question/potential snag: I currently have the schemas in a shared `Catalog<AnyComponentAPI>` so that we can reuse them for UIKit or AppKit (macOS) and only need to keep one copy of the schema up to date. To get to the world where each component file includes everything it needs to pick up and go, we would need to bring the schema into the view file as well, which means keeping a copy for each framework we want to support.

How are you thinking about the balance of duplication vs portability here?

I think I'm leaning towards moving the implementation definitions but continuing to share the schema.

swift/swiftui/Sources/BasicCatalog/A2UIRow.swift
pinieb: 🥳

swift/swiftui/Sources/BasicCatalog/A2UIButton.swift
pinieb: 💃

swift/core/Sources/BasicCatalog/BasicCatalog.swift
pinieb: It is now private :)

swift/swiftui/Sources/BasicCatalog/BasicCatalogImplementation.swift
jacobsimionato: Oh yes I totally agree with you! The split you have mirrors web_core structure too - we declare the APIs once at the core layer, then define implementations that extend them in each framework adapter!

I purely wanted the ComponentImplementation (which is already SwiftUI specific) to be with each View implementation, so I can define my own catalog and drop in Components from the basic catalog without needing to really instantiate a basic catalog or anything.

pinieb: Sweet! They've been rehomed.

PR #2407, opened 2026-08-26
docs(ecosystem): add an agent-side libraries section

quotentiroler:
The ecosystem page covers renderers thoroughly. There is no home for projects on the producing side, which is where `agent_sdks/` sits and where TypeScript currently has no entry.

This adds an "Agent-Side Libraries" section with one row, for a TypeScript library that emits A2UI rather than rendering it.

If you would rather it went in `community.md`, or you want it held until more than one project qualifies, say so and I will close it.

Announcement discussion: https://github.com/a2ui-project/a2ui/discussions/2406


quotentiroler:
CLA signed.

jacobsimionato (approved):
Maybe we should rename this page, but I'll approve this now so that we have your link to include in any future pages.

Your package looks interesting!!

PR #2438, opened 2026-08-28
Formalize catalog versioning and deprecated fields

wrenj:
## Summary of Changes

This pull request formalizes the versioning and deprecation strategy for A2UI catalogs. By establishing clear rules for schema evolution and providing standardized annotations for deprecation, the changes ensure that agents, renderers, and templates can evolve without disrupting existing sessions or historical data. The updates improve documentation and provide concrete examples for developers to follow when managing catalog lifecycles.

### Highlights

* **Catalog Evolution Rules**: Formalized seven key rules for catalog schema evolution, emphasizing additive changes, deprecation over deletion, and strict type invariance to ensure backward and forward compatibility.
* **Deprecation Strategy**: Introduced a standardized approach for deprecating components and properties using the `deprecated` boolean flag and an `x-deprecated-reason` string for improved developer clarity.
* **Versioning Guidance**: Clarified the usage of major version bumps in `catalogId` URIs, reserving them strictly for breaking structural changes or major cleanups.


docs/public/concepts/catalogs.md
```diff
@@ -372,46 +372,50 @@ The `catalogId` is a unique text identifier used for negotiation between the cli
...
+To maintain backward and forward compatibility, the following rules apply:
 
-- **Non-Breaking Changes (Allowable under Major Version)**  
-  Changes that can be safely ignored or degrade gracefully without breaking the layout or data model can stay at the current version.
-    - **Adding a leaf component (non-container):** e.g., adding `Badge` or `Tooltip`. If ignored, the layout remains intact.
-    - **Adding an optional property:** e.g., adding `subtitle` to a Card.
-    - **Removing a property:** Safe for the client to ignore if the agent stops sending it.
-    - **Adding new functions or styles:** These can generally be ignored without changing the semantic meaning of the component.
-    - **Metadata Changes:** Updating `description` fields or fixing typos in docs requires no version bump and has no impact on runtime.
+1. **Additive Only**: It's safe to add new components, functions, or optional properties.
```
jacobsimionato, line 381: Let say `MyCatalog` has `FooComponent`.

Then I add `BarComponent` and update my agent to use it. But some of the clients, e.g. mobile clients, don't know about `BarComponent`.

How should agents know that the client does not support `BarComponent`, so they can avoid using it? I know we don't have great answers for this right now and the solution is probably deeper changes to spec and SDKs, but is there an interim story?

Alternative: I guess if instead we made each minor version of the catalog also change the catalog ID, then there would at least be enough information exchanged between renderer and agent to negotiate these issues somewhat. E.g. probably the future looks like one catalog document with annotations on each api like `@since('v1.2')` then the clientCapabilities can automatically declare support for a specific minor version, and the agent can prune its catalog based on that.

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -700,6 +710,25 @@ Below is an annotated, fully compliant `catalog.json` schema template (written i
...
+
+A2UI catalogs should be forward and backward compatible across versions so that agents, renderers, and templates can evolve without breaking running sessions or stored transcripts.
+
+#### Catalog Schema Evolution Rules
+
+1. **Additive Only**: Adding new components, functions, or optional properties is allowed without incrementing the catalog version.
+2. **Deprecate Rather than Delete**: Never delete existing components, functions, or properties. Mark them `deprecated: true` along with an `x-deprecated-reason` instead. Renderers should retain handling logic for deprecated components and properties to maintain backward compatibility for historical transcripts, cached states, and templates.
+3. **Strict Type Invariance**: Never alter the data type of an existing field.
+4. **Open Enums**: Treat enum definitions as open so older renderers do not fail when new enum variants are introduced upstream.
+5. **Graceful Degradation**: Renderers should provide a graceful fallback for unknown components instead of failing the entire view hierarchy or surface.
```
jacobsimionato: I believe this would require updating all our renderer SDKs. I think it makes sense though.

jacobsimionato (commented):
Overall I like it, but curious what others think too.

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -700,6 +710,25 @@ Below is an annotated, fully compliant `catalog.json` schema template (written i
 }
 ```
 
+### Catalog Versioning
+
+A2UI catalogs should be forward and backward compatible across versions so that agents, renderers, and templates can evolve without breaking running sessions or stored transcripts.
+
+#### Catalog Schema Evolution Rules
+
+1. **Additive Only**: Adding new components, functions, or optional properties is allowed without incrementing the catalog version.
```
jacobsimionato: I think some of these rules don't need to be in this document, because they are best practices that don't affect the semantics of the protocol at all.

Let's just include the points that affect the contract e.g. graceful degradation, round trip preservation.

specification/v1_0/docs/a2ui_protocol.md
jacobsimionato: I think we should specify this in https://github.com/a2ui-project/a2ui/blob/main/blueprints/modules/a2ui_core.blueprint.md and https://github.com/a2ui-project/a2ui/blob/main/blueprints/modules/a2ui_framework_adapter.blueprint.md also.

Though it also belongs here, because it is related to the contract between agent and client.

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -700,6 +710,25 @@ Below is an annotated, fully compliant `catalog.json` schema template (written i
...
+2. **Deprecate Rather than Delete**: Never delete existing components, functions, or properties. Mark them `deprecated: true` along with an `x-deprecated-reason` instead. Renderers should retain handling logic for deprecated components and properties to maintain backward compatibility for historical transcripts, cached states, and templates.
+3. **Strict Type Invariance**: Never alter the data type of an existing field.
+4. **Open Enums**: Treat enum definitions as open so older renderers do not fail when new enum variants are introduced upstream.
```
jacobsimionato: I think we need to be more specific about this. Could this mean:

1. Every catalog should specify every enum as open like:

```json
    {
      "anyOf": [
        { "type": "string", "enum": ["primary", "secondary", "accent"] },
        { "type": "string" }
      ]
    }

```

2. Catalogs declare enums normally, but renderer logic needs to not fail if the received strings don't match any of the elements. E.g. catalog implementations might receive an "unknown value" sentinel or something from the SDK layer and need to handle that.


I thiiiiink I vote for the latter because it's simpler to author, though I haven't thought very hard about it. It's important we decide, because there are many places where deeply analyse these schemas, for example in the Express format converters, in the generic binder logic in renderers.

docs/public/concepts/catalogs.md
wrenj: So in this scheme it doesn't. You add BarComponent, wait for it to deploy to your mobile clients, then tell agents they can use it. If a mobile client is old then it would render some fallback. 

Alternative makes sense, and at IG we would have feature flags for a client to say they support this new feature. We could think about it as a follow-on. I don't have a good solution right now though since nobody is incrementing the catalog ID.

specification/v1_0/docs/a2ui_protocol.md
wrenj: we would do option 2

the schema should be strict for what the LLM generates. 

The non-strict stuff (open enums, ignore unknown fields) is for a2ui from another source where we dont want to crash on version skew

docs/public/concepts/catalogs.md
jacobsimionato: I think on top of our existing protocol definition, you can already do either the versioning approach described here (approach A), or the alternative that I described (approach B). B gives the agent more information, which it could use to do polyfills etc. Polyfills are totally possible using the new macros feature that we're working on, which does not need protocol changes to work.

So, I feel reluctant to prescribe A) at the protocol level, when its possible that developers could start doing a decent versioning approach on top of this. E.g. if clients report minor versions, agents can still define one catalog for inference purposes and just ensure that the agent accepts clients reporting higher or lower versions of the same catalog. 

How about:
- In the public docs, we recommend A) for now, because it's easy and matches what people are doing
- At the protocol level, we don't prescribe anything in particular
- At the SDK level, we start adding support for graceful rendering of unrecognized components, e.g. with a placeholder

Then in the future, we can add better SDK support  for representing catalogs that in a way that spans multiple "versions", without needing to change the protocol.

wrenj: sounds good, all I put in the protocol is we use x-deprecated-reason since that's an extension to JSON Schema (versus people making up different names for deprecated reason). ptal

jacobsimionato (approved):
Thank you for iterating!!

Issue #2420, opened 2026-08-27
web_core: DataModel accepts unbounded list indices, amplifying the serialized client data model


polina-c:
`web_core`'s `DataModel` accepts an arbitrarily large list index, such as `/items/999999999`. The write itself is cheap, because JavaScript arrays are sparse. The cost appears later, when that array is turned into JSON.

## Measured

Against the built `web_core` `DataModel`:

```
set('/items/1000000', 'x')
  heap growth:       ~14 KB
  array length:       1000001
  serialized length:  5,000,002 chars
```

`JSON.stringify` expands every hole to `null`. At index 10⁹ that is roughly 5 GB of JSON from a 14 KB write.

## Why it matters

The index does not have to come from the page's own code. It can come from the agent:

1. The agent creates a surface with `sendDataModel: true`. That opts the surface into sending its data model along with anything the client sends back, so the agent can see form state.
2. The agent sends `updateDataModel` with `path: "/items/999999999"`. It looks ordinary and `web_core` accepts it. Cost so far: ~14 KB.
3. Later the user does something that the client reports, for example tapping a `Button` with an `action`. Before sending, the client gathers its data model via [`MessageProcessor.getClientDataModel()`](https://github.com/a2ui-project/a2ui/blob/57491139217943f57e5e0debec75ff5dbe0485de/renderers/web_core/src/v0_9/processing/message-processor.ts#L238), which reads `dataModel.get('/')` for every surface with `sendDataModel` enabled.
4. The transport JSON-encodes that object into the [`a2uiClientDataModel`](https://github.com/a2ui-project/a2ui/blob/57491139217943f57e5e0debec75ff5dbe0485de/specification/v0_9_1/json/client_data_model.json) metadata field of the outgoing message.
5. The array reports `length: 1000000000`, so `JSON.stringify` writes `null` a billion times. The outgoing message is roughly 5 GB.

So one small, well-formed message from the agent makes the client build an enormous payload on its next turn. The client pays the cost, and the agent controls the index.

## Why the two implementations differ

[...]

PR #2408, opened 2026-08-26
Define the Dart a2ui_agent API and its tests, limited to protocol v0.9

polina-c:
Prerequisite: https://github.com/a2ui-project/a2ui/pull/2538

Contributes to https://github.com/a2ui-project/a2ui/issues/2356, https://github.com/a2ui-project/a2ui/issues/2373.

Prerequisite for https://github.com/a2ui-project/a2ui/pull/2439.

## Prompt

Implement API and tests for [Dart a2ui\_agent](https://github.com/a2ui-project/a2ui/tree/main/dart/a2ui_agent):

1. Define API that mostly throws unimplemented exceptions when invoked, unless implementation is trivial. Limit API and implementation to a2ui protocol of version 0.9 \- methods should throw an exception for messages with other or unspecified versions.  
2. Create tests that exercise this API. The tests should be skipped with the parameter [skip](https://api.flutter.dev/flutter/flutter_test/test.html):  
   1. At least one e2e test that validates primary use case defined in blueprint for a2ui\_agent  
   2. Unit tests to test cover all API endpoints

Notes:

1. Put what belongs to a2ui\_core, to a2ui\_core. If it breaks a2ui\_core tests, update tests and/or code of a2ui\_core. Update a2ui\_core API and implementation minimally, to make it possible to implement needed changes in a2ui\_agent.  
2. The tests should exercise the agent SDK against [the protocol schema of the basic catalog](https://github.com/a2ui-project/a2ui/tree/main/specification/v0_9_1/catalogs/basic), not the catalog implemented in a2ui\_core.  
3. Put data for the tests into [https://github.com/a2ui-project/a2ui/tree/main/conformance](https://github.com/a2ui-project/a2ui/tree/main/conformance/core), so that other packages could use them.  
4. Check tests for other packages in the repository. For tests that are applicable for a2ui\_agent, move the data to [https://github.com/a2ui-project/a2ui/tree/main/conformance](https://github.com/a2ui-project/a2ui/tree/main/conformance/core), so that all packages use the same datasets for testing. Example of a candidate for such test: [https://github.com/a2ui-project/a2ui/blob/delta/renderers/web\_core/src/v0\_9/state/data-model.test.ts](https://github.com/a2ui-project/a2ui/blob/delta/renderers/web_core/src/v0_9/state/data-model.test.ts) 

Sources of information:

- Protocol:  
[...]

PR #2439, opened 2026-08-28
[dart] Extend a2ui_core for agent SDKs, limited to protocol v0.9

polina-c:
Contributes to https://github.com/a2ui-project/a2ui/issues/2356, https://github.com/a2ui-project/a2ui/issues/2373.

Prerequisite for **#2408** (Dart `a2ui_agent` API), split out of it in response to [review feedback](https://github.com/a2ui-project/a2ui/pull/2408#discussion_r3876643015) asking that the agent PR carry agent work only. Everything here is a change to `a2ui_core` and necessary support in infrastructure.

### Catalog

- **[`Catalog<C extends ComponentApi, F extends FunctionApi>`](../blob/dart-a2ui-core/dart/a2ui_core/lib/src/core/catalog.dart)** — breaking, `0.1.1` → `0.2.0`. Agents parameterise with `CatalogFunction` (signature only), renderers with `FunctionImplementation`. `SchemaCatalog` aliases the agent shape.
- **`Catalog.fromJson` / `catalogSchema` / `copyWith`**, plus schema-only `CatalogComponent` and `CatalogFunction`. Catalog documents round-trip through core, and a narrowed catalog renders a narrowed document with `$defs/anyComponent` and `$defs/anyFunction` narrowed to match — so the schema knowledge lives in one place and an agent's transformers stay trivial.

### Capabilities, versions, errors

- **[`A2uiRendererCapabilities`](../blob/dart-a2ui-core/dart/a2ui_core/lib/src/core/renderer_capabilities.dart)**, mirroring [`client_capabilities.json`](../blob/main/specification/v0_9_1/json/client_capabilities.json) and web_core's `A2uiClientCapabilities`.
- **[`A2uiProtocolVersion`](../blob/dart-a2ui-core/dart/a2ui_core/lib/src/primitives/protocol_version.dart)** and the [`A2uiParseError` / `A2uiCompileError` / `A2uiCatalogError` / `A2uiIntegrityError` / `A2uiRecursionError`](../blob/dart-a2ui-core/dart/a2ui_core/lib/src/primitives/errors.dart) categories.
- **[`A2uiValidator`](../blob/dart-a2ui-core/dart/a2ui_core/lib/src/validation/validator.dart)**, with the v0.9 version gate implemented and the structural / catalog-schema checks declared but stubbed.

### One bug fixed

[`DataModel.set`](../blob/dart-a2ui-core/dart/a2ui_core/lib/src/core/data_model.dart) silently dropped a write whose parent path resolved to a primitive (`/user/name/first` where `/user/name` is a string). It now throws `A2uiDataError`, matching web_core. The shared dataset below surfaced it.

Notification on an unchanged value is fixed the same way: `DataModel` hands the signal a copy of a container rather than bypassing the equality check, so an observer whose own value did not change is no longer woken. Both behaviours moved *into* the shared suite rather than being excluded from it.

## `conformance/` — shared data

- **[`core/data_model.yaml`](../blob/dart-a2ui-core/conformance/core/data_model.yaml)** (new, `data_model` action): 37 cases migrated from [`renderers/web_core/src/v0_9/state/data-model.test.ts`](../blob/main/renderers/web_core/src/v0_9/state/data-model.test.ts).
- **[`core/message_processor.yaml`](../blob/dart-a2ui-core/conformance/core/message_processor.yaml)** (new, `process_messages` action): the [v1_0 branch's suite](../blob/v1_0/conformance/core/message_processor.yaml) of the same name is the primary one, with 51 cases. Only the two it does not cover are here — component and data-model **isolation between surfaces**, which it opens two surfaces for twice and asserts in neither — written in the v1_0 case vocabulary (`messages`, `catalogPaths`, `expect.surfaces`) so the two files concatenate instead of conflicting when that branch lands.
- [`conformance_schema.json`](../blob/dart-a2ui-core/conformance/conformance_schema.json) gains the two actions and the `DataError` category; the file's existing formatting is preserved.

## `renderers/web_core` — consuming the shared data

- [`tests/conformance/harness.ts`](../blob/dart-a2ui-core/renderers/web_core/tests/conformance/harness.ts) (new) locates `conformance/` by walking up from the test file and loads a suite by name.
- [`data-model.conformance.test.ts`](../blob/dart-a2ui-core/renderers/web_core/tests/conformance/data-model.conformance.test.ts) and [`message-processor.conformance.test.ts`](../blob/dart-a2ui-core/renderers/web_core/tests/conformance/message-processor.conformance.test.ts) run the two suites. Both are **additive**: the hand-written [`data-model.test.ts`](../blob/main/renderers/web_core/src/v0_9/state/data-model.test.ts) and [`message-processor.test.ts`](../blob/main/renderers/web_core/src/v0_9/processing/message-processor.test.ts) are untouched.
- Adds the `yaml` devDependency.

Each harness builds catalogs natively from the case's catalog id, because neither renderer builds a catalog from JSON Schema.

### Divergences the migration surfaced

Comparing two live implementations turned up disagreements that are decisions, not formatting:

1. **List index caps.** Dart lists are dense, so `/items/999999999` must be rejected; JavaScript arrays are sparse, so the same write is cheap and allowed. Sparse arrays protect web_core's heap but not its serialized size, which makes this an amplification vector when `sendDataModel` ships the model back to the agent — filed as **#2420** rather than changed here, since capping accepted input for a shipped renderer is a product decision, not a side effect of a test migration.
2. Language-level differences stay excluded from the suite: prototype-pollution guards, `null`/`undefined` path arguments, `undefined` versus a removed key, leading-zero indices.

**These want a maintainer decision.**

## `blueprints/modules/a2ui_core.blueprint.md`

The gaps that produced the review findings on #2408, written down so the next generated implementation does not repeat them:

- **Package Boundary & Non-Goals** — `a2ui_agent` depends on `a2ui_core` and never the reverse; core does not own catalog narrowing, prompt generation, response parsing or capability negotiation; a core change made to unblock an agent should be minimal and land as its own reviewable unit (this PR).
- **Catalog** — catalogs are immutable, and `protocolVersion` is a property of each catalog rather than of the process.
- **Capabilities Objects** (new) — both directions are version-keyed maps with a normative JSON Schema, and the filenames differ per version (`client_capabilities.json` / `server_capabilities.json` in v0.9, `renderer_capabilities.json` / `agent_capabilities.json` in v1.0, a third spelling in v0.8).
- **Exceptions** — completes the hierarchy, which listed five classes while implementations use ten, and adds the rule that parsing wire JSON raises from it rather than throwing a raw `TypeError` / `ClassCastException`.
- **Conformance Test Plan** (new — core had none) — the `core/` vs `agent/` split is about ownership, not subject matter, and an existing suite must be looked for on every active spec branch before a new one is authored.

## CI

[`flutter_packages_test.yml`](../blob/dart-a2ui-core/.github/workflows/flutter_packages_test.yml) now discovers packages under `dart/` as well as `samples/`. The `dart/` packages were not built, analyzed or tested by any workflow before this.

## `dart/a2ui_agent`

Two lines, both consequences of the above rather than agent work:

- the `a2ui_core` constraint follows the major bump (`^0.1.1` → `^0.2.0`);
- the unnecessary `library;` directive in the stub is removed, which the new CI coverage would otherwise fail on.

## Verification

All run locally:

| Check | Result |
| --- | --- |
| `a2ui_core`: format, `dart analyze --fatal-infos`, `flutter test` | clean, 261 passed / 25 for v0.8 skipped |
| `a2ui_agent` (stub): `dart analyze --fatal-infos`, `flutter test` | clean, 1 passed |
| `web_core`: `yarn test`, `yarn lint` | 389 passed, 0 errors |
| `conformance`: `pytest` (suite self-validation) | 10 passed |
| `yarn build:all` | pass |
| `prettier --check`, `validate_blueprints.py` | clean |


dart/a2ui_core/lib/src/core/data_model.dart
```diff
@@ -182,9 +190,15 @@ class DataModel {
...
-    sig.set(newValue, force: true);
+    // A container mutated in place keeps its identity, so the live object
+    // would compare equal and suppress the notification. Hand over a copy,
+    // and let the signal's equality check suppress genuinely unchanged
+    // values; notifying unconditionally would wake unaffected observers.
+    sig.set(switch (newValue) {
+      final Map<String, Object?> map => Map<String, Object?>.of(map),
+      final List<Object?> list => List<Object?>.of(list),
+      _ => newValue,
+    });
```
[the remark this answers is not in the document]
polina-c: Applied, with one change to the suggestion and one correction to the rationale.

Confirmed by probe before the fix: writing a `Map<dynamic, dynamic>` fell through to `_ => newValue`, so the signal saw the same live object and **the notification was silently lost** — `fired=0`. That makes this the worst of the five: no exception, just a missed observer.

Two refinements:

- **`Map<String, Object?>` was never affected.** `Object?` and `dynamic` are mutual subtypes in Dart, so `Map<String, Object?>` already matched the old pattern (`fired=1`). Only `Map<dynamic, dynamic>` — a bare `{}` literal, or YAML — missed. The `List<Object?>` arm was never broken either, by the same covariance.
- **The bare `Map` / `List` patterns fail this repo's analyzer.** `strict_raw_type` is enabled, so `final Map map` is a warning and `dart analyze --fatal-infos` rejects it. Used `Map<Object?, Object?>` and `List<Object?>` instead, which match everything the raw form does.

Also kept the `Map<String, Object?>` arm ahead of the new one so the common case still copies to a String-keyed map, and avoided `map.cast<String, Object?>()` — that throws on a map with non-String keys, which `DataModel` does not otherwise forbid.

[data_model.dart](https://github.com/a2ui-project/a2ui/blob/67373ba2/dart/a2ui_core/lib/src/core/data_model.dart#L197-L208) — after the fix all three map shapes notify.

dart/a2ui_core/lib/src/core/messages.dart
```diff
@@ -101,6 +95,86 @@ abstract class A2uiMessage {
...
+Map<String, dynamic> _body(Map<String, dynamic> json, String key) {
+  final Object? body = json[key];
+  if (body is! Map<String, dynamic>) {
+    throw A2uiValidationError(
+      "Message body '$key' must be an object.",
+      details: json,
+    );
+  }
+  return body;
+}
```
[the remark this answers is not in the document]
polina-c: Applied as suggested.

Confirmed by probe: a `Map<dynamic, dynamic>` body was rejected with

```
A2uiValidationError: Message body 'createSurface' must be an object.
```

for a value that *is* an object — a false rejection, and a misleading message. Now `body is! Map` followed by `body.cast<String, dynamic>()`.

One correction to the rationale: `Map<String, Object?>` was never affected, since `Object?` and `dynamic` are mutual subtypes and Dart's `Map` is covariant. Only `Map<dynamic, dynamic>` missed.

Verified the negative case still holds — a non-map body is rejected exactly as before.

[messages.dart](https://github.com/a2ui-project/a2ui/blob/67373ba2/dart/a2ui_core/lib/src/core/messages.dart#L99-L108)

dart/a2ui_core/lib/src/core/messages.dart
```diff
@@ -101,6 +95,86 @@ abstract class A2uiMessage {
...
+      if (entry is Map<String, dynamic>)
+        entry
+      else
+        throw A2uiValidationError(
+          "Field '$messageType.components' must hold objects, got "
+          '${entry.runtimeType}.',
+          details: body,
+        ),
+  ];
+}
```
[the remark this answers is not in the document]
polina-c: Applied as suggested.

Confirmed by probe: a component entry given as `Map<dynamic, dynamic>` was rejected with

```
A2uiValidationError: Field 'updateComponents.components' must hold objects, got _Map<dynamic, dynamic>.
```

The error text naming the map type it had just refused to accept as a map made the defect fairly self-evident. Now `entry is Map` followed by `entry.cast<String, dynamic>()`.

Verified the negative case still holds — `components: ['nope']` is still rejected, now reporting `got String`.

[messages.dart](https://github.com/a2ui-project/a2ui/blob/67373ba2/dart/a2ui_core/lib/src/core/messages.dart#L164-L175)

dart/a2ui_core/lib/src/core/catalog.dart
```diff
@@ -60,18 +65,334 @@ abstract class FunctionImplementation extends FunctionApi {
...
+        "Catalog 'functions' must be an object or a list of definitions.",
+        catalogId: catalogId,
+      );
+    }
+    final functions = <CatalogFunction>[];
+    for (final MapEntry<Object?, Object?> entry in raw.entries) {
+      final Map<String, Object?> schema = _asSchemaMap(entry.value);
+      final Map<String, Object?> properties =
+          schema['properties'] as Map<String, Object?>? ??
+          const <String, Object?>{};
```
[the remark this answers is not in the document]
polina-c: Applied, and taken a step further.

Confirmed by probe — this one escaped the error hierarchy entirely:

```
_TypeError: type '_Map<dynamic, dynamic>' is not a subtype of type 'Map<String, Object?>?' in type cast
```

That is exactly what the rule this PR adds to the core blueprint forbids: parsing a catalog document must raise from `A2uiError`, never a raw language-level cast failure. A caller catching `A2uiError` around `Catalog.fromJson` would have missed it.

Rather than defaulting a malformed `properties` to `const {}`, which hides a bad catalog, the three cases are now distinguished:

- absent → empty, as before;
- any `Map` → `cast<String, Object?>()`;
- anything else → `A2uiCatalogError` naming the function and the offending type.

That matches `_asSchemaMap` directly below it, which already used the `is Map` + `cast` idiom this comment asks for. Probe after the fix: a dynamic-typed `properties` parses, and `'properties': 'nope'` raises `A2uiCatalogError: Catalog function 'required' has a non-object 'properties' (got String).`

Used `Map<Object?, Object?>` rather than a bare `Map` in the pattern — `strict_raw_type` is enabled here, so the raw form fails `dart analyze --fatal-infos`.

[catalog.dart](https://github.com/a2ui-project/a2ui/blob/67373ba2/dart/a2ui_core/lib/src/core/catalog.dart#L244-L259)

dart/a2ui_core/lib/src/validation/validator.dart
```diff
@@ -0,0 +1,340 @@
...
+        final Map<String, Object?> document = catalog.catalogSchema;
+        return {
+          for (final MapEntry<String, C> entry in catalog.components.entries)
+            entry.key: Schema.fromMap(
+              resolveSchemaRefs(
+                entry.value.schema.value,
+                document,
+                commonTypes: commonTypesSchema,
+              ),
+            ),
```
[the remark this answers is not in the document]
polina-c: Not applied — this one does not reproduce, unlike the other four in this review.

`entry.value.schema.value` is not dynamically typed. `Schema.value` is declared `Map<String, Object?>` in `json_schema_builder`, so the call to `resolveSchemaRefs` is statically sound and there is no cast to fail.

Probed it directly with a catalog whose component schemas are `Map<dynamic, dynamic>` all the way down — the shape a bare `{}` literal and YAML both produce:

```
schema.value runtimeType: CastMap<dynamic, dynamic, String, Object?>
validate: OK
```

`Catalog._asSchemaMap` already calls `.cast<String, Object?>()` when parsing the document, so what reaches the validator is a `CastMap` that both statically and dynamically **is** a `Map<String, Object?>`. The suggested `(… as Map).cast<String, Object?>()` would wrap that `CastMap` in a second one and change nothing.

Happy to revisit with a concrete failing input if you have one.

[validator.dart](https://github.com/a2ui-project/a2ui/blob/67373ba2/dart/a2ui_core/lib/src/validation/validator.dart#L326-L339)

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -193,6 +204,12 @@ export interface Catalog<TComponent extends ComponentApi, TFunction extends Func
 }
 ```
 
+A `Catalog` is immutable once constructed.
+
+`protocolVersion` is a property of **each catalog, not of the process**. A renderer or an agent may hold catalogs for several protocol versions at once.
```
nan-yu: @jacobsimionato has a proposal to make catalogs version-agnostic by removing the protocol version from catalog in both the specification and SDKs.

It looks feasible overall, but here are a couple of caveats to consider:

- Identifier Validation: v1.0 validates catalog identifiers (component and function names) against UAX#31. If catalogs are version-agnostic, we’d need to apply this check across all catalogs uniformly.
- Feature Parity & Serialization: If a catalog uses newer schema constructs that cannot be represented in older wire formats, serializing to legacy protocol versions will need clear fallback or error-handling behavior.

I'd suggest not adding this to the blueprint.

dart/a2ui_agent/lib/a2ui_agent.dart
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-library;
```
nan-yu, line 15: nit: Do we want to check in the a2ui_agent change in this PR?

dart/a2ui_core/lib/src/core/catalog.dart
```diff
@@ -60,18 +65,341 @@ abstract class FunctionImplementation extends FunctionApi {
...
+/// [C] is the component representation and [F] the function representation:
+/// renderers use [FunctionImplementation], agents [CatalogFunction].
+class Catalog<C extends ComponentApi, F extends FunctionApi> {
+  /// The catalog id, from the document's `catalogId` field.
   final String id;
-  final Map<String, T> components;
-  final Map<String, FunctionImplementation> functions;
+
+  /// The protocol version this catalog conforms to.
+  final A2uiProtocolVersion protocolVersion;
```
nan-yu: Suggest not adding this. If we have to associate the protocol version with the catalog, we need to re-consider if catalogs can be version-agnostic.

dart/a2ui_core/lib/src/core/renderer_capabilities.dart
```diff
@@ -0,0 +1,147 @@
...
+
+/// The rendering capabilities a renderer advertises, mirroring
+/// `a2uiClientCapabilities` in `client_capabilities.json` and web_core's
+/// `A2uiClientCapabilities`.
+///
+/// A capabilities object without a `v0.9` entry is rejected. Other versions
+/// are kept in [unsupportedVersions] but never negotiated against.
+class A2uiRendererCapabilities {
+  /// The capabilities declared for v0.9.
+  final A2uiVersionCapabilities v0_9;
```
nan-yu: I understand this is currently built on v0.9, but we should make it flexible enough to handle multiple protocol versions (v0.9, v1.0, and beyond).

In Python and TypeScript, we use versioned packages to manage protocol differences, so aligning this work with that multi-version approach will save us from major refactors later.

dart/a2ui_core/lib/src/validation/validator.dart
```diff
@@ -0,0 +1,340 @@
...
+        await _validateComponent(component, catalog);
+      }
+    }
+  }
+
+  /// Validates a complete payload: envelopes, then structure, then catalog
+  /// schemas.
+  ///
+  /// Returns the parsed messages, and throws as the individual steps do.
+  Future<List<A2uiMessage>> validate(List<Map<String, Object?>> payload) async {
```
nan-yu: We should update MessageProcessor to serve as the primary entry point for these validations.

While working on the Python SDK, I found that validation naturally occurs in fragmented phases rather than a single pass:
- Envelope validation runs first when parsing the raw protocol message.
- Theme validation executes while handling createSurface messages. 
- Component validation executes while handling updateComponents messages.

Because of this flow, we can eliminate the top-level `validate()` method from a standalone central validator and let MessageProcessor drive the validation lifecycle directly. An architectural advantage is that MessageProcessor is stateful—it already retains existing surfaces and components, which are required for verifying structural integrity (such as catching orphan components, dangling references, etc.).

dart/a2ui_agent/lib/a2ui_agent.dart
polina-c: nope: it is response to lint.

dart/a2ui_core/lib/src/core/catalog.dart
polina-c: Nan, I think catalogs schemas are different by version. 

Do we consider them to be both backward and forward compatible? If yes, we need to say it somewhere: in glossary or in blueprint.

nan-yu: If we want to make catalogs version-agnostic, we may not always be able to support backward or forward compatibility across versions.

The workflow in my mind is like:
### On the agent side:
1. Agent developers define a list of supported catalogs  (version-agnostic).
2. Agent developers define a list of supported protocol versions. (This is a new step because protocol versions were implied in the catalogs previously.)
3. Agent developers define a list of examples (version-agnostic). The examples do have the protocol version (v0.9, v1.0) in the A2UI messages, but we can mutate the version field with the supported protocol versions. This helps developers avoid maintaining version-specific but duplicated examples.
4. `A2uiGenerator` validates if catalogs conform to the catalog definition schema with the protocol versions. If not, fail fast. During the conversion, there're the following scenarios:
    - Unrecognized properties (components, component properties, functions, function args) should be dropped, for example, the `theme` field in v0.9 should be dropped when targeting at v1.0, or `allowed_parents` field in v1.0 should be dropped when targeting at v0.9.
    - Known incompatibilities are converted seamlessly, for examples, `styles` in v0.8 should be converted to `theme` automatically when targeting at v0.9.
    - Unknown incompatibilities should fail the validation.
5. When receiving renderer request, resolve catalogs based on the protocol version and supportedIds from the renderer capabilities. This requires updating the `resolve_catalogs` helper function to handle both catalogIds and protocol versions.
6. `A2uiRequestProcessor` updates the `version` field in the examples to match the resolved protocol version, and validates if examples conform to the resolved catalogs.
7. Inferring, parsing, validating and delivering messages to the renderer.

### On the renderer side:
1. App developers define a list of supported catalogs (version-agnostic).
2. App developers define a single protocol version. (Question: single or multiple versions? Do we allow a surface to have components in different protocol versions?). This is also a new step.
3. Do we want to validate if the catalogs conform to the specific protocol version on the client side? Probably yes, and the meta validation should happen in the MessageProcessor initialization stage.
4. MessageProcessor builds the renderer capabilities.
5. App developers send request to agent with the renderer capabilities built in step 4.
6. MessageProcessor processes messages returned from the agent, processing, validating, and mutating state.
7. Renderer framework renders the A2UI messages.


@jacobsimionato @polina-c @gspencergoog Please let me know if the workflow looks good in terms of making catalogs version-agnostic? This requires additional updates in the agent SDK.

polina-c: Yes, let's go with version agnostyic catalog for now. In future we can switch to declaring range of versions (like ">=9.0 <1.9.9").

dart/a2ui_core/lib/src/core/catalog.dart
```diff
@@ -60,18 +65,341 @@ abstract class FunctionImplementation extends FunctionApi {
...
+
+  final Map<String, C> components;
+  final Map<String, F> functions;
   final Schema? themeSchema;
 
+  /// The document this catalog was parsed from, if any.
```
nan-yu: This caches the source schema. https://github.com/a2ui-project/a2ui/issues/1626 proposes to remove the cache, but compute it dynamically. 

https://github.com/a2ui-project/a2ui/pull/2458 is the implementation for the Python version. We need to handle it in a similar way in Dart.

dart/a2ui_core/lib/src/core/renderer_capabilities.dart
polina-c: Agreed — reshaped so the version is data rather than structure.

The hardcoded `v0_9` field is gone. `A2uiRendererCapabilities` now holds a
[`Map<A2uiProtocolVersion, A2uiVersionCapabilities>`](https://github.com/a2ui-project/a2ui/blob/5440a986/dart/a2ui_core/lib/src/core/renderer_capabilities.dart#L98),
so a renderer can advertise several versions at once:

- [`fromJson`](https://github.com/a2ui-project/a2ui/blob/5440a986/dart/a2ui_core/lib/src/core/renderer_capabilities.dart#L126)
  parses every version key it recognises instead of reaching for `v0.9`. A key
  it does not recognise is not an error — it goes to `unsupportedVersions`, and
  the object is rejected only when *no* implemented version is present.
- [`forVersion`](https://github.com/a2ui-project/a2ui/blob/5440a986/dart/a2ui_core/lib/src/core/renderer_capabilities.dart#L166)
  returns a nullable rather than switching over a closed set.
- Added [`A2uiProtocolVersion.tryParse`](https://github.com/a2ui-project/a2ui/blob/5440a986/dart/a2ui_core/lib/src/primitives/protocol_version.dart#L64)
  for the non-throwing lookup, and routed `fromJson` through it so the two do
  not each scan `values`.

dart/a2ui_core/lib/src/validation/validator.dart
polina-c: Right now it is not possible - falidation is async in Dart at the moment.

Drafted [PR for json_schema_builder](https://github.com/flutter/genui/pull/1029) to address this.

blueprints/modules/a2ui_core.blueprint.md
jacobsimionato: I think my proposal requires more thought and that it will create too much confusion and churn to rethink how we do catalogs prior to 1.0. So if we have protocolVersion on the catalog in Python, I'm fine with formalizing it here and leaving it everywhere.

Edit: reposting this below on the main comment thread

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -20,15 +22,24 @@ The A2UI Core SDK acts as the central state coordinator. It is designed to repre
...
 5. **JSON Pointer Scope:** Standardize relative pointer evaluation and reactivity via scoped context managers.
 6. **Validation:** Performs structural JSON Schema checks, reference checks, loop/recursion analysis, and layout integrity checks.
 7. **Resolution:** Resolves bound context paths and binds state variables to components for local evaluation.
 8. **Multi-Version Protocol Branching:** Supports multiple versions of the protocol.
 
+#### Package Boundary & Non-Goals
+
+Not in core:
+
+- `CatalogTransformer`, `ComponentPruningTransformer`, `FunctionPruningTransformer`, and the `CatalogConfig` pipeline that applies them
```
jacobsimionato: I worry it won't be maintainable to exhaustively list everything that won't be in core here. Instead, how about just having a link to the agent SDK and framework adapter blueprints and mention that core should not contain functionality that is designated for those layers.

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -333,7 +350,7 @@ myCustomCatalog = Catalog(
 
 #### `MessageProcessor`
 
-The "Controller" that accepts the raw stream of A2UI messages, parses them, and mutates the Models. It also handles the aggregation of renderer state for synchronization.
+The "Controller" that accepts the raw stream of [**A2UI messages**](../../docs/public/concepts/glossary.md#a2ui-message), parses them, and mutates the Models. It also handles the aggregation of renderer state for synchronization.
```
jacobsimionato, line 349: I like that we're going to encourage agents to read the glossary! I could imagine that improving their performance!

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -370,6 +387,12 @@ When a surface is created with `sendDataModel: true`, the renderer is responsibl
...
 - **Component Lifecycle**: If an `updateComponents` message provides an existing `id` but a _different_ `type`, the processor MUST remove the old component and create a fresh one to ensure framework renderers correctly reset their internal state.
 
+#### [Capabilities Objects](../../docs/public/concepts/glossary.md#capabilities-object)
+
+Both sides advertise their capabilities to each other.
+
+Schemas live in `specification/<version>/json/`. v1.0 names the pair [`renderer_capabilities.json`](../../specification/v1_0/json/renderer_capabilities.json) and [`agent_capabilities.json`](../../specification/v1_0/json/agent_capabilities.json). v0.9 and v0.9.1 name the same pair [`client_capabilities.json`](../../specification/v0_9_1/json/client_capabilities.json) and [`server_capabilities.json`](../../specification/v0_9_1/json/server_capabilities.json), carried as `a2uiClientCapabilities` and `a2uiServerCapabilities`. v0.8 spells it differently again ([`a2ui_client_capabilities_schema.json`](../../specification/v0_8/json/a2ui_client_capabilities_schema.json)) and publishes no server-side counterpart.
+
 #### Generating Renderer Capabilities and Schema Types
```
jacobsimionato, line 392: I think we could ignore v0.8

conformance/core/data_model.yaml
jacobsimionato, line 1: This is a fantastic test!!!

dart/a2ui_core/lib/src/core/catalog.dart
```diff
@@ -60,18 +64,322 @@ abstract class FunctionImplementation extends FunctionApi {
   ]);
 }
 
+/// A [ComponentApi] backed by a catalog document's JSON schema.
+///
+/// Produced by [Catalog.fromJson]; carries no rendering behaviour, so it is
+/// the agent-side representation.
+class CatalogComponent implements ComponentApi {
```
jacobsimionato: Why do we need this and CatalogFunction, when they are identical to ComponenApi and FunctionApi?

dart/a2ui_core/lib/src/core/catalog.dart
jacobsimionato: +1 - let's not cache this here. The component and function objects should be the single source of truth.

blueprints/modules/a2ui_core.blueprint.md
polina-c: Dropped and explained in "Package Boundary & Non-Goals".

dart/a2ui_core/lib/src/validation/validator.dart
polina-c: Okay, merged it.

Done — `MessageProcessor` now drives validation, in the phases you described:

- [`processPayload`](https://github.com/a2ui-project/a2ui/blob/3483953b/dart/a2ui_core/lib/src/processing/processor.dart#L63)
  parses a raw payload and validates envelopes as it goes.
- [Creating a surface](https://github.com/a2ui-project/a2ui/blob/3483953b/dart/a2ui_core/lib/src/processing/processor.dart#L101)
  checks its theme against the catalog's theme schema.
- [Each arriving component](https://github.com/a2ui-project/a2ui/blob/3483953b/dart/a2ui_core/lib/src/processing/processor.dart#L145)
  is checked against its surface's catalog before the batch is applied.

Validation is on by default; pass a `validator` only to configure it. The
conformance harness now builds a real catalog instead of an empty one, so the suite exercises this.

I kept `A2uiValidator.validate` rather than removing it: agents validate
outbound payloads and have no `MessageProcessor`, and `conformance_schema.json` still declares `validate` as an action alongside `process_messages`.

dart/a2ui_core/lib/src/core/catalog.dart
polina-c: Done.

- `Catalog.fromJson` now copies the document's own `#/...` definitions into
  each component and function schema as it parses, the way `inline_local_refs` does in #2458. Only `#/...` references are copied in. The `common_types.json` ones are left for the validator.
- `_sourceSchema` is gone. `catalogSchema` rebuilds the document from the components and functions the catalog holds, anyComponent` and `anyFunction` included, so removing a component removes it from the document too.

dart/a2ui_core/lib/src/core/catalog.dart
polina-c: They are not identical. `ComponentApi` and `FunctionApi` are abstract and declare getters only — no fields, no constructor. `Catalog.fromJson` parses a document at runtime and has to instantiate something, and an abstract class cannot be instantiated. `CatalogComponent` and `CatalogFunction` are the concrete holders: `final` fields plus a constructor.

Making the abstract classes concrete instead would break the existing implementations, which compute rather than store. `MinimalTextApi.schema` builds a `Schema.object(...)` on each access and holds no field, and every hand-written component in the repo is written that way.

`CatalogFunction` also carries a member `FunctionApi` does not: `description`, read from the document.

The split is load-bearing for the second type parameter, too. `CatalogFunction` deliberately does not implement `FunctionImplementation`, so `SchemaCatalog` is statically "signatures, nothing runnable" — what an agent works with. A renderer takes `Catalog<C, FunctionImplementation>`, so the compiler stops a catalog parsed from JSON from reaching code that will call `execute()`.

Does it help?

.claude/CLAUDE.md
nan-yu, line 1: Do we want to keep a skills folder for each coding agent?

blueprints/modules/a2ui_core.blueprint.md
nan-yu: We don't have any Dart/Flutter users on v0.8, so it is okay to drop it for Dart. However, the removal of v0.8 from the blueprint also impacts other SDKs. The Python and TypeScript SDKs need to support v0.8 for backward compatibility. Should we make the drop limited to Dart SDK?

dart/a2ui_core/lib/src/core/catalog.dart
nan-yu: I don't quite understand `Making the abstract classes concrete instead would break the existing implementations, which compute rather than store`.

I feel like it is fine to make these abstract classes `ComponentApi` and `FunctionApi` concrete classes with standard generative constructor. Both Python and TypeScript work in a similar way.

dart/a2ui_core/lib/src/core/catalog.dart
```diff
@@ -60,18 +65,289 @@ abstract class FunctionImplementation extends FunctionApi {
...
+    }
+
+    // Local references are expanded here, once, so each component and function
+    // schema stands alone afterwards. The document is then no longer needed,
+    // and [catalogSchema] rebuilds it from the parts rather than caching it.
+    final document = inlineLocalRefs(json, json)! as Map<String, Object?>;
+
+    return SchemaCatalog(
+      id: rawId,
+      components: _parseComponents(document['components'], rawId),
```
nan-yu, line 173: The Python and TypeScript version also filters those components defined under `$defs/anyComponent`. That's the list referenced by https://github.com/a2ui-project/a2ui/blob/main/specification/v0_9/json/server_to_client.json#L68.

Any components are not in anyComponent.oneOf should be ignored.

Similar for the functions.

dart/a2ui_core/lib/src/core/catalog.dart
```diff
@@ -60,18 +65,289 @@ abstract class FunctionImplementation extends FunctionApi {
...
+  /// The catalog document for this catalog, as JSON.
+  ///
+  /// The catalog as a document, rebuilt from the components and functions it
+  /// currently holds.
+  ///
+  /// Nothing is cached: a pruned catalog renders a pruned document, with the
+  /// `anyComponent` and `anyFunction` unions covering exactly what is left.
+  /// Component and function schemas carry their local definitions inline, so
+  /// the document needs no `$defs` beyond the theme and those two unions.
+  Map<String, Object?> get catalogSchema => {
```
nan-yu, line 284: I'm curious to see how closely the computed catalog schema matches the basic catalog schema.

https://github.com/a2ui-project/a2ui/blob/v1_0/conformance/core/catalog.yaml#L672-L691 has conformance tests. The python conformance test runner runs the test and confirms the computed version (https://github.com/a2ui-project/a2ui/blob/v1_0/conformance/test_data/converted_basic_catalog_v09.json) is self-contained and matches the v0.9 basic catalog mostly.

dart/a2ui_core/lib/src/validation/common_types.g.dart
```diff
@@ -0,0 +1,326 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Generated by tool/generate_common_types.dart. Do not edit by hand.
+//
+// A verbatim copy of specification/v0_9/json/common_types.json.
+
+/// The `common_types.json` document for protocol v0.9, as published.
+const String commonTypesV0_9Json = r'''
```
nan-yu, line 20: Instead of JSON string, can we use Dart classes to represent common types?

dart/a2ui_core/lib/src/validation/component_graph.dart
```diff
@@ -0,0 +1,233 @@
...
+
+/// Checks component ids and references within one surface.
+///
+/// [requireRoot] and [allowDangling] distinguish a full render, which must
+/// declare every component it names, from an incremental update, which may
+/// reference components the client already holds.
+///
+/// Throws [A2uiIntegrityError] for duplicate ids, a missing root, or a
+/// reference to a component that does not exist.
+void checkComponentIntegrity(
```
nan-yu, line 47: The integrity check isn't invoked by the message processor.

dart/a2ui_core/lib/src/validation/validator.dart
```diff
@@ -0,0 +1,400 @@
...
+        validateComponent(component, catalog);
+      }
+    }
+  }
+
+  /// Validates a complete payload: envelopes, then structure, then catalog
+  /// schemas.
+  ///
+  /// Returns the parsed messages, and throws as the individual steps do.
+  List<A2uiMessage> validate(List<Map<String, Object?>> payload) {
```
nan-yu: This top-level validate method isn't invoked by the message processor.

dart/a2ui_core/lib/src/core/catalog.dart
polina-c: Very close — Dart round-trips the published v0.9 catalog exactly: 18 components, 14 functions, same names, and the rebuilt document is a fixed point.

One thing worth flagging: `converted_basic_catalog_v09.json` has 25 functions, but the published v0.9 catalog and Dart both have 14. The extra 11 (`add`, `divide`, `contains`, `equals`, …) are not in the spec document — Python builds them from its own `BasicCatalog` class. The `$defs` differ too: Python hoists 19 shared types where Dart inlines them. So that file is a Python expectation rather than a shared one.

I factored the test out so both SDKs can run it. The cases name the source document instead of an SDK's own bundled catalog, so every implementation converts the same input, and they assert behaviour rather than presentation — same components and functions as the source, `anyComponent`/`anyFunction` covering exactly what is present, every local reference resolving, and rebuilding twice giving the same document. That leaves each SDK free on `$defs` layout.

- `catalog_schema` action: [`conformance/conformance_schema.json` L549](blueprints) (`CatalogSchemaTest`, following the existing per-action `$defs` convention)
- Cases: [`conformance/core/catalog.yaml` L479-L595](blueprints) — `test_v09_basic_catalog_schema`, `test_catalog_schema_keeps_declared_metadata`, `test_catalog_schema_after_pruning`
- Dart runner: [`dart/a2ui_core/test/conformance/catalog_schema_conformance_test.dart`](blueprints)

dart/a2ui_core/lib/src/validation/common_types.g.dart
polina-c: For this file I'd keep the string. Its only consumer resolves `$ref`s by JSON Pointer into it, so any Dart representation gets serialized straight back to a map — and the string is provably right: `common_types_test.dart` compares it byte for byte against the spec, so it cannot drift.

We already have Dart classes for these types in `CommonSchemas`, and they have drifted: `anyOf` where the spec says `oneOf`, and no `additionalProperties: false` on the `ChildList` template. A `MinimalCatalog` processor accepts `{componentId, path, bogus: 1}` where the basic catalog rejects it.

Worth fixing, but in the other direction — generate `CommonSchemas` from this document so there is one source. 

It is out of scope of my effort to fix it - I change a2ui_core minimally to enable agent SDK development.

dart/a2ui_core/lib/src/validation/component_graph.dart
polina-c: Good point. Thanks. Added.

dart/a2ui_core/lib/src/validation/validator.dart
polina-c: Correct, the two paths check different things rather than duplicating each other.

`validate` is agent side, payload-scoped: it requires a `root` on a full render, rejects components unreachable from it, and cannot check references on an incremental payload because a payload does not carry the surface's history. That is the question an agent asks before sending — is this payload a complete, coherent render?

`MessageProcessor` is renderer side, surface-scoped: it resolves references against the components the surface already holds, measures cycles over the merged graph, and does not require a root because one may arrive in a later message. That is the question only a client can ask — does this batch fit the surface I have?

Neither subsumes the other.

Makes sense?

I updated doc comments to make it clearer.

dart/a2ui_core/lib/src/core/catalog.dart
nan-yu: Good to know that they're very close and good findings on the Python output. Agreed that we should remove those functions. The caveat is that the extra 11 functions are also implemented in TypeScript and serve as commonly known functions - components should be able to use them in the function calls  even though they're not in the catalog definition. I'll follow it up in a separate PR.

dart/a2ui_core/lib/src/validation/common_types.g.dart
nan-yu: I'm good with following up separately. Would you mind creating a issue to track the followup work?

dart/a2ui_core/lib/src/validation/validator.dart
nan-yu: I'm thinking we could have the Agent call MessageProcessor to handle and validate messages, too.

This way, the agent can actually keep track of the surface history. If a message comes in to update a surface component, the agent will already understand the active component tree and can validate against it on the fly. What do you think?

As you said, this is not a blocker for this PR. We can address it later.

conformance/core/catalog.yaml
```diff
@@ -475,3 +475,121 @@
...
+            text: {type: string}
+          required: [component, text]
+        Image:
+          type: object
+          properties:
+            component: {const: Image}
+            url: {type: string}
+          required: [component, url]
+  args:
+    allowed_components: [Text]
```
nan-yu: Why do we need `allowed_components` in the Catalog.catalog_schema method? It is supposed to be a concept in the agent when trying to remove some components to reduce input tokens for better inference performance.

dart/a2ui_core/lib/src/core/catalog.dart
nan-yu: This comment is marked as resolved, but I'm not seeing how it is resolved.

I'm okay with following up later.

dart/a2ui_core/lib/src/processing/processor.dart
```diff
@@ -20,21 +20,69 @@ import '../core/messages.dart';
...
+  final A2uiValidator<T, FunctionImplementation> validator;
 
   MessageProcessor({
     required this.catalogs,
+    A2uiValidator<T, FunctionImplementation>? validator,
     void Function(A2uiClientAction)? onAction,
-  }) : groupModel = SurfaceGroupModel<T>() {
+  }) : validator =
+           validator ??
+           A2uiValidator<T, FunctionImplementation>(catalogs: catalogs),
```
nan-yu, line 65: This is problematic. It takes all catalogs to initialize the validator. However, we should use the particular catalog from the messages to run validation.

A message processor defines a list of catalogs to support all surfaces and components. However, a component is only from one particular catalog.

For example,
- component_1 in surface_1 uses catalog1.
- component_2 in surface_2 uses catalog2.

We should use catalog1 to validate component_1, and use catalog2 to validate component_2.

dart/a2ui_core/lib/src/validation/validator.dart
polina-c: Good suggestion! Updated agent blueprint to contain it.

dart/a2ui_core/lib/src/processing/processor.dart
polina-c: The constructor takes all catalogs, but that map is not what components are validated against. Every call site passes the surface's own catalog: `validateTheme(message.theme, catalog)` on create, and `validateComponent(compJson, surface.catalog)` and `validateComponentBatch(..., surface.catalog)` on update. Each `SurfaceModel` stores the catalog named at `createSurface`, so your example already behaves as you describe.

I checked it with two catalogs and two surfaces:

```
s1 (cat1) + Alpha: ACCEPTED     s2 (cat2) + Beta:  ACCEPTED
s1 (cat1) + Beta:  REJECTED     s2 (cat2) + Alpha: REJECTED
```

Added that as a test, since nothing was pinning it.

But, you were right about the standalone validator, though. `A2uiValidator` has no surface state, and v0.9 puts `catalogId` on `createSurface` only, so for a payload that just updates a surface it could not tell which catalog applied. It picked the sole catalog when there was one and otherwise skipped those components and returned success. A bogus component was rejected with one catalog and silently accepted with two.

Fixed two ways: `validate` now takes an optional `surfaceCatalogs` map naming the catalog per surface id, and when it still cannot tell it throws `A2uiCatalogError` instead of skipping. Reporting a payload valid that nothing checked was the worse failure.

This is also a good argument for your other suggestion about agents driving `MessageProcessor` — the processor knows the mapping already, so the question does not arise there.

conformance/core/catalog.yaml
polina-c: You're right, and it's removed. Pruning is an agent concern and it already has its own action — `test_with_pruning_components_v09` covers exactly what the case did, including `anyComponent` narrowing down to the components that survive. Borrowing `allowed_components` into `catalog_schema` conflated two things and added no coverage.

`catalog_schema` now just parses a document and rebuilds it: same components and functions, metadata preserved, every local reference resolving, and rebuilding twice giving the same document. `args` is gone from `CatalogSchemaTest` too.

No library change either way — `allowed_components` was never a parameter of `Catalog.catalogSchema`; the case pruned through `copyWith` before rebuilding.

dart/a2ui_core/lib/src/processing/processor.dart
nan-yu: Instead of passing a map of available catalogs to every validation method, what about scoping A2uiValidator to a single catalog?

Under this model, the MessageProcessor would instantiate a fresh validator instance scoped to the target catalog every time it processes incoming messages.

My reasoning comes from our message-handling lifecycle. While the validator is structurally designed to handle messages containing multiple surfaces, [Line 293 in message-processor.ts](https://github.com/a2ui-project/a2ui/blob/main/renderers/web_core/src/v0_9/processing/message-processor.ts#L293) reveals that the processor actively blocks the mixing of multiple update types within a single batch.

Since MessageProcessor already filters and splits messages down to specific, surface-isolated payloads before invoking the validation step, A2uiValidator only ever needs to validate against one active catalog context. Transitioning it to a single-catalog scope would significantly simplify our signature contracts and dependency passing.

nan-yu: This requires updating MessageProcessor first to block mixed update types. As you mentioned, you'd like to get this PR in to unblock other work. I'm okay with addressing them as followups.

nan-yu (approved):

dart/a2ui_core/lib/src/processing/processor.dart
polina-c: Yes, will address.

polina-c: A2uiValidator is scoped to many catalogs in the blueprint.

Updated blueprint to scope it to single catalog and implemented: https://github.com/a2ui-project/a2ui/pull/2538

PR #2457, opened 2026-08-31
Make reusable catalog function for calling MCP Tools

wrenj:


sugoi-yuzuru (commented):
The overall architecture looks reasonably good, but it feels like there is an assumption that the Host is only connected to a single MCP server.  Especially, the catalog definition that establishes the callMcpTool binding to an MCP client will likely be not incrementally executed as users add new MCP connection to their host. 

How do we support multiple MCP server connection that might have tool name collision?

sugoi-yuzuru (approved):

PR #2466, opened 2026-09-01
feat(spec,sdk): use relative common_types references in v1.0 catalogs

jacobsimionato:
## Overview & Rationale

This PR updates the **A2UI v1.0 protocol specification**, the **Core SDK Architecture Blueprint**, and **Agent SDK schema validation** to use unversioned, relative references to standard types (`common_types.json#/$defs/<TypeName>`) in component catalogs.

While breaking changes between v0.9 and v1.0 prevent v0.9 catalogs from running directly on v1.0 runtimes, using relative references in v1.0 catalogs establishes forward compatibility so that v1.0 catalogs can work with future protocol versions (such as v1.1) without requiring catalog authors to rewrite reference URLs.

### Key Changes:

1. **Relative `$ref` Targets in Catalogs**:
   - Updated external standard type references in `specification/v1_0/catalogs/basic/catalog.json` from version-qualified URLs (`https://a2ui.org/specification/v1_0/common_types.json#/$defs/...`) to unversioned relative targets: `"common_types.json#/$defs/<TypeName>"`.
   - At runtime, schema validators dynamically resolve `"common_types.json"` against the active protocol version's schema.

2. **Catalog Protocol Versioning**:
   - Catalogs retain explicit protocol version declarations via `"protocolVersion": "1.0"` in catalog documents (`catalogs/basic/catalog.json` and `test/testing_catalog.json`).
   - Retained `protocolVersion` metadata definition in `specification/v1_0/json/catalog_definition.json`.

3. **Protocol Specification & Evolution Documentation**:
   - Updated Rule 3 in `specification/v1_0/docs/a2ui_protocol.md` to require relative targets (`common_types.json#/$defs/...`) and added a note on catalog evolution and future protocol compatibility.
   - Updated `specification/v1_0/docs/evolution_guide.md` with notes for catalog authors detailing the relative reference format, versioning requirements, and forward-compatibility design.

4. **Core SDK Architecture Blueprint (`blueprints/modules/a2ui_core.blueprint.md`)**:
   - Documented the canonical public authoring API (`@a2ui/core`, `a2ui.catalog`, `package:a2ui_core`) re-exporting canonical primitives.
   - Documented the `A2uiValidator` rule requiring `common_types.json` to be registered under the root relative key in the schema referencing registry.
   - Restored `protocolVersion` property on the `Catalog` interface.

5. **Python Agent SDK Validation**:
   - Registered `"common_types.json"` as a root relative URI in `A2uiValidatorWrapperV10` referencing registry.

---

## Verification & Testing

- **Specification Test Runner (`specification/v1_0/test/run_tests.py`)**: PASSED (167/167 passed, 0 failed).
- **Multi-Version Specification Validator (`specification/scripts/validate.py`)**: PASSED across v0.8, v0.9, and v1.0.
- **Python Agent Tests (`uv run pytest`)**: PASSED (849/849 passed, 0 failed).
- **Blueprint Validation (`blueprints/validate_blueprints.py`)**: PASSED.
- **Prettier Code Formatting (`yarn prettier --check`)**: PASSED.


specification/v1_0/json/common_types.json
```diff
@@ -390,6 +390,16 @@
...
+    "FunctionCommon": {
+      "type": "object",
+      "description": "Common properties for function definitions, such as optional function-level catalogId overrides.",
+      "properties": {
+        "catalogId": {
+          "type": "string",
+          "description": "The catalog ID for this function, overriding any surface-level default catalogId."
+        }
+      }
     }
```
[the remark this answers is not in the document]
jacobsimionato: Fixed this - thanks!

gspencergoog (approved):

specification/v1_0/json/catalog_definition.json
```diff
@@ -10,8 +10,7 @@
     "protocolVersion": {
```
nan-yu, line 10: This was added recently. Can we just delete it completely?

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -186,13 +184,34 @@ export enum A2uiProtocolVersion {
...
-  readonly protocolVersion: A2uiProtocolVersion;
   readonly components: ReadonlyMap<string, TComponent>;
   readonly functions?: ReadonlyMap<string, TFunction>;
   readonly themeSchema?: Record<string, any>;
 }
 ```
 
+#### Canonical Common Types Authoring Layer (`a2ui.core.catalog.common_types`)
+
+To allow developers to author components and functions without coupling to specific protocol versions or migrating import paths as A2UI evolves, the Core SDK exposes canonical, version-agnostic common types directly from the root catalog namespace (`@a2ui/core`, `a2ui.catalog`, `package:a2ui_core`).
```
nan-yu: Are common types version-agnostic? There is a `common_types` file under each protocol version, which makes it version-specific.

The version-agnostic catalog schema references the common_types via `common_types.json#/$defs/<TypeName`. It appears referencing a version-agnostic common_types.json file, but I think it is just a relative path under the version folder.

specification/v1_0/catalogs/basic/catalog.json
nan-yu, line 1: I'm curious about the basic catalog schema. Are we going to have a `basic/catalog.json` for each future version? If not, should we move the basic catalog away from the `v1_0` folder?

nan-yu: If we move the basic catalog from the version folder, how can it reference a relative common_types.json file?

polina-c (approved):

blueprints/modules/a2ui_core.blueprint.md
gspencergoog: I think they are version-specific.  Perhaps we can refactor things so that the types used in a catalog are part of the catalog, but there are a bunch of places where the SDKs make assumptions about the structure and names of the common types in order to hook them up to actual code (e.g. dynamic paths to the data model need to actually be hooked up to the data model, so the code needs to know that theres a "path" member of an object that it needs to look for in a specific place).

Probably the right answer is a separate schema language for defining catalogs, like Jacob has proposed.  Then the "common" parts can be just part of that language, and can be polymorphic based on version or something.

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -119,12 +119,10 @@ The core modular components are organized within the `a2ui.core` namespace. Publ
...
 ├── exceptions                      # Root exception hierarchy
-├── basic_catalog/                  # Bundled default components and operators
-│   ├── v0_8/                       # Conforms to spec v0.8
-│   ├── v0_9/                       # Conforms to spec v0.9, v0.9.1
-│   └── v1_0/                       # Conforms to spec v1.0
+├── basic_catalog/                  # Bundled default components & operators (Protocol-version-agnostic)
 ├── catalog/                        # Catalog declarations
-│   ├── catalog                     # Catalog base class
+│   ├── catalog                     # Catalog base class (Version-agnostic component/function container)
+│   ├── common_types                # Canonical public authoring primitives (DynamicString, DataBinding, Action, etc.)
```
gspencergoog: How is this different from the schema's version-specific `common_types` referenced below?

specification/v1_0/json/catalog_definition.json
gspencergoog: If we actually succeed in making catalogs version agnostic, I think we should remove this completely, since it was added for v1.0.

I added it because we needed to know which version of `common_types.json` the catalog was expecting to conform to, since we have things like `ComponentCommon`, which could be different for different versions, and which all components are supposed to implement.

blueprints/modules/a2ui_core.blueprint.md
jacobsimionato: The way I was hoping this could work is that the catalog references common types in a version agnostic way, through a magical "common_types.json" reference. Then at runtime, when an SDK is actually parsing or rendering content, it uses the correct `common_types.json` for the active protocol version, for the message that it is generating / processing / rendering.

This is a bit like the pattern we already use in the relationship between `agent_to_renderer.json` and the catalog, which is referenced through a magical `catalog.json` alias.

specification/v1_0/catalogs/basic/catalog.json
jacobsimionato: Well, I think catalogs can still be versioned themselves and this is a different problem, that wrenj@ is solving. 

In the future, I think we should totally decouple the catalogs we own from the spec folder, and update them on a separate schedule, with separate version names to avoid confusion.

But seeing as we have actually changed the basic catalog between v0.9 and v1.0, I think we should include this new version here.

blueprints/modules/a2ui_core.blueprint.md
jacobsimionato: This just reexports the schemas for the latest A2UI version that the SDK supports. Ultimately, they're basically sentinel values that the schema analyzer in the node layer uses to do data binding automatically etc, so the most important thing is to just have all the possible common type symbols there to use.

blueprints/modules/a2ui_core.blueprint.md
nan-yu: > Then at runtime, when an SDK is actually parsing or rendering content, it uses the correct common_types.json for the active protocol version,

How does the runtime resolve the **active protocol version**? I'd imagine both renderer app and agent would need a list of supported protocol versions, and a list of supported catalogs. It is most likely the renderer app only sets a single protocol version, so there is only one active protocol version after the negotiation. It that's the case, it sounds like app developers and agent developers will need to maintain a compatibility matrix between protocol versions and catalog versions, for example:

| Protocol Version | Catalog Version | Catalog Name | CatalogId |
| --- | --- | --- |  --- |
| v0.9 | 0.1 | Basic Catalog | https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json 
| v1.0 | 1.0 | Basic Catalog | https://a2ui.org/specification/v1_0/catalogs/basic/catalog.json
| v1.0 | 2.0 | Basic Catalog | https://a2ui.org/catalogs/basic/2_0/catalog.json |
| v1.0 | 1.0 | Custom Catalog | custom::1.0::catalog.json |
| v1.0 | 2.0 | Custom Catalog | custom::2.0::catalog.json |
| v2.0 | 1.0 | Basic Catalog | https://a2ui.org/specification/v1_0/catalogs/basic/catalog.json | 
| v2.0 | 3.0 | Basic Catalog | https://a2ui.org/catalogs/basic/3_0/catalog.json | 

Note: Basic Catalog 1.0 works for both protocol version v1.0 and v2.0.

If protocol version v2.0 introduces breaking changes that make Basic Catalog 1.0 become incompatible, the SDK should provide a way to fail fast, so that developers can 
1. Remove the v2.0 and Basic Catalog 1.0 combination from the matrix.
2. Update Basic Catalog 1.0 to 3.0.
3. Add a v2.0 and Basic Catalog 3.0 combination to the matrix.
4. Set up with protocol v2.0 and Basic Catalog 3.0.

How can we make catalog-to-protocol compatibility resolution more automated or ergonomic?

jacobsimionato: Yes, this is a great point - I agree!!

In clientCapabilities, we already basically report a matrix of protocolversion/catalog support, because there is a different entry for each protocol version, with a list of catalog Ids. It's kind of duplicative, but it is what we have. So I was thinking that the client can report this, and then the server can choose a catalog/protocol combination that is supported.

Ultimately, the client dictates which combinations are supported. So whatever it supports, the agent should be able to use on its side.

In https://docs.google.com/document/d/1cNrDTy8fG9ogBnlPjI07tjfBq1ZXSag-IJdjkGA6VCk/edit?resourcekey=0-QBnb6JgDUozM2xDc4XbN4Q&tab=t.n1ne9051bt5r, I propose that we have a required `maxProtocolVersion` configuration on CatalogImplementation that ensures that if protocol version v2.0 comes out, a client which implements basic catalog 1.0 will not report that this catalog can be used with protocol version 2.0 unless the developer explicitly changes the configuration in their client application code.

WDYT? I thiiiiiink there's a workable model here, but it is definitely confusing to reason about. In practice, there may only be a small number of catalogs that many people use (e.g. material catalog), so maybe it's okay?

gspencergoog (approved):

nan-yu (approved):

PR #2473, opened 2026-09-01
feat(python): implement Step 3.1 A2UI v1.0 Agent-to-Renderer callRendererFunction RPC

nan-yu:
## Summary
- Extract `requiresUserActivation` metadata from v1.0 catalog JSON schemas and emit `requires_user_activation` on Python `FunctionApi` definitions.
- Implement Agent-to-Renderer Function RPC (`callRendererFunction`) in Python `MessageProcessor` and version adapter.
- Enforce `allowedCallers` (`agentOnly` / `rendererOrAgent`) and `requiresUserActivation` guards.
- Enable and verify all `callRendererFunction` test cases in `conformance/core/rpc_functions.yaml`.

TAG=agy

### DIFF
--- a/python/a2ui_core/src/a2ui/core/processing/message_processor.py
+++ b/python/a2ui_core/src/a2ui/core/processing/message_processor.py
@@ -32,6 +32,11 @@ from ..exceptions import (
     A2uiValidationError,
 )
 from ..schema import AgentToRendererMessage, ProtocolVersion
+from ..schema.v1_0 import (
+    FunctionResponse,
+    FunctionResponseError,
+    RendererFunctionResponseMessage,
+)
 from .adapters import VersionAdapterFactory
 from .operations import (
     InternalCallRendererFunctionOp,
@@ -40,6 +45,7 @@ from .operations import (
     InternalOperation,
     InternalUpdateComponentsOp,
     InternalUpdateDataModelOp,
+    RpcErrorCode,
 )
 
 
@@ -78,6 +84,15 @@ class MessageProcessor:
             if resp:
                 responses.append(resp)
         return responses
+
+    def _resolve_catalog(self, catalog_id: str | None = None) -> Any | None:
+        """Finds catalog by catalog_id or returns default (first registered) catalog."""
+        if catalog_id is not None:
+            for cat in self.catalogs:
+                if getattr(cat, "catalog_id", None) == catalog_id:
+                    return cat
+            return None
+        return self.catalogs[0] if self.catalogs else None
 
     def get_renderer_capabilities(
         self,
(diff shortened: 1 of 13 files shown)


python/a2ui_core/src/a2ui/core/processing/message_processor.py
```diff
@@ -126,6 +131,102 @@ def _process_operation(self, op: InternalOperation) -> None:
...
+                    "functionCallId": call_id,
+                    "error": {
+                        "code": "INVALID_FUNCTION_CALL",
+                        "message": f"Function not found: {op.call}",
+                    },
+                },
+            }
+
+        allowed_callers = getattr(fn, "allowed_callers", "rendererOnly")
+        if allowed_callers not in ("agentOnly", "rendererOrAgent"):
```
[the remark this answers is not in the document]
nan-yu: The A2UI v1.0 JSON schema explicitly defines "rendererOnly" as the default for allowedCallers in `specification/v1_0/json/catalog_definition.json` whenever it is omitted in a function definition.

```json
"allowedCallers": {
  "type": "string",
  "enum": ["rendererOnly", "agentOnly", "rendererOrAgent"],
  "default": "rendererOnly",
  "description": "Specifies which roles are authorized to invoke this function."
}
```

python/a2ui_core/src/a2ui/core/processing/message_processor.py
```diff
@@ -67,12 +74,28 @@ def process_messages(
             | Mapping[str, Any]
             | Sequence[Mapping[str, Any]]
         ),
-    ) -> None:
+        user_activation_present: bool = False,
```
nan-yu: I don't like the fact that we have to pass `user_activation_present` all the way down to the version adapters, but not sure if there're better ways to indicate an activation has been performed from the user.

gspencergoog: You could create an execution context object that you pass down, and that can hold stuff like this that is set at one layer and used at another, without having to change the signatures in the future.  I mean, it's still passing something down, but at least it's only one thing, and it is easier to change.  We do this for Flutter with the `BuildContext` object.

E.g.:

```py
@dataclass
class ExecutionContext:
    user_activation_present: bool = False
    abort_signal: Any | None = None
    metadata: dict[str, Any] = field(default_factory=dict)
```

nan-yu: That's a great idea. I'll update the processMessages method with the execution context. Thanks!

python/a2ui_core/tests/test_conformance.py
```diff
@@ -694,3 +696,83 @@ def validate_get_renderer_data_model_case(case: dict[str, Any]) -> None:
...
+        )
+
+    cat_id = (
+        "media_catalog"
+        if "playMedia" in fn_metadata
+        else "system_catalog"
+        if "internalStateReset" in fn_metadata
+        else "store_catalog"
+        if "queryInventory" in fn_metadata
+        else "basic"
```
gspencergoog: Just a nit: Is there any way to not hard code these function names?  If we change the names in the conformance suite, then this will be wrong.

I asked, and Gemini suggested:

```suggestion
    cat_id = (
        args.get("catalogId")
        or (
            message.get("callRendererFunction", {})
            .get("callFunction", {})
            .get("catalogId")
            if isinstance(message, dict) and "callRendererFunction" in message
            else None
        )
        or (
            outbound_call.get("callFunction", {}).get("catalogId")
            if isinstance(outbound_call, dict)
            else None
        )
        or "basic"
```

gspencergoog (approved):

python/a2ui_core/src/a2ui/core/processing/message_processor.py
nan-yu: Addressed! Introduced `ExecutionContext` dataclass containing `user_activation_present: bool = False`, and updated `MessageProcessor.process_messages` and `VersionAdapter.extract_operations` to accept `context: ExecutionContext | None = None`.

python/a2ui_core/tests/test_conformance.py
nan-yu: Addressed! Updated `test_conformance.py` catalog resolution logic to dynamically extract `catalogId` from test case `args`, `message`, or `outboundCall`, eliminating hardcoded function names.

PR #2478, opened 2026-09-01
feat(python): implement Step 3.2 Renderer-to-Agent callAgentFunction & agentFunctionResponse

nan-yu:
Stacked on #2473

Support renderer-initiated remote function calls sent to the agent (`callAgentFunction`) and handling of inbound `agentFunctionResponse` messages in Python `MessageProcessor`.

- Add `create_call_agent_function_message` helper to format outbound `callAgentFunction` requests.
- Add `_pending_agent_calls` mapping and `register_pending_agent_call` method to `MessageProcessor` to route inbound `agentFunctionResponse` results/errors to pending `asyncio.Future` objects or callbacks.
- Update `V1Point0Adapter` and operations to process inbound `agentFunctionResponse` messages using `FunctionResponse` and `FunctionResponseError` models.
- Update conformance test runner and `rpc_functions.yaml` test suite to cover inbound `agentFunctionResponse` processing.

TAG=agy

### DIFF
--- a/python/a2ui_core/src/a2ui/core/processing/message_processor.py
+++ b/python/a2ui_core/src/a2ui/core/processing/message_processor.py
@@ -12,10 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
+import concurrent.futures
 import copy
+import logging
 from collections.abc import Mapping, Sequence
 from typing import Any, Callable, cast
 
+logger = logging.getLogger(__name__)
+
+from ..common.events import EventSource
 from ..state import SurfaceGroupModel, SurfaceModel, ComponentModel
 from ..validation import (
     PayloadValidator,
@@ -33,12 +39,17 @@ from ..exceptions import (
 )
 from ..schema import AgentToRendererMessage, ProtocolVersion
 from ..schema.v1_0 import (
+    AgentFunctionResponseMessage,
+    CallAgentFunction,
+    CallAgentFunctionMessage,
     FunctionResponse,
     FunctionResponseError,
     RendererFunctionResponseMessage,
 )
+from ..schema.v1_0.common_types import FunctionCall
 from .adapters import VersionAdapterFactory
 from .operations import (
+    InternalAgentFunctionResponseOp,
     InternalCallRendererFunctionOp,
     InternalCreateSurfaceOp,
     InternalDeleteSurfaceOp,
@@ -53,7 +64,7 @@ from .execution_context import ExecutionContext
 
 
 class MessageProcessor:
-    """Core state engine that validates payloads, manages surfaces, and applies mutation ops."""
+    """Core processor for handling A2UI messages, updating state, and executing operations."""
 
     def __init__(
         self,
@@ -66,8 +77,16 @@ class MessageProcessor:
         self.catalogs = catalogs
         self.model = SurfaceGroupModel()
         self.validation_config = validation_config
+        self.on_agent_function_response = EventSource()
+        self._pending_agent_calls: dict[str, Any] = {}
         if action_handler:
             self.model.on_action.subscribe(action_handler)
+
+    def register_pending_agent_call(
+        self, function_call_id: str, future_or_callback: Any
+    ) -> None:
+        """Registers a pending async future or callback for an outbound callAgentFunction invocation."""
+        self._pending_agent_calls[function_call_id] = future_or_callback
 
     def process_messages(
         self,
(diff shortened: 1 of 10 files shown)


python/a2ui_core/src/a2ui/core/processing/message_processor.py
```diff
@@ -88,14 +107,40 @@ def process_messages(
                 responses.append(resp)
         return responses
 
+    def create_call_agent_function_message(
+        self,
+        surface_id: str,
+        function_call_id: str,
+        call: str,
+        version: str,
+        catalog_id: str | None = None,
```
gspencergoog, line 211: This probably shouldn't default to None.

If `catalog_id` is None, passing `catalogId=None` marks the field as explicitly set in Pydantic v2. Calling `msg.model_dump` then emits `"catalogId": null`, and the schema in `common_types.json` doesn't allow a null catalog ID.

python/a2ui_core/src/a2ui/core/processing/message_processor.py
```diff
@@ -151,8 +196,56 @@ def _process_operation(self, op: InternalOperation) -> dict[str, Any] | None:
...
+    ) -> None:
+        """Processes an inbound agentFunctionResponse from the agent."""
+        resp_error = (
+            FunctionResponseError.model_validate(op.error) if op.error else None
+        )
+        response_obj = FunctionResponse(
+            functionCallId=op.function_call_id,
+            value=op.value,
+            error=resp_error,
+        )
```
gspencergoog: Passing both the op and the error to this sets them both, even if they are null, and the response doesn't accept null values, only strings.

python/a2ui_core/src/a2ui/core/processing/message_processor.py
```diff
@@ -151,8 +196,56 @@ def _process_operation(self, op: InternalOperation) -> dict[str, Any] | None:
...
+                        concurrent.futures.InvalidStateError,
+                    ) as exc:
+                        logger.debug(
+                            "Ignored agentFunctionResponse for call %s: pending future"
+                            " already done or cancelled (%s)",
+                            op.function_call_id,
+                            exc,
+                        )
+            elif callable(pending):
+                pending(op.value, op.error)
```
gspencergoog: If someone registers a 1-parameter callback, then this will raise a `TypeError`. Also, unhandled exceptions inside the callback will crash `process_messages`.

python/a2ui_core/src/a2ui/core/processing/message_processor.py
```diff
@@ -63,9 +74,17 @@ def __init__(
         self.catalogs = catalogs
         self.model = SurfaceGroupModel()
         self.validation_config = validation_config
+        self.on_agent_function_response = EventSource()
+        self._pending_agent_calls: dict[str, Any] = {}
         if action_handler:
             self.model.on_action.subscribe(action_handler)
 
+    def register_pending_agent_call(
+        self, function_call_id: str, future_or_callback: Any
```
gspencergoog: It would be nice to avoid the `Any` on `future_or_callback`.

python/a2ui_core/tests/test_conformance.py
```diff
@@ -755,24 +755,48 @@ def execute(fn_args: dict[str, Any], *args: Any, **kwargs: Any) -> Any:
...
+
+        import asyncio
+
+        loop = asyncio.new_event_loop()
+        fut = loop.create_future()
+        processor.register_pending_agent_call(outbound_call["functionCallId"], fut)
+        processor.process_messages(inbound_response)
+        assert fut.done()
+        assert fut.result() == case.get("expect", {}).get("result")
+        loop.close()
```
gspencergoog: Best to put this inside of a `finally:` so that the loop closes even if an assertion fires.

gspencergoog (commented):
Some suggestions:

- Add a typed `RpcError(A2uiError)` with `code` and `function_call_id` so errors aren't so generic.
- Validate arguments against the function schema before calling `execute` so you can separate `INVALID_FUNCTION_CALL` from `EXECUTION_ERROR`.

In the TS implementation, I have an async method to call agent functions with timeout handling and disposal cleanup: I don't see where you are cleaning up pending calls that haven't been responded to.

You might want to make the agent-to-renderer functions calls accept async functions and return a promise, since otherwise they have to be sync functions.

python/a2ui_core/src/a2ui/core/processing/message_processor.py
```diff
@@ -151,8 +191,40 @@ def _process_operation(self, op: InternalOperation) -> dict[str, Any] | None:
...
+        if pending is not None:
+            if hasattr(pending, "set_result") and hasattr(pending, "set_exception"):
+                if op.error:
+                    err_code = op.error.get("code", "UNKNOWN_ERROR")
+                    err_msg = op.error.get("message", "Agent function execution failed")
+                    pending.set_exception(
+                        A2uiError(f"Agent function error [{err_code}]: {err_msg}")
+                    )
+                else:
+                    pending.set_result(op.value)
```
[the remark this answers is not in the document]
nan-yu: Fixed! Guarded futures with  and wrapped set_result / set_exception in try-except for InvalidStateError.

python/a2ui_core/src/a2ui/core/processing/message_processor.py
nan-yu: Fixed. I added `exclude_none=True` to `msg.model_dump` so that null values are completely omitted .

python/a2ui_core/src/a2ui/core/processing/message_processor.py
nan-yu: Fixed!  is now instantiated conditionally with either  or  so no  properties are emitted.

python/a2ui_core/src/a2ui/core/processing/message_processor.py
nan-yu: Fixed! Implemented  helper which uses  to support both 1-parameter and 2-parameter callbacks and safely catches/logs unhandled callback exceptions.

python/a2ui_core/src/a2ui/core/processing/message_processor.py
nan-yu: Fixed!  is now strictly typed to a concrete callback signature . Added a helper  using a generic .

python/a2ui_core/tests/test_conformance.py
nan-yu: Fixed! Wrapped event loop execution in .

python/a2ui_core/src/a2ui/core/processing/message_processor.py
nan-yu: Fixed! `_pending_agent_calls` is now strictly typed to a concrete callback signature `PendingAgentCallCallback = Callable[[Any, Optional[dict[str, Any]]], None]`. Added a helper `register_pending_future(function_call_id, future: asyncio.Future[T] | concurrent.futures.Future[T])` using a generic `TypeVar("T")`.

python/a2ui_core/tests/test_conformance.py
nan-yu: Fixed! Wrapped event loop execution in 'try...finally: loop.close()'.

gspencergoog (approved):

PR #2480, opened 2026-09-01
feat(python): implement Step 3.3 Multi-Catalog Resolution Engine

nan-yu:
Stacked on #2478

Implement multi-catalog source resolution and protocol version validation across mixed catalog surfaces in Python core.

- Rename `python/a2ui_core/src/a2ui/core/rendering` directory to `python/a2ui_core/src/a2ui/core/resolution` and move `node_graph.py` into `resolution/`.
- Implement catalog resolution precedence: (1) explicit `catalogId`, (2) surface default `catalogId`, (3) throw `A2uiCatalogError` (no fallback to capabilities list).
- Implement `SurfaceModel.validate_catalog_versions` to verify that all active catalogs on a surface share the exact same `protocolVersion`.
- Enable and route `multi_catalog.yaml` conformance tests in `test_conformance.py`.

TAG=agy


python/a2ui_core/src/a2ui/core/state/surface_model.py
```diff
@@ -22,19 +22,29 @@
...
+
+
 class SurfaceModel(Generic[TComponent, TFunction]):
     """Represents a single active UI Surface state tree."""
 
     def __init__(
         self,
         surface_id: str,
         default_catalog: Catalog[TComponent, TFunction],
+        available_catalogs: Sequence[Catalog[TComponent, TFunction]] | None = None,
```
gspencergoog: Is this list of available catalogs now being passed in the message processor?

gspencergoog (approved):

python/a2ui_core/src/a2ui/core/state/surface_model.py
nan-yu: Thanks for capturing this. Ideally no, `SurfaceModel` should only hold the default catalog and component-specific catalogs can be resolved directly via the components model.

However, I have a quick question regarding the lifecycle of function calls: Are all function calls strictly triggered from within a component?
- If yes: We can easily traverse the component tree to resolve the component-specific catalog and verify the validity of the function’s catalogId.
- If no: We may need to pass `available_catalogs` down from MessageProcessor to SurfaceModel to ensure we can validate any standalone function calls.

nan-yu: Never mind. Bases on the renderer_to_agent.json protocol, function calls are not strictly triggered from within a component. Protocol messages like `callAgentFunction` or `callRendererFunction` invoke functions directly at the top level of the message payload. They are not tied to any component in the UI tree.

Because standalone RPC function calls exist outside the component hierarchy, catalog resolution cannot rely on traversing component nodes. Thus, passing `available_catalogs` seems needed.

nan-yu: Updated MessageProcessor to pass the available_catalogs for function catalog resolution.

gspencergoog (approved):

Issue #2463, opened 2026-09-01
[ci] Is dart format doing what we want (in CI)?


ditman:
While working on an unrelated task, I noticed the following output in the dart format check:

```
Running Dart format...
Warning: Package resolution error when reading "analysis_options.yaml" file for "samples/client/flutter/restaurant_finder/app/lib/main.dart":
Failed to resolve package URI "package:lints/recommended.yaml" in include at "/home/runner/work/a2ui/a2ui/analysis_options.yaml".
Warning: Package resolution error when reading "analysis_options.yaml" file for "samples/client/flutter/restaurant_finder/app/lib/src/main.dart":
Failed to resolve package URI "package:lints/recommended.yaml" in include at "/home/runner/work/a2ui/a2ui/analysis_options.yaml".
Warning: Package resolution error when reading "analysis_options.yaml" file for "samples/client/flutter/restaurant_finder/app/test/widget_test.dart":
Failed to resolve package URI "package:lints/recommended.yaml" in include at "/home/runner/work/a2ui/a2ui/analysis_options.yaml".
Warning: Package resolution error when reading "analysis_options.yaml" file for "samples/client/flutter/restaurant_finder/e2e_test/test/infra_test.dart":
Failed to resolve package URI "package:lints/recommended.yaml" in include at "/home/runner/work/a2ui/a2ui/analysis_options.yaml".
Warning: Package resolution error when reading "analysis_options.yaml" file for "samples/client/flutter/restaurant_finder/e2e_test/test/test_infra/ai_client.dart":
Failed to resolve package URI "package:lints/recommended.yaml" in include at "/home/runner/work/a2ui/a2ui/analysis_options.yaml".
Formatted 13 files (0 changed) in 0.02 seconds.
```

> **Failed to resolve package URI `"package:lints/recommended.yaml"` in include at `".../analysis_options.yaml"`.**

See [this build task](https://github.com/a2ui-project/a2ui/actions/runs/33453710575/job/99689136987?pr=1935#step:7:51).

Are we applying the lints that we expect?

jacobsimionato:
Gemini ideas:
### Root Cause & Proposed Solution

  #### 🔍 Root Cause
  • scripts/fix_format.sh https://github.com/a2ui-project/a2ui/blob/main/scripts/fix_format.sh runs dart pub get
  before formatting so dart format can resolve package:lints/recommended.yaml referenced in analysis_options.yaml.
[...]

PR #2494, opened 2026-09-03
ci(dart): analyze, format and test the packages under dart/

### LINKED ISSUE #2493: [BUG]: the Dart packages under dart/ are not covered by any CI job

## Describe the Bug

Nothing in CI builds, analyzes, formats or tests `dart/a2ui_core` and `dart/a2ui_agent`. They are the two packages this repository ships to pub.dev, and they are the only source tree here with no CI job at all.

The Flutter workflow looks like it would cover them, but its matrix is built from `samples` only:

```yaml
# .github/workflows/flutter_packages_test.yml
DIRS_TO_TEST=$(find samples -name pubspec.yaml -not -path "*/.dart_tool/*" -not -path "*/e2e_test/*" -exec dirname {} \;)
```

so `dart/` never enters it, and its `dart pub get` / `dart format` / `flutter analyze --fatal-infos` / `flutter test` steps never run against those packages. `cycle-check` in the same workflow scans `samples` too. No other workflow runs `dart analyze` or `dart test` — `grep -rn "dart analyze\|dart test" .github/workflows/` comes back with only those `samples` lines.

Formatting has the same gap from the other direction: `scripts/fix_format.sh`, which `presubmit-lint.yml` runs, formats `samples/client/flutter renderers/flutter`. `dart/` is not in that list, and `renderers/flutter` currently holds only a README, so the Dart formatting check effectively covers the samples alone.

## Steps to Reproduce

Observable on merged and open PRs that change these packages:

1. Open a PR touching `dart/a2ui_core/lib` and `dart/a2ui_core/test` — #2374 (merged) and #2491 (open) both do.
2. Look at the checks. Every Dart-related job that runs belongs to the samples (`samples_client_flutter_restaurant_finder_app`, `cycle-check`) or to formatting.
3. Nothing compiled the package, ran its analyzer, or ran its 74 tests.

#2374 changed `MessageProcessor` and was merged without CI ever running that package's test suite.

## Expected Behavior

The packages published from this repository are held to the bar the samples are held to: formatting, analysis with `--fatal-infos`, and their own tests, on every PR that touches them.

`--fatal-infos` matters specifically for the question raised in #2463. The analyzer reports lint violations as *infos*, so a Dart job without that flag exits 0 on a file that breaks every rule in `analysis_options.yaml`.

Formatting is not a substitute: `dart format` reads `analysis_options.yaml` only for its `formatter:` section (`page_width`), never for lint rules. This repository sets no `formatter:` section, so the unresolved include reported in #2463 leaves formatting output unchanged. What does change it is the language version, which `fix_format.sh` already pins for that reason.

## Environment Details

- **OS**: macOS
- **Browser/Platform**: Dart 3.12.2
- **SDK/Package Name & Version**: `a2ui_core` 0.1.1, `a2ui_agent` 0.0.1-wip001, at main (676a899)
- **Protocol Version**: v0.9

## Additional Context

Both packages pass all four steps today, so the gate can be turned on without fixing anything first — bar three `unnecessary_library_directive` infos, which #2439 is already resolving as part of its own work.

One constraint worth recording for whoever picks this up: a Dart-only SDK cannot resolve these packages. They use `resolution: workspace`, and the root `pubspec.yaml` declares `flutter: ">=3.35.7 <4.0.0"` and lists the Flutter samples in its workspace, so `dart pub get` fails there — the same root cause discussed in #2463. A job for `dart/` therefore needs the repository's `setup-dart` action (which installs Flutter) until that root pubspec is split.

I have a PR ready that adds the workflow.

diegolopezrm:
# Description

Fixes #2493, and answers the lint half of #2463.

`dart/a2ui_core` and `dart/a2ui_agent` — the two packages this repository publishes to pub.dev — are not covered by any CI job. The Flutter workflow's matrix comes from `find samples -name pubspec.yaml`, so `dart/` never enters it; no other workflow runs `dart analyze` or `dart test`; and `fix_format.sh` formats `samples/client/flutter renderers/flutter`, which does not include `dart/` either. #2374 changed `MessageProcessor` and was merged without CI ever running that package's tests.

This adds `dart_packages_test.yml`, modelled on `flutter_packages_test.yml`, holding those packages to the same bar as the samples: `dart pub get`, `dart format --output=none --set-exit-if-changed`, `dart analyze --fatal-infos`, `dart test`. The matrix is generated the same way the Flutter one is, so a new package under `dart/` is picked up with no workflow edit.

Both packages pass all four steps today: `a2ui_core` 74 tests, `a2ui_agent` 1, and the analyzer is clean in both after the first commit here.

## Why `--fatal-infos`, and why it answers #2463

The analyzer reports lint violations as *infos*. A Dart job without `--fatal-infos` exits 0 on a file that breaks every rule in `analysis_options.yaml`, so the flag is what makes this a lint gate rather than a compile check. The samples already use it (`flutter analyze --fatal-infos`).

For the question in #2463 — "are we applying the lints that we expect?" — this is the part that was actually missing. The warnings in that issue come from `dart format`, which reads `analysis_options.yaml` only for its `formatter:` section (`page_width`) and never for lint rules; this repository sets no `formatter:` section, so the unresolved include leaves formatting output unchanged. What does change that output is the language version, which `fix_format.sh` already pins for exactly that reason. Neither path applies a single lint to `dart/`, because no job analyzes it.

## Why Flutter and not a Dart-only SDK

The packages use `resolution: workspace`, and the root `pubspec.yaml` declares `flutter: ">=3.35.7 <4.0.0"` and lists the Flutter samples in its workspace, so `dart pub get` fails under a Dart-only SDK — the root cause described in #2463. The job therefore uses the repository's `setup-dart` action, which installs Flutter. If the root pubspec is later split as proposed there, the job can drop to a lightweight Dart SDK by swapping that one step; there is a comment at the step saying so.

## First commit

`dart analyze --fatal-infos` needs the three `unnecessary_library_directive` infos resolved first. #2439 is already resolving the same three, and I matched its approach rather than inventing a second one: the two libraries in `a2ui_core` get the doc comment they were missing, `a2ui_agent`'s bare directive is removed. `reactivity.dart` and `a2ui_agent.dart` come out byte-identical to that PR, so they merge with no conflict; `a2ui_core.dart` conflicts only in the wording of the doc comment. If #2439 lands first I will rebase and keep its wording — the workflow itself does not conflict with anything.

## Verification

I cannot run GitHub Actions locally, so I ran the workflow's own steps, including the matrix-generation script verbatim:

```
matrix=[{"name":"dart_a2ui_agent","path":"dart/a2ui_agent"},{"name":"dart_a2ui_core","path":"dart/a2ui_core"}]

dart/a2ui_agent | format:OK | analyze --fatal-infos:OK | 00:00 +1: All tests passed!
dart/a2ui_core  | format:OK | analyze --fatal-infos:OK | 00:00 +74: All tests passed!
```

The workflow file parses as YAML, carries the license header (`fix_licenses.py --check` does not flag it) and is Prettier-clean.

## A note on the checklist below

As in #2491, I have not touched the packages' CHANGELOG.md files, so the changelog health check will flag this PR. #2439 already adds a `## 0.2.0` section to both, and a competing section here would only give that PR a conflict to resolve. The change in this PR is a doc comment and a removed directive — no API or behaviour change — but I am glad to add entries wherever you prefer.

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].

For this PR:

- [ ] I have updated the relevant CHANGELOG.md file. (see the note above)
- [x] I updated/added relevant documentation. (library doc comments; the workflow explains its own Flutter dependency)
- [x] My code changes (if any) have tests. (the change *is* test coverage; the packages' existing suites now run in CI)
- [ ] If my branch is on a fork, I have verified that [scripts/e2e_test.sh](https://github.com/a2ui-project/a2ui/blob/main/scripts/e2e_test.sh) passes. (no product code changes; both Dart suites pass)


.github/workflows/dart_packages_test.yml
```diff
@@ -0,0 +1,100 @@
...
+      matrix: ${{ steps.generate_matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
+
+      - name: Generate testing matrix
+        id: generate_matrix
+        run: |
+          DIRS_TO_TEST=$(find dart -name pubspec.yaml -not -path "*/.dart_tool/*" -exec dirname {} \;)
```
gspencergoog, line 52: This will skip dart files in other parts of the tree.  There currently aren't any, but that doesn't mean there couldn't be (imagine a tool implemented in Dart, for instance), and it is a monorepo, so the top level directory is also excluded by this.  I'd search the whole tree.

Issue #2496, opened 2026-09-03
[BUG]: Dart and web_core expression parsers disagree on 408 of 2,592 templates


diegolopezrm:
- [x] I have searched the existing issues to make sure this bug has not already been reported.

## Describe the Bug

The client-side expression parser behind `formatString` is implemented once per language, each implementation a close port of the others, and nothing compares them. I ran the Dart (`dart/a2ui_core`) and TypeScript (`renderers/web_core`) parsers over the same 2,592 templates and diffed their output. They disagree on 408 of them.

Grouped by cause:

| divergence | templates | Dart | web_core |
| --- | --- | --- | --- |
| the `null` keyword | 127 | `null` | `""` |
| literal text after an escaped `\${` | 236 | one literal part | split into several |
| adjacent literal parts | 43 | joined | kept separate |
| invalid number literal | 1 | `FormatException` | `NaN` |
| mixed (more than one of the above) | 1 | | |

Two of these are only about how the parts array is split — join adjacent literals on both sides and 272 of the 408 agree. The remaining **137 are genuine disagreements about what a template means**, and 136 of those are the `null` keyword.

### 1. The `null` keyword — the specification contradicts itself

The implementation guide lists it as a literal the parser must recognise:

> **Literals:** Quoted strings (`'...'` or `"..."`), numbers, and keywords (`true`, `false`, `null`).
> — `specification/v0_9/docs/basic_catalog_implementation_guide.md`

But `DynamicValue` in `specification/v0_9/json/common_types.json` is `oneOf` string, number, boolean, array, `DataBinding`, `FunctionCall` — there is no null branch. So a parser cannot both recognise the keyword and produce a value of the declared type.

The two implementations resolved that differently. Dart returns `null`; web_core returns `''` ([expression_parser.ts#L152](https://github.com/a2ui-project/a2ui/blob/676a8999936b17070195230d396c18d14a22d64d/renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts#L152)). At the top level it does not show, because both filter empty strings out of the parts array. Inside a function call it does:

```
${check(value: null)}

  Dart:     {call: check, args: {value: null},  returnType: any}
  web_core: {call: check, args: {value: ""},    returnType: any}
```

`required` is specified as "true if `args.value` is strictly not `null`, not `undefined`, not an empty string" — so that particular function agrees either way, but any function that distinguishes absent from empty does not, and neither does an agent reading a reported data model back.

[...]

PR #2182, opened 2026-08-06
feat: Add initial conformance tests to the a2ui core SDK (#2150)

jacobsimionato:
### Summary & Rationale
Introduces a language-agnostic conformance testing architecture for A2UI Client Framework and Core SDKs under a top-level `/conformance` directory. This complements the server-side Agent SDK conformance suite in `agent_sdks/conformance/`, providing standardized YAML test definitions, a JSON validation schema, and cross-platform verification for `DataModel` reactive state management across both Python (`a2ui_core`) and TypeScript (`@a2ui/web_core`).

### Key Changes
- **Core Conformance Directory Layout (`/conformance/`)**:
  - Added `/conformance/README.md`: Explaining Core SDK conformance testing architecture, execution model, and harness design.
  - Added `/conformance/conformance_schema.json`: JSON Schema for validating Core Framework YAML test suites.
  - Added `/conformance/suites/data_model.yaml`: Comprehensive test suite covering test scenarios for `DataModel` (`get`, `set`, auto-vivification, array handling, subscriptions/container semantics, RFC 6901 JSON Pointer escaping, prototype pollution protection, and edge case error handling).
  - Added `/conformance/tests/test_conformance_yaml.py`: Pytest suite verifying all YAML test files match `conformance_schema.json`.
- **Python Harness (`a2ui_core`)**:
  - Added `agent_sdks/python/a2ui_core/tests/test_data_model_conformance.py`: Pytest harness executing `data_model.yaml` against Python `DataModel`.
  - Enforced security parity against prototype pollution (`__proto__`, `constructor`, `prototype`).
- **TypeScript Harness (`@a2ui/web_core`)**:
  - Added `renderers/web_core/src/v0_9/state/data-model-conformance.test.ts`: Node test harness executing `data_model.yaml` against TypeScript `DataModel`.
- **Documentation & Skills**:
  - Created `.agents/skills/a2ui-conformance-testing/SKILL.md` (and linked via `blueprints/link_skills.sh`).
  - Updated `AGENTS.md`, root `README.md`, and SDK implementation skill to reference `/conformance/`.

### Verification & Testing
- **TypeScript (`@a2ui/web_core`)**: Executed unit and conformance test runner (`corepack yarn workspace @a2ui/web_core run test`). All 291 tests passed, including all 41 `DataModel` tests in `data-model-conformance.test.ts`.
[...]

PR #2497, opened 2026-09-03
test(conformance): add a cross-language conformance suite for the expression parser

diegolopezrm:
# Description

Addresses part of #2496.

The client-side expression parser behind `formatString` exists once per language, each implementation a close port of the others, and nothing compares them. Running the Dart and TypeScript parsers over the same 2,592 templates, they disagree on 408 — 137 of those on what a template actually *means* rather than on how the parts array is split.

This PR does two things: it fixes the one disagreement that is unambiguously a bug in both, and it adds a conformance suite so the next drift is caught by CI instead of by someone diffing two parsers by hand.

## 1. Invalid number literals

A literal the scanner accepts but neither language can parse — `${1.2.3}`, `${1..2}` — ended somewhere different in each:

| | before | after |
| --- | --- | --- |
| Dart | `FormatException` from `num.parse` | `A2uiExpressionError` |
| web_core | `Number()` returns `NaN`, handed back as a parsed value | `A2uiExpressionError` |

Both were wrong in their own way. `FormatException` is outside the `A2uiError` hierarchy consumers catch, and one `avoid_catching_errors` discourages catching at all; `NaN` is not a JSON value and not a member of `DynamicValue`, so web_core was placing a value in the parse tree that the protocol cannot express, silently. Every other parse failure in both raises `A2uiExpressionError`, and now so does this one.

## 2. `conformance/core/expressions.yaml`

A shared suite of 30 cases — literals, data bindings, function calls, nested interpolation, escaped markers, whitespace, and the five parse errors — expressed once and run by both clients. It follows the existing structure in `conformance/`, and `conformance/tests/test_conformance_yaml.py` validates it against `conformance_schema.json` with no change to that test.

Supporting changes:

- `conformance_schema.json`: adds the `parse_expression_template` action and its case shape, alongside the existing ones.
- `conformance/README.md`: documents the suite, the two client harnesses, and how to write cases for the new action.
- Dart harness: `dart/a2ui_core/test/conformance/expressions_conformance_test.dart`.
- TypeScript harness: `renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.conformance.test.ts`.

Both harnesses locate the suite by walking up from the test, so neither needs a configured path, and both map the suite's language-agnostic error categories onto their own types (`ParseError` → `A2uiExpressionError`). js-yaml comes in as a web_core dev dependency, matching the approach in #2182.

As far as I can tell these are the first conformance suites run by the *client* implementations — `conformance/` is currently consumed by the Python and Kotlin agent SDKs — so the same pattern is available to Swift and Kotlin clients next.

### What the suite deliberately does not decide

Two of the divergences in #2496 are open questions rather than bugs, and this PR does not quietly settle either:

- **Literal splitting.** The harnesses join adjacent literal parts before comparing, so a case fixes what a template means, not how an implementation splits the literal text around its values. Both implementations conform today despite splitting differently. If you would rather canonicalise the split, the joining step is where that decision goes.
- **The `null` keyword.** No case uses it. The guide lists `null` as a literal the parser must recognise while `DynamicValue` has no null branch, so any expectation I wrote would be me picking a side in a specification contradiction. #2496 lays out the options.

## Verification

- `conformance`: `pytest conformance/tests/test_conformance_yaml.py` — 9 passed, the new suite validating with the rest.
- Dart: `dart test` in `dart/a2ui_core` — 104 passing (74 before, plus the 30 conformance cases). `dart analyze` reports the same two pre-existing `unnecessary_library_directive` infos as main and nothing new; `dart format` clean.
- TypeScript: `node --test "dist/**/*.test.js"` in `renderers/web_core` — 447 passing, 0 failing. `eslint` reports no new problems (the two `no-explicit-any` warnings are pre-existing); `prettier --check` clean on everything touched.
- Differential run, before and after, over the same 2,592 templates: semantic disagreements went from 137 to 136, and every remaining one is the `null` keyword left to #2496.

## A note on the checklist below

As in #2491 and #2494, I have not touched `dart/a2ui_core/CHANGELOG.md`; #2439 already adds a `## 0.2.0` section there and a competing one would only give that PR a conflict. `renderers/web_core/CHANGELOG.md` I did leave alone deliberately as well — the web_core change here is test-only plus the number-literal fix, and I would rather add the entry in the wording you prefer. Happy to add both on request.

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].

For this PR:

- [ ] I have updated the relevant CHANGELOG.md file. (see the note above)
- [x] I updated/added relevant documentation. (`conformance/README.md`)
- [ ] If my branch is on a fork, I have verified that [scripts/e2e_test.sh](https://github.com/a2ui-project/a2ui/blob/main/scripts/e2e_test.sh) passes. (both package suites and the conformance YAML validation pass)


renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
```diff
@@ -244,7 +244,12 @@ export class ExpressionParser {
     while (!scanner.isAtEnd() && (this.isDigit(scanner.peek()) || scanner.peek() === '.')) {
       scanner.advance();
     }
-    return Number(scanner.input.substring(start, scanner.pos));
+    const text = scanner.input.substring(start, scanner.pos);
+    const value = Number(text);
+    if (!Number.isFinite(value)) {
+      throw new A2uiExpressionError(`Invalid number literal: '${text}'`);
+    }
+    return value;
```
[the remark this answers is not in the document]
diegolopezrm: Adopted the regex in both implementations, though the premise needs a correction.

`num.tryParse("1.")` returns 1.0 on the Dart VM, not null, and `.5` returns 0.5. I also compiled the same code with dart2js and ran it under Node to check the VM/web split specifically — the two agree on every input tried:

| input | Dart VM | dart2js | `Number()` |
| --- | --- | --- | --- |
| `1.` | 1.0 | 1 | 1 |
| `.5` | 0.5 | 0.5 | 0.5 |
| `""` | null | null | 0 |
| `1.2.3` | null | null | NaN |

So the drift you describe is not there today. `.5` and `""` also cannot reach this function — the scanner only enters it on a digit, and stops at anything that is not a digit or a point.

The underlying point stands anyway, and it is the better design: the accepted shape should be stated, not inherited from whichever number parser the platform ships. Both now use the grammar you suggested. The visible consequence is that `${1.}` moves from accepted-by-accident to rejected in both, so I added a conformance case for it in `core/expressions.yaml` — that way the answer lives in the shared suite instead of in two platform parsers, and changing our minds means editing one case.

dart/a2ui_core/lib/src/processing/expressions.dart
```diff
@@ -235,7 +235,12 @@ class ExpressionParser {
         (_isDigit(scanner.peek()) || scanner.peek() == '.')) {
       scanner.advance();
     }
-    return num.parse(scanner.input.substring(start, scanner.pos));
+    final String text = scanner.input.substring(start, scanner.pos);
+    final num? value = num.tryParse(text);
+    if (value == null) {
+      throw A2uiExpressionError("Invalid number literal: '$text'");
+    }
+    return value;
```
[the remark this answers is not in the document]
diegolopezrm: Same change applied here; see the reply on the TypeScript side for the measurements, including a dart2js run that shows the VM and web behave the same on these inputs.

gspencergoog (approved):
This is great!  Thank you for implementing the conformance tests for this.

dart/a2ui_core/lib/src/processing/expressions.dart
```diff
@@ -14,6 +14,12 @@
 
 import '../primitives/errors.dart';
 
+/// Digits, an optional decimal point, and optional further digits.
+///
+/// Every client implementation accepts a trailing point (`1.`) today and none
+/// accepts a second point (`1.2.3`), so the grammar is written to keep that.
+final RegExp _numberLiteral = RegExp(r'^\d+\.?\d*$');
```
gspencergoog, line 21: This doesn't allow negative number literals.  Or exponential notation.  Or commas instead of decimals (locale support).  (I don't think we actually support commas elsewhere yet either, but we shouldn't rule it out.)

gspencergoog: Oh, actually, commas can't work anyhow because of the parser (arguments are comma-separated).  We'd have to do that in the presentation layer.

diegolopezrm: You're right that none of those work, and the comma point answers itself as you say — the argument separator takes it.

Worth separating what this PR changed from what it inherited, though. The limits are in the scanner, not in the validation: `_parseNumberLiteral` is only entered when the current character is a digit, and it consumes only digits and `.`. So the regex validates what the scanner already collected, and the same inputs behave identically with and without this PR:

| template | main | this PR |
| --- | --- | --- |
| `${-1}` | `{path: "-1"}` | `{path: "-1"}` |
| `${-1.5}` | `{path: "-1.5"}` | `{path: "-1.5"}` |
| `${1e3}` | `A2uiExpressionError` | `A2uiExpressionError` |
| `${1,5}` | `A2uiExpressionError` | `A2uiExpressionError` |
| `${f(a: -1)}` | `args: {a: {path: "-1"}}` | `args: {a: {path: "-1"}}` |

The negative case is the one I'd flag: `${-1}` does not fail, it parses as a **data binding** to the path `-1`, because `-` is one of the characters `scanPathOrIdentifier` accepts. `${f(a: -1)}` therefore hands the function a path rather than the number the author wrote, and nothing reports it. Silent wrong value rather than an error.

Supporting negatives and exponents means changing the scanner: enter `parseNumberLiteral` on `-` when a digit follows, and consume `e`/`E` with an optional sign. The negative half needs care precisely because `-` is currently a valid path character, so `${-1}` and `${a-b}` have to keep meaning different things.

Happy to do it as its own PR across the four implementations, with conformance cases for the negative, exponent and `${a-b}` forms so the behaviour is pinned rather than assumed. It is a behaviour change on inputs that today resolve to a path, so it seemed wrong to fold into this one.

gspencergoog: Oooh, nice catch.  Okay, sure, let's do a separate PR to fix the path problem and the negative parsing, and exponentials.  I was focusing on this regex but you're right that the parser needs to support it too.

Issue #2498, opened 2026-09-03
[BUG]: Dart and web_core DataModel diverge on 221 of 1,392 operation programs


diegolopezrm:
- [x] I have searched the existing issues to make sure this bug has not already been reported.

## Describe the Bug

Following the same method as #2496, but pointed at `DataModel` — the state a client reports back to the agent when `sendDataModel` is on — I ran the Dart (`dart/a2ui_core`) and TypeScript (`renderers/web_core`) implementations over the same 1,392 generated operation programs and compared the resulting state, reads and errors. They disagree on 221.

Reduced to minimal cases:

| # | operations | Dart | web_core |
| --- | --- | --- | --- |
| 1 | `set('/a', 1)` then `set('/a', null)` | key removed → `{}` | `{a: null}` |
| 2 | `set('/a/b', 's')` then `set('/a/b/c', 1)` | **silently writes nothing** | `A2uiDataError` |
| 3 | `set('/', 1)` then `set('/a', 2)` | **silently writes nothing** | **`TypeError`** |
| 4 | `set('/', false)` then `set('/a', 2)` | silently writes nothing | **root discarded** → `{a: 2}` |
| 5 | `set('/__proto__/x', 1)` | stored as an ordinary key | `A2uiDataError` |
| 6 | `set('/items/100000', 1)` | `A2uiDataError` (capped at 10,000) | array grown to 100,001 |

Cases 2, 3 and 4 are one rule broken three ways, and I have sent a PR for them. The rest are decisions rather than defects, so they are here rather than in that PR.

### The `null` write (case 1) — 137 of the 221 programs

Dart removes the key; web_core stores the null. The v0.9 schema for `updateDataModel` distinguishes the two intents by whether `value` is present at all, so on the wire `{"path": "/a", "value": null}` and `{"path": "/a"}` are different messages — but Dart's `DataModel` has no way to express the difference, since `set(path, null)` is its removal path. `test/data_model_test.dart` pins this deliberately:

```dart
test('removes keys when setting null', () {
  final model = DataModel({'foo': 'bar'});
  model.set('/foo', null);
  expect(model.get('/'), isEmpty);
});
```

So an agent sending `value: null` clears the key on Flutter and sets it to null on web, and a subsequent client data model report differs accordingly. This is the same underlying gap as #1896, one layer down and in this repository's Dart core rather than in `package:genui`.

[...]

PR #2499, opened 2026-09-03
fix(web_core): reject writes through a primitive value consistently

diegolopezrm:
# Description

Addresses the web_core half of #2498. Rewritten after rebasing: this PR originally fixed the same bug in Dart and TypeScript, and #2439 has since landed the Dart fix, so what is left is web_core plus the Dart tests for the fix that arrived without any.

## What this PR changes

**One branch in `DataModel.set`**, which is why the two rows below are a single diff hunk:

```ts
-    if (!this.data) {
+    if (this.data === undefined || this.data === null) {
       this.data = {};
+    } else if (typeof this.data !== 'object') {
+      throw new A2uiDataError(
+        `Cannot set path '${path}': the data model root is a primitive value.`,
+        path,
+      );
     }
```

The old guard was a truthiness test, so a primitive root behaved differently depending on whether it happened to be truthy:

| operations | before | after |
| --- | --- | --- |
| `set('/', 1)`, `set('/a', 2)` | `TypeError: Cannot create property 'a' on number '1'` | `A2uiDataError`, root preserved |
| `set('/', false)`, `set('/a', 2)` | root discarded, result `{a: 2}` | `A2uiDataError`, root preserved |

A truthy primitive root fell through to `current[lastSegment] = value` on a number or a string; modules run in strict mode, so that throws a raw `TypeError` — outside the `A2uiError` hierarchy consumers catch, and not what the method's own doc comment promises:

```ts
 * @throws {A2uiDataError} If path is null, undefined, invalid for
 *     arrays/primitives, or contains forbidden segments
```

A falsy primitive root took the other path: `!this.data` was true, so a root holding `false`, `0` or `''` — a value the caller put there — was silently replaced with a fresh object.

Verified against the built code, all six primitive roots now behave the same way and keep their value:

```
set('/', 1)      then set('/a', 2) -> A2uiDataError, root still 1
set('/', "text") then set('/a', 2) -> A2uiDataError, root still "text"
set('/', true)   then set('/a', 2) -> A2uiDataError, root still true
set('/', false)  then set('/a', 2) -> A2uiDataError, root still false
set('/', 0)      then set('/a', 2) -> A2uiDataError, root still 0
set('/', "")     then set('/a', 2) -> A2uiDataError, root still ""
```

**Two Dart tests**, covering the fix #2439 landed rather than one of mine. They assert the error type rather than its message, so they pass against that wording unmodified, and the case they pin — a write when the root itself is a primitive — had no test on main. Confirmed with @gspencergoog in review that keeping them here is fine.

## What this PR no longer changes

The Dart fix in `DataModel.set`. #2439 landed the same correction at the same branch, raising the same `A2uiDataError`; I kept its wording and dropped mine.

## What is not addressed here

Three findings from #2498 are decisions rather than defects, and none is in this PR:

- `set(path, null)` removes the key in Dart and stores the null in web_core — 137 of the 221 diverging programs. The v0.9 schema distinguishes the two intents by whether `value` is present, but Dart's `DataModel` has no way to express that, and `test/data_model_test.dart` pins the removal deliberately. Fixing it means splitting the setter from a remover and having `MessageProcessor` branch on presence.
- `__proto__` is rejected by web_core and stored as an ordinary key by Dart.
- Unbounded list indices in web_core, already filed as #2420.

## Tests

Two per language, all four failing on `main` before the respective fix:

- web_core — a primitive root rejected with `A2uiDataError` rather than `TypeError`, and each falsy root (`false`, `0`, `''`) surviving a rejected write.
- Dart — a write through a primitive, and a write when the root is a primitive, each expecting `A2uiDataError` with the untouched value still readable.

`node --test "dist/**/*.test.js"` in `renderers/web_core`: 421 passing. `eslint src/v0_9/state/` reports 33 warnings, exactly as on main. `dart test` in `dart/a2ui_core`: 324 passing; `dart analyze` and `dart format` clean.

## Compatibility

Both changes turn silence into an error, so a client that was quietly losing writes will now hear about them. No existing test in either package expected the old behaviour.

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].

For this PR:

- [ ] I have updated the relevant CHANGELOG.md file. (glad to add entries; say where you would like them)
- [x] I updated/added relevant documentation. (the root-guard comment explains what it protects)
- [ ] If my branch is on a fork, I have verified that [scripts/e2e_test.sh](https://github.com/a2ui-project/a2ui/blob/main/scripts/e2e_test.sh) passes. (both package suites pass in full)


dart/a2ui_core/test/data_model_test.dart
```diff
@@ -155,6 +155,20 @@ void main() {
       expect(model.get('/'), isEmpty);
     });
 
+    test('rejects writes through a primitive value', () {
```
gspencergoog, line 158: This is a Dart test, but the rest of the code changes are web_core changes.  Did you mean to separate it like that?

diegolopezrm: Fair question — it did not start that way, and the split is a leftover from a rebase.

The PR originally fixed the same bug in both: Dart's `DataModel.set` walked to the container for the last segment, found neither a map nor a list, and returned normally having written nothing, so the agent got no error and the value simply was not there. web_core had two variants of the same hole — a raw `TypeError` on a primitive root, and a root holding `false`, `0` or `''` being silently discarded, because the guard was `if (!this.data)`.

Then #2439 landed with the same Dart fix, at the same branch, raising the same `A2uiDataError`. I dropped my version of that hunk and kept its wording, which left what you are looking at: the web_core fix, plus the Dart tests, which now cover #2439's fix rather than my own. They assert the error type rather than the message, so they passed against that wording unmodified — and the case they pin, a write when the root itself is a primitive, has no test on main.

So the Dart test is deliberate but not original intent. Happy to split it into its own PR if you would rather review them separately; it is two tests and it moves cleanly.

gspencergoog: No, as long as it is adding to the Dart test corpus, and not a mistake, then this is fine to keep here.

gspencergoog (approved):

PR #2504, opened 2026-09-03
Rename label from 'in-discussion' to 'needs-team-input'

polina-c:


ditman (approved):
LGTM if everybody else agrees!

PR #2527, opened 2026-09-04
fix(expressions): support non-ASCII data model keys in templates (#2500)

### LINKED ISSUE #2500: [BUG]: non-ASCII data model keys are unreachable from templates in Dart, web_core and Python

## Describe the Bug

A data model key containing any non-ASCII character cannot be referenced from a template. `${señor}`, `${café/precio}` and `${日本}` are parse errors in Dart, TypeScript and Python, and parse correctly in Swift.

| template | Dart | TypeScript | Python | Swift |
| --- | --- | --- | --- | --- |
| `${señor}` | error | error | error | `{path: "señor"}` |
| `${café/precio}` | error | error | error | `{path: "café/precio"}` |
| `${日本}` | error | error | error | `{path: "日本"}` |
| `hola ${señor} qué tal` | error | error | error | parses, literals intact |
| `${a_b-c.d}` | ok | ok | ok | ok |

So an agent that names its data with anything but English ASCII produces a surface that renders on iOS and fails everywhere else.

The data model itself has no such limit, which makes the state reachable but undisplayable:

```dart
final m = DataModel(<String, Object?>{});
m.set('/señor', 'Diego');   // fine
m.get('/señor');            // 'Diego'
ExpressionParser().parse(r'${señor}');  // A2uiExpressionError
```

An agent can write the key, and read it back through `getClientDataModel`, but cannot put it on screen.

## Root cause

The identifier scanners disagree, and the pattern is worth noting: the three that fail carry a hand-written ASCII range, while the one that works uses its language's own Unicode-aware predicate.

- `dart/a2ui_core/lib/src/processing/expressions.dart` — `_isAlnum` tests code units `0x30–0x39`, `0x41–0x5A`, `0x61–0x7A`.
- `renderers/web_core/.../expression_parser.ts` — `isAlnum` tests `c >= 'a' && c <= 'z'` and friends.
- `agent_sdks/python/a2ui_core/.../expression_parser.py` — `is_alnum` tests `"a" <= c <= "z"` and friends.
- `swift/core/Sources/BasicCatalog/Functions/ExpressionParser.swift` — `scanPathOrIdentifier` tests `c.isLetter || c.isNumber`, which is Unicode-aware in Swift.

The ASCII range looks like a literal port of the same three lines between the first three implementations; Swift's reads like the idiomatic translation, and is the one that comes out right.

## Expected Behavior

I would suggest aligning the other three with Swift rather than the reverse:

- The data model is JSON, and JSON object keys are Unicode strings.
- Paths are JSON Pointer (RFC 6901), which places no restriction on the characters in a key.
- Nothing in the v0.9 specification or in `basic_catalog_implementation_guide.md` says an identifier inside `${...}` must be ASCII.
- Restricting it makes part of a legal data model unreachable from the UI, which is the asymmetry shown above.

Each fix is a couple of lines: Dart has `RegExp(r'\p{L}|\p{N}', unicode: true)` or `String.fromCharCode` category checks, TypeScript has `/[\p{L}\p{N}]/u`, Python has `str.isalnum()`.

If instead the intent is that identifiers are ASCII-only, then that belongs in the guide, and Swift should be tightened — but then agents need a documented way to reach a key that ASCII cannot spell.

## Steps to Reproduce

```
${señor}
```

through each implementation's `ExpressionParser.parse`. Full table above.

## Environment Details

- **OS**: macOS
- **Browser/Platform**: Dart 3.12.2, Node.js 25.2.1, Python 3.13, Swift 6.2.1
- **SDK/Package Name & Version**: all four at main (676a899)
- **Protocol Version**: v0.9

## Additional Context

Found by running all four parser implementations over the same corpus, the method used for #2496 and #2498, now extended from two implementations to four. Two other results from the same run:

- On the `null` keyword left open in #2496, Dart is alone: TypeScript, Python and Swift all produce `''`. That is worth weighing in that discussion, though the guide's wording still points the other way.
- The Python parser raises bare `ValueError` for every parse failure, rather than the `A2uiParseError` its own SDK defines in `a2ui/core/exceptions.py`. Since the conformance schema's error categories map onto that hierarchy, a Python harness for a parser suite cannot currently distinguish a protocol error from any other `ValueError`. Filing separately if useful.

Varun-S10:
# Description

Fixes #2500.

### What is changing and why
Templates with non-English or accented words like `${señor}`, `${café/precio}`, and `${日本}` crashed with an error in Dart, TypeScript (web_core), and Python. At the same time, they worked fine in Swift.

The DataModel can store any language characters, but the template parser only checked for English letters (`a-z`, `A-Z`, `0-9`). When it saw a character like `ñ` or `é`, it stopped reading early and threw an error.

This PR fixes Dart, TypeScript, and Python so they support all Unicode letters and numbers, matching Swift:
- **TypeScript (`web_core`)**: Updated the parser to allow Unicode letters and numbers, with a fast check for English letters so performance stays fast.
- **Python**: Updated the parser to use Python's built-in `isalnum()` to allow Unicode letters.
- **Dart**: Updated the parser to allow Unicode letters and numbers.
- **Tests**: Added unit tests for `${señor}`, `${café/precio}`, and `${日本}` across all three test suites.

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].

For this PR:

### DIFF
--- a/agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
+++ b/agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
@@ -50,7 +50,7 @@ class Scanner:
     def matches_keyword(self, keyword: str) -> bool:
         if self.input.startswith(keyword, self.pos):
             next_char = self.peek(len(keyword))
-            if not re.match(r"[a-zA-Z0-9_]", next_char):
+            if not (next_char.isalnum() or next_char == "_"):
                 self.advance(len(keyword))
                 return True
         return False
@@ -264,7 +264,12 @@ class ExpressionParser:
         return int(num_str)
 
     def is_alnum(self, c: str) -> bool:
-        return ("a" <= c <= "z") or ("A" <= c <= "Z") or ("0" <= c <= "9")
+        return (
+            ("a" <= c <= "z")
+            or ("A" <= c <= "Z")
+            or ("0" <= c <= "9")
+            or c.isalnum()
+        )
 
     def is_digit(self, c: str) -> bool:
         return "0" <= c <= "9"
--- a/agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
+++ b/agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
@@ -264,12 +264,7 @@ class ExpressionParser:
         return int(num_str)
 
     def is_alnum(self, c: str) -> bool:
-        return (
-            ("a" <= c <= "z")
-            or ("A" <= c <= "Z")
-            or ("0" <= c <= "9")
-            or c.isalnum()
-        )
+        return c.isalnum()
 
     def is_digit(self, c: str) -> bool:
         return "0" <= c <= "9"
--- a/agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
+++ b/agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
 from typing import Any, Dict, List, Union
 
 
--- a/dart/a2ui_core/lib/src/processing/expressions.dart
+++ b/dart/a2ui_core/lib/src/processing/expressions.dart
@@ -240,7 +240,7 @@ class ExpressionParser {
 
   static final RegExp _unicodeAlnum = RegExp(r'\p{L}|\p{N}', unicode: true);
 
-  bool _isAlnum(String c) {
+  static bool _isAlnum(String c) {
     if (c.isEmpty) return false;
     final int u = c.codeUnitAt(0);
     if ((u >= 0x30 && u <= 0x39) || // 0-9
@@ -314,21 +314,7 @@ class _Scanner {
     return input.substring(start, end);
   }
 
-  static bool _isWordChar(String c) {
-    if (c.isEmpty) return false;
-    final int u = c.codeUnitAt(0);
-    if ((u >= 0x30 && u <= 0x39) || // 0-9
-        (u >= 0x41 && u <= 0x5A) || // A-Z
-        (u >= 0x61 && u <= 0x7A) || // a-z
-        u == 0x5F) {
-      // _
-      return true;
-    }
-    if (u < 128) {
-      return false;
-    }
-    return ExpressionParser._unicodeAlnum.hasMatch(c);
-  }
+  static bool _isWordChar(String c) => c == '_' || ExpressionParser._isAlnum(c);
 
   static bool _isWhitespace(int u) {
     return u == 0x20 || u == 0x09 || u == 0x0A || u == 0x0D; // space/tab/LF/CR
--- a/dart/a2ui_core/lib/src/processing/expressions.dart
+++ b/dart/a2ui_core/lib/src/processing/expressions.dart
@@ -238,12 +238,21 @@ class ExpressionParser {
     return num.parse(scanner.input.substring(start, scanner.pos));
   }
 
+  static final RegExp _unicodeAlnum = RegExp(r'\p{L}|\p{N}', unicode: true);
+
   bool _isAlnum(String c) {
     if (c.isEmpty) return false;
     final int u = c.codeUnitAt(0);
-    return (u >= 0x30 && u <= 0x39) || // 0-9
+    if ((u >= 0x30 && u <= 0x39) || // 0-9
         (u >= 0x41 && u <= 0x5A) || // A-Z
-        (u >= 0x61 && u <= 0x7A); // a-z
+        (u >= 0x61 && u <= 0x7A)) {
+      // a-z
+      return true;
+    }
+    if (u < 128) {
+      return false;
+    }
+    return _unicodeAlnum.hasMatch(c);
   }
 
   bool _isDigit(String c) {
@@ -287,7 +296,7 @@ class _Scanner {
   bool matchesKeyword(String keyword) {
     if (input.startsWith(keyword, pos)) {
       final String next = peek(keyword.length);
-      if (next.isEmpty || !_isWordChar(next.codeUnitAt(0))) {
+      if (next.isEmpty || !_isWordChar(next)) {
         advance(keyword.length);
         return true;
       }
@@ -305,11 +314,20 @@ class _Scanner {
     return input.substring(start, end);
   }
 
-  static bool _isWordChar(int u) {
-    return (u >= 0x30 && u <= 0x39) || // 0-9
+  static bool _isWordChar(String c) {
+    if (c.isEmpty) return false;
+    final int u = c.codeUnitAt(0);
+    if ((u >= 0x30 && u <= 0x39) || // 0-9
         (u >= 0x41 && u <= 0x5A) || // A-Z
         (u >= 0x61 && u <= 0x7A) || // a-z
-        u == 0x5F; // _
+        u == 0x5F) {
+      // _
+      return true;
+    }
+    if (u < 128) {
+      return false;
+    }
+    return ExpressionParser._unicodeAlnum.hasMatch(c);
   }
 
   static bool _isWhitespace(int u) {
--- a/renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
+++ b/renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
@@ -247,8 +247,16 @@ export class ExpressionParser {
     return Number(scanner.input.substring(start, scanner.pos));
   }
 
+  private static readonly UNICODE_ALNUM = /\p{L}|\p{N}/u;
+
   private isAlnum(c: string): boolean {
-    return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9');
+    if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')) {
+      return true;
+    }
+    if (c.charCodeAt(0) < 128) {
+      return false;
+    }
+    return ExpressionParser.UNICODE_ALNUM.test(c);
   }
 
   private isDigit(c: string): boolean {
@@ -297,7 +305,7 @@ class Scanner {
   matchesKeyword(keyword: string): boolean {
     if (this.input.startsWith(keyword, this.pos)) {
       const next = this.peek(keyword.length);
-      if (!/[a-zA-Z0-9_]/.test(next)) {
+      if (!/[a-zA-Z0-9_]/.test(next) && !/\p{L}|\p{N}/u.test(next)) {
         this.advance(keyword.length);
         return true;
       }
--- a/renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
+++ b/renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
@@ -249,7 +249,7 @@ export class ExpressionParser {
 
   private static readonly UNICODE_ALNUM = /\p{L}|\p{N}/u;
 
-  private isAlnum(c: string): boolean {
+  static isAlnum(c: string): boolean {
     if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')) {
       return true;
     }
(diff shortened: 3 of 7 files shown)


agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
```diff
@@ -264,7 +264,12 @@ def parse_number_literal(self, scanner: Scanner) -> Union[int, float]:
         return int(num_str)
 
     def is_alnum(self, c: str) -> bool:
-        return ("a" <= c <= "z") or ("A" <= c <= "Z") or ("0" <= c <= "9")
+        return (
+            ("a" <= c <= "z")
+            or ("A" <= c <= "Z")
+            or ("0" <= c <= "9")
+            or c.isalnum()
+        )
```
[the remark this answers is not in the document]
Varun-S10: Applied. Simplified is_alnum to return c.isalnum() directly and cleaned up unused imports.

dart/a2ui_core/lib/src/processing/expressions.dart
```diff
@@ -238,12 +238,21 @@ class ExpressionParser {
...
-        (u >= 0x61 && u <= 0x7A); // a-z
+        (u >= 0x61 && u <= 0x7A)) {
+      // a-z
+      return true;
+    }
+    if (u < 128) {
+      return false;
+    }
+    return _unicodeAlnum.hasMatch(c);
   }
```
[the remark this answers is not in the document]
Varun-S10: Applied. Made _isAlnum static so it can be shared with _Scanner.

dart/a2ui_core/lib/src/processing/expressions.dart
```diff
@@ -305,11 +314,20 @@ class _Scanner {
...
-        u == 0x5F; // _
+        u == 0x5F) {
+      // _
+      return true;
+    }
+    if (u < 128) {
+      return false;
+    }
+    return ExpressionParser._unicodeAlnum.hasMatch(c);
   }
```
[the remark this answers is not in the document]
Varun-S10: Applied. Updated isWordChar' to reuse c == '_' || ExpressionParser._isAlnum(c).

renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
```diff
@@ -247,8 +247,16 @@ export class ExpressionParser {
...
   private isAlnum(c: string): boolean {
-    return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9');
+    if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')) {
+      return true;
+    }
+    if (c.charCodeAt(0) < 128) {
+      return false;
+    }
+    return ExpressionParser.UNICODE_ALNUM.test(c);
   }
```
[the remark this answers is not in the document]
Varun-S10: Applied. Removed the private modifier and exposed UNICODE_ALNUM as static readonly.

renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
```diff
@@ -297,7 +305,7 @@ class Scanner {
   matchesKeyword(keyword: string): boolean {
     if (this.input.startsWith(keyword, this.pos)) {
       const next = this.peek(keyword.length);
-      if (!/[a-zA-Z0-9_]/.test(next)) {
+      if (!/[a-zA-Z0-9_]/.test(next) && !/\p{L}|\p{N}/u.test(next)) {
         this.advance(keyword.length);
         return true;
       }
```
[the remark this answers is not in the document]
Varun-S10: Applied. Reused the shared ExpressionParser.UNICODE_ALNUM regex in matchesKeyword.

agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
```diff
@@ -50,7 +49,7 @@ def matches_string(self, expected: str) -> bool:
     def matches_keyword(self, keyword: str) -> bool:
         if self.input.startswith(keyword, self.pos):
             next_char = self.peek(len(keyword))
-            if not re.match(r"[a-zA-Z0-9_]", next_char):
+            if not (next_char.isalnum() or next_char == "_"):
```
gspencergoog, line 52: Instead of this, we should be using Unicode UAX `#31` identifiers, not just isalnum + _.  That is what is specified in the 1.0 specification. See https://www.unicode.org/reports/tr31/

See https://github.com/a2ui-project/a2ui/blob/main/specification/v1_0/docs/a2ui_protocol.md#architectural-principles

Varun-S10: Done. Updated to use Unicode UAX `#31` identifiers across Python, TypeScript, and Dart, and added tests. Could you please check it? @gspencergoog

agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
```diff
@@ -12,10 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
 from typing import Any, Dict, List, Union
 
 
+def _is_id_continue(c: str) -> bool:
```
gspencergoog, line 18: This seems like it is fragile.  How about this instead?

```py
import regex

# Compiled pattern matching exactly one XID_Continue character
XID_CONTINUE_PATTERN = regex.compile(r'^\p{XID_Continue}$')

def is_xid_continue_regex(char: str) -> bool:
    return bool(XID_CONTINUE_PATTERN.match(char))
```

dart/a2ui_core/lib/src/processing/expressions.dart
```diff
@@ -259,12 +258,22 @@ class ExpressionParser {
...
+        (u >= 0x61 && u <= 0x7A) || // a-z
+        u == 0x5F) {
+      // _
+      return true;
+    }
+    if (u < 128) {
+      return false;
+    }
+    return _xidContinue.hasMatch(c);
   }
```
gspencergoog, line 277: This also seems fragile.

Try this:

```dart
final xidContinueRegex = RegExp(r'^\p{XID_Continue}$', unicode: true);

bool _isIdContinue(String char) {
  if (char.runes.length != 1) {
    return false;
  }

  return xidContinueRegex.hasMatch(char);
}
```

dart/a2ui_core/lib/src/processing/expressions.dart
```diff
@@ -326,12 +335,7 @@ class _Scanner {
     return input.substring(start, end);
   }
 
-  static bool _isWordChar(int u) {
-    return (u >= 0x30 && u <= 0x39) || // 0-9
-        (u >= 0x41 && u <= 0x5A) || // A-Z
-        (u >= 0x61 && u <= 0x7A) || // a-z
-        u == 0x5F; // _
-  }
+  static bool _isWordChar(String c) => ExpressionParser._isIdContinue(c);
```
gspencergoog, line 338: You might not need this function at all, since you're just calling it in one place.  Just replace the call with a call to `_isIdContinue`.

renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
```diff
@@ -266,8 +266,23 @@ export class ExpressionParser {
     return Number(text);
   }
 
+  static readonly XID_CONTINUE = /\p{XID_Continue}/u;
+  /** @deprecated Kept for backwards compatibility. */
+  static readonly UNICODE_ALNUM = ExpressionParser.XID_CONTINUE;
+
+  static isIdContinue(c: string): boolean {
```
gspencergoog, line 273: This is also fragile.  Use this:

```ts
const XID_CONTINUE_REGEX = /^\p{XID_Continue}$/u;

function isIdContinue(input: string): boolean {
  if (Array.from(input).length !== 1) {
    return false;
  }

  return XID_CONTINUE_REGEX.test(input);
}
```

renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
gspencergoog, line 304: This doesn't seem like it will handle Unicode characters correctly. It will split multi-byte characters incorrectly. It should be something like this:

```ts
  peek(offset = 0): string {
    const targetPos = this.pos + offset;
    if (targetPos >= this.input.length) return '\0';

    // 1. Get the numeric 32-bit code point at the target index
    const codePoint = this.input.codePointAt(targetPos);
    if (codePoint === undefined) return '\0';

    // 2. Convert that code point back to its full string representation.
    // String.fromCodePoint correctly reconstructs surrogate pairs automatically.
    return String.fromCodePoint(codePoint);
  }
```

dart/a2ui_core/lib/src/processing/expressions.dart
gspencergoog, line 297: You'll probably need a better peek here to handle Unicode:

```dart
  // Unicode-safe peek: Returns a full String representation of the code point,
  // correctly handling surrogate pairs.
  String peek([int offset = 0]) {
    // 1. Convert the input to a list of 32-bit runes (code points)
    final runesList = input.runes.toList();

    if (pos + offset >= runesList.length) return '';

    // 2. Safely extract the full 32-bit code point
    final codePoint = runesList[pos + offset];

    // 3. Convert it back to a valid String
    return String.fromCharCode(codePoint);
  }
```

agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
gspencergoog, line 41: This peek is fine: because Python is smart about Unicode.

gspencergoog (commented):
I'd also make sure that you're only trying to pass a single Unicode character at a time to the "is XID_Continue" tests.

Note that in Unicode encodings there a lot of kinds of "character":

- A byte is a single ASCII character, but only part of a Unicode character.
- A 16-bit word covers most Unicode characters, but not all. It's what a lot of languages use, though.
- A full Unicode code point (32-bit) which can also describe a surrogate pair.
- A multi-byte Grapheme Cluster which can also describe emoji and other complex characters.

agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
Varun-S10: Done. Added the regex package dependency and updated _is_id_continue to use regex.compile(r'^\p{XID_Continue}$').

dart/a2ui_core/lib/src/processing/expressions.dart
Varun-S10: Updated to RegExp(r'^\p{XID_Continue}$', unicode: true) and added the char.runes.length != 1 check.

dart/a2ui_core/lib/src/processing/expressions.dart
Varun-S10: Done, removed _isWordChar and called _isIdContinue directly in matchesKeyword.

renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
Varun-S10: Updated isIdContinue to use /^\p{XID_Continue}$/u and check Array.from(input).length !== 1.

renderers/web_core/src/v0_9/basic_catalog/expressions/expression_parser.ts
Varun-S10: Updated peek() to use codePointAt and String.fromCodePoint, and updated advance() to step by the character's UTF-16 length.

dart/a2ui_core/lib/src/processing/expressions.dart
Varun-S10: Updated peek() to extract the full 32-bit code point and updated advance() to step by the character's UTF-16 length so surrogate pairs are handled properly.

agent_sdks/python/a2ui_core/src/a2ui/core/basic_catalog/expression_parser.py
Varun-S10: Kept Python peek() as is.

Varun-S10:
@gspencergoog, could you please check?

gspencergoog (approved):

gspencergoog:
Thanks, Varun! Resolve the conflicts and we can commit it.

Varun-S10:
> Thanks, Varun! Resolve the conflicts and we can commit it.

Thanks @gspencergoog. I resolved the merge conflict in dart/a2ui_core/CHANGELOG.md. could you please check?

Issue #846, opened 2026-03-13
Strict Type Coercion in DataContext


jacobsimionato:
# Proposal: Strict Type Coercion in DataContext

## Reason
The A2UI protocol defines standard coercion rules (e.g., `null` resolving to `""` for strings). Without central enforcement, component authors must manually handle these edge cases, leading to bugs like `[object Object]` appearing in text labels.

## Design
1. **Coercion Utils**: Implement the coercion table from the `renderer_guide.md` in a shared utility.
2. **Context Integration**: Update `DataContext.resolveDynamicValue` and `resolveSignal` to apply these rules.
3. **Primitive Safety**:
   - String targets: Convert any non-string to localized string; map `null/undefined` to `""`.
   - Number targets: Parse strings; map `null/undefined` to `0`.
   - Boolean targets: Map `"true"` (case-insensitive) and non-zero numbers to `true`.
4. **GenericBinder Alignment**: Ensure the `GenericBinder` uses these coerced values when populating `ResolvedProps`.

## Key Files
- `renderers/web_core/src/v0_9/rendering/data-context.ts`
- `renderers/web_core/src/v0_9/rendering/generic-binder.ts`

jacobsimionato:
I believe we should still consider investigating and fixing this issue.

polina-c:
This issue should be considered for a P1 priority together with other P2 issues in the next planning cycle after we burn through the triage queue.

PR #2529, opened 2026-09-04
fix(web_core): make the setter for a binding-only prop callable

### LINKED ISSUE #2528: [BUG]: GenericBinder generates an uncallable setter for a binding-only property

### Background

`@a2ui/web_core`'s `GenericBinder` (`renderers/web_core/src/v0_9/rendering/generic-binder.ts`) is
the engine that turns a component's raw A2UI JSON config plus its Zod schema into the resolved,
strongly typed props a renderer hands to a component implementation. The React and Lit renderers
both build their component props on it (Angular does not, since it rolls its own binding).

Two exported utility types in that file describe the result:

- `ResolveA2uiProp<T>` types what a component **reads**: an `Action` becomes `() => void`, a
  `DynamicString` becomes `string`, and so on.
- `GenerateSetters<T>` types what a component **writes**: for a `value: DynamicString` property
  it generates a `setValue(v: string)` method for two-way binding.

Both start from the same idea. A dynamic property is declared as a union of literal branches plus
the binding shapes, for example `DynamicString = string | DataBinding | FunctionCall`. To find the
resolved type, both subtract the binding shapes: `Exclude<T, DataBinding | FunctionCall>`.

### The bug

A property can be declared with **only** binding branches and no literal one, for example
`z.union([DataBindingSchema, FunctionCallSchema])`. For such a property the subtraction removes
everything and leaves nothing, and the two types disagree about what to do next.

`ResolveA2uiProp` guards that case and falls back to `any` (lines 175-177):

```ts
: Exclude<T, DynamicTypes> extends never
  ? any
  : Exclude<T, DynamicTypes>;
```

`GenerateSetters` has no equivalent guard (same file, line 185):

```ts
value: Exclude<NonNullable<T[K]>, DynamicTypes>,
```

So for a property declared as `DataBinding | FunctionCall` the read side falls back to `any`,
while the write side resolves to `(value: never) => void`. `never` is uninhabited, so no value
can be passed and the generated setter cannot be called at all. The asymmetry is the whole
defect: the read path has a fallback and the write path was never given one.

No component in the basic catalog declares such a property, and `DataBinding` is absent from
`COMMON_TYPE_SCHEMAS` in `src/v0_9/catalog/schema_loader.ts`, so JSON-defined catalogs cannot
declare one either. The case is reachable from hand-written Zod component schemas, which is
presumably why it has gone unnoticed.

## Steps to Reproduce

1. Declare a component whose schema has a property with only binding branches.
2. Read the generated setter's parameter type.

```ts
import {z} from 'zod';
import {DataBindingSchema, FunctionCallSchema} from '@a2ui/web_core/v0_9';
import type {ResolveA2uiProps} from '@a2ui/web_core/v0_9';

const BindingOnly = z.object({sort: z.union([DataBindingSchema, FunctionCallSchema])});
declare const props: ResolveA2uiProps<z.infer<typeof BindingOnly>>;

props.sort; // any - the read-side fallback applies
props.setSort; // (value: never) => void - uncallable

props.setSort({path: '/sortOrder'});
```

3. Run `tsc --noEmit`. The call on the last line is rejected.

## Expected Behavior

The generated setter is callable. The read side already decides what to do when the subtraction
leaves nothing; the write side should make the same decision rather than producing a parameter
no value can satisfy.

## Screenshots / Video / Logs

```
error TS2345: Argument of type '{ path: string; }' is not assignable to parameter of type 'never'.
```

## Environment Details

- **OS**: macOS 15
- **Browser/Platform**: Node.js, TypeScript 5.9.3 (compile-time only, no runtime behaviour involved)
- **SDK/Package Name & Version**: `@a2ui/web_core` 0.10.7 (reproduced against `main` @ `0dc4a30f`)
- **Protocol Version**: v0.9, v0.9.1
- **Agent Framework & LLM Model**: n/a

## Additional Context

**Suggested fix** - give the write side the fallback the read side already has:

```ts
value: [Exclude<NonNullable<T[K]>, DynamicTypes>] extends [never]
  ? unknown
  : Exclude<NonNullable<T[K]>, DynamicTypes>,
```

`unknown` rather than `any`: it accepts every argument while keeping the value checked at the
call site, and does not add to the `any` count tracked in #1296.

This is non-breaking. A parameter of type `never` admits no argument today, so widening it can
only permit calls that previously failed to compile; no existing call site can change meaning.

retz8:
# Description

`GenerateSetters` types a generated setter's parameter as `never` when the property is declared as a binding with no literal branch, making the setter uncallable. Fixes #2528.

## Fix

Give the write side the fallback the read side already has:

```diff
-    value: Exclude<NonNullable<T[K]>, DynamicTypes>,
+    value: [Exclude<NonNullable<T[K]>, DynamicTypes>] extends [never]
+      ? unknown
+      : Exclude<NonNullable<T[K]>, DynamicTypes>,
```

`unknown` rather than `any`: it accepts every argument while keeping the value checked at the call site, and does not add to the `any` count tracked in #1296.

**Non-breaking.** A parameter of type `never` admits no argument, so widening it can only permit calls that previously failed to compile. No existing call site can change meaning.

**No runtime change.** The edit is a type annotation inside a type alias. 

## Verification

- `@a2ui/web_core`: 458/458 existing tests pass, `tsc --noEmit` clean, `format:check` clean, lint
  reports 0 errors (402 pre-existing warnings, unchanged).
- `@a2ui/react`: `typecheck` clean.
- `@a2ui/lit`, `@a2ui/react`: one pre-existing failure each, `Live Invitation Builder should
  render date`, which asserts on a locale/timezone-formatted date and fails identically on a
  clean checkout of `main` in this environment. Unrelated to this change, which emits no
  JavaScript.

## Scope

No component in the basic catalog declares a binding-only property, and `DataBinding` is absent
from `COMMON_TYPE_SCHEMAS` in `src/v0_9/catalog/schema_loader.ts`, so JSON-defined catalogs
cannot declare one. The case is reachable from hand-written Zod component schemas.

## Related, not addressed here

`ResolveA2uiProp` types a bound property by its literal branches, so a `DynamicString` property whose path holds a non-string is still typed `string`. That is a separate problem with a different fix, and is closer to #846 than to this PR.

## Update after review (09.10.26)

Added a compile-time test in `generic-binder.test.ts` covering this type, per review feedback: it asserts the setter accepts a value when the property has no literal branch, and that a literal branch still constrains the setter (`@ts-expect-error`).

Dropped the `ResolveA2uiProp` comparison from the `GenerateSetters` doc comment, per review feedback.

Verified the test fails without the fix: reverting the type edit makes `tsc` exit 2 with `error TS2345: Argument of type '"anything"' is not assignable to parameter of type 'never'`, and the same for the `unknown` value. With the fix restored, `@a2ui/web_core` is 496/496 tests passing, `tsc --noEmit` clean, `format:check` clean, lint 0 errors (402 warnings, unchanged). The library JavaScript in `dist` is still byte-identical to `main`; only the test file is new.

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].

For this PR:

      No test accompanies this change. It is a type-only edit with no runtime behaviour to
      exercise, and the repository has no existing type-level test harness.

### DIFF
--- a/renderers/web_core/src/v0_9/rendering/generic-binder.ts
+++ b/renderers/web_core/src/v0_9/rendering/generic-binder.ts
@@ -179,10 +179,14 @@ export type ResolveA2uiProp<T> = [NonNullable<T>] extends [Action]
 /**
  * Automatically generates two-way binding setters for dynamic properties.
  * If a schema has a `value: DynamicString`, this type generates a `setValue(val: string)` method.
+ * A property declared as a binding with no literal branch has no such value type, so its setter
+ * falls back to `unknown`, mirroring the fallback `ResolveA2uiProp` applies to the read side.
  */
 export type GenerateSetters<T> = {
   [K in keyof T as IsDynamic<T[K]> extends true ? `set${Capitalize<string & K>}` : never]-?: (
-    value: Exclude<NonNullable<T[K]>, DynamicTypes>,
+    value: [Exclude<NonNullable<T[K]>, DynamicTypes>] extends [never]
+      ? unknown
+      : Exclude<NonNullable<T[K]>, DynamicTypes>,
   ) => void;
 };
 
--- a/renderers/web_core/src/v0_9/rendering/generic-binder.ts
+++ b/renderers/web_core/src/v0_9/rendering/generic-binder.ts
@@ -180,7 +180,7 @@ export type ResolveA2uiProp<T> = [NonNullable<T>] extends [Action]
  * Automatically generates two-way binding setters for dynamic properties.
  * If a schema has a `value: DynamicString`, this type generates a `setValue(val: string)` method.
  * A property declared as a binding with no literal branch has no such value type, so its setter
- * falls back to `unknown`, mirroring the fallback `ResolveA2uiProp` applies to the read side.
+ * falls back to `unknown`.
  */
 export type GenerateSetters<T> = {
   [K in keyof T as IsDynamic<T[K]> extends true ? `set${Capitalize<string & K>}` : never]-?: (
--- a/renderers/web_core/CHANGELOG.md
+++ b/renderers/web_core/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- (v0_9) Fix the setter `GenericBinder` generates for a property declared as a binding with no literal branch: its parameter resolved to `never`, making the setter uncallable. It now falls back to `unknown`, matching the fallback `ResolveA2uiProp` already applies to the resolved property ([#2528](https://github.com/a2ui-project/a2ui/issues/2528)).
 - (v0_9) Deprecate `injectBasicCatalogStyles`, `computeColorVariant`, `ColorVariantLightDarkOptions`, and `ColorVariantHoverOptions` exports from `@a2ui/web_core/v0_9`. Consumers should import them from `@a2ui/web_core/v0_9/basic_catalog` instead.
 - (v0_9) An invalid number literal in an expression, such as `${1.2.3}`, now throws `A2uiExpressionError` instead of being handed back as `NaN`. The accepted shape — digits, an optional decimal point and optional further digits — is stated in the parser rather than inherited from `Number()`, so every implementation accepts the same literals. ([#2497](https://github.com/a2ui-project/a2ui/pull/2497))
 - (v0_9) The expression parser now runs the shared conformance suite at `conformance/core/expressions.yaml`, alongside the Dart client. Adds `js-yaml` as a dev dependency to read it. ([#2497](https://github.com/a2ui-project/a2ui/pull/2497))
--- a/renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
+++ b/renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
@@ -17,12 +17,12 @@
 import * as assert from 'node:assert';
 import {describe, it} from 'node:test';
 import {z} from 'zod';
-import {GenericBinder, scrapeSchemaBehavior} from './generic-binder.js';
+import {GenericBinder, scrapeSchemaBehavior, type GenerateSetters} from './generic-binder.js';
 import {ComponentContext} from './component-context.js';
 import {SurfaceModel} from '../state/surface-model.js';
 import {Catalog} from '../catalog/types.js';
 import {ComponentModel} from '../state/component-model.js';
-import {CommonSchemas} from '../schema/common-types.js';
+import {CommonSchemas, type DataBinding, type FunctionCall} from '../schema/common-types.js';
 
 describe('GenericBinder Checkable Trait', () => {
   const mockCatalog = new Catalog('test', [], []);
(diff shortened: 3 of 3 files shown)


renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -179,10 +179,14 @@ export type ResolveA2uiProp<T> = [NonNullable<T>] extends [Action]
...
  * If a schema has a `value: DynamicString`, this type generates a `setValue(val: string)` method.
+ * A property declared as a binding with no literal branch has no such value type, so its setter
+ * falls back to `unknown`, mirroring the fallback `ResolveA2uiProp` applies to the read side.
  */
 export type GenerateSetters<T> = {
   [K in keyof T as IsDynamic<T[K]> extends true ? `set${Capitalize<string & K>}` : never]-?: (
-    value: Exclude<NonNullable<T[K]>, DynamicTypes>,
+    value: [Exclude<NonNullable<T[K]>, DynamicTypes>] extends [never]
+      ? unknown
+      : Exclude<NonNullable<T[K]>, DynamicTypes>,
```
andrewkolos, line 189: This is a great fix for this type, so I'd love to get it under test so it can't be undone accidentally. My suggestion is as follows:

In `generic-binder.test.ts`, add `GenerateSetters` to the binder import and `DataBinding` and `FunctionCall` to the common-types import, all as type imports:

```ts
it('keeps the setter callable for a property declared with no literal branch', () => {
  const bindingOnly: GenerateSetters<{value: DataBinding | FunctionCall}> = {
    setValue: () => {},
  };
  bindingOnly.setValue('anything');
  const value: unknown = {selected: true};
  bindingOnly.setValue(value);

  const withLiteral: GenerateSetters<{value: string | DataBinding | FunctionCall}> = {
    setValue: () => {},
  };
  withLiteral.setValue('ok');
  // @ts-expect-error a literal branch must still constrain the setter.
  withLiteral.setValue(123);
});
```

renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -179,10 +179,14 @@ export type ResolveA2uiProp<T> = [NonNullable<T>] extends [Action]
 /**
  * Automatically generates two-way binding setters for dynamic properties.
  * If a schema has a `value: DynamicString`, this type generates a `setValue(val: string)` method.
+ * A property declared as a binding with no literal branch has no such value type, so its setter
+ * falls back to `unknown`, mirroring the fallback `ResolveA2uiProp` applies to the read side.
```
andrewkolos: Nit:

I'm not sure the comparison to `ResolveA2uiProp` is warranted here.

```suggestion
 * A property declared as a binding with no literal branch has no such value type, so its setter
 * falls back to `unknown`.
```

I'm happy to reconsider.

andrewkolos (changes requested):
Looks like a great contribution--thank you! I left just one blocking comment

renderers/web_core/src/v0_9/rendering/generic-binder.ts
retz8: Thank you for the review! You are right that this should not be able to regress silently. Added it in commit 98d356fe.

I also checked that it actually guards the fix: reverting the type edit makes `tsc` exit 2 with `error TS2345: Argument of type '"anything"' is not assignable to parameter of type 'never'`, and the same for the `unknown` value.

renderers/web_core/src/v0_9/rendering/generic-binder.ts
retz8: Good point, and happy to take it. Applied in d0ec9c68.

It also ties this comment to another type's behaviour without anything enforcing that they stay in step, so if `ResolveA2uiProp` changes, the comment goes stale silently. Better left to the issue.

josemontespg:
@andrewkolos retz8 addressed both of your comments on 2026-09-11: the type-level test is in 98d356fe and the doc comment change in d0ec9c68. Ready for another look.

@retz8 the branch conflicts with main again, only in renderers/web_core/CHANGELOG.md (a new entry landed under "Unreleased" on 2026-09-14). A re-merge of main should be enough.

PR #2531, opened 2026-09-04
fix(web_core): classify dynamic unions nested inside wider unions as DYNAMIC

### LINKED ISSUE #2530: [BUG]: GenericBinder misclassifies nested dynamic unions as STATIC

### Background

`@a2ui/web_core`'s `GenericBinder` decides per property how to treat the incoming JSON value by scraping the component's Zod schema (`getFieldBehavior` in `renderers/web_core/src/v0_9/rendering/generic-binder.ts`). A property classified `DYNAMIC` gets the full data-binding treatment: `{path}` and `{call}` values are resolved through the `DataContext`, subscribed for updates, and a runtime setter is generated. A property classified `STATIC` is passed through untouched.

`getFieldBehavior` recognizes a dynamic property two ways:

- a `REF:` marker in the schema description (`#/$defs/DynamicString` and friends)
- a structural check that scans a `ZodUnion`'s options for the `DataBindingSchema` object shape
  (an object with `path` and no `componentId`).

Both checks look exactly one level deep.

### The bug

A schema that composes a `Dynamic*` schema into a wider union defeats both checks. The basic catalog's own `DateTimeInput` does this for `min` and `max` (`renderers/web_core/src/v0_9/basic_catalog/components/basic_components.ts`):

```ts
'min': z
  .union([DynamicStringSchema, z.string().date(), z.string().time(), z.string().datetime()])
  .describe('The minimum allowed date/time in ISO 8601 format.')
  .optional(),
```

~~The outer `.describe()` replaces the description, so the union carries no `REF:` marker.~~

> **Correction (per review on #2531):** `.describe()` is not what hides the marker. `z.union([...])` creates a new schema with no description of its own, so the outer union never had a `REF:` marker to replace. The marker stays on the nested `DynamicStringSchema`; `getFieldBehavior` simply does not read the descriptions of a union's options. Removing the `.describe()` would not change the outcome.

And `DynamicStringSchema` is itself a `ZodUnion`, not a `ZodObject`, so the structural scan never sees the `{path}` branch nested inside it. Both properties scrape as `STATIC`, while the sibling `value`, a bare `DynamicStringSchema`, scrapes as `DYNAMIC`.

The consequences, all silent:

- A bound `min` reaches the component implementation as the raw `{path: '...'}` object: never resolved, never subscribed. The React implementation's defensive check (`typeof props.min === 'string' ? props.min : undefined`) then discards the object, so the constraint is simply dropped.
- The type layer and the runtime disagree about setters. `GenerateSetters` sees the `DataBinding` branch in the union and declares `setMin`/`setMax` on the resolved props type, but the binder only generates runtime setters for `DYNAMIC` properties, so the typed method is `undefined` at runtime.

The trap is not specific to `DateTimeInput`: any catalog schema that wraps a `Dynamic*` schema inside a wider union (hand-written Zod catalogs can) gets the same misclassification.

## Steps to Reproduce

1. Scrape the shipped `DateTimeInput` schema:

```ts
import {scrapeSchemaBehavior} from '@a2ui/web_core/v0_9';
import {DateTimeInputApi} from '@a2ui/web_core/v0_9/basic_catalog';

const behavior = scrapeSchemaBehavior(DateTimeInputApi.schema);
console.log(behavior.shape.min);   // {type: 'STATIC'}  - wrong
console.log(behavior.shape.value); // {type: 'DYNAMIC'} - correct, same DynamicStringSchema
```

2. Or observe it end to end: render a surface with a `DateTimeInput` whose `min` is `{"path": "/limits/min"}` and any value at `/limits/min` in the data model. The rendered input has no minimum constraint, and updating `/limits/min` has no effect.

## Expected Behavior

`min` and `max` classify as `DYNAMIC` like their sibling `value`: a bound value is resolved, subscribed, and delivered to the component as a string, and the generated setters exist at runtime as the props type promises.

## Environment Details

- **OS**: macOS 15
- **Browser/Platform**: Node.js 22
- **SDK/Package Name & Version**: `@a2ui/web_core` 0.10.7 (reproduced against `main` @ `0dc4a30f`)
- **Protocol Version**: v0.9,v0.9.1
- **Agent Framework & LLM Model**: n/a

## Additional Context

**Suggested fix** - make the structural union check recurse: a union any of whose branches is itself dynamic (nested `Dynamic*` union, or a branch carrying a `Dynamic*`/`DataBinding` `REF:` description) is dynamic. The fix belongs in the scraper rather than in the `DateTimeInput` schema, because third-party catalogs composing `Dynamic*` schemas into wider unions hit the same misclassification.

Prior art on the `v1_0` branch, where this is already solved twice over:

- the v1.0 basic catalog flattened `min`/`max` to a bare `DynamicStringSchema` with the `REF:` marker preserved, removing the trigger
- #2353 rewrites the classifier so the union check reads each option's `REF:` definition name (`getRefDefName(o).startsWith('Dynamic')`), which classifies this exact shape as `DYNAMIC`

retz8:
# Description

`GenericBinder`'s schema scraper checks a union's options for the binding shape one level deep, so a `Dynamic*` schema composed into a wider union scrapes as `STATIC` and its bindings are silently ignored. The basic catalog's `DateTimeInput` `min`/`max` are declared exactly this way. Full mechanism and reproduction in the issue. Fixes #2530!

## Fix

Make the structural union check recurse into nested union options:

```diff
-    // Dynamic strings/values are unions containing DataBindingSchema { path: ... } but NOT { componentId: ... }
-    const isDynamic = options.some(
-      o => o._def.typeName === 'ZodObject' && o._def.shape().path && !o._def.shape().componentId,
-    );
-    if (isDynamic) return {type: 'DYNAMIC'};
+    // Dynamic strings/values are unions containing DataBindingSchema { path: ... } but NOT
+    // { componentId: ... }. The binding branch may sit inside a nested union, e.g. a Dynamic*
+    // schema composed into a wider union of literal formats.
+    const hasDynamicBranch = (opts: z.ZodTypeAny[]): boolean =>
+      opts.some(o =>
+        o._def.typeName === 'ZodUnion'
+          ? hasDynamicBranch(o._def.options as z.ZodTypeAny[])
+          : o._def.typeName === 'ZodObject' && o._def.shape().path && !o._def.shape().componentId,
+      );
+    if (hasDynamicBranch(options)) return {type: 'DYNAMIC'};
```

The predicate cannot misfire on the other special unions: `ActionSchema`'s branches are `{event}` / `{functionCall}` objects with no `path`, and `ChildList`'s dynamic-template object carries `componentId`, which the predicate excludes.

**Behavior change, and an intended one.** `DateTimeInput` `min`/`max` bindings now resolve, subscribe, and update reactively, and the runtime `setMin`/`setMax` setters exist as the resolved props type has always promised. A catalog schema that composes a `Dynamic*` schema into a wider union gets the same correction.

## Verification

- `@a2ui/web_core`: 459/459 tests pass (458 pre-existing plus a new regression test covering the `min`/`max` shape and a description-less nested union). `format:check` clean; lint reports 0 errors (404 warnings vs 402 on `main`. Two new ones are `(behavior as any)` casts in the new test, matching the idiom of the surrounding scraper tests).
- Scraping the shipped `DateTimeInputApi.schema` now yields `DYNAMIC` for `min`, `max`, and `value`; on `main` it yields `STATIC` for `min`/`max`.
- `@a2ui/react`: 131/132, `@a2ui/lit`: 137/138. The single failure in each is the pre-existing `Live Invitation Builder should render date` test, which asserts on a locale/timezone formatted date and fails identically on a clean checkout of `main` in this environment.

## Pre-launch Checklist

One time:

- [x] I signed the [CLA].

For this PR:

### DIFF
--- a/renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
+++ b/renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
@@ -383,6 +383,19 @@ describe('GenericBinder Checkable Trait', () => {
       assert.deepStrictEqual((behavior as any).shape.min, {type: 'DYNAMIC'});
       assert.deepStrictEqual((behavior as any).shape.deep, {type: 'DYNAMIC'});
     });
+
+    it('should see dynamic branches through Optional/Nullable/Default wrappers', () => {
+      const schema = z.object({
+        a: z.union([DynamicStringSchema.nullable(), z.number()]),
+        b: z.union([DynamicStringSchema.optional(), z.number()]),
+        c: z.union([DynamicStringSchema.default('fallback'), z.number()]),
+      });
+
+      const behavior = scrapeSchemaBehavior(schema);
+      assert.deepStrictEqual((behavior as any).shape.a, {type: 'DYNAMIC'});
+      assert.deepStrictEqual((behavior as any).shape.b, {type: 'DYNAMIC'});
+      assert.deepStrictEqual((behavior as any).shape.c, {type: 'DYNAMIC'});
+    });
   });
 
   describe('Static behavior for unannotated schemas', () => {
--- a/renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
+++ b/renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
@@ -22,7 +22,7 @@ import {ComponentContext} from './component-context.js';
 import {SurfaceModel} from '../state/surface-model.js';
 import {Catalog} from '../catalog/types.js';
 import {ComponentModel} from '../state/component-model.js';
-import {CommonSchemas} from '../schema/common-types.js';
+import {CommonSchemas, DynamicStringSchema} from '../schema/common-types.js';
 
 describe('GenericBinder Checkable Trait', () => {
   const mockCatalog = new Catalog('test', [], []);
@@ -362,6 +362,26 @@ describe('GenericBinder Checkable Trait', () => {
       assert.strictEqual(nestedBehavior.type, 'OBJECT');
       assert.strictEqual((nestedBehavior as any).shape.value.type, 'OBJECT');
       assert.strictEqual((nestedBehavior as any).shape.text.type, 'ARRAY');
+    });
+  });
+
+  describe('scrapeSchemaBehavior nested dynamic unions', () => {
+    it('should classify a union containing a Dynamic* schema as DYNAMIC', () => {
+      // The basic catalog's DateTimeInput declares min/max exactly like this: the
+      // outer describe() replaces the REF: marker and DynamicStringSchema is a
+      // nested union, so neither detection path may rely on one level only.
+      const schema = z.object({
+        min: z
+          .union([DynamicStringSchema, z.string().date(), z.string().time(), z.string().datetime()])
+          .describe('The minimum allowed date/time in ISO 8601 format.')
+          .optional(),
+        deep: z.union([z.union([z.string(), z.object({path: z.string()})]), z.number()]),
+      });
+
+      const behavior = scrapeSchemaBehavior(schema);
+      assert.strictEqual(behavior.type, 'OBJECT');
+      assert.deepStrictEqual((behavior as any).shape.min, {type: 'DYNAMIC'});
+      assert.deepStrictEqual((behavior as any).shape.deep, {type: 'DYNAMIC'});
     });
   });
 
--- a/renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
+++ b/renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
@@ -367,9 +367,8 @@ describe('GenericBinder Checkable Trait', () => {
 
   describe('scrapeSchemaBehavior nested dynamic unions', () => {
     it('should classify a union containing a Dynamic* schema as DYNAMIC', () => {
-      // The basic catalog's DateTimeInput declares min/max exactly like this: the
-      // outer describe() replaces the REF: marker and DynamicStringSchema is a
-      // nested union, so neither detection path may rely on one level only.
+      // Matches DateTimeInput's min/max schema, where DynamicStringSchema
+      // is nested inside another union.
       const schema = z.object({
         min: z
           .union([DynamicStringSchema, z.string().date(), z.string().time(), z.string().datetime()])
--- a/renderers/web_core/src/v0_9/rendering/generic-binder.ts
+++ b/renderers/web_core/src/v0_9/rendering/generic-binder.ts
@@ -66,6 +66,18 @@ const ACTION_REF = 'REF:common_types.json#/$defs/Action';
 const DATA_BINDING_REF = 'REF:common_types.json#/$defs/DataBinding';
 const DYNAMIC_REF_PREFIX = '#/$defs/Dynamic';
 
+function unwrapModifiers(type: z.ZodTypeAny): z.ZodTypeAny {
+  let current = type;
+  while (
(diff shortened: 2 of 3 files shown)


renderers/web_core/src/v0_9/rendering/generic-binder.ts
```diff
@@ -114,11 +114,16 @@ function getFieldBehavior(type: z.ZodTypeAny, propertyName?: string): BehaviorNo
...
-    if (isDynamic) return {type: 'DYNAMIC'};
+    // Dynamic strings/values are unions containing DataBindingSchema { path: ... } but NOT
+    // { componentId: ... }. The binding branch may sit inside a nested union, e.g. a Dynamic*
+    // schema composed into a wider union of literal formats.
+    const hasDynamicBranch = (opts: z.ZodTypeAny[]): boolean =>
+      opts.some(o =>
+        o._def.typeName === 'ZodUnion'
+          ? hasDynamicBranch(o._def.options as z.ZodTypeAny[])
+          : o._def.typeName === 'ZodObject' && o._def.shape().path && !o._def.shape().componentId,
+      );
```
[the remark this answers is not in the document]
retz8: Modifier-unwrapping point is adopted in fe4b716c: each union option is now unwrapped through `ZodOptional`/`ZodNullable`/`ZodDefault` before inspection, using the same modifier list `getFieldBehavior` already unwraps at the top level, with a test covering all three wrappers.

The suggested code sample is not applied, for two reasons:

1. `isComponentIdSchema` does not exist in `web_core`.
2. Replacing the `path`-without-`componentId` check with a ComponentId-based one changes what is being detected. This predicate looks for the `DataBinding` branch (`{path: string}`, a plain string property), which contains no ComponentId-typed property, so the suggested predicate would classify no property as `DYNAMIC` and break existing behavior.

renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
```diff
@@ -365,6 +365,39 @@ describe('GenericBinder Checkable Trait', () => {
     });
   });
 
+  describe('scrapeSchemaBehavior nested dynamic unions', () => {
+    it('should classify a union containing a Dynamic* schema as DYNAMIC', () => {
+      // The basic catalog's DateTimeInput declares min/max exactly like this: the
+      // outer describe() replaces the REF: marker and DynamicStringSchema is a
+      // nested union, so neither detection path may rely on one level only.
```
andrewkolos: Does the outer `describe()` replace a `REF:` marker? I'm not seeing it.

```suggestion
      // Matches DateTimeInput's min/max schema, where DynamicStringSchema
      // is nested inside another union.
```

andrewkolos (changes requested):
One nit, otherwise LGTM!

renderers/web_core/src/v0_9/rendering/generic-binder.test.ts
retz8: Thank you for catching this! Applied your wording in e5b45e84.

I checked it against the schemas to be sure: `z.union([...])` creates a new schema with no description, so there is no `REF:` marker for `.describe()` to replace. The marker stays on the nested `DynamicStringSchema`; the scraper just does not read union options' descriptions. Dropping the `.describe()` would not change the outcome.

josemontespg:
@andrewkolos the nit is applied (e5b45e84, 2026-09-11), so this only needs your approval now.

@retz8 the branch conflicts with main again, only in renderers/web_core/CHANGELOG.md (a new entry landed under "Unreleased" on 2026-09-14). A re-merge of main should be enough.

PR #2451, opened 2026-08-31
release: prepare web SDK packages for release

jacobsimionato:
### Summary

Prepares the A2UI Web/TypeScript SDK packages for public release by updating package version numbers and moving unreleased items into version release sections in `CHANGELOG.md`.

### Packages Released

- **`@a2ui/web_core`**: `0.10.6` -> `0.10.7`
- **`@a2ui/lit`**: `0.10.3` -> `0.10.4`
- **`@a2ui/angular`**: `0.10.5` -> `0.10.6`
- **`@a2ui/react`**: `0.10.2` -> `0.10.3`

### Verification

- Package versions incremented using `./renderers/scripts/increment_version.mjs`.
- All `CHANGELOG.md` headings updated cleanly.

PR #2190, opened 2026-08-07
refactor(web_core, lit): move Lit basic catalog component implementations to web_core

josemontespg:
## Overview

Moves the 18 Basic Catalog Web Component implementations from `@a2ui/lit` into `@a2ui/web_core`, making them available as framework-agnostic universal custom elements that can be shared across all web renderers.

## Details

- **Universal Components**: Relocates all 18 Basic Catalog components (`AudioPlayer`, `Button`, `Card`, `CheckBox`, `ChoicePicker`, `Column`, `DateTimeInput`, `Divider`, `Icon`, `Image`, `List`, `Modal`, `Row`, `Slider`, `Tabs`, `Text`, `TextField`, `Video`) into `@a2ui/web_core/v0_9/basic_catalog/components/`.
- **Framework-Agnostic Element Base**: Provides `BasicCatalogA2uiLitElement` and controller utilities in `@a2ui/web_core/v0_9/basic_catalog` with Light DOM rendering and shared styling.
- **Backwards Compatibility**: Re-exports all components, controllers, context helpers, and catalog definitions from `@a2ui/lit` with JSDoc notices to ensure existing code and consumers continue working without breaking changes.
- **TypeScript Typing**: Strongly types controllers, lifecycle hooks (`willUpdate`), and custom element tags.

## Pre-launch Checklist

- [x] I signed the [CLA].
- [x] I read the [Contributors Guide].
- [x] I read the [Style Guide].
- [x] I have built and run at least one [sample].
- [x] I have updated the relevant CHANGELOG.md file.
- [x] I updated/added relevant documentation.
- [x] My code changes (if any) have tests.

[CLA]: https://cla.developers.google.com/
[Contributors Guide]: https://github.com/a2ui-project/a2ui/blob/main/CONTRIBUTING.md
[Style Guide]: https://github.com/a2ui-project/a2ui/blob/main/CONTRIBUTING.md#coding-style
[sample]: https://github.com/a2ui-project/a2ui/blob/main/samples/README.md#samples

Related to #1270

PR #2535, opened 2026-09-04
release: bump web SDK packages (web_core 0.10.8, lit 0.11.0, angular 0.10.7, react 0.11.1)

andrewkolos:
(note: PR description now out of date; not updating since this is closed)

### Summary

Bumps the web packages that have unreleased changes and moves their `## Unreleased` changelog entries into release sections.

`@a2ui/lit` gets a minor bump instead of a patch because #2205 changes the DOM structure and CSS class names of several Basic Catalog components.

### Version bumps

- **`@a2ui/web_core`**: `0.10.7` -> `0.10.8`
- **`@a2ui/lit`**: `0.10.4` -> `0.11.0`
- **`@a2ui/angular`**: `0.10.6` -> `0.10.7`
- **`@a2ui/react`**: `0.11.0` -> `0.11.1`

### Changelog fixes

- `renderers/web_core/CHANGELOG.md`: restores the `## 0.10.7` header that #2190 removed. Without it, the six entries #2451 listed under that header sat under `## Unreleased`.
- `renderers/angular/CHANGELOG.md`: restores the `## 0.10.6` header that #2449 removed. Without it, the entry #2451 listed under that header sat under `## Unreleased`.
- Adds an entry under `web_core` 0.10.8 for the `DataModel.set` change in #2499, which shipped without one.
- Drops entries under `lit` 0.11.0 and `web_core` 0.10.8 for changes with no effect on the published packages: test coverage, and an export cleanup and a getter change to code that was itself unreleased.
- Rewrites the `lit` entry for #2190. `@a2ui/lit/v0_9` does not re-export the component classes and never did; what it gains is `BasicCatalogA2uiLitElement`. The `A2uiLitElement.controller` getter is a break for `@a2ui/lit` subclasses, so its entry lives there.

### Verification

- Versions incremented with `./renderers/scripts/increment_version.mjs`.
- `npm view @a2ui/<package> version` matches the left-hand number of each bullet above.

### DIFF
--- a/renderers/web_core/CHANGELOG.md
+++ b/renderers/web_core/CHANGELOG.md
@@ -1,9 +1,15 @@
 ## Unreleased
 
+## 0.10.8
+
 - (v0_9) Add unit test coverage for all basic catalog Web Component implementations. [#2357](https://github.com/a2ui-project/a2ui/pull/2357)
 - (v0_9) Replace `A2uiLitElement.controller` property with a read-only getter to disallow external reassignment, simplify style root target resolution, and replace basic catalog barrel wildcard exports with explicit exports.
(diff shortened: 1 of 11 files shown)


renderers/web_core/CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
 ## Unreleased
 
-- (v0_9) Add unit test coverage for all basic catalog Web Component implementations. [#2357](https://github.com/a2ui-project/a2ui/pull/2357)
-- (v0_9) Replace `A2uiLitElement.controller` property with a read-only getter to disallow external reassignment, simplify style root target resolution, and replace basic catalog barrel wildcard exports with explicit exports.
+## 0.10.8
```
josemontespg: What do you think about leaving an advisory note here

```
- Downstream consumers using `@a2ui/lit` must upgrade both `@a2ui/lit` and `@a2ui/web_core` to `0.11.0` together. Using `@a2ui/web_core@0.11.0` alongside `@a2ui/lit <= 0.10.4` will result in a runtime `NotSupportedError: This name has already been registered in the registry` due to colliding custom element registrations.
```

PR #2537, opened 2026-09-05
test(conformance): expand v1.0 protocol conformance suites and align core engines

gspencergoog:
## Summary

Expands the shared, language-agnostic conformance corpus for protocol v1.0, then fixes the places where the TypeScript, Python and Kotlin engines disagreed with it. The TypeScript runner is now wired into `yarn test` and web CI, so the corpus is enforced instead of only being present.

Addresses #2496

### Conformance corpus (`conformance/`)

New suites:

- `core/message_processor_v1_0.yaml`: v1.0 message lifecycle, component
  replacement and surface management.
- `core/accessibility.yaml`: accessibility attributes and live region
  announcements.
- `core/data_deletion.yaml`: key and element deletion semantics.
- `core/data_model_pointers.yaml`: RFC 6901 pointer navigation and relative
  scoping.
- `extensions/adk/adk_extensions.yaml`: ADK subagent surface mapping and SSRF
  address policies.

Expanded suites: `core/functions.yaml`, `core/expressions.yaml`, `core/data_model.yaml`, `core/rpc_functions.yaml`, `core/multi_catalog.yaml`, `core/composition_constraints.yaml`, `core/message_processor_v0_9.yaml`, `agent/inference_format.yaml`, `agent/parser.yaml`, `agent/streaming_parser.yaml`.

### Test harnesses

- The TypeScript runner (`tests/conformance/conformance_test.mjs`) now resolves a protocol version per case, from the case's `protocolVersion`, the `version` field on its messages, or the catalog id, instead of assuming v1.0. It also honors `catalogPaths` and orders explicitly named catalogs first. The Python harness already did all of this. Making the two agree is what let the v0.9 message-processor cases run under TypeScript at all.
- Actions the TypeScript engine does not implement are listed in a named `UNIMPLEMENTED_ACTIONS` allowlist with a reason. Any other unknown action still throws.
- The Python harness now fails rather than skipping in three cases: a suite that will not parse, a case with no `name` or `action`, and an action with no handler. A new `test_all_suites_load` reports parse failures directly.
- Suites the core library cannot run are named in `UNRUNNABLE_SUITES` with a reason. `accessibility.yaml` is listed there because the core libraries have no access to the UI frameworks that set the attributes. The renderers will own that suite.

### Catalog reference resolution

The published catalogs reference the shared types across documents, for example `common_types.json#/$defs/ChildList`. `inline_local_refs` only resolves references that start with `#/`, so those survived into the loaded schema and `PayloadValidator` compensated by reading `specification/**/common_types.json` off disk at validation time. That tree is not part of the wheel, so the fallback worked only from a source checkout.

- `Catalog.from_json` accepts a `common_types_schema` argument, rewrites cross-document references into local `#/$defs/...` pointers, and satisfies them from that document or from the built-in definitions.
- `A2uiCatalog.core_catalog` passes its `common_types_schema` through.
- `PayloadValidator` no longer reads schema files. It does now report `referencing.exceptions.Unresolvable` as an `invalid_reference` validation error rather than swallowing it.
- `test_specification_drift.py` compares the built-in shared types against the published document. Four definitions differ; each is listed with its reason, and the test asserts the listed ones still differ, so the list cannot rot.

### Engine fixes

TypeScript (`typescript/web_core`):

- The expression parser accepts signed numbers and exponent notation.
- `GenericBinder` preserves `functionCall` and `event` wrappers instead of flattening them at bind time, and resolves `event.userMessage` before dispatch. `userMessage` is a `DynamicString`, and `renderer_to_agent.json` states it is sent already resolved.
- Catalog loading validates component property names, function names and function argument names against UAX `#31`. Only component names were checked before.
- The message processor gained a headless data context for standalone RPC execution, and normalizes catalog version strings by stripping a leading `v`.
- Deleting an array index past the end is a no-op.
- `IntegrityChecker` no longer treats a bare `child` string property as a component pointer.

Python (`python/a2ui_core`, `python/a2ui_agent`):

- The expression parser accepts signed numbers and exponent notation, keeping `float` for decimal literals.
- `DataModel.delete(path)` was added for parity with TypeScript. Deleting a list index past the end is a no-op.
- Registering a second pending agent call with an in-use `functionCallId` raises an `A2uiRpcError` with a `DUPLICATE` code. `TIMEOUT` and `DUPLICATE` were added to `RpcErrorCode`.
- The v0.8 child-property heuristic in `component_model.py` rejects a list of known scalar property names before applying its naming rules.
- `DirectJsonStreamParser` guards non-string list entries and excludes properties that are not yet complete.
- Function argument validation merges `$defs` from the function schema with those from the catalog schema.

Kotlin (`kotlin/agent_sdk_legacy`):

- `BlockLexer` was rewritten as an explicit state machine. It tracks string literals, escapes, triple quotes and comments. It also matches the open and close tags by regex, so attributes on the open tag are tolerated.
- Matching changes in `Parser`, `PayloadFixer` and `A2uiConstants`.
- `ktfmt` is ordered after the ANTLR source generation task.

### CI

- New `test:conformance` wireit target in `typescript/web_core`; `test` depends on it.
- `conformance/**` added to the path filter for the `renderers` job in `web_ci.yml`, so a change to the corpus triggers the TypeScript run.

## Impact & Risks

Behavior changes a consumer could notice:

- `PayloadValidator` no longer reads from `specification/` at runtime. A catalog that references the shared types across documents must now have its `common_types_schema` passed to `Catalog.from_json`; otherwise those references resolve against the built-in definitions.
- Catalog loading in TypeScript is stricter for v1.0 and later. A catalog with a property, function or argument name that does not satisfy UAX `#31` now throws where it previously loaded.
- The TypeScript error text for an `updateComponents` entry with no `id` changed, to match Python and the phrase the shared suite asserts on.
- Deleting an out-of-range array or list index is a no-op in both languages. It previously extended the container: deleting `/items/10` from a three-element array produced an eleven-element array. Deleting an in-range element still leaves a hole rather than splicing; the suite does not yet pin that down.
- `event.userMessage` is resolved before dispatch. An agent that previously received a raw `{path}` object will now receive the resolved string.
- Registering a duplicate `functionCallId` now raises instead of overwriting the earlier callback.
- `yarn test` in `web_core` now runs the conformance corpus, so it takes longer and can fail for conformance reasons.

No schema changes. The v0.9 and v1.0 fixtures still load.

## Testing

1. Python, from the repository root:

   ```bash
   uv run pytest conformance/tests python/a2ui_core/tests python/a2ui_agent/tests -q
   uv run pyink --check .
   uv run --all-packages mypy .
   ```

   1097 passed, 34 skipped. `pyink` and `mypy` clean.

2. TypeScript:

   ```bash
   yarn workspace @a2ui/web_core test
   yarn workspace @a2ui/web_core format:check
   ```

   680 unit tests, plus 397 of 397 conformance cases and 6 skipped.

3. To confirm the CI gate is real, break an expectation in
   `conformance/core/data_model.yaml` and re-run step 2. `yarn test` should exit
   non-zero. Wireit caches `test:conformance`, so touch the file rather than
   re-running an unchanged tree.

Test changes in this PR:

- The conformance suites listed above.
- `python/a2ui_core/tests/test_specification_drift.py`, new.
- `python/a2ui_agent/tests/conformance/test_adk_extensions.py`, new.
- Unit tests for the schema loader, both expression parsers, the generic binder
  and the message processor.
- The two golden catalog fixtures under `conformance/test_data/`, regenerated
  for the restored descriptions.

## Pre-launch Checklist

- [x] I updated the relevant CHANGELOG.md file.


conformance/agent/inference_format.yaml
```diff
@@ -360,3 +363,106 @@
                   - id: "root"
                     component: "Text"
                     text: "Welcome to A2UI v1.0"
+            - version: "v0.9"
```
nan-yu: `v0.9` -> `v1.0`

conformance/agent/streaming_parser.yaml
```diff
@@ -3039,3 +3039,200 @@
...
+              createSurface:
+                surfaceId: "main"
+                catalogId: "basic"
+                components:
+                  - id: "root"
+                    component: "Text"
+                    text: "Streaming v1.0"
+                dataModel:
+                  count: 10
+            - version: "v0.9"
```
nan-yu: Should the assertion fail with the unexpected v0.9 version?

conformance/core/data_model.yaml
```diff
@@ -652,3 +652,54 @@
       path: "/k"
       value: "w"
       expect_notified: ["/k"]
+
+- name: test_data_model_reactivity_child_deletion_notifies_parent_and_child
+  description: >-
+    Deleting a child property notifies both the child path and ancestor containers.
+  catalog:
+    version: "0.9"
```
nan-yu: Other test specs use `protocolVersion` as the property key, and the value should have a leading `v`.

typescript/web_core/src/catalog/schema_loader.ts
```diff
@@ -477,9 +484,18 @@ export function loadCatalogFromSchema(
   const anyComp = defs?.anyComponent as Record<string, unknown> | undefined;
   const permittedNames = extractPermittedNames(anyComp?.oneOf, '#/components/');
 
+  const protocolVersion = catalogSchema.protocolVersion as string | undefined;
+  const isV10 = protocolVersion === 'v1.0' || protocolVersion === '1.0';
+
   const components: ComponentApi[] = [];
   const componentsMap = (catalogSchema.components as Record<string, unknown>) ?? {};
   for (const [name, rawCompSchema] of Object.entries(componentsMap)) {
+    if (isV10) {
```
nan-yu: This is not forward compatible. Instead of `isV10`, we should check `isAtLeastV10`.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -557,8 +551,16 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
...
+      const targetCatalog =
+        (op.catalogId ? this.catalogs.find(c => c.id === op.catalogId) : undefined) ??
+        this.catalogs[0];
+      const surface =
+        (op.catalogId
+          ? Array.from(this.model.surfacesMap.values()).find(s => s.catalog?.id === op.catalogId)
+          : undefined) ?? this.model.surfacesMap.values().next().value;
+      const dataContext = surface
+        ? new DataContext(surface, '/')
+        : new DataContext(new SurfaceModel('_rpc_fallback', targetCatalog), '/');
```
nan-yu: Do we need to make it unique?

conformance/agent/inference_format.yaml
gspencergoog: Good catch. Switched to 1.0.

conformance/agent/streaming_parser.yaml
gspencergoog: Yes, it should.  Also fixed the parser to record inline components from `createSurface` as yielded, eliminating the spurious `updateComponents` emission, and update the test assertion to expect only the `createSurface` message.

conformance/core/data_model.yaml
gspencergoog: Fixed.

typescript/web_core/src/catalog/schema_loader.ts
gspencergoog: Good point. Updated this to `isAtLeastV10`.

typescript/web_core/src/processing/message-processor.ts
gspencergoog: Updated to use `_rpc_fallback_${op.functionCallId}` so that the fallback has a unique ID.

typescript/web_core/src/catalog/schema_loader.ts
gspencergoog: Turns out that we had several places where it was comparing versions as a float, not semver, so I updated it to actually parse semver.

conformance/agent/inference_format.yaml
```diff
@@ -360,3 +363,99 @@
...
+  format: express
+  input: |
+    <a2ui>
+    root = Column([helloText])
+    helloText = Text("Welcome to Express")
+    </a2ui>
+  expect:
+    - text: ""
+      a2ui:
+        - version: "v1.0"
```
nan-yu, line 381: Nit: the output version is `v1.0` because the test runner sets the protocol version to be v1.0. Can we configure it explicitly in the conformance spec, so nothing needs to be inferred from the runner?

python/a2ui_core/src/a2ui/core/validation/payload_validator.py
```diff
@@ -626,6 +667,117 @@ def validate_theme(self, theme: dict[str, Any]) -> None:
...
+
+        Args:
+            rel_path: Relative specification file path.
+
+        Returns:
+            Parsed schema dictionary if found, or None otherwise.
+        """
+        return _load_spec_file_cached(rel_path)
+
+    def _get_registry(self, target_cat: Any = None) -> Any:
```
nan-yu: Why do we need the cross-schema $ref resolution? My intent is that the catalog schema should be self-contained, for example, https://github.com/a2ui-project/a2ui/blob/v1_0/conformance/test_data/converted_basic_catalog_v10.json.

Then we don't need a resource registry to handle cross-schema reference.

conformance/agent/inference_format.yaml
gspencergoog: Sure, that makes sense.

python/a2ui_core/src/a2ui/core/validation/payload_validator.py
gspencergoog: I would rather we use the specification files directly.

If we use the converted schemas, every time a component or type is updated in `specification/`, developers must remember to run `assemble_catalog.py` to re-generate and commit `converted_basic_catalog_v10.json`. If they forget, tests continue to pass against obsolete fixtures while the actual spec diverges.  We could add some CI to enforce this, but why not read it directly from `specification/`?

If the conformance suites test only against `test_data/converted_basic_catalog_v10.json`, they stop testing whether the engines can consume the actual spec files. Spec errors or regressions in `specification/` could go unnoticed. Pointing language-neutral tests to Python-specific serialized artifacts couples the shared test suite to Python's serialized output rather than the spec files themselves.

conformance/agent/inference_format.yaml
gspencergoog: I configured `catalog: { protocolVersion: 'v1.0' }` explicitly for each format, and updated the runner to extract the protocol version from the spec rather than assuming v1.0.

python/a2ui_core/src/a2ui/core/validation/payload_validator.py
gspencergoog: How about this as a solution: instead of reading spec files from disk or doing resolution,  we could have`Catalog.from_json` inline referenced types in memory using the Pydantic models already defined in `a2ui.core.schema.v1_0.common_types`.

When encountering a `$ref` pointing to `common_types.json#/$defs/<Type>`, it:

- Rewrites the reference to a local pointer: `"#/$defs/<Type>"`.
- Populates `"$defs"` with the schema generated from the corresponding Pydantic model via `TypeAdapter` (extending the pattern `Catalog.catalog_schema` already uses for dynamic types).

This makes the in-memory catalog schema self-contained before it reaches `PayloadValidator`, eliminates the need for `_get_registry`, requires no disk I/O lookups, and works in installed packages.

Then at least it's using the common_types definitions that Python has defined, instead of the ones in the conformance suite, which might be out of date.  I mean the Python ones could be out of date too, but that's at least one fewer thing that needs to be kept up to date when the specs change.

Sound OK?

nan-yu: We are not using the ones in the conformance suite. It is just an example to show we handle pointers to the common types, `"$ref": "#/$defs/DynamicString",` v.s. `"$ref": "common_types.json#/$defs/DynamicString"`, which is exactly what you suggested.

`Catalog.catalog_schema` already uses `_get_dynamic_types_defs()` to generate schema definitions for common/dynamic types directly from `a2ui.core.schema.v1_0.common_types` (the Pydantic models) and inject them into `schema['$defs']`.

Because these types are populated into `$defs` in memory, the catalog schema is self-contained when passed to PayloadValidator, requiring zero disk I/O lookups or external reference resolution.

kotlin/agent_sdk_legacy/src/main/kotlin/com/google/a2ui/parser/BlockLexer.kt
```diff
@@ -16,44 +16,165 @@
...
+          i = match.range.last + 1
+          state = LexerState.IN_A2UI
+          currentRaw.clear()
+          continue
+        } else {
+          currentText.append(content[i])
+          i++
+          continue
+        }
+      }
```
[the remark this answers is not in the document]
gspencergoog: Optimized the tag scanning loop in BlockLexer to search directly for the next openTagPattern match and advance the cursor directly, reducing normal text tokenization from O(N^2) to linear time.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -628,113 +639,122 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
...
+  ): void {
+    const {id, component, ...properties} = comp;
+    const rawCatalogId = (comp as any).catalogId ?? (comp as any).catalogID;
 
-    const surface = this.model.getSurface(op.surfaceId);
-    if (!surface) {
-      throw new A2uiStateError(`Surface not found for message: ${op.surfaceId}`);
+    if (!id) {
+      throw new A2uiValidationError(`Component '${component}' is missing an 'id'.`);
     }
```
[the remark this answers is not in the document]
gspencergoog: Enforced `typeof id !== 'string' || !id` in validateComponentProperties to ensure non-string or empty component IDs are rejected during validation.

python/a2ui_agent/src/a2ui/inference_formats/direct_json/streaming_v09.py
```diff
@@ -229,19 +240,32 @@ def _sniff_partial_data_model(self, messages: list[ResponsePart]) -> None:
...
+    def _record_inline_components(self, sid: str, components: Any) -> None:
+        """Records inline components from createSurface as already yielded."""
+        if not isinstance(components, list):
+            return
+        for comp in components:
+            if isinstance(comp, dict) and 'id' in comp:
+                cid = comp['id']
+                self._seen_components[cid] = comp
+                self._yielded_ids.setdefault(sid, set()).add(cid)
+                self._yielded_contents[(sid, cid)] = json.dumps(comp, sort_keys=True)
```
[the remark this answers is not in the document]
gspencergoog: Updated `_record_inline_components` to access `_components_by_surface.setdefault(sid, {})` directly using the provided `sid` parameter rather than relying on `self.surface_id`.

python/a2ui_core/src/a2ui/core/validation/payload_validator.py
```diff
@@ -12,19 +12,49 @@
...
 from jsonschema import Draft202012Validator
+import jsonschema.exceptions
+import referencing.exceptions
 from ..exceptions import A2uiValidationError, A2uiErrorDetail, A2uiCatalogError
 from ..catalog.catalog import Catalog, TComponent, TFunction
 from ..processing.format_pydantic_error import format_validation_error
 
 
+@functools.lru_cache(maxsize=32)
+def _load_spec_file_cached(rel_path: str) -> dict[str, Any] | None:
```
nan-yu: We've decided to not load from the spec, but use the in-memory Pydantic common objects. This seems not updated yet.

gspencergoog: Updated to not load from the spec, but to include a test to make sure that the spec hasn't diverged from the pydantic definitions.

python/a2ui_core/tests/test_specification_drift.py
```diff
@@ -0,0 +1,131 @@
...
+def test_published_catalog_resolves_without_specification_files() -> None:
+    """The published basic catalog loads with every cross-document reference localised."""
+    with open(BASIC_CATALOG_PATH, "r", encoding="utf-8") as f:
+        catalog_schema = json.load(f)
+    with open(COMMON_TYPES_PATH, "r", encoding="utf-8") as f:
+        common_types_schema = json.load(f)
+
+    catalog = Catalog.from_json(
+        catalog_schema=catalog_schema,
+        protocol_version="v1.0",
```
nan-yu: This is a great safeguard! Just one minor comment, what about updating it to dynamically detect all version directories under specification/ and test them all?

This would save us from having to manually update the latest protocol version whenever we bump or add protocol versions in the future.

gspencergoog: Done!

python/a2ui_core/src/a2ui/core/catalog/catalog.py
nan-yu, line 71: This method returns more than dynamic types definitions now. I'm thinking whether we can generate the dict from the Pydantic model classes directly, instead of hard-coding those common types in the method.

python/a2ui_core/src/a2ui/core/catalog/catalog.py
```diff
@@ -396,12 +591,22 @@ def __init__(
         functions: list[TFunction] | None = None,
         theme_schema: dict[str, Any] | None = None,
         instructions: str | None = None,
+        defs: dict[str, Any] | None = None,
+        common_types_defs: dict[str, Any] | None = None,
```
nan-yu, line 595: Do we allow users to provide any arbitrary common types definitions? I feel like `common_types_defs` should be inferred with the given protocol_version. The Catalogs's constructor and the `from_json` method should only take `protocol_version`, but not both `protocol_version` and `common_types`.

gspencergoog: Well, yeah, this would allow that I guess, but there's a narrower reason for it.

Before this PR, `PayloadValidator` located common types by walking parent directories from the file's location looking for spec files. Which means in a published package, it wouldn't find the spec files. Taking this document argument is what let us remove the disk reads.

I agree with the direction. It's more than just the signature, though: `_get_dynamic_types_defs()` takes no `version` argument and imports `DynamicValue` from `schema.v1_0`, so the built-ins are v1.0-shaped whatever `protocol_version` says. Supplying `common_types_defs` is currently the only thing that makes it possible to validate a v0.9 catalog.

I could do version-aware built-in common types. I'd have to bundle `common_types.json` per protocol version as package data, select by `protocol_version`, and extend `test_specification_drift.py` to check every version instead of only the latest one. This is what would let `common_types_defs` be removed entirely.

python/a2ui_core/src/a2ui/core/catalog/catalog.py
gspencergoog: Yeah, we should rename it to `_get_common_types_defs `, probably.

Four of the six generated Pydantic definitions already are different from `common_types.json`, and they're pinned with a reason in `ACCEPTED_DEVIATIONS` in this PR. This is because the specification uses `not` guards and `allOf` composition that Pydantic doesn't support, so converting the hand-written six would make that list longer (which is OK, but ideally it gets shorter).  We could try and embed the common_types.json into the package data and compare those directly for each version, but I feel like that's a bigger change we should put in a separate PR.

python/a2ui_core/src/a2ui/core/catalog/catalog.py
nan-yu: Talked offline. We've decided to go as-is and update the Pydantic classes for common_types to be version specific and match with the common_types.json file.

nan-yu (approved):

PR #2538, opened 2026-09-05
[blueprint + dart] Scope validator to single catalog.

polina-c:
Contributes to https://github.com/a2ui-project/a2ui/issues/2356, https://github.com/a2ui-project/a2ui/issues/2373.

Follow up for: https://github.com/a2ui-project/a2ui/pull/2439#discussion_r3927367290

## What this PR changes

`A2uiValidator` becomes [`PayloadValidator`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/validation/validator.dart#L52) and holds [one catalog](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/validation/validator.dart#L58) instead of a map of them:

```diff
-  final Map<String, Catalog<C, F>> catalogs;
+  final Catalog<C, F> catalog;
```

A component belongs to exactly one [`Catalog`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/core/catalog.dart#L101), and v0.9 declares `catalogId` on `createSurface` alone, so a validator holding several cannot tell which one applies to a payload that only updates a surface. #2439 worked around that with a `surfaceCatalogs` map threaded through `validate`, `validateStructure` and `validateAgainstCatalogs`, plus a `_catalogFor` ladder that fell back to the sole catalog and otherwise threw. Scoping the validator to one catalog removes the question, and all of that with it.

What remains is three single-item checks, none of which takes a `Catalog` parameter: [`validateComponent`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/validation/validator.dart#L142), [`validateFunction`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/validation/validator.dart#L177) and [`validateTheme`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/validation/validator.dart#L204). A payload that creates a surface against a catalog the processor does not support raises [`A2uiCatalogError`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/primitives/errors.dart#L80).

### MessageProcessor owns catalog resolution

[`MessageProcessor`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/processing/processor.dart#L63) keeps one validator per catalog, built on first use and reused, reachable through [`validatorFor(catalog)`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/processing/processor.dart#L111). The resolved-schema cache lives on the validator, so building a fresh one per batch would re-resolve every component schema on every message.

The catalog for each item is settled by [`catalogFor`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/processing/processor.dart#L127), in this order: the `catalogId` the item names for itself (v1.0 allows one surface to mix catalogs), then the surface's default recorded at `createSurface`, then the sole [supported catalog](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/processing/processor.dart#L65), which is the agent case. `A2uiCatalogError` is raised when none of those settles it, rather than the item being skipped, since skipping would report a payload valid that nothing had checked.

[`processMessages`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/processing/processor.dart#L156) is the single entry point for validation as well as for processing, and it validates as it applies. The `validator:` constructor parameter is replaced by [`protocolVersion`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/processing/processor.dart#L72) and [`commonTypesSchema`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/processing/processor.dart#L80): injecting a single validator means nothing once there are several, and taking the configuration directly keeps the [version](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/primitives/protocol_version.dart#L21) consistent between envelope parsing and catalog checks.

### Envelope parsing moves off the validator

Envelope parsing is [`A2uiMessage.parseAll(payload, protocolVersion:)`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/core/messages.dart#L38). The version tag and the single-update-type rule read no catalog, so parsing belongs to the [message model](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/core/messages.dart#L19) rather than to a validator. A caller holding a raw payload parses it first, which is what lets a payload be read before each message is matched to a surface, and so to a catalog.

### Reference resolution moves to checkSurfaceComplete

`processMessages` checks each batch as a graph against the surface it joins, so duplicate ids, cycles and over-deep chains throw. Whether a reference resolves is deliberately [not checked there](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/processing/processor.dart#L398). A payload may declare a parent before its child, as the basic catalog's [`00_incremental`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/specification/v0_9/catalogs/basic/examples/00_incremental.json) example does, so at arrival time a reference to nothing cannot be told apart from a reference whose target is still in flight.

[`checkSurfaceComplete(surfaceId)`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/processing/processor.dart#L178) is where that is settled. It asserts that a `root` component exists, that every reference resolves, and that every component is reachable from the root. Completeness is something the caller declares: an agent runs it over the surfaces its turn built before sending the payload, and a renderer can run it once a stream ends. A surface carrying only a theme is exempt, since nothing was rendered and there is no root to require.

### On the "block mixed update types" prerequisite

https://github.com/a2ui-project/a2ui/pull/2439#discussion_r3928080732 asks for this first. Dart already enforces it, in [`A2uiMessage.fromJson`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/core/messages.dart#L63), which throws [`A2uiValidationError`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/lib/src/primitives/errors.dart#L26) when an envelope carries more than one of `createSurface`, `updateComponents`, `updateDataModel`, `deleteSurface`. `processMessages` takes sealed typed messages, so it cannot mix by construction. Nothing pinned the rule, so this adds a test.

What actually had to change first was making the envelope stage catalog-independent, since that is the step that has to happen before a message can be matched to its catalog. Batch-level grouping by surface turned out to be unnecessary: the processor dispatches per message, and each message names one surface, hence one catalog.

## Tests

[`processMessages rejects an envelope mixing update types`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/processor_test.dart#L273) pins the rule above.

[`MessageProcessor catalog scope`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/processor_test.dart#L28) pins per-surface catalog isolation, which nothing covered:

```
s1 (cat1) + Alpha: ACCEPTED     s2 (cat2) + Beta:  ACCEPTED
s1 (cat1) + Beta:  REJECTED     s2 (cat2) + Alpha: REJECTED
```

That is [`checks each surface against the catalog it was created with`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/processor_test.dart#L73) and [`rejects a component from another surface's catalog`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/processor_test.dart#L81), plus [`builds one validator per catalog and reuses it`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/processor_test.dart#L86), since the schema cache depends on it.

[`PayloadValidator single-item checks`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/validator_test.dart#L248) replaces the `surfaceCatalogs` group, and [`MessageProcessor catalog resolution`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/validator_test.dart#L961) covers the ladder above. `checkSurfaceComplete` is pinned by [`a child reference that names no component leaves it incomplete`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/validator_test.dart#L391), [`a surface with no root component is incomplete`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/validator_test.dart#L413) and [`a surface with an unreachable component is incomplete`](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/validator_test.dart#L435).

The [conformance harness](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/dart/a2ui_core/test/conformance/validator_conformance_test.dart#L186) derives one catalog per case instead of registering the document under every id the payload names.

## Blueprint

New [Catalog Scope](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/blueprints/modules/a2ui_core.blueprint.md#L839) section under the validation layer, written as instructions for implementers in other languages: scope the validator to one catalog, expose catalog-free envelope parsing, build one validator per catalog in `MessageProcessor` and reuse it, record the catalog on the surface, check each surface against the catalog it was created with. The [`PayloadValidator` signature](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/blueprints/modules/a2ui_core.blueprint.md#L747), the [renderer/agent table](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/blueprints/modules/a2ui_core.blueprint.md#L708) and the [validation matrix](https://github.com/a2ui-project/a2ui/blob/8a48e96c47d08fa4a014fb1ed8be098b536c112b/blueprints/modules/a2ui_core.blueprint.md#L859) are updated, with two new Catalog Scope rows.

## Verification

- `dart analyze` in `dart/a2ui_core` and `dart/a2ui_agent`: no issues
- `dart test` in `dart/a2ui_core`: 338 passed, 25 skipped (pre-existing v0.8 cases)
- `dart format`: clean


blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -482,18 +485,37 @@ export class A2uiValidator {
...
 | :----------------------- | :------------------------------------------------------------------------------------------------ | :------------------------------------------------------------------------ | :-------------------- |
-| **Protocol Envelope**    | Single update type per message (`createSurface`, `updateComponents`, etc.)                        | `A2uiValidator` (Zod envelope schema)                                     | `A2uiValidationError` |
-| **Protocol Envelope**    | Valid `version` tag (`v0.8`, `v0.9`, `v1.0`) & required envelope keys                             | `A2uiValidator` (Zod envelope schema)                                     | `A2uiValidationError` |
+| **Protocol Envelope**    | Single update type per message (`createSurface`, `updateComponents`, etc.)                        | `A2uiValidator.parseMessages()` (envelope schema, no catalog needed)      | `A2uiValidationError` |
+| **Protocol Envelope**    | Valid `version` tag (`v0.8`, `v0.9`, `v1.0`) & required envelope keys                             | `A2uiValidator.parseMessages()` (envelope schema, no catalog needed)      | `A2uiValidationError` |
```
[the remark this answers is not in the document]
polina-c: Fixed by aligning the implementation with the blueprint.

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -464,9 +464,12 @@ export interface ValidationConfig {
...
 }
 
-/** Stateless validator executing envelope structure, component property schema, theme schema, and path syntax checks. */
+/** Stateless validator executing envelope structure, component property schema, theme schema, and path syntax checks. Scoped to a single catalog. */
 export class A2uiValidator {
-  constructor(catalogs: Catalog<any, any>[], validationConfig?: ValidationConfig);
+  constructor(catalog: Catalog<any, any>, validationConfig?: ValidationConfig);
+
+  /** Parses envelopes without a catalog: version tag and single update type. */
+  static parseMessages(payload: object[], targetVersion?: string): AgentToRendererMessage[];
```
nan-yu: FYI, I'm going to update the a2ui_core blueprint with the following changes:

- Validation Entry Point: `MessageProcessor.processMessages` becomes the primary entry point for message validation.
- PayloadValidator: `A2uiValidator` will be renamed to `PayloadValidator`, scoped to validate a single component, a single function, and theme against its corresponding catalog.
This cleanly justifies making the validator single-catalog: multi-catalog dispatch happens at the processor level, while PayloadValidator only ever evaluates a single component against its own catalog.

dart/a2ui_core/lib/src/processing/processor.dart
```diff
@@ -36,49 +37,86 @@ import '../validation/validator.dart';
...
+  /// types unchecked.
+  final Map<String, Object?> commonTypesSchema;
+
+  /// One validator per catalog, built on first use.
+  final Map<String, A2uiValidator<T, FunctionImplementation>> _validators = {};
 
   MessageProcessor({
     required this.catalogs,
-    A2uiValidator<T, FunctionImplementation>? validator,
+    this.protocolVersion = A2uiProtocolVersion.v0_9,
```
nan-yu: Can we remove the default protocol version and make it required?
Otherwise, we'd update the default value when a new protocol version is available.

dart/a2ui_core/lib/src/validation/validator.dart
```diff
@@ -425,60 +436,46 @@ class A2uiValidator<C extends ComponentApi, F extends FunctionApi> {
...
+  /// Throws [A2uiCatalogError] for a surface created against another catalog.
+  void _checkDeclaredCatalogs(List<A2uiMessage> messages) {
+    for (final message in messages) {
+      if (message is! CreateSurfaceMessage) continue;
+      if (message.catalogId == catalog.id) continue;
       throw A2uiCatalogError(
-        "Unknown catalog '$declared'. This validator holds: "
-        '${catalogs.keys.join(', ')}.',
-        catalogId: declared,
+        "Surface '${message.surfaceId}' is created against catalog "
```
nan-yu: This basically disallowed components from mixed catalogs, which is a key feature in v1.0.

We should support that. Therefore, the validator can only validate a single component, but not all components overall.

blueprints/modules/a2ui_core.blueprint.md
polina-c: **`PayloadValidator`** (renamed from `A2uiValidator`) holds one catalog and checks one item against it: `validateComponent`, `validateFunction`, `validateTheme`. No payload-level entry point, and it no longer rejects a payload for naming another catalog — it cannot know which catalog an item belongs to, so that is not its decision to make. `parseMessages` is static, because envelopes read no catalog.

**`MessageProcessor`** holds every supported catalog and does the routing. It resolves the catalog per item, in the order: the item's own `catalogId` (v1.0's override), the surface's default from `createSurface`, then the sole supported catalog. `A2uiCatalogError` when none of those settles it — skipping would report a payload valid that nothing had looked at. Two entry points: `processMessages` applies a payload and checks each message against the surface it joins, and `validatePayload` checks a payload on its own with nothing applied and no surface required.

The blueprint's "Catalog Scope" section and the validation matrix are rewritten around that, with a new **Function Arguments** row for `validateFunction`.

Three things:

1. **`validatePayload` is a second entry point, not a replacement for `processMessages`.** `processMessages` needs the surfaces to exist, and a fair number of payloads are checked before that is true: the whole `conformance/core/validator.yaml` suite is standalone payloads, several of which are incremental with no `createSurface`, and an agent checking its own output has no surface state either. So "the processor is the entry point for validation" is satisfied, but through two methods rather than one.
2. **`validateFunction` has no caller yet in Dart.** Function-call arguments are currently checked through the component schema, which reaches the catalog's functions via `anyFunction` in `common_types.json` — the `a call to a function the catalog does not declare` test in `validator_basic_catalog_test.dart` passes that way. `validateFunction` is implemented and tested as a per-item entry point because you named it, but nothing routes to it. Want it wired in as its own pass, or left as public API until v1.0 needs per-call catalog resolution?

dart/a2ui_core/lib/src/processing/processor.dart
polina-c: Done, and applied to `PayloadValidator` and `PayloadValidator.parseMessages` too, since the same argument holds for each: a default there would silently move every caller onto the next version the day one lands.

dart/a2ui_core/lib/src/validation/validator.dart
polina-c: fixed. The check you flagged is gone, along with the payload-wide walk it belonged to.

What replaces it: catalog resolution is per item and lives on `MessageProcessor`, which is the only thing holding every supported catalog. Each component resolves through its own `catalogId` first, then the surface's default, then the sole supported catalog, and is checked by the validator for whatever that resolves to. `PayloadValidator` never sees the question.

So a payload spanning several catalogs is now checked against each of them rather than rejected, and the v1.0 per-component override works today rather than needing another pass later. New coverage in `validator_test.dart` under `MessageProcessor catalog resolution`:

| case                                                        | result                                 |
| ----------------------------------------------------------- | -------------------------------------- |
| `s1` → `cat1` + Alpha, `s2` → `cat2` + Beta, in one payload | accepted                               |
| `s1` → `cat1`, component `Beta` with `catalogId: cat2`      | accepted (v1.0 per-component override) |
| `s1` → `cat1`, component `Beta` with no override            | rejected                               |
| `s1` → `cat2` on a processor supporting only `cat1`         | `A2uiCatalogError`                     |
| incremental payload, several catalogs supported, none named | `A2uiCatalogError`                     |

The last row is the one case that still refuses rather than resolves: v0.9 declares `catalogId` on `createSurface` alone, so an incremental payload arriving with no surface state and more than one candidate catalog genuinely cannot be attributed. Validating it against an arbitrary catalog would report a payload valid that nothing had checked. Note this only affects `validatePayload` — through `processMessages` the surface exists and carries its catalog, so it never comes up.

Graph checks follow the same rule: `checkComponentTopology` and `checkComponentIntegrity` now take the reference fields merged across every catalog a surface draws on, so a mixed-catalog surface has its whole graph walked rather than the part one catalog happens to describe.

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -459,13 +459,15 @@ const agentProcessor = new MessageProcessor({
   catalogs: [negotiatedCatalog], // Single negotiated catalog enforces catalog compliance
   actionHandler: undefined, // Agent does not render DOM elements or handle clicks
 });
-agentProcessor.processMessages(generatedLlmPayloadMessages);
+// Checks the payload on its own: no surface state, nothing applied.
+agentProcessor.validatePayload(generatedLlmPayload);
```
nan-yu: Why not reusing the `processMessages` method? We can leverage the surface state to do more checks, like parent-child references. State can be updated internally on the agent side.

blueprints/modules/a2ui_core.blueprint.md
nan-yu: Why would we need a static `parseMessages` method? From the implementation, it converts the JSON list into A2uiMessage[]. This should be covered by the Parser on the agent SDK. The `parse_response` method converts the LLM output into `List[ResponsePart]`, which is already in the format of  `A2uiMessage[]`. It also feels weird for a validator to parse messages.

dart/a2ui_core/lib/src/processing/processor.dart
```diff
@@ -33,52 +36,119 @@ import '../validation/validator.dart';
...
+  /// must declare a protocol version this SDK implements and carry exactly one
+  /// update type. Neither check needs a catalog, which is what lets a payload
+  /// be parsed before each message is matched to the surface, and so the
+  /// catalog, it belongs to. Returns the parsed messages.
   ///
-  /// Throws [A2uiValidationError] for a malformed envelope, before any message
-  /// reaches the models.
+  /// Throws [A2uiValidationError] for a malformed envelope, including one
+  /// mixing update types, before any message reaches the models.
   List<A2uiMessage> processPayload(List<Map<String, Object?>> payload) {
```
nan-yu: When would we call processPayload to process raw payload?

dart/a2ui_core/lib/src/processing/processor.dart
```diff
@@ -90,6 +160,164 @@ class MessageProcessor<T extends ComponentApi> {
...
+  /// Checks a payload on its own: envelopes, then structure, then catalogs.
+  ///
+  /// Nothing is applied and no surface has to exist, so this is what an agent
+  /// runs over its own output before sending it, and what checks a payload
+  /// captured on the wire. A renderer holding surface state gets a stricter
+  /// check from [processPayload], which resolves references against the
+  /// components each surface already holds.
+  ///
+  /// Returns the parsed messages, and throws as the individual stages do.
+  List<A2uiMessage> validatePayload(List<Map<String, Object?>> payload) {
```
nan-yu: I'd propose to reuse processMessages for validation.

blueprints/modules/a2ui_core.blueprint.md
polina-c: Agreed on the direction, and `validatePayload` is gone: the agent keeps a processor for the session and checks its own output through `processMessages`, exactly as a renderer checks what it receives.

Two things came out of doing it. `processMessages` never required a root or rejected an unreachable component (`requireRoot: false, allowOrphans: true`), correctly, since either may be settled by a later message — so completeness is now an explicit `checkSurfaceComplete(surfaceId)`, which an agent runs over the surfaces its turn built.

The second is a real bug this surfaced: `processMessages` checks each batch against what the surface already holds plus the batch itself, so a forward reference inside one payload reads as dangling. That rejects our own published `00_incremental.json`, where `root` points at `restaurant_card` one message before it is declared. `validatePayload` never hit it because it merged the payload per surface first. I am moving reference resolution out of the per-batch check and into `checkSurfaceComplete`; duplicate ids and cycles stay per batch.

dart/a2ui_core/lib/src/processing/processor.dart
polina-c: dropped

thanks

dart/a2ui_core/lib/src/processing/processor.dart
polina-c: Done — validatePayload is gone and processMessages is the only entry point, for agents as well as renderers.

blueprints/modules/a2ui_core.blueprint.md
polina-c: Agreed, and it has moved to `A2uiMessage.parseAll(payload, protocolVersion:)`. `A2uiMessage.fromJson` already existed and already parsed the version, so turning a list of envelopes into messages is the message model's job, and the validator no longer parses anything.

It cannot live only in the agent SDK's `Parser`, though: a renderer receives raw payloads over the wire with no agent SDK present, and still has to turn them into messages before `processMessages` can route them. Keeping it on the message model serves both sides.

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -459,13 +459,20 @@ const agentProcessor = new MessageProcessor({
   catalogs: [negotiatedCatalog], // Single negotiated catalog enforces catalog compliance
   actionHandler: undefined, // Agent does not render DOM elements or handle clicks
 });
-agentProcessor.processMessages(generatedLlmPayloadMessages);
+// Same entry point as the renderer: the processor is kept for the session, so
+// each payload is checked against the state the previous ones built.
+agentProcessor.processMessages(
+  AgentToRendererMessage.parseAll(generatedLlmPayload, protocolVersion),
```
nan-yu, line 465: The generated LlmPayload may not be the AgentToRendererMessages. It might be in other inference format, e.g. Express DSL.

What about something like `agentProcessor.processMessages(parsedLlmPayloadMessages)` to hide the parsing details?

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -459,13 +459,20 @@ const agentProcessor = new MessageProcessor({
...
+);
+// Once the turn is built, assert each surface it created is a finished render.
+agentProcessor.checkSurfaceComplete('surface-1');
```
nan-yu, line 468: Is `checkSurfaceComplete` a new method? It is no in the API design doc.

The `processMessages` should cover all validations. What additional functionalities does the `checkSurfaceComplete` method provide?

dart/a2ui_core/lib/src/processing/processor.dart
```diff
@@ -33,64 +35,235 @@ import '../validation/validator.dart';
...
   ///
-  /// Defaults to a validator over [catalogs], resolving the shared types
-  /// against the `common_types.json` this package publishes. Supply one to
-  /// configure it — to override that document, or to accept a different
-  /// protocol version.
-  final A2uiValidator<T, FunctionImplementation> validator;
+  /// Defaults to the copy this package publishes for [protocolVersion]; pass
+  /// a different document to override it, or an empty map to leave the shared
+  /// types unchecked.
+  final Map<String, Object?> commonTypesSchema;
```
nan-yu, line 80: Why do we need to get commonTypesSchema? Isn't it already inlined in the Catalogs?

dart/a2ui_core/lib/src/processing/processor.dart
```diff
@@ -33,64 +35,235 @@ import '../validation/validator.dart';
...
+  /// — an agent runs it over the surfaces its turn built before sending the
+  /// payload, and a renderer can run it once a stream ends.
+  ///
+  /// Every reference must resolve within the surface, a component with id
+  /// `root` must exist, and every component must be reachable from it.
+  ///
+  /// Throws [A2uiStateError] if no such surface exists, [A2uiIntegrityError]
+  /// for a missing root or a reference to no component, and
+  /// [A2uiRecursionError] for a cycle or an over-deep chain.
+  void checkSurfaceComplete(String surfaceId) {
```
nan-yu, line 178: This can be consolidated into processMessages.

dart/a2ui_core/lib/src/processing/processor.dart
nan-yu: Never mind. This is a move to make the catalog protocol version-agnostic. Having an extra commonTypesSchema allows catalogs to be compatible with multiple protocol versions.

blueprints/modules/a2ui_core.blueprint.md
polina-c: done

blueprints/modules/a2ui_core.blueprint.md
polina-c: Yes, thank you. Corrected.

nan-yu (approved):

PR #2540, opened 2026-09-05
[blueprints] Link glossary terms.

polina-c:
Link glossary.md in blueprints, when the term is used first time in a doc.

### DIFF
--- a/blueprints/codebases/agent_sdks/python/a2ui_agent/codebase.blueprint.md
+++ b/blueprints/codebases/agent_sdks/python/a2ui_agent/codebase.blueprint.md
@@ -13,17 +13,17 @@ local_development:
 
 ## **Architecture & Ecosystem Map**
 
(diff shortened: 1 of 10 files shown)


jacobsimionato (commented):
These docs are primarily for agent consumption. The glossary is a fantastic agent resource, because it's compact enough for the agent to read in full in many/most conversations, but should provide a lot of value in ensuring the agent understands the other materials properly. I suspect the section links e.g. `#compilation` aren't necessarily useful to agent, because it is able to read line-by-line rather than by section, as far as I know. Also, all of these blueprints include links to multiple glossary entries.

So, I was wondering about just including one link to the glossary near the top of each blueprint with an explanation that it should be read in conjunction to properly understand the blueprint, or even link to the glossary in AGENTS.md which is read for every conversation.

That way, we don't need to try to keep these links up to date, which is more complex given that new glossary terms get added, and sections of documentation move around.

PR #2546, opened 2026-09-07
refactor(deps): standardize @a2ui/markdown-it workspace dependencies

jacobsimionato:
### Summary & Rationale

Standardizes `@a2ui/markdown-it` dependency declarations across all internal workspace packages to use `workspace:*`. Previously, some packages used `*` while others used `workspace:*`, causing syncpack mismatch warnings when syncpack linting was run.

### Key Changes

- **Syncpack Configuration**: Added a `.syncpackrc` version group rule pinning `@a2ui/markdown-it` to `workspace:*`.
- **Package Manifest Updates**: Updated 12 `package.json` files across `renderers/` and `samples/` to use `"@a2ui/markdown-it": "workspace:*"`.
- **Lockfile**: Updated `yarn.lock`.

### Verification & Testing

- Verified dependency version compliance with `yarn lint:deps` (syncpack) — 0 errors.
- Verified all workspace builds and tests pass.

### DIFF
--- a/renderers/angular/package.json
+++ b/renderers/angular/package.json
@@ -41,7 +41,7 @@
     "zod": "^3.25.76"
   },
   "peerDependencies": {
-    "@a2ui/markdown-it": "*",
+    "@a2ui/markdown-it": "workspace:*",
     "@angular/common": "^21.2.5",
     "@angular/core": "^21.2.5",
     "@angular/platform-browser": "^21.2.5"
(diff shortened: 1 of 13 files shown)


renderers/angular/package.json
```diff
@@ -41,7 +41,7 @@
     "zod": "^3.25.76"
   },
   "peerDependencies": {
-    "@a2ui/markdown-it": "*",
+    "@a2ui/markdown-it": "workspace:*",
```
ditman, line 44: This looks good to me, does this work well with our publish scripts? (I think it does but I haven't touched them in a while)

ditman (approved):
LGTM! Does this work well with our publish script?

PR #2554, opened 2026-09-08
feat(spec): add legal document diff viewer example with code block formatting

agarwaldipali-hash:
## Description
This PR adds a new specification example (`37_legal-document-diff.json`) to both the v0.9 and v1.0 basic catalogs.

> [!NOTE]
> Design discussion open at #2620 — I'd like maintainer input on whether Markdown-in-`Text` is the right vehicle for this before anyone spends review time here. See [Open question](#open-question) below.

### Key Highlights
- **Enterprise Agent Scenario:** Showcases automated legal agreement and product licensing comparison between two releases (v1.2 March 2023 vs v2.0 September 2024), highlighting changes in deployment limits, AI telemetry clauses, and liability caps.
- **Markdown Diff Blocks:** Renders fenced Markdown diff blocks (```` ```diff ... ``` ````) inside the standard `Text` component to present the redline.
- **Interactive Acknowledgement:** The action `Row` emits an `acknowledge_diff` event carrying `sourceVersion` / `targetVersion`, so the accept action is meaningful rather than decorative.
- **Specification Compliance:** Validated with 0 errors against JSON Schema Draft 2020-12 and `catalog.json` rules.

### Open question

To present the redline I leaned on Markdown inside `Text`. The basic catalog documents `Text.text` as supporting *simple* Markdown and notes that "utilizing dedicated UI components is generally preferred", so it's worth confirming this usage is in scope:

1. Are multi-line fenced code blocks inside `Text` in-scope for "simple Markdown"?
2. Is fenced-block rendering (and language hints like `diff`) expected to be consistent across renderers, or renderer-dependent in practice?
3. If Markdown-in-`Text` is the wrong vehicle, would a diff/comparison presentation be wanted as a dedicated component — or is this out of scope?

Happy to rework, scope down, or close depending on the answer.

### Note on word-level markers

The summary lines use `++...++` / `--...--` inside inline code spans to annotate which words changed. These are **presentational literals**, not a Markdown or A2UI diff feature — they render as literal text. Flagging explicitly to avoid implying renderer support that doesn't exist.

### DIFF
--- a/specification/v0_9/catalogs/basic/examples/37_legal-document-diff.json
+++ b/specification/v0_9/catalogs/basic/examples/37_legal-document-diff.json
@@ -1,0 +1,174 @@
+{
+  "name": "Legal Document Diff Viewer",
+  "description": "Compares two versions of an enterprise licensing agreement and highlights clause modifications and word-level diffs using Markdown diff blocks.",
+  "messages": [
+    {
+      "version": "v0.9",
+      "createSurface": {
+        "surfaceId": "gallery-legal-document-diff",
+        "catalogId": "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json",
+        "sendDataModel": true
+      }
+    },
+    {
+      "version": "v0.9",
+      "updateComponents": {
+        "surfaceId": "gallery-legal-document-diff",
+        "components": [
+          {
+            "id": "root",
+            "component": "Card",
+            "child": "main_column"
+          },
+          {
+            "id": "main_column",
+            "component": "Column",
+            "children": [
+              "title_text",
+              "subtitle_text",
+              "divider_top",
+              "summary_card",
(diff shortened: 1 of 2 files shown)


josemontespg:
Thanks for the example and for raising the design question in #2620. @andrewkolos, you requested review on this on Sep 8; can you answer the open question (fenced diff blocks inside Text, which the basic catalog documents as supporting only simple Markdown) so the author knows whether to continue, or hand it back to triage? Two notes for whoever picks it up: the canonical examples are kept identical across specification/v0_9, v0_9_1 and v1_0 and this PR only adds the v0_9 and v1_0 copies, and the spec validation workflow did not run on this PR because its catalogs/** trigger landed on Sep 22 (the messages do validate locally with the validate.py configuration).

PR #2555, opened 2026-09-08
Update shared mcp catalog to support multiple mcp servers, databinding, mcp callback handler

wrenj:
This pull request enhances the shared MCP catalog to support more complex and dynamic tool execution scenarios. Key improvements include: 

1. Host client supplies a callMcpTool function that can dial an MCP server by tool name. Client host decides which MCP servers can be called.
2. Dynamic Data Binding: The callMcpTool function now supports dynamic path-based resolution for tool names, server names, and arguments, allowing for more flexible UI interactions. 
3. Result Handling: Added an optional McpToolResultHandler callback, providing a hook to process tool execution results. 
4. Schema and API Updates: The mcp_catalog.json schema and associated TypeScript API definitions were updated to incorporate these new features. 
5. Sample Integration: The a2ui-over-mcp-recipe sample has been updated to utilize these new capabilities, including corresponding changes to the client engine and server-side catalog configuration.


sugoi-yuzuru (commented):
Thanks for working on this! Moving MCP tool calls to first-class catalog functions is a great direction.

However, now that I saw how this architecture is making the MCP server's intent tightly coupled to the Host UI's McpClientGetter implementation, I am not too keen on the `targetServer` concept without it being a canonical URI so that the MCP server can continue to be host-UI agnostic. 

I thought may be this can be resolved by some hand-shake protocol to exchange "allowed server" information, I also now have concerns regarding the multi-server routing design.

### 1. Security boundary and cross-server execution risks

Allowing client UI payloads to direct tool calls to arbitrary servers introduces a dangerous attack vector (confused deputy):

* **Local privilege escalation:** In real-world environments, users connect to a mix of remote/untrusted MCP servers (e.g., third-party web services) and sensitive local MCP servers (e.g., `filesystem`, `sqlite`, `bash` running over stdio). Local servers generally have no internal authorization because they assume the host process is the sole caller.
* **Sandbox escape:** If a remote server (`evil.com`) returns an A2UI payload that specifies `server: "filesystem"` to invoke `read_file`, it can attempt to exfiltrate local data or execute code.
* **Why handshake gating is insufficient:** Relying on host UIs to gate cross-server calls places the entire security burden on host implementations, which can suffer from permission fatigue, origin confusion, or buggy allow-lists. 


### 2. Architectural coupling: host-dependent servers

Even setting security aside, this design breaks the host-agnostic nature of MCP:

In `recipe_form.json`, the button action specifies:
```json
"action": {
  "functionCall": {
    "call": "callMcpTool",
    "args": {
      "server": "a2ui-mcp-recipe-demo",
      "name": "get_recipe_a2ui",
      ...
    }
  }
}
```

This pattern was used in the earlier recipe demo via custom `event` actions, but this PR formalizes it into the shared catalog schema (`mcp_catalog.json`) and SDK API (`createMcpCatalog`).

This creates tight coupling:
* The MCP server payload must anticipate what nickname the host assigned to it. If the host config registers the connection as `"recipes"` or an auto-generated ID, `"a2ui-mcp-recipe-demo"` fails.
* The fallback in the sample engine—iterating `this.mcpClients.values()` and taking the first client—is non-deterministic when multiple servers are connected.
* MCP servers should be host-agnostic and have no awareness of host registration keys.

### Suggested solution

#### A. Enforce strict origin isolation (`self` only)
Drop the `server` argument from `mcp_catalog.json` entirely. 

When an A2UI surface is created from an MCP tool result or template resource, the runtime engine binds that surface to its originating server connection. When `callMcpTool` is invoked, it routes strictly to that same connection:
```json
"action": {
  "functionCall": {
    "call": "callMcpTool",
    "args": {
      "name": "get_recipe_a2ui",
      "arguments": { ... }
    }
  }
}
```
This eliminates the host coupling completely and neutralizes the cross-server security threat.

#### B. Handle multi-server interactions via host/agent events
If a UI surface legitimately needs to trigger actions involving other servers (for example, taking recipe ingredients and adding them to an external cart service):
* The UI should not execute direct `functionCall`s to foreign servers.
* Instead, it should emit an A2UI **`event`** back to the Host/Agent:
  ```json
  "action": {
    "event": {
      "name": "add_to_cart",
      "context": { "items": { "path": "/ingredients" } }
    }
  }
  ```
* The Host Agent then orchestrates the cross-server workflow with proper policy checks, user awareness, and reasoning oversight.

sugoi-yuzuru (approved):

PR #2566, opened 2026-09-08
refactor(web_core): support configurable RPC version in RpcHandler

nan-yu:
## Summary
Updates TypeScript `RpcHandler` to support configurable RPC protocol versions instead of hardcoding `v1.0`.

## Changes
- Updated `CallOptions`, `NormalizedAgentCall`, `callAgentFunction`, `dispatchAgentCall`, `sendOutboundMessage`, `executeFunctionSafely`, and `createResponseError` to pass and handle protocol version dynamically.
- Made `version` parameter required in `createResponseError` and `executeFunctionSafely`.
- Accessed `message.version` directly in `validateInboundMessage` and `handleCallRendererFunction` without defaulting to `v1.0`.
- Updated unit tests in `rpc.test.ts` to pass explicit version in `callAgentFunction` test options.

## Testing
- Ran TypeScript Web Core test suite (`yarn --cwd typescript/web_core test`): 611/611 passed.
- Verified TypeScript build and lint (`yarn --cwd typescript/web_core build && yarn --cwd typescript/web_core lint`): 0 errors.

TAG=agy

### DIFF
--- a/typescript/web_core/src/rpc/rpc-handler.ts
+++ b/typescript/web_core/src/rpc/rpc-handler.ts
@@ -82,6 +82,8 @@ export interface CallOptions {
   functionCallId?: string;
   /** Timeout in milliseconds before rejecting with TIMEOUT. */
   timeoutMs?: number;
+  /** Protocol version for the outbound message. */
+  version?: string;
 }
 
 /**
@@ -99,6 +101,7 @@ interface NormalizedAgentCall {
   functionCallId: string;
   call: FunctionCall;
   effectiveTimeoutMs: number;
+  version?: string;
 }
 
 /**
@@ -154,15 +157,17 @@ export class RpcHandler {
       return validationError;
     }
 
-    const {functionCallId, callFunction} = message.callRendererFunction;
+    const {version, callRendererFunction} = message;
+    const {functionCallId, callFunction} = callRendererFunction;
     const {call, catalogId, args} = callFunction;
 
-    const resolved = this.resolveFunctionImplementation(catalogId, call, context, message.version);
+    const resolved = this.resolveFunctionImplementation(catalogId, call, context, version);
     if ('error' in resolved) {
       const errorResponse = this.createResponseError(
         functionCallId,
         RpcErrorCode.INVALID_FUNCTION_CALL,
         resolved.error,
+        version,
       );
       await this.emitOutboundResponse(errorResponse);
       return errorResponse;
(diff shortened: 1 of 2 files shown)


typescript/web_core/src/rpc/rpc-handler.ts
```diff
@@ -312,20 +321,24 @@ export class RpcHandler {
         'unknown',
         RpcErrorCode.INVALID_FUNCTION_CALL,
         'Inbound message is null or undefined.',
+        'unknown',
       );
     }
+    const version = message.version;
```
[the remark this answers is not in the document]
nan-yu: message.version should always be defined. If not, it will fail at the validation step before.

typescript/web_core/src/rpc/rpc-handler.ts
```diff
@@ -536,9 +552,10 @@ export class RpcHandler {
     functionCallId: string,
     code: RpcErrorCode | string,
     message: string,
+    version: string,
   ): RendererFunctionResponseMessage {
     return {
-      version: 'v1.0',
+      version: version as any,
```
gspencergoog: Casting to `any` seems suspect.  Why is that needed?

typescript/web_core/src/rpc/rpc-handler.ts
```diff
@@ -502,17 +517,18 @@ export class RpcHandler {
     surfaceId: string,
     functionCallId: string,
     call: FunctionCall,
+    version: string | undefined,
     cleanup: () => void,
     reject: (reason?: any) => void,
   ): void {
     const outboundMsg: CallAgentFunctionMessage = {
-      version: 'v1.0',
+      ...(version !== undefined ? {version: version as any} : {}),
```
gspencergoog: Why do you need to cast to `any`?

typescript/web_core/src/rpc/rpc-handler.ts
nan-yu: It was casted to any probably because there was no default value.

I removed `as any` and updated to use `(version || 'v1.0') as RendererFunctionResponseMessage['version']` to preserve strict type checking for the surrounding message object.

typescript/web_core/src/rpc/rpc-handler.ts
nan-yu: Removed as any. Replaced with `(version || 'v1.0') as CallAgentFunctionMessage['version']`.

gspencergoog (approved):

PR #2544, opened 2026-09-07
fix(react): allow React 18 & 19 peerDependencies in @a2ui/react

jacobsimionato:
### Summary & Rationale

Restores `@a2ui/react`'s `peerDependencies` range for `react` and `react-dom` to support React 18 (`^18.0.0 || ^19.0.0`) alongside React 19. Previously, syncpack collapsed `@a2ui/react`'s `peerDependencies` down to `^19.2.7`, breaking installation for downstream React 18 applications.

Additionally, added dual React 18 & 19 unit test matrix execution to `@a2ui/react` per code review feedback.

### Key Changes

- **Restored Peer Dependencies**: Restored `peerDependencies` range in `renderers/react/package.json` to `"^18.0.0 || ^19.0.0"`.
- **Syncpack Configuration**: Added an `isIgnored: true` version group rule in `.syncpackrc` for `@a2ui/react`'s `react` and `react-dom` peer dependencies.
- **Markdown-IT Syncpack Exemption**: Added an `isIgnored: true` rule in `.syncpackrc` for `@a2ui/markdown-it` so this PR remains laser-focused on React 18 dependencies.
- **React 18 & 19 Unit Test Matrix**: Added npm package alias devDependencies (`react-18`, `react-dom-18`, `@testing-library/react-18`), configured conditional module aliasing in `vitest.config.ts`, and added a `test:unit:react18` target so the full 475-test unit suite runs under both React 19 and React 18.

### Verification & Testing

- Verified dependency version compliance with `yarn lint:deps` (syncpack) — 0 errors.
- Built `@a2ui/react` with `yarn workspace @a2ui/react build`.
- Executed and passed all 475 unit tests under React 19 (`yarn workspace @a2ui/react test:unit`).
- Executed and passed all 475 unit tests under React 18 (`yarn workspace @a2ui/react test:unit:react18`).
- Verified all GitHub Actions CI checks pass on PR #2544.

Closes #2543

PR #2568, opened 2026-09-08
refactor(react): replace synthetic module aliasing with dynamic React 18 matrix script

jacobsimionato:
### Summary & Rationale

Follow-up to #2544 addressing code review feedback from @ditman:
1. **Reverted Browser Sandbox Configuration**: Reverted `ChromeHeadlessNoSandbox` in `renderers/react/a2ui_explorer/karma.conf.cjs` back to standard `browsers: [ChromeHeadless]`.
2. **Removed Synthetic Module Aliasing**:
   - Removed synthetic npm alias devDependencies (`react-18`, `react-dom-18`, `@types/react-18`, `@types/react-dom-18`, `@testing-library/react-18`).
   - Removed `createRequire`, `require.resolve`, `ssr.alias`, and regex import replacement from `vitest.config.ts`.
   - Removed `Module._resolveFilename` CommonJS hook from `tests/setup.ts`.
   - Removed `react18Aliases` from `karma.conf.cjs`.
3. **React 18 Compatibility Matrix**:
   - Added `renderers/react/scripts/test-react18.sh` which temporarily switches `@a2ui/react` and `@a2ui/react-explorer` to React 18 (`react@^18.3.1`, `react-dom@^18.3.1`, `@testing-library/react@^14.3.1`), executes the unit and integration test suites, and restores original React 19 dependencies upon completion.
   - Includes an upfront git preflight check to prevent running with uncommitted dependency changes, resets traps on cleanup, and restores original files atomically via `git checkout`.
   - Swaps dependencies using standard `yarn add` commands.
   - Added a `Test React 18 Compatibility` step in `.github/workflows/web_ci.yml` under the `ci-renderers` job.
   - Added script target `"test:react18": "./scripts/test-react18.sh"` in `renderers/react/package.json`.
4. **Tooling & Formatting**:
   - Added `renderers/react/a2ui_explorer/src/generated/**` to `.prettierignore`, matching Angular and Lit explorer generated paths.

### Verification

- Ran standard test suite: `yarn workspace @a2ui/react test` (475 unit tests, 132 integration tests under React 19).
- Ran React 18 matrix test: `yarn workspace @a2ui/react test:react18` (475 unit tests, 132 integration tests under React 18).
- Ran dependency lint: `yarn lint:deps` (0 errors).
- Ran lint and format checks across `@a2ui/react` and `@a2ui/react-explorer`.

Addresses feedback on #2544

### DIFF
--- a/.github/workflows/web_ci.yml
+++ b/.github/workflows/web_ci.yml
@@ -104,7 +104,7 @@ jobs:
       - name: Test Renderers
         run: eval "yarn workspaces foreach -A $RENDERERS_INC -p run test"
       - name: Test React 18 Compatibility
-        run: ./renderers/react/scripts/test-react18.sh
+        run: yarn workspace @a2ui/react run test:react18
       - name: Verify NPM Publish Script
         run: node --test renderers/scripts/publish_npm.test.mjs
 
--- a/.github/workflows/web_ci.yml
+++ b/.github/workflows/web_ci.yml
@@ -103,6 +103,8 @@ jobs:
         run: eval "yarn workspaces foreach -A $RENDERERS_INC -p --topological-dev run build"
       - name: Test Renderers
         run: eval "yarn workspaces foreach -A $RENDERERS_INC -p run test"
+      - name: Test React 18 Compatibility
+        run: ./renderers/react/scripts/test-react18.sh
       - name: Verify NPM Publish Script
         run: node --test renderers/scripts/publish_npm.test.mjs
 
--- a/renderers/react/scripts/test-react18.sh
+++ b/renderers/react/scripts/test-react18.sh
@@ -19,6 +19,9 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REACT_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
 REPO_ROOT="$(cd "${REACT_DIR}/../.." && pwd)"
 
+# Allow Yarn to update the lockfile in CI during temporary dependency swapping
+export YARN_ENABLE_IMMUTABLE_INSTALLS=false
+
 REACT_PKG="${REACT_DIR}/package.json"
 EXPLORER_PKG="${REACT_DIR}/a2ui_explorer/package.json"
 ROOT_LOCK="${REPO_ROOT}/yarn.lock"
--- a/renderers/react/scripts/test-react18.sh
+++ b/renderers/react/scripts/test-react18.sh
@@ -19,9 +19,6 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REACT_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
 REPO_ROOT="$(cd "${REACT_DIR}/../.." && pwd)"
 
-# Allow Yarn to update the lockfile in CI during temporary dependency swapping
-export YARN_ENABLE_IMMUTABLE_INSTALLS=false
-
 REACT_PKG="${REACT_DIR}/package.json"
 EXPLORER_PKG="${REACT_DIR}/a2ui_explorer/package.json"
 ROOT_LOCK="${REPO_ROOT}/yarn.lock"
@@ -53,32 +50,21 @@ trap cleanup EXIT INT TERM
 echo "=== Building @a2ui/react ==="
 yarn --cwd "${REACT_DIR}" build
 
-# Temporarily swap dependencies to React 18 in a single pass
+# Temporarily swap devDependencies to React 18
 echo "=== Swapping to React 18 dependencies ==="
-node -e '
-const fs = require("fs");
-[
-  [process.argv[1], "devDependencies", {
-    "react": "^18.3.1",
-    "react-dom": "^18.3.1",
-    "@types/react": "^18.3.1",
-    "@types/react-dom": "^18.3.1",
-    "@testing-library/react": "^14.3.1"
-  }],
-  [process.argv[2], "dependencies", {
-    "react": "^18.3.1",
-    "react-dom": "^18.3.1",
-    "@types/react": "^18.3.1",
-    "@types/react-dom": "^18.3.1"
-  }]
-].forEach(([file, key, deps]) => {
-  const pkg = JSON.parse(fs.readFileSync(file, "utf8"));
-  pkg[key] = { ...pkg[key], ...deps };
-  fs.writeFileSync(file, JSON.stringify(pkg, null, 2) + "\n");
-});
-' "${REACT_PKG}" "${EXPLORER_PKG}"
+yarn --cwd "${REACT_DIR}" add -D \
+  "react@^18.3.1" \
+  "react-dom@^18.3.1" \
+  "@types/react@^18.3.1" \
+  "@types/react-dom@^18.3.1" \
+  "@testing-library/react@^14.3.1"
 
-yarn --cwd "${REPO_ROOT}" install
+yarn --cwd "${REACT_DIR}/a2ui_explorer" add \
+  "react@^18.3.1" \
+  "react-dom@^18.3.1" \
+  "@types/react@^18.3.1" \
+  "@types/react-dom@^18.3.1"
+
 DEPS_SWAPPED=true
 
 # Run unit tests under real React 18
--- a/renderers/react/scripts/test-react18.sh
+++ b/renderers/react/scripts/test-react18.sh
@@ -19,25 +19,30 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REACT_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
 REPO_ROOT="$(cd "${REACT_DIR}/../.." && pwd)"
 
+REACT_PKG="${REACT_DIR}/package.json"
+EXPLORER_PKG="${REACT_DIR}/a2ui_explorer/package.json"
+ROOT_LOCK="${REPO_ROOT}/yarn.lock"
+
+TRACKED_FILES=("${REACT_PKG}" "${EXPLORER_PKG}" "${ROOT_LOCK}")
+
 echo "=== Testing @a2ui/react with React 18 ==="
 
-# Save backups so we restore the exact starting state
-cp "${REACT_DIR}/package.json" "${REACT_DIR}/package.json.bak"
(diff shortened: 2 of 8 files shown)


renderers/react/scripts/test-react18.sh
```diff
@@ -0,0 +1,71 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REACT_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+REPO_ROOT="$(cd "${REACT_DIR}/../.." && pwd)"
+
+echo "=== Testing @a2ui/react with React 18 ==="
+
+# Save backups so we restore the exact starting state
+cp "${REACT_DIR}/package.json" "${REACT_DIR}/package.json.bak"
+cp "${REACT_DIR}/a2ui_explorer/package.json" "${REACT_DIR}/a2ui_explorer/package.json.bak"
+cp "${REPO_ROOT}/yarn.lock" "${REPO_ROOT}/yarn.lock.bak"
```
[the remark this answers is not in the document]
jacobsimionato: Addressed in c8a37b1c. Instead of managing `.bak` files on disk, we now rely on Git for atomic restoration (`git checkout -- <files>`). We also added an upfront safety preflight check (`git diff --quiet HEAD -- <files>`) that verifies the working tree is clean before running, ensuring no local developer edits can be accidentally lost. This completely eliminates backup file management, leftover artifacts, and potential corruption on interrupted runs.

renderers/react/scripts/test-react18.sh
```diff
@@ -0,0 +1,71 @@
...
+    mv -f "${REACT_DIR}/package.json.bak" "${REACT_DIR}/package.json"
+  fi
+  if [[ -f "${REACT_DIR}/a2ui_explorer/package.json.bak" ]]; then
+    mv -f "${REACT_DIR}/a2ui_explorer/package.json.bak" "${REACT_DIR}/a2ui_explorer/package.json"
+  fi
+  if [[ -f "${REPO_ROOT}/yarn.lock.bak" ]]; then
+    mv -f "${REPO_ROOT}/yarn.lock.bak" "${REPO_ROOT}/yarn.lock"
+  fi
+  yarn --cwd "${REPO_ROOT}" install
+}
```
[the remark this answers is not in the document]
jacobsimionato: Addressed in c8a37b1c. Added `trap - EXIT INT TERM` at the start of `cleanup()` to prevent double execution, initialized `DEPS_SWAPPED=false`, and guarded the `yarn install` call so it only runs if dependencies were actually modified.

renderers/react/scripts/test-react18.sh
```diff
@@ -0,0 +1,71 @@
...
+  "react-dom@^18.3.1" \
+  "@types/react@^18.3.1" \
+  "@types/react-dom@^18.3.1" \
+  "@testing-library/react@^14.3.1"
+
+yarn --cwd "${REACT_DIR}/a2ui_explorer" add \
+  "react@^18.3.1" \
+  "react-dom@^18.3.1" \
+  "@types/react@^18.3.1" \
+  "@types/react-dom@^18.3.1"
```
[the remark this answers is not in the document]
jacobsimionato: Addressed in c8a37b1c. Replaced sequential `yarn add` calls with a single-pass inline Node script that patches both `package.json` files directly, followed by a single root `yarn install`. This eliminates a full Yarn resolution/link cycle and significantly speeds up execution.

.github/workflows/web_ci.yml
```diff
@@ -103,6 +103,8 @@ jobs:
         run: eval "yarn workspaces foreach -A $RENDERERS_INC -p --topological-dev run build"
       - name: Test Renderers
         run: eval "yarn workspaces foreach -A $RENDERERS_INC -p run test"
+      - name: Test React 18 Compatibility
+        run: ./renderers/react/scripts/test-react18.sh
```
ditman: You've given this script a yarn script name, so this may be: `yarn workspace @a2ui/react run test:react18`, or similar? I forget the exact syntax, but there's no need to call the sh directly here since it's a yarn script within the workspace right?

renderers/react/scripts/test-react18.sh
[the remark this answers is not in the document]
jacobsimionato: Update in a1648be3: Reverted back to standard `yarn add` commands. In practice, `yarn add` handles dependency resolution natively and cleanly within seconds, avoiding custom inline JSON manipulation and unnecessary environment overrides (`YARN_ENABLE_IMMUTABLE_INSTALLS`), keeping the shell script simple and standard.

.github/workflows/web_ci.yml
jacobsimionato: Good point! Updated in 9b06768c to `yarn workspace @a2ui/react run test:react18`.

renderers/react/scripts/test-react18.sh
```diff
@@ -0,0 +1,78 @@
...
+  git -C "${REPO_ROOT}" checkout -- "${TRACKED_FILES[@]}"
+  if [[ "${DEPS_SWAPPED}" == "true" ]]; then
+    yarn --cwd "${REPO_ROOT}" install
+  fi
+}
+trap cleanup EXIT INT TERM
+
+# Ensure dependencies are built first
+echo "=== Building @a2ui/react ==="
+yarn --cwd "${REACT_DIR}" build
```
ditman: I find this odd, don't we have to replace the dependencies *before* running "build" (at least for ${REACT_DIR})?

renderers/react/scripts/test-react18.sh
```diff
@@ -0,0 +1,78 @@
...
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REACT_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+REPO_ROOT="$(cd "${REACT_DIR}/../.." && pwd)"
+
+REACT_PKG="${REACT_DIR}/package.json"
+EXPLORER_PKG="${REACT_DIR}/a2ui_explorer/package.json"
+ROOT_LOCK="${REPO_ROOT}/yarn.lock"
+
+TRACKED_FILES=("${REACT_PKG}" "${EXPLORER_PKG}" "${ROOT_LOCK}")
```
ditman: Do we still need to track the package.json of the packages?

renderers/react/scripts/test-react18.sh
```diff
@@ -0,0 +1,78 @@
...
+  echo "Please commit or stash your changes before running the React 18 test matrix."
+  exit 1
+fi
+
+DEPS_SWAPPED=false
+
+cleanup() {
+  trap - EXIT INT TERM
+  echo "=== Restoring original React 19 dependencies ==="
+  git -C "${REPO_ROOT}" checkout -- "${TRACKED_FILES[@]}"
```
ditman: I wonder if there's any need to keep track of files if it's only for this cleanup, we can just `git reset --hard` or similar at the end of the script, maybe?

renderers/react/scripts/test-react18.sh
```diff
@@ -0,0 +1,78 @@
...
+  "react@^18.3.1" \
+  "react-dom@^18.3.1" \
+  "@types/react@^18.3.1" \
+  "@types/react-dom@^18.3.1"
+
+DEPS_SWAPPED=true
+
+# Run unit tests under real React 18
+echo "=== Running unit tests under React 18 ==="
+yarn --cwd "${REACT_DIR}" test:unit
```
ditman: I think this should just run `test`, to make it similar to what the CI does. That way the contents of the `test` script drive this one, instead of having to keep this updated if we change the test types later?

ditman (approved):
This is much simpler and cleaner than before, thanks for addressing the comments! I left some extra questions but at this stage it feels like nitpicking. Tackle what you want and ship it!

renderers/react/scripts/test-react18.sh
jacobsimionato: Good catch! Updated in 82a0ae8a to swap dependencies to React 18 before running the test suite. Because Wireit configures `test` to depend on `build`, running `yarn test` automatically builds `@a2ui/react` against React 18 first, ensuring it typechecks and bundles against React 18 prior to test execution.

renderers/react/scripts/test-react18.sh
jacobsimionato: Implemented in 82a0ae8a. We now simply invoke `TZ=UTC yarn test`, which lets the workspace's Wireit test graph drive `build`, `test:unit`, and `test:integration` under React 18 automatically without having to duplicate test targets in the script.

renderers/react/scripts/test-react18.sh
jacobsimionato: Running `yarn add` writes the React 18 versions directly into `renderers/react/package.json` (devDependencies) and `renderers/react/a2ui_explorer/package.json` (dependencies), as well as `yarn.lock`, so all three need to be restored to return the workspace to React 19.

Targeting these specific files (renamed to `SWAPPED_FILES` in 82a0ae8a for clarity) also ensures the preflight check and `git checkout` only inspect and restore the dependency manifests, meaning developers can safely run the test locally without clobbering any unstaged code edits in other files.

renderers/react/scripts/test-react18.sh
jacobsimionato: While `git reset --hard` works well in clean CI containers, it is dangerous for local development. If a developer has uncommitted work in progress on a component (e.g. `Button.tsx`) and runs `yarn test:react18`, `git reset --hard` in the cleanup trap would wipe out their uncommitted work.

By scoping the preflight check and `git checkout` strictly to `SWAPPED_FILES` (the 3 dependency manifests), developers can safely test locally without risking data loss in unrelated files.

renderers/react/scripts/test-react18.sh
ditman: > "safely run the test locally without clobbering any unstaged code edits in other files."

Ah ok, fair point, I keep forgetting that this is runnable by the developers as well, not just CI.

renderers/react/scripts/test-react18.sh
ditman: I wonder if a better solution would be to keep a copy of the tracked files (at least the package.json) and just copy back? git checkout -- would clobber uncommitted changes to the files, while copying a `.bak` file wouldn't?

jacobsimionato: I had a version of this previously, but I didn't like that it polluted the workspace with `.bak` files. 

The current version at least checks whether you have local changes of the tracked files and aborts if you do, providing some safety around clobbering. Obviously if you edit the  files while the test is running, it won't protect you, but I think that's getting into the weeds!

PR #2572, opened 2026-09-09
docs(spec,blueprint): support cross-version surface updates and catalog compatibility

jacobsimionato:
## Overview & Rationale

This PR refines the **A2UI v1.0 protocol specification**, the **Evolution Guide**, and the **Core SDK Architecture Blueprint** to:
1. Clarify that component catalog compatibility is evaluated at the **message level** rather than enforcing that every catalog on a surface must use the exact same specification version.
2. Explicitly permit a single UI surface (`surfaceId`) to receive updates across different protocol versions over its lifetime (e.g. a surface created by a `v0.9` message can be updated by `v1.0` messages).
3. Specify conditional surface default fallback: components in an update message can fall back to the surface's default catalog only if that default catalog is *compatible* with the message's protocol version. If incompatible, components must explicitly provide a `catalogId`.

---

### Key Changes:

1. **Protocol Specification (`specification/v1_0/docs/a2ui_protocol.md`)**:
   - Updated **Mixable catalogs and component resolution logic**:
     - Clarified that all catalogs referenced within an `agentToRenderer` message must be compatible with that message's protocol version.
     - Documented that a surface created with one protocol version may be updated by subsequent messages of a different protocol version.
     - Updated resolution order: fallback to the surface default catalog is only permitted when the default catalog is compatible with the message's protocol version; otherwise, an explicit compatible `catalogId` is required.

2. **Evolution Guide (`specification/v1_0/docs/evolution_guide.md`)**:
   - Updated resolution logic in the overview and Section 2.7 (Processing rules).
   - Updated Section 3 (`For renderers`) to instruct renderers to support multi-version surface handling in `SurfaceGroupModel`.

3. **Core SDK Architecture Blueprint (`blueprints/modules/a2ui_core.blueprint.md`)**:
   - Updated the **Surface Lifecycle** section to note that surfaces are long-lived state containers that may receive updates across messages of different protocol versions.

---

### Verification:

- `python3 blueprints/validate_blueprints.py` passed.
- `python3 specification/scripts/validate.py` passed across v0.8, v0.9, and v1.0.
- Prettier formatting verified via `yarn prettier --config .prettierrc --check`.

### DIFF
--- a/specification/v1_0/docs/a2ui_protocol.md
+++ b/specification/v1_0/docs/a2ui_protocol.md
@@ -492,16 +492,22 @@ This structure is designed to be both flexible and strictly validated.
 
 #### Mixable catalogs and component resolution logic
 
-Renderers can support components and functions from multiple catalogs simultaneously within a single surface (mixable catalogs). When a renderer advertises `supportedCatalogIds` in its capabilities, components from any of those catalogs can be combined in the same UI tree. The set of available catalogs for a surface includes both `supportedCatalogIds` and the `catalogId` of any inline catalog declared in `inlineCatalogs` (when supported by the agent). All catalog IDs specified at the component and function-call levels and at the surface-level must refer to catalogs which use the same A2UI specification version.
+Renderers can support components and functions from multiple catalogs simultaneously within a single surface (mixable catalogs). When a renderer advertises `supportedCatalogIds` in its capabilities, components from any of those catalogs can be combined in the same UI tree. The set of available catalogs for a surface includes both `supportedCatalogIds` and the `catalogId` of any inline catalog declared in `inlineCatalogs` (when supported by the agent).
+
+All catalog IDs referenced within a single `agentToRenderer` message (whether at the component, function-call, or surface level) must refer to catalogs that are compatible with that message's protocol version. While protocol v0.9 and v1.0 catalogs are incompatible due to structural changes, future protocol versions will endeavour to maintain backward compatibility with older catalog versions (for example, protocol v1.1 _may_ accept v1.0 catalogs).
+
+Furthermore, a surface created with one protocol version may be updated by subsequent messages using a different protocol version (for example, a surface created with a v0.9 message may receive updates from a v1.0 message). In such cases, components in the updating message must resolve against a catalog compatible with that message's protocol version.
 
 When resolving a component (or function call), the renderer evaluates catalog identity using the following strict resolution order:
 
-1. **Explicit Component/Function-Level `catalogId`**: The renderer checks if the component or function call explicitly specifies a `catalogId`. If provided, the component or function is resolved against that catalog.
-2. **Surface Default `catalogId`**: If the component or function call does not specify a `catalogId`, the renderer checks if a default `catalogId` was specified on the surface in the `createSurface` message. If provided, the component or function is resolved against that surface default catalog.
-3. **Resolution Error**: If neither an explicit component/function-level `catalogId` nor a surface default `catalogId` is present, resolution fails immediately with an error and the component is not rendered (or the function call is rejected).
+1. **Explicit Component/Function-Level `catalogId`**: The renderer checks if the component or function call explicitly specifies a `catalogId`. If provided, the component or function is resolved against that catalog (which must be compatible with the message's protocol version).
+2. **Surface Default `catalogId`**: If the component or function call does not specify a `catalogId`, the renderer checks if a default `catalogId` was specified on the surface in the `createSurface` message:
+   - If the surface default catalog is **compatible** with the message's protocol version, the component or function is resolved against that surface default catalog.
+   - If the surface default catalog is **incompatible** with the message's protocol version (such as a v1.0 message updating a surface whose default catalog is v0.9), fallback to the surface default is disallowed. The component or function call must explicitly specify a `catalogId` compatible with the message's protocol version.
+3. **Resolution Error**: If neither an explicit component/function-level `catalogId` nor a compatible surface default `catalogId` is present, resolution fails immediately with an error and the component is not rendered (or the function call is rejected).
 
 > [!IMPORTANT]
-> There is **no fallback** to the list of catalogs declared in `rendererCapabilities` (even if the renderer only advertises a single supported catalog). Every component and function call must resolve through either its explicit `catalogId` or the surface default `catalogId`.
+> There is **no fallback** to the list of catalogs declared in `rendererCapabilities` (even if the renderer only advertises a single supported catalog). Every component and function call must resolve through either its explicit `catalogId` or a compatible surface default `catalogId`.
 
 ### Catalog-Agnostic Accessibility Requirements
 
--- a/blueprints/modules/a2ui_core.blueprint.md
+++ b/blueprints/modules/a2ui_core.blueprint.md
@@ -380,7 +380,7 @@ When a surface is created with `sendDataModel: true`, the renderer is responsibl
 3.  The **Transport Layer** (e.g., A2A, MCP) calls `getRendererDataModel()` before sending any message to the agent.
 4.  If a non-empty data model map is returned, it is included in the transport's metadata field (e.g., `A2uiRendererDataModel` in A2A metadata).
 
-- **Surface Lifecycle**: It is an error to receive a `createSurface` message for a `surfaceId` that is already active; `surfaceId` must be globally unique per client session. The processor MUST throw an error or report a validation failure if this occurs.
+- **Surface Lifecycle**: It is an error to receive a `createSurface` message for a `surfaceId` that is already active; `surfaceId` must be globally unique per client session. The processor MUST throw an error or report a validation failure if this occurs. Surfaces are long-lived state containers that may receive updates across messages of different protocol versions (e.g., a v1.0 `updateComponents` updating a surface created via v0.9), provided each component resolves to a catalog compatible with the incoming message's protocol version.
 - **Component Lifecycle**: If an `updateComponents` message provides an existing `id` but a _different_ `type`, the processor MUST remove the old component and create a fresh one to ensure framework renderers correctly reset their internal state.
 
 #### [Capabilities Objects](../../docs/public/concepts/glossary.md#capabilities-object)
--- a/specification/v1_0/docs/evolution_guide.md
+++ b/specification/v1_0/docs/evolution_guide.md
@@ -12,7 +12,7 @@ Version 1.0 differs from 0.9 in the following ways:
 - Added an optional `catalogId` property to `ComponentCommon` and `FunctionCall` to allow individual components and function calls to explicitly declare their source catalog.
 - Retained `catalogId` on `createSurface` as an optional parameter that defines the default catalog for that surface.
 - Added extensibility metadata support via `$defs/Extensions` in `common_types.json`, allowing `ComponentCommon`, `createSurface`, and `ComponentDefinition` to convey arbitrary extension key-value pairs (with Unicode UAX #31 keys and reserved `a2ui_` namespace).
-- Defined explicit component and function call resolution logic: the renderer checks the component-level (or function call-level) `catalogId` first, then falls back to the surface default `catalogId`. If neither is defined, the renderer errors out and does not render the component (or rejects the function call). There is no fallback to catalogs declared in capabilities. Available catalogs for a surface include both `supportedCatalogIds` and any negotiated `inlineCatalogs`, and all mixed catalogs must use the same A2UI specification version.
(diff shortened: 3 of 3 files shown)


specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -492,16 +492,22 @@ This structure is designed to be both flexible and strictly validated.
...
+Renderers can support components and functions from multiple catalogs simultaneously within a single surface (mixable catalogs). When a renderer advertises `supportedCatalogIds` in its capabilities, components from any of those catalogs can be combined in the same UI tree. The set of available catalogs for a surface includes both `supportedCatalogIds` and the `catalogId` of any inline catalog declared in `inlineCatalogs` (when supported by the agent).
+
+All catalog IDs referenced within a single `agentToRenderer` message (whether at the component, function-call, or surface level) must refer to catalogs that are compatible with that message's protocol version. While protocol v0.9 and v1.0 catalogs are incompatible due to structural changes, future protocol versions will endeavour to maintain backward compatibility with older catalog versions (for example, protocol v1.1 _may_ accept v1.0 catalogs).
```
nan-yu, line 497: > future protocol versions will endeavour to maintain backward compatibility with older catalog versions (for example, protocol v1.1 _may_ accept v1.0 catalogs).

Patch and minor versions should maintain backward compatibility. How about major version update?

nan-yu, line 497: How do we determine if catalogs are compatible with the message's protocol version? Do we assume all major version updates are incompatible, e.g. v0.8 and v0.9, v0.9 and v1.0 are not compatible. v1.0 and v1.1 should be compatible. Do we expect v1.0 and v2.0 to be compatible?

specification/v1_0/docs/a2ui_protocol.md
```diff
@@ -492,16 +492,22 @@ This structure is designed to be both flexible and strictly validated.
...
-3. **Resolution Error**: If neither an explicit component/function-level `catalogId` nor a surface default `catalogId` is present, resolution fails immediately with an error and the component is not rendered (or the function call is rejected).
+1. **Explicit Component/Function-Level `catalogId`**: The renderer checks if the component or function call explicitly specifies a `catalogId`. If provided, the component or function is resolved against that catalog (which must be compatible with the message's protocol version).
+2. **Surface Default `catalogId`**: If the component or function call does not specify a `catalogId`, the renderer checks if a default `catalogId` was specified on the surface in the `createSurface` message:
+   - If the surface default catalog is **compatible** with the message's protocol version, the component or function is resolved against that surface default catalog.
```
nan-yu, line 505: @gspencergoog FYI, this requires changing the validation logic in the a2ui-core SDK.

gspencergoog: Yeah.  I'm looking at that logic right now.

gspencergoog: I think this means that the new flowchart is:

```mermaid
flowchart TD
    A[Inbound Message] --> B[Adapter extracts operation with message version]
    B --> C{Component specifies catalogId?}
    C -- Yes --> D{Is catalog compatible with message version?}
    D -- Yes --> E[Resolve to Component Catalog]
    D -- No --> F[Resolution Error: Catalog incompatible with message version]
    C -- No --> G{Was default catalogId specified in createSurface?}
    G -- No --> H[Resolution Error: No default catalog on surface]
    G -- Yes --> I{Is surface default catalog compatible with message version?}
    I -- Yes --> J[Resolve to Surface Default Catalog]
    I -- No --> K[Resolution Error: Fallback disallowed for incompatible message version]
```

nan-yu: Exactly. The only thing I am not sure is how we actually determine if a catalog is compatible with the message's protocol version.

Should we just assume a standard SemVer approach where major version changes break compatibility, but minor/patch updates should be compatible?

specification/v1_0/docs/a2ui_protocol.md
gspencergoog: Yes, this is the semver major version question we should solve.  When we bump a major version (other than 0.9 to 1.0), do we expect incompatibility?

I would say that the protocols are incompatible then (e.g. a v2.0 message with a v1.0 catalogId would fail), but that the renderers themselves will still support rendering all protocol versions. You wouldn't be able to mix and match v1.x messages with v2.x messages using the same catalog ID. I would assume that any a v1.x message with a v1.x catalog ID would still handle it without failing (forward and backward compatible). Here's a compatibility matrix for a catalog for a particular protocol version and its compatibility with incoming update messages of other protocol versions:

(from Gemini):

| Catalog protocolVersion | Msg 0.8 | Msg 0.9 | Msg 1.0 | Msg 1.1 | Msg 2.0 |
| :--- | :---: | :---: | :---: | :---: | :---: |
| **0.8** | ✅ **Yes** | ❌ No | ❌ No | ❌ No | ❌ No |
| **0.9** | ❌ No | ✅ **Yes** | ❌ No\* | ❌ No\* | ❌ No\* |
| **1.0** | ❌ No | ✅ **Yes** | ✅ **Yes** | ✅ **Yes** | ❌ No\* |
| **1.1** | ❌ No | ✅ **Yes** | ✅ **Yes** | ✅ **Yes** | ❌ No\* |
| **2.0** | ❌ No | ❌ No | ❌ No | ❌ No | ✅ **Yes** |

> [!NOTE]
> **\*** Indicates cases where **fallback to the surface default catalog is disallowed** due to major SemVer incompatibility ($1.x/0.9 \leftrightarrow 2.x$). However, a cross-version update can be placed on the surface if **every component and function call in the updating message provides an explicit compatible `catalogId`**.

### The Two Runtime Compatibility Checks

When a message arrives at a renderer, runtime resolution strictly evaluates **two questions**:

1. **Can the renderer parse this message?** The incoming message's protocol version must be included in the renderer's supported versions.
2. **Can the resolved catalog validate this component?** The resolved catalog's `protocolVersion` must be compatible with the incoming message's protocol version.

Neither the surface container itself nor the catalog's internal release version participates in this check.

I found this helpful:

A2UI distinguishes between **7 distinct version concepts**:

| # | Version Concept | Where It Lives | Runtime Role & Checked Against |
| :-: | :--- | :--- | :--- |
| **1** | **Incoming Message Protocol Version** | In the message root: `{"version": "v1.0", ...}` | **Primary runtime driver.** Checked against (A) Renderer Supported Versions and (B) Resolved Catalog `protocolVersion`. |
| **2** | **Catalog Target `protocolVersion`** | In `catalog.json` under `"protocolVersion": "1.0"` | **Schema compatibility anchor.** Checked against the incoming message's protocol version. |
| **3** | **Catalog Domain / Release Version** | Inside `catalogId` (e.g. `acme.com/catalog:v2.4`) or catalog metadata | **Opaque identifier.** Purely for catalog maintainers; the protocol does not evaluate this semver. |
| **4** | **Renderer Supported Protocol Versions** | Top-level version keys in `a2uiRendererCapabilities` & registered `VersionAdapter`s | **Gatekeeper.** Ensures the renderer possesses an adapter capable of parsing the incoming message. |
| **5** | **Surface Creation Version (Colloquial)** | *Implicit only* (the `version` of the initial `createSurface` message) | **Informal shorthand.** Surfaces in memory are version-agnostic containers; they hold components from mixable catalogs. |
| **6** | **Agent Supported Protocol Versions** | Declared in agent metadata / handshake | **Negotiation.** The protocol versions the agent knows how to generate. |
| **7** | **SDK / Software Library Release Version** | In `package.json` / `pyproject.toml` (e.g. `v0.2.1`) | **Code distribution.** Library release tag, orthogonal to wire protocol versioning. |

nan-yu: Thanks. The version table is very clear. A few clarification questions:
- `#4`, RendererCapabilities does not have `supportedProtocolVersions` now. It currently only has `supportedCatalogIds`. Are we planning to add it to https://github.com/a2ui-project/a2ui/blob/v1_0/specification/v1_0/json/renderer_capabilities.json?
- I thought we've decided v0.9 Catalogs are not compatible with v1.x protocol versions due to the breaking changes. Should we update the compatibility matrix?

gspencergoog: Sorry, for `#4`, it was hallucinated.  I edited to have the right things.

> I thought we've decided v0.9 Catalogs are not compatible with v1.x protocol versions due to the breaking changes. Should we update the compatibility matrix?

Yeah, good point, we did decide that.  Updated.

nan-yu: I'm still not clear about [RendererCapabilities](https://github.com/a2ui-project/a2ui/blob/v1_0/specification/v1_0/json/renderer_capabilities.json). It currently only has a single protocol version. Do we allow multiple protocol versions? If not, how do we handle messages with multiple versions?
```json
{  "messages": [
    {
      "version": "v1.0",
      "createSurface": {
        "surfaceId": "s1",
        "catalogId": "https://a2ui.org/specification/v1_0/catalogs/basic/catalog.json"
      }
    },
    {
      "version": "v1.1",
      "updateComponents": {
        "surfaceId": "s1",
        "components": [],
      }
    }
]}
```

nan-yu: Catalog 1.0 and 1.1 should also be incompatible with message protocol version 0.9.

jacobsimionato: These tables are amazing and highlight the difficulty of the problems that we've invented for ourselves :-D.

I think that the table still overstates the compatibility between v0.9 and v1.0 a little - it says that `Msg 0.9` *is* compatible with protocol version `1.0`, and `1.1`.

jacobsimionato: Looking closely at the matrix and thinking through future iterations, I think there is an important nuance regarding **directional compatibility** within a major version line (i.e. why `messageVersion >= catalogVersion`).

In standard Semantic Versioning, minor versions guarantee **backward compatibility**, but **not forward compatibility**:

- A newer message (`v1.1`) can consume an older catalog (`v1.0`) because the `v1.1` specification is a superset of `v1.0`.
- An older message (`v1.0`) cannot be guaranteed to consume a newer catalog (`v1.1`), because the newer catalog might leverage schema features or types that did not exist in `v1.0`.

### A Concrete Example

Suppose in protocol `v1.1`, we introduce plummable component outputs:
1. `common_types.json` in `v1.1` adds a new type, e.g. `StringOutput`.
2. A `v1.1` catalog's `TextField` declares an output property of type `StringOutput`.
3. In `v1.1`, a `Text` component instance can plumb that value to a `DynamicString` input using a new expression construct:
   ```json
   "text": {
     "fromOutput": {
       "componentId": "myTextField",
       "path": "text"
     }
   }
   ```

A `v1.0` protocol message has no grammar or schema vocabulary to express `fromOutput`, nor does a `v1.0` validator recognize `StringOutput`. Therefore, a `v1.0` message cannot reliably consume or bind components from a `v1.1` catalog. 

Conversely, a `v1.1` message can easily use a `v1.0` catalog without any issues.

### Adjusted Compatibility Matrix

Factoring in bidirectional incompatibility for `v0.9` and directional backward compatibility for `v1.x`:

| Catalog `protocolVersion` | Msg 0.8 | Msg 0.9 | Msg 1.0 | Msg 1.1 | Msg 2.0 |
| :--- | :---: | :---: | :---: | :---: | :---: |
| **0.8** | ✅ **Yes** | ❌ No | ❌ No | ❌ No | ❌ No |
| **0.9** | ❌ No | ✅ **Yes** | ❌ No\* | ❌ No\* | ❌ No\* |
| **1.0** | ❌ No | ❌ No | ✅ **Yes** | ✅ **Yes** *(backward compatible)* | ❌ No\* |
| **1.1** | ❌ No | ❌ No | ❌ **No** *(forward incompatible)* | ✅ **Yes** | ❌ No\* |
| **2.0** | ❌ No | ❌ No | ❌ No | ❌ No | ✅ **Yes** |

*\* Fallback to surface default catalog is disallowed across major version mismatches, but explicit compatible `catalogId`s can still be used.*

### Proposed Specification Rule

This keeps runtime checking very clean and standard without needing complex range syntax:
1. **Across major versions** (e.g. `0.9` ↔ `1.x` ↔ `2.x`): Mutually incompatible by default.
2. **Within a major version line** (`1.x`): Standard directional backward compatibility:  
   `Major(message) == Major(catalog)` and `Minor(message) >= Minor(catalog)`

Does this model make sense to you both? If you agree, I can update the wording in this PR to capture this directional SemVer principle.

specification/v1_0/docs/a2ui_protocol.md
jacobsimionato: Yes, but I think that additionally protocol version >= catalog version - see my proposed update to Greg's table above.

specification/v1_0/docs/a2ui_protocol.md
gspencergoog: Yes, that makes sense to me. I originally thought we were trying to go for more forward compatibility within a major version, so that 1.1 catalogs would support 1.0 messages, but I see what you mean about how that's not really feasible.

PR #2583, opened 2026-09-09
feat(swift): v1.0 support

pinieb:
## Summary

This pull request implements A2UI Protocol v1.0 support in the Swift codebase (`swift/core`, `swift/swiftui`, and `swift/sample`) while maintaining backward compatibility with v0.9 and v0.9.1 catalogs and payloads. It also modernizes wire models and error types to use typed enums and updated naming conventions.

### Protocol models & versioning
* Added `A2UIProtocolVersion` enum (`v0.9`, `v0.9.1`, `v1.0`) with standard raw-value `Codable` decoding and encoding.
* Modernized wire message models: renamed types to `AgentToRendererMessage`, `RendererToAgentMessage`, `RendererAction`, and `RendererError` without legacy type aliases.
* Refactored `ValidationFailedError` to use a string-backed `Code` enum (`VALIDATION_FAILED`, `UNALLOWED_PARENT`, `UNALLOWED_CHILD`) with compiler-synthesized `Codable`.

### Remote procedure calls (RPC)
* Added v1.0 RPC wire message types: `CallRendererFunctionMessage`, `RendererFunctionResponseMessage`, `CallAgentFunctionMessage`, and `AgentFunctionResponseMessage`.
* Added `RPCHandler` to `MessageProcessor` for bidirectional function execution and correlation with timeout and error handling.

### Catalogs & validation
* Embedded JSON Schema Draft 2020-12 `common_types.json` and `catalog_definition.json` in `A2UICommonSchema.swift`.
* Added `AllowedCallers`, `ValidationResult`, `ValidationSeverity`, and `UnicodeIdentifierValidator`.
* Extended `ComponentAPI` with `allowedParents` and `allowedChildren`, and `FunctionAPI` with `allowedCallers` and `requiresUserActivation`.
* Updated `GraphTopologyValidator` to enforce component composition rules.

### State engine updates
* Implemented explicit null deletion in `DataModel`.
* Added dynamic `@index` evaluation in `DataContext` for templates.
* Updated path resolution for empty relative pointers (`path: ""`) and template child instance IDs (`<child>_<parent>_<index>`).

### Basic catalog v1.0
* Added `IndexFunction` for accessing iteration index in templates.
* Updated validation functions (`EmailFunction`, `NumericFunction`, `RequiredFunction`, `RegexFunction`, `LengthFunction`) to return `ValidationResult`.
* Registered `v10Catalog` and `v10Components` with Draft 2020-12 schema compatibility.

### Testing & conformance
* Added `V10ConformanceTests` with embedded test YAML data in `EmbeddedV10YAML.swift`, ensuring test suites run self-contained within the repository without requiring unmerged upstream files on disk.
* Added unit test suites for `RPCHandler`, `A2UIProtocolVersion`, `CatalogV10`, `StateEngineV10`, and `IndexFunction`.

## Verification

From repository root:
* `swift build`: builds cleanly without errors.
* `swift test` in `swift/core`: 225 core tests and 17 conformance tests pass.
* `swift test` in `swift/swiftui`: 58 SwiftUI integration tests pass.
* `swift-format lint -r swift/`: 0 warnings, 0 errors.


jacobsimionato (commented):
## Code Review: A2UI Protocol v1.0 Swift Support

Thank you for putting this together! The v1.0 protocol coverage, bidirectional RPC implementation, and backward compatibility for existing Swift clients are great additions. 

Before merging, there are a few important issues—including a critical runtime file resolution bug in production binaries and several architectural edge cases—that should be addressed:

---

### 1. Replace `#filePath` runtime file traversal with build-time code-gen or static constants
* **File:** [`swift/core/Sources/A2UIJSON/A2UICommonSchema.swift` (lines 84–139)](https://github.com/a2ui-project/a2ui/blob/pr-2583/swift/core/Sources/A2UIJSON/A2UICommonSchema.swift#L84-L139)
* **Problem:** `A2UICommonSchema` is production library code (`Sources/A2UIJSON/`). Using `#filePath` with 5x `deletingLastPathComponent()` to resolve the repository root and read `specification/` JSON files from disk at runtime only works in local development environments. When this package is consumed as a dependency in an iOS/macOS application, deployed to a physical device, or distributed as a compiled binary framework, the repository root and `specification/` folder do not exist on the device filesystem. This causes `assertionFailure` in debug and returns an empty schema `.object([:])` in release, breaking schema validation and `$ref` resolution in production.
* **Suggested Fix:** Inject the schema text as static constants at build time (e.g., via an automated generation script or SPM build plugin) or embed the canonical JSON string literals so schemas initialize synchronously and reliably without disk I/O in production binaries.

---

### 2. Declare v1.0 component schemas explicitly instead of runtime string hacking
* **File:** [`swift/core/Sources/BasicCatalog/BasicCatalog.swift` (lines 54–135)](https://github.com/a2ui-project/a2ui/blob/pr-2583/swift/core/Sources/BasicCatalog/BasicCatalog.swift#L54-L135)
* **Problem:** `migrateSchemaToV10` serializes v0.9 schemas to JSON strings, performs string replacements on base URIs, and runs a hardcoded `switch name` with ad-hoc dictionary mutations for `Video`, `Slider`, `Text`, `Icon`, and `Tabs`. This is fragile, difficult to maintain, and violates the repository principle to keep catalog definitions schema-driven.
* **Suggested Fix:** Define explicit v1.0 component declarations (e.g. in `BasicCatalogV10Components`) that directly specify the v1.0 schemas and composition rules. They can map to the exact same underlying SwiftUI view implementations (`ComponentImplementation`).

---

### 3. Manage async RPC tasks and pass active `DataContext`
* **File:** [`swift/core/Sources/A2UICore/Processing/MessageProcessor.swift` (lines 435–447)](https://github.com/a2ui-project/a2ui/blob/pr-2583/swift/core/Sources/A2UICore/Processing/MessageProcessor.swift#L435-L447)
* **Problem:** In `processCallRendererFunction`:
  1. Spawning a detached `Task { ... }` leaves incoming RPC executions unmanaged and non-cancellable when surfaces or the processor are deallocated.
  2. Passing `dataContext: nil` causes `RPCHandler` to fall back to an empty `DataModel()`. Any dynamic data bindings (`{"path": "/..."}`) in function call arguments will evaluate to `null`.
* **Suggested Fix:** Retain the spawned `Task` in a cancellable collection on `MessageProcessor` and resolve the appropriate surface's `DataContext` (or surface data model) before delegating to `RPCHandler`.

---

### 4. Prevent ID collisions in nested template repeaters
* **File:** [`swift/core/Sources/A2UICore/Resolution/NodeResolver.swift` (lines 909–923)](https://github.com/a2ui-project/a2ui/blob/pr-2583/swift/core/Sources/A2UICore/Resolution/NodeResolver.swift#L909-L923)
* **Problem:** Template children are currently assigned `let itemID = "\(templateID)_\(index)"`. In nested repeaters, inner template items will produce colliding IDs (`Item_0`, `Item_1`) across different outer rows, breaking view identity in SwiftUI `ForEach`. Furthermore, the PR description mentions `<child>_<parent>_<index>`, but the code does not incorporate the parent instance ID.
* **Suggested Fix:** Incorporate the parent component instance ID into the generated child ID: `let itemID = "\(templateID)_\(instanceID)_\(index)"`.

---

### 5. Base component reference detection on schema structure, not property names
* **File:** [`swift/core/Sources/A2UICore/Validation/GraphTopologyValidator.swift` (lines 418–456)](https://github.com/a2ui-project/a2ui/blob/pr-2583/swift/core/Sources/A2UICore/Validation/GraphTopologyValidator.swift#L418-L456)
* **Problem:** Checking `propertyName == "child"` or `"children"` causes regular data properties with those names to be misidentified as component references (throwing dangling reference errors), while non-standard slot properties (`header`, `leadingItem`) without those exact names are skipped.
* **Suggested Fix:** Remove the `propertyName` checks and identify component references strictly by inspecting the schema definition (`$ref` targeting `Child` or `ChildList`).

---

### 6. Update PR description to match current branch contents
* **Problem:** Commit `94a9b84b` removed `EmbeddedV10YAML.swift` and `V10ConformanceTests.swift`, yet the PR description still references them as verification mechanisms. It also states schemas are embedded in `A2UICommonSchema.swift` and mentions `<child>_<parent>_<index>`.
* **Suggested Fix:** Please update the PR summary and testing sections to reflect the current state of the codebase.

swift/core/Sources/A2UICore/Models/A2UIProtocolVersion.swift
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+/// Supported A2UI protocol versions.
+public enum A2UIProtocolVersion: String, Codable, Sendable, CaseIterable, Comparable {
+  case v09 = "v0.9"
+  case v091 = "v0.9.1"
+  case v10 = "v1.0"
+
+  /// The default active protocol version for outgoing messages.
+  public static let `default`: A2UIProtocolVersion = .v10
```
jacobsimionato, line 22: Shall we make the developer pass this explicitly? @nan-yu WDYT? My concern here is that if we have a default, then it's a breaking change for us to change the default when we introduce v1.1 etc. If we ask developers to pass this explicitly, then there's an obvious thing for them to change.

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -102,6 +136,19 @@ public final class MessageProcessor: ObservableObject {
...
     }
 
+    /// Creates a capabilities option instance with an `A2UIProtocolVersion`.
+    ///
+    /// - Parameters:
+    ///   - includeInlineCatalogs: Whether to include inline catalog definitions.
+    ///   - protocolVersion: The `A2UIProtocolVersion` enum case.
+    public init(
+      includeInlineCatalogs: Bool = false,
+      protocolVersion: A2UIProtocolVersion
```
jacobsimionato, line 146: @nan-yu can we make it so that surfaces are not tied to one protocol version?

jacobsimionato (commented):
(Moved to inline review comments)

swift/core/Sources/A2UIJSON/A2UICommonSchema.swift
```diff
@@ -35,294 +36,105 @@ public enum A2UICommonSchema {
...
+        v10CatalogDefinitionDocument,
+    ]
+  }
+
+  private static func loadSpecDocument(
+    relativePath: String,
+    overrideID: String? = nil,
+    filePath: String = #filePath
+  ) -> JSONValue {
+    let repoRoot = URL(fileURLWithPath: filePath)
```
jacobsimionato, line 89: ### Replace `#filePath` runtime file traversal with build-time code-gen or static constants

Relying on `#filePath` and 5x `deletingLastPathComponent()` to read `specification/` at runtime works during local SPM test execution, but it will fail when compiled as a dependency in an iOS/macOS application or on a physical device where the repository source tree does not exist. This will cause `assertionFailure` in debug and return an empty schema `.object([:])` in release, breaking schema validation in production.

**Suggested Fix:** Inject the schema text as static constants at build time (e.g. via an automated generation script or SPM build plugin) or embed the canonical JSON string literals so schemas initialize synchronously and reliably without disk I/O in production binaries.

swift/core/Sources/BasicCatalog/BasicCatalog.swift
```diff
@@ -48,21 +51,140 @@ public enum BasicCatalog: Sendable {
       """
   )
 
-  private static func makeCatalog(id: String) -> AnyCatalog {
+  private static func migrateSchemaToV10(name: String, _ schema: Schema) -> Schema {
```
jacobsimionato, line 54: ### Declare v1.0 component schemas explicitly instead of runtime string hacking

`migrateSchemaToV10` serializes v0.9 schemas to JSON strings, performs string replacements on base URIs, and runs a hardcoded `switch name` with ad-hoc dictionary mutations for `Video`, `Slider`, `Text`, `Icon`, and `Tabs`. This is fragile, difficult to maintain, and violates the repository principle to keep catalog definitions schema-driven.

**Suggested Fix:** Define explicit v1.0 component declarations (e.g. in `BasicCatalogV10Components`) that directly specify the v1.0 schemas and composition rules. They can map to the exact same underlying SwiftUI view implementations (`ComponentImplementation`).

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -337,32 +388,106 @@ public final class MessageProcessor: ObservableObject {
...
+    if let components = msg.components {
+      try validateComponentsBatch(components, on: vm)
+      applyComponentsBatch(components, to: vm)
+    }
+
     surfaceGroupModel.addSurface(vm)
   }
 
+  private func processCallRendererFunction(_ msg: CallRendererFunctionMessage) throws {
+    Task { @MainActor [weak self] in
```
jacobsimionato, line 436: ### Manage async RPC tasks and pass active `DataContext`

In `processCallRendererFunction`:
1. Spawning a detached `Task { ... }` leaves incoming RPC executions unmanaged and non-cancellable when surfaces or the processor are deallocated.
2. Passing `dataContext: nil` causes `RPCHandler` to fall back to an empty `DataModel()`. Any dynamic data bindings (`{"path": "/..."}`) in function call arguments will evaluate to `null`.

**Suggested Fix:** Retain the spawned `Task` in a cancellable collection on `MessageProcessor` and resolve the appropriate surface's `DataContext` (or surface data model) before delegating to `RPCHandler`.

swift/core/Sources/A2UICore/Resolution/NodeResolver.swift
```diff
@@ -748,10 +871,12 @@ public final class NodeResolver: Sendable {
       var resolvedNodes: [Node] = []
       for item in arr {
         guard let childID = item.stringValue else { continue }
+        let childInstanceID = index.map { "\(childID)_\($0)" } ?? childID
```
jacobsimionato, line 874: ### Prevent ID collisions in nested template repeaters

Template children are assigned `let childInstanceID = index.map { "\(childID)_\($0)" } ?? childID`. In nested repeaters, inner template items will produce colliding IDs (`Item_0`, `Item_1`) across different outer rows, breaking view identity in SwiftUI `ForEach`.

The PR description mentions `<child>_<parent>_<index>`, but the code does not incorporate the parent instance ID.

**Suggested Fix:** Incorporate the parent component instance ID into the generated child ID: e.g. `let itemID = "\(templateID)_\(instanceID)_\(index)"`.

swift/core/Sources/A2UICore/Validation/GraphTopologyValidator.swift
```diff
@@ -167,52 +332,145 @@ public enum GraphTopologyValidator {
...
+    }
+    return false
+  }
+
+  private static func isSingleChildSchema(_ schema: JSONValue, propertyName: String = "") -> Bool {
+    if let ref = schema["$ref"]?.stringValue {
+      let refName = ref.split(separator: "/").last.map(String.init)
+      if refName == "Child" || refName == "ComponentId" { return true }
+    }
+    if propertyName == "child", schema["type"]?.stringValue == "string" {
```
jacobsimionato, line 444: ### Base component reference detection on schema structure, not property names

Checking `propertyName == "child"` or `"children"` causes regular data properties with those names to be misidentified as component references (throwing dangling reference errors), while non-standard slot properties (`header`, `leadingItem`) without those exact names are skipped.

**Suggested Fix:** Remove the `propertyName` checks and identify component references strictly by inspecting the schema definition (`$ref` targeting `Child` or `ChildList`).

swift/core/Sources/A2UICore/Messages/RendererToAgentMessage.swift
```diff
@@ -0,0 +1,101 @@
...
+        codingPath: container.codingPath,
+        debugDescription: "Internal error: version key was not filtered out"
+      )
+      throw DecodingError.dataCorrupted(context)
+    }
+  }
+
+  public func encode(to encoder: Encoder) throws {
+    var container = encoder.container(keyedBy: CodingKeys.self)
+    try container.encode(A2UIProtocolVersion.default, forKey: .version)
```
jacobsimionato, line 83: ### Support emitting v0.9 / v0.9.1 messages in `RendererToAgentMessage`

`RendererToAgentMessage.encode(to:)` hardcodes `try container.encode(A2UIProtocolVersion.default, forKey: .version)`, where `A2UIProtocolVersion.default` is an immutable constant (`let`) set to `.v10`. Outbound messages sent to existing v0.9.1 agents (which strictly validate `client_to_server.json` with `"version": {"const": "v0.9.1"}`) will be rejected by the agent's schema check.

**Suggested Fix:** Parameterize or store the negotiated protocol version on `RendererToAgentMessage` (or provide an encoder configuration hook) so the renderer can speak v0.9.1 to older agents.

<details>
<summary>Suggested Implementation Details</summary>

```swift
public struct RendererToAgentMessage: Equatable, Codable, Sendable {
  public var version: A2UIProtocolVersion
  public var payload: Payload

  public enum Payload: Equatable, Sendable {
    case action(RendererAction)
    case callAgentFunction(CallAgentFunctionMessage)
    case rendererFunctionResponse(RendererFunctionResponseMessage)
    case error(RendererError)
  }

  public init(_ payload: Payload, version: A2UIProtocolVersion = .default) {
    self.version = version
    self.payload = payload
  }

  public func encode(to encoder: Encoder) throws {
    var container = encoder.container(keyedBy: CodingKeys.self)
    try container.encode(version, forKey: .version)
    // Encode payload based on case...
  }
}
```
</details>

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -25,6 +25,12 @@ public final class MessageProcessor: ObservableObject {
   /// The surface group model owning all active surfaces.
   public let surfaceGroupModel: SurfaceGroupModel
 
+  /// The RPC handler managing function invocations.
+  public let rpcHandler: RPCHandler
```
jacobsimionato, line 29: ### Make `rpcHandler` internal to `MessageProcessor`

Exposing `public let rpcHandler: RPCHandler` directly on `MessageProcessor` introduces several API hazards:
1. It creates competing entry points with `processor.callAgentFunction(...)` (which is already exposed on `MessageProcessor` at line 464).
2. It exposes internal protocol methods like `handleIncomingCall` and `handleAgentResponse` that should only ever be invoked by `MessageProcessor.process(message:)` to prevent bypassing envelope validation and `DataContext` provisioning.

**Suggested Fix:**
Change `rpcHandler` from `public` to `internal`:
```swift
internal let rpcHandler: RPCHandler
```
If callers need cancellation or lifecycle management, expose `processor.cancelPendingCalls()` directly on `MessageProcessor`.

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -79,6 +97,16 @@ public final class MessageProcessor: ObservableObject {
     return .object(result)
   }
 
+  /// Returns the data model for a specific surface ID, if it exists.
+  public func getRendererDataModel(surfaceID: String) -> JSONValue? {
+    surfaceGroupModel.surfacesMap[surfaceID]?.dataModel.data
+  }
+
+  /// Returns the `SurfaceViewModel` for the specified surface ID.
+  public func surface(id: String) -> SurfaceViewModel? {
```
jacobsimionato, line 106: ### Swift-idiomatic `SurfaceGroupModel` subscripting and computed properties

`processor.surface(id:)` and `processor.getRendererDataModel()` are procedural helper methods ported directly from TypeScript that duplicate or bypass Swift collection idioms.

**Suggested Fix:**
1. Add subscripting to `SurfaceGroupModel`: `surfaceGroupModel[id]`.
2. Move data aggregation to a computed property on `SurfaceGroupModel`: `surfaceGroupModel.rendererDataModel`.
3. Access a single surface's data model directly: `surfaceGroupModel[id]?.dataModel.data`.

<details>
<summary>Suggested Implementation Details</summary>

```swift
extension SurfaceGroupModel {
  /// Enables subscripting: processor.surfaceGroupModel["main"]
  public subscript(surfaceID: String) -> SurfaceViewModel? {
    surfacesMap[surfaceID]
  }

  /// Aggregated data models for surfaces with sendDataModel: true
  public var rendererDataModel: JSONValue? {
    var result: OrderedDictionary<String, JSONValue> = [:]
    for (id, surface) in surfacesMap where surface.sendDataModel {
      result[id] = surface.dataModel.data
    }
    return result.isEmpty ? nil : .object(result)
  }
}
```
</details>

jacobsimionato (commented):
Detailed line-by-line review comments addressing runtime safety, architecture, and protocol versioning.

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -25,6 +25,12 @@ public final class MessageProcessor: ObservableObject {
   /// The surface group model owning all active surfaces.
   public let surfaceGroupModel: SurfaceGroupModel
 
+  /// The RPC handler managing function invocations.
+  public let rpcHandler: RPCHandler
+
+  /// Listener for outbound messages destined to the agent.
+  public var outboundListener: (@Sendable (RendererToAgentMessage) -> Void)?
```
jacobsimionato, line 32: ### Unify outbound transport via `outboundListener`

Currently, outbound communication from the renderer is bifurcated across two separate mechanisms:
1. RPC messages (`callAgentFunction` and `rendererFunctionResponse`) go out through `outboundListener`.
2. User UI actions (`action`) and client validation errors (`error`) still go through the legacy `ActionHandling` delegate protocol.

Because [`RendererToAgentMessage`](https://github.com/a2ui-project/a2ui/blob/pr-2583/swift/core/Sources/A2UICore/Messages/RendererToAgentMessage.swift#L18) unifies all four cases (`.action`, `.error`, `.callAgentFunction`, `.rendererFunctionResponse`), host applications should not need to configure two separate channels to communicate with the agent.

**Suggested Fix:**
Route UI actions and errors through `outboundListener` as `RendererToAgentMessage.action` and `.error`. `ActionHandling` can remain as an optional local interceptor for apps that want to handle actions locally before/instead of transmitting them.

<details>
<summary>Suggested Implementation Details</summary>

```swift
// In MessageProcessor.swift:
public func dispatchOutboundAction(_ action: RendererAction) {
  outboundListener?(.action(action))
}

public func dispatchOutboundError(_ error: RendererError) {
  outboundListener?(.error(error))
}
```
</details>

swift/core/Sources/A2UICore/Models/A2UIProtocolVersion.swift
nan-yu: Agreed. We shouldn't define a global default protocol version in the SDK. Both Python and TypeScript implementations require developers to either pass the version explicitly or declare it directly in the JSON catalog schema.

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
nan-yu: Agreed. A renderer may support multiple protocol versions simultaneously. Therefore, MessageProcessor's capabilities generation should accept a list of versions rather than being bound to a single version.

Currently, in Python, `get_renderer_capabilities` supports multiple versions by taking loose keyword arguments (versions=..., include_inline_catalogs=...).
In TS, `CapabilitiesOptions` took a single `version?: ProtocolVersion`, restricting capability generation to a single version per call.

We've opened #2786 to align both implementations.
- TS: Updated `CapabilitiesOptions` to take `versions?: ProtocolVersion[]`.
- Python: Introduced `CapabilitiesOptions(versions=..., include_inline_catalogs=...)` and updated `get_renderer_capabilities(options=...)` to match.
- Blueprint: Added the `CapabilitiesOptions` interface and documented the multi-version return mapping.

nan-yu (commented):
Still working my way through the rest of the changes—I'll follow up with another pass soon.

swift/core/Sources/A2UICore/RPC/RPCHandler.swift
```diff
@@ -0,0 +1,321 @@
...
+    )
+    let message = CallAgentFunctionMessage(
+      surfaceID: surfaceID,
+      functionCallID: callID,
+      callFunction: payload
+    )
+    let outbound = RendererToAgentMessage.callAgentFunction(message)
+
+    return try await withTaskCancellationHandler {
+      try await withCheckedThrowingContinuation { continuation in
```
nan-yu: ### 🚨 [CRITICAL] Data Race & Double-Resume in `RPCHandler`

### Issue:
An unstructured background Task is spawned inside `withCheckedThrowingContinuation` to handle timeouts. Because this Task is not actor-isolated, it executes on a cooperative background thread pool and accesses `self.pendingCalls` concurrently with `@MainActor` processor methods.

```swift
// ❌ Problematic Pattern:
return try await withCheckedThrowingContinuation { continuation in
  self.pendingCalls[callID] = continuation
  Task {
    try? await Task.sleep(nanoseconds: UInt64(timeoutSeconds * 1_000_000_000))
    if let pending = self.pendingCalls.removeValue(forKey: callID) {
      pending.resume(throwing: FunctionError.timeout(callID: callID))
    }
  }
}
```

### Failure Mode:

- **Thread Data Race**: `self.pendingCalls` dictionary mutation across threads without locks or actor isolation causes heap corruption.

- **Fatal Double Resume**: If a valid response arrives on the main thread right as the sleep timer expires on a background thread, both threads can attempt to remove/resume the continuation, triggering a fatal Swift runtime crash (`Fatal error: SWIFT TASK CONTINUATION MISUSE: tried to resume continuation twice`).

### Fix:
Isolate the timeout task to `@MainActor` and capture [weak self]:

```swift
Task { @MainActor [weak self] in
  try? await Task.sleep(nanoseconds: UInt64(timeoutSeconds * 1_000_000_000))
  guard let self else { return }
  if let pending = self.pendingCalls.removeValue(forKey: callID) {
    pending.resume(throwing: FunctionError.timeout(callID: callID))
  }
}
```

swift/core/Sources/A2UIJSON/A2UICommonSchema.swift
```diff
@@ -35,294 +36,105 @@ public enum A2UICommonSchema {
...
+                  .string("any"),
+                  .string("void"),
+                  .string("validationResult"),
+                ]),
+              ]),
+            ]),
+            "required": .array([.string("call")]),
+            "additionalProperties": .boolean(false),
+          ])
+          dict["$defs"] = .object(defs)
```
nan-yu: These lines hardcode an ad-hoc runtime mutation of `$defs["FunctionCall"]` inside loadSpecDocument that injects a v0.9-style `returnType` enum, overriding the actual v1.0 FunctionCall definition (`allOf: [FunctionCommon, oneOf: [anyFunction, IndexSystemFunction]]`).

Suggest using the specification as the single source of truth and remove the add-hoc override.

swift/core/Sources/A2UIJSON/A2UICommonSchema.swift
nan-yu: +1 on using build-time code-gen.

swift/core/Sources/A2UICore/Extensions/JSONValue+Path.swift
```diff
@@ -177,37 +180,41 @@ extension JSONValue {
...
-                array[index] = newValue
-              }
-            } else if index < array.count {
-              // Setting an array index to nil preserves the array
-              // length (sparse array), matching the blueprint's
-              // JSON Pointer Implementation Rules.
-              array[index] = .null
+        if isLastComponent {
+          if let newValue {
+            while array.count < index {
```
nan-yu: It replaces the previous `index <= array.count` check with `while array.count < index { array.append(.null) } `in four places (leaf and non-leaf for both existing and auto-vivified arrays) without any upper bound on `index`. An untrusted payload with a path like `{"path": "/items/999999999", "value": "x"}` will block `@MainActor` and crash the client with an Out-Of-Memory error.

Recommendation: Enforce a maximum array index bound (e.g., `maxArrayIndex = 10_000`, matching Python/TypeScript SDKs) before padding with `.null`.

swift/core/Sources/A2UICore/Processing/MessageProcessor.swift
```diff
@@ -337,32 +388,106 @@ public final class MessageProcessor: ObservableObject {
...
     surfaceGroupModel.addSurface(vm)
   }
 
+  private func processCallRendererFunction(_ msg: CallRendererFunctionMessage) throws {
+    Task { @MainActor [weak self] in
+      guard let self else { return }
+      let response = await self.rpcHandler.handleIncomingCall(
+        msg,
+        catalogs: self.catalogs,
+        defaultCatalogID: nil,
```
nan-yu: `processCallRendererFunction` passes `defaultCatalogID: nil, dataContext: nil` into `rpcHandler.handleIncomingCall(call)`. Inside `RPCHandler`, `let context = dataContext ?? DataContext(dataModel: DataModel(), path: "/")` creates a brand-new empty `DataModel`. Any `{"path": "/..."}` data binding inside `callRendererFunction` arguments will always resolve to `.null`.

It spawns an unstructured Task { @MainActor [weak self] in ... } that is never stored or cancelled when MessageProcessor is deallocated or reset.

Recommendation: Resolve the appropriate surface/catalog context (or pass the surface's DataContext when applicable) and track in-flight RPC tasks so they can be cancelled on teardown.

swift/core/Sources/A2UICore/Validation/A2UIValidator.swift
nan-yu: Python and TS refactored the validator to be PayloadValidator, which is single-catalog scoped. This can be updated to match the other 2 implementations.

swift/core/Sources/A2UICore/Messages/AgentToRendererMessage.swift
```diff
@@ -14,41 +14,37 @@
...
 import Foundation
 
-/// A container message enclosing one of the supported incoming
-/// server-to-client commands.
+/// A container message enclosing one of the supported incoming agent-to-renderer commands.
 ///
-/// Matches `specification/v0_9_1/json/server_to_client.json`.
-public enum ServerToClientMessage: Codable, Sendable, Equatable {
+/// Matches `specification/v1_0/json/agent_to_renderer.json`.
+public enum AgentToRendererMessage: Codable, Sendable, Equatable {
```
nan-yu: In Python and TypeScript, the core state and resolution engine (`MessageProcessor`, `SurfaceModel`, `DataModel`, `DataContext`, `NodeResolver`, `RPCHandler`) is shared across protocol versions, while wire schemas (`v0_9/schema` vs. `v1_0/schema`) and basic catalog definitions (`v0_9/basic_catalog` vs. `v1_0/basic_catalog`) are separated by version.

In this PR, Swift keeps a single unversioned set of message structs and catalog definitions, relying on runtime dictionary patching and superset models to handle both v0.9 and v1.0. While sharing the SwiftUI view layer (`A2UISwiftUI`) across v0.9 and v1.0 makes sense because both versions resolve into the same dictionary-backed `Node` tree, avoiding versioned schemas and catalog definitions in `A2UICore` and `BasicCatalog` introduces several issues:

#### 1. Runtime schema patching in `BasicCatalog.migrateSchemaToV10` produces invalid v1.0 schemas
Instead of defining v1.0 component and function schemas directly (`BasicCatalog/v0_9` vs. `BasicCatalog/v1_0`), `BasicCatalog.swift` mutates the v0.9 `ComponentAPI` schemas in place at runtime. This leaves several v1.0 specification changes incomplete:
* `Icon`: Adds `svgPath` to `customProps` without removing `path` or updating `required: ["path"]` to `required: ["svgPath"]`. Valid v1.0 `Icon` components using `{"name": {"svgPath": "..."}}` fail schema validation because `path` is still required.
* `Slider`: Adds `"default": 0` to `min` without removing `"min"` from `required: ["component", "value", "min", "max"]`. Valid v1.0 `Slider` components that omit `min` fail validation.
* `OpenURLFunction`: In v1.0 `basic/catalog.json`, `openUrl` sets `"requiresUserActivation": true`, which is missing from `OpenURLFunction.swift`.

#### 2. Unversioned wire message structs weaken validation and hardcode `"v1.0"` on encode
Because `CreateSurfaceMessage`, `AgentToRendererMessage`, and `RendererToAgentMessage` are single unversioned types shared across versions:
* `CreateSurfaceMessage` and `A2UIValidator` accept `theme` in v1.0 payloads, even though `theme` was removed from `CreateSurfaceMessage` in v1.0 (`additionalProperties: false`).
* `RendererToAgentMessage.encode(to:)` and `AgentToRendererMessage.encode(to:)` hardcode `"version": A2UIProtocolVersion.default.rawValue` (`"v1.0"`), which prevents encoding v0.9 or v0.9.1 outbound messages when communicating with v0.9/v0.9.1 agents.

#### 3. Missing v1.0 catalog registration and property handling in `A2UISwiftUI`
* `swift/swiftui/Sources/BasicCatalog/BasicCatalogImplementation.swift` only registers `v09Catalog` and `v091Catalog` in `allCatalogs` and was not updated to include `v10Catalog` (`BasicCatalog.v10CatalogURI`). In addition, each `ComponentImplementation` binds directly to the v0.9 `BasicCatalogComponents.<name>` schema.
* `A2UIIcon.swift` checks `dict["svgPath"]` (v1.0) and ignores `dict["path"]` (v0.9.1).
* `A2UIVideo.swift` reads `url` and ignores the v1.0 `posterUrl` property.

#### Recommendation
Keep the shared core engine and shared SwiftUI views (updating `BasicCatalogImplementation` to register `v10Catalog` and handling `posterUrl` and `path`/`svgPath`), but separate the v0.9 and v1.0 schema and `BasicCatalog` definitions (similar to TypeScript's `v0_9/basic_catalog` and `v1_0/basic_catalog`) rather than patching v0.9 JSON schemas at runtime. Ensure message encoding and validation preserve and enforce the active surface's `A2UIProtocolVersion`.

PR #2595, opened 2026-09-10
refactor(lit): clean up Lit basic catalog exports

josemontespg:
## Summary

This PR cleans up the unnecessary exports in the `@a2ui/lit` package.

Historically (in PR #2190), when the basic catalog components were migrated to `@a2ui/web_core`, the individual components and `BasicCatalogA2uiLitElement` were re-exported from `@a2ui/lit/v0_9/catalogs/basic` in an attempt to provide backwards compatibility. 

However, since deep imports are strictly blocked by the `exports` field in `@a2ui/lit/package.json`, these symbols were never actually accessible to external consumers via that path.

This PR:
1. Removes the redundant re-exports of individual components (e.g. `A2uiButton`) and `BasicCatalogA2uiLitElement` from `renderers/lit/src/v0_9/catalogs/basic/index.ts`.
2. Removes `BasicCatalogA2uiLitElement` from the root `@a2ui/lit/v0_9` entrypoint entirely, as it is an internal base class not meant for external use.
3. Redefines `basicCatalog` locally in Lit in terms of the individual imported components from `web_core`. This makes it possible to wrap the web_core implementation of the components in Lit wrappers.
4. Updating the Changelog entries to mention that the move to web_core was a breaking change. Purposely not documenting the removal of the individual component exports or BasicCatalogA2uiLitElement, because those were previously not exported by the currently published version of lit (before #2190).

## Pre-launch Checklist


renderers/lit/CHANGELOG.md
```diff
@@ -1,7 +1,6 @@
 ## Unreleased
 
-- (v0_9) Replace wildcard re-exports in `@a2ui/lit/v0_9` with explicit named exports.
-- (v0_9) Move universal basic catalog component implementations (`A2uiText`, `A2uiButton`, `A2uiCard`, etc.) to `@a2ui/web_core/v0_9/basic_catalog` and re-export them from `@a2ui/lit/v0_9` and `@a2ui/lit/v0_9/catalogs/basic` for backwards compatibility. [#2190](https://github.com/a2ui-project/a2ui/pull/2190)
+- (v0_9) Move universal basic catalog component implementations (`A2uiText`, `A2uiButton`, `A2uiCard`, etc.) to `@a2ui/web_core/v0_9/basic_catalog`. [#2190](https://github.com/a2ui-project/a2ui/pull/2190)
```
josemontespg: Mark this as a breaking change, because downstream apps that depend on lit should also update web_core, otherwise this would break.

josemontespg: Done! I've prefixed the changelog entry with `**BREAKING CHANGE**:` and added an explicit note stating that apps must update to the latest version of `@a2ui/web_core` to upgrade this package.

renderers/lit/src/v0_9/catalogs/basic/index.ts
```diff
@@ -14,30 +14,55 @@
...
   A2uiTabs,
-  A2uiText,
-  A2uiTextField,
-  A2uiVideo,
+  A2uiModal,
 } from '@a2ui/web_core/v0_9/basic_catalog';
+import type {LitComponentApi} from '../../types.js';
+
+/**
+ * The single canonical basic catalog of A2UI components implemented via Web Components (Custom Elements)
```
josemontespg: Use the same jsdoc from before PR 2190

 * The basic catalog for A2UI components in Lit.
 *
 * This catalog includes a wide range of components such as list, image, icon,
 * video, audio player, card, divider, checkbox, slider, date-time input, choice
 * picker, tabs, and modal. It also includes the basic functions from package
 * @a2ui/web_core.

Adapting it to the current code situation.

josemontespg: Done! I've restored the original JSDoc comment, updated the component enumeration to accurately reflect all 18 components, and removed the implementation detail about `@a2ui/web_core`.

renderers/lit/src/v0_9/catalogs/basic/index.ts
```diff
@@ -14,30 +14,59 @@
  * limitations under the License.
  */
 
-/**
- * Re-exported for backwards compatibility.
- */
-export {
-  basicCatalog,
```
ditman, line 21: I think we could remove `basicCatalog` from the `web_core` package as well:

https://github.com/a2ui-project/a2ui/blob/981e82f1a3cef88456416fa6fd80d8490964df01/renderers/web_core/src/v0_9/basic_catalog/index.ts#L44

(It doesn't even *need* to be created there? Maybe only for tests?)

renderers/lit/src/v0_9/catalogs/basic/index.ts
```diff
@@ -14,30 +14,59 @@
...
-  BasicCatalogA2uiLitElement,
-  type ResolvedChildRef,
-  type ResolvedChildList,
-  A2uiAudioPlayer,
+import {Catalog} from '@a2ui/web_core/v0_9';
+import {
+  BASIC_FUNCTIONS,
+  A2uiText,
```
ditman: The changes in this import statement are weird, maybe sort the imports alphabetically? (A2uiText should not show up before A2uiButton, for example)

ditman (approved):
Small comment for web_core cleanup. Also, please sort the imports of A2uiXXXXs alphabetically, and their usage when creating the newCatalog as well (A2uiButton should not come after A2uiText)

renderers/lit/src/v0_9/catalogs/basic/index.ts
josemontespg: Good point. I think at this moment the renderers all create a `new Catalog()` of some sort, so the basicCatalog symbol from web_core is not actually used.

On the other hand, it would look strange if web_core would not export basicCatalog, because it does export a `Catalog` symbol and all the components in the basic catalog. I think it is expected that it would also export an instance of `Catalog` with the components bundled in.

Open to reconsidering this. Anyways out of scope from this PR, we can do this in a follow up if we need to

renderers/lit/src/v0_9/catalogs/basic/index.ts
josemontespg: Done! Sorted both the `A2ui*` imports and the components array passed to `new Catalog` alphabetically.

PR #2272, opened 2026-08-14
feat(web_core): add global markdown renderer for basic catalog text component

josemontespg:
## Why

Basic catalog web components implemented in Lit within `@a2ui/web_core` (such as `a2ui-basic-text`) are designed to be shared across multiple web framework adapters (Angular, React, Lit) as part of #1270.

Currently, `a2ui-basic-text` retrieves its markdown renderer exclusively via `@lit/context`. When `a2ui-basic-text` is rendered within non-Lit host adapters (such as Angular or React), `@lit/context` is not available across framework boundaries, leaving basic text components unable to render markdown even when a markdown renderer package is installed.

## What

- **Global Default Markdown Renderer**: Added `setMarkdownRenderer(renderer?: MarkdownRenderer)` and `getMarkdownRenderer()` in `directives/markdown.ts`, exported from `@a2ui/web_core/v0_9/basic_catalog`.
- **Renderer Fallback Precedence**: `MarkdownDirective` resolves the effective markdown renderer in order:
  1. Per-component explicit renderer (passed via `@lit/context` or prop)
  2. Global default renderer (registered via `setMarkdownRenderer(...)`)
  3. Plain text fallback wrapped in `<span class="no-markdown-renderer">${value}</span>` accompanied by a one-time `console.warn`
- **Unit Tests**:
  - `markdown.test.ts`: Added tests verifying synchronous fallback rendering with async update, global markdown renderer resolution, and unconfigured fallback span with one-time warning logging.
  - `Text.test.ts`: Added test verifying `<a2ui-basic-text>` successfully renders markdown using `setMarkdownRenderer`.

Related to #1270

PR #2596, opened 2026-09-10
feat(web_core)!: add `element` to WebComponentImplementation and element registration helpers

josemontespg:
### Summary

`WebComponentImplementation` now carries the custom element constructor that implements it, and `@a2ui/web_core/v0_9/universal` gets the helpers a renderer needs to detect such implementations and register their elements on demand. Nothing changes at runtime yet: the basic catalog elements still register themselves on import. #2698 (next in the stack) removes that side effect and uses these helpers from `renderA2uiNode`.

### Changes

`@a2ui/web_core/v0_9/universal`:

* BREAKING CHANGE: `WebComponentImplementation` requires a new `element: CustomElementConstructor` field.
* `isWebComponentImplementation(api)` is a type guard that checks for a string `tagName` and a constructor `element`.
* `registerUniversalElement(component)` wraps `customElements.define`. It no-ops when the tag is already bound to the same constructor, throws when the tag is bound to a different constructor, and no-ops when `customElements` is undefined (SSR).
* `A2uiWebComponentElement` is the DOM contract of an A2UI custom element: an `HTMLElement` with an optional `context` property. Renderers that only hold an `HTMLElement` use it to assign the context.

`@a2ui/web_core/v0_9/basic_catalog`:

* The 18 `WebComponentImplementation` constants (`A2uiCard`, `A2uiButton`, ...) set `element` to their element class and are typed against their own schema (for example `WebComponentImplementation<typeof CardApi.schema>`). The classes and their `@customElement` decorators are untouched.

Tests:

* New suites for `isWebComponentImplementation` and `registerUniversalElement`.
* `render-a2ui-node.test.ts` mocks the element with a real `class extends HTMLElement`, so the JSDOM globals are installed at module load.
* The Lit `custom-catalogs.test.ts` fixture supplies an `element`.

### Verification

* `yarn --cwd renderers/web_core test` (537 tests passing)
* `yarn --cwd renderers/lit test:unit` (76 tests passing)
* `yarn workspaces foreach -A --include 'renderers/**' -p --topological-dev run build` at the top of the stack
* `./scripts/fix_format.sh --check`


renderers/web_core/src/v0_9/catalog/types.ts
```diff
@@ -111,6 +111,21 @@ export interface WebComponentImplementation<
...
+ * @returns True if the object defines a valid string `tagName`.
+ */
+export function isWebComponentImplementation(api: unknown): api is WebComponentImplementation {
+  return (
+    typeof api === 'object' &&
+    api !== null &&
+    'tagName' in api &&
+    typeof (api as {tagName?: unknown}).tagName === 'string'
+  );
+}
```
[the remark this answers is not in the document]
josemontespg: Agent: Implement this improvement

renderers/web_core/src/v0_9/catalog/create_component_implementation.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Creates a WebComponentImplementation pairing a ComponentApi with a custom element constructor.
+ * Resolves the custom element tagName from the registry (via customElements.getName) or defaults to `a2ui-${api.name.toLowerCase()}`.
+ *
+ * @param api The ComponentApi defining the component name and Zod schema.
+ * @param element The CustomElementConstructor implementing the component.
+ * @returns A WebComponentImplementation ready to be registered in a Catalog.
+ */
+export function createComponentImplementation<Schema extends z.ZodTypeAny = z.ZodTypeAny>(
```
josemontespg: To make this analogous to isWebComponentImplementation, let's rename this to createWebComponentImplementation, and update the new name across the stack too.

renderers/web_core/src/v0_9/catalog/types.ts
```diff
@@ -111,6 +111,23 @@ export interface WebComponentImplementation<
...
 
+export {createComponentImplementation} from './create_component_implementation.js';
+
+/**
+ * Type guard to check if a component API or implementation is a WebComponentImplementation.
+ *
+ * @param api The ComponentApi or component declaration to check.
+ * @returns True if the object defines a valid string `tagName`.
+ */
+export function isWebComponentImplementation(api: unknown): api is WebComponentImplementation {
```
josemontespg: Please create a new file for this function, along with a test file, to not bloat this generic types file.

renderers/web_core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## Unreleased
 
 - (v0_9) Deprecate `injectBasicCatalogStyles`, `computeColorVariant`, `ColorVariantLightDarkOptions`, and `ColorVariantHoverOptions` exports from `@a2ui/web_core/v0_9`. Consumers should import them from `@a2ui/web_core/v0_9/basic_catalog` instead.
+- (v0_9) Add `createComponentImplementation` helper for Web Components / Lit and refactor all basic catalog components to use it. [#2313](https://github.com/a2ui-project/a2ui/pull/2313)
```
josemontespg: Merge these entries into one, now that they are in the same PR.

renderers/web_core/src/v0_9/catalog/types.ts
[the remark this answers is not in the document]
josemontespg: Done. I've updated the type guard to validate `name` and `schema` properties as well.

renderers/web_core/src/v0_9/catalog/create_component_implementation.ts
josemontespg: Done. I've renamed the function and file to `createWebComponentImplementation` and updated all usages across the stack.

renderers/web_core/src/v0_9/catalog/types.ts
josemontespg: Done. I've extracted the function to `is_web_component_implementation.ts` and moved/expanded its tests to `is_web_component_implementation.test.ts`.

renderers/web_core/CHANGELOG.md
josemontespg: Done. I've merged the entries into a single line item.

renderers/web_core/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - (v0_9) Deprecate `injectBasicCatalogStyles`, `computeColorVariant`, `ColorVariantLightDarkOptions`, and `ColorVariantHoverOptions` exports from `@a2ui/web_core/v0_9`. Consumers should import them from `@a2ui/web_core/v0_9/basic_catalog` instead.
+- (v0_9) Add WebComponentImplementation helpers (`isWebComponentImplementation` type guard and `createWebComponentImplementation` helper). [#2273](https://github.com/a2ui-project/a2ui/pull/2273) [#2313](https://github.com/a2ui-project/a2ui/pull/2313)
```
josemontespg: Agent: Only link the current PR here

renderers/web_core/CHANGELOG.md
josemontespg, line 1: Agent: Please update the PR description and title with the changes in this PR.

renderers/web_core/src/v0_9/universal/create_web_component_implementation.ts
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+import type {z} from 'zod';
+import type {ComponentApi} from '../catalog/types.js';
+import type {WebComponentImplementation} from './web_component_implementation.js';
+
+/**
+ * Creates a WebComponentImplementation pairing a ComponentApi with a custom element constructor.
+ * Resolves the custom element tagName from the registry (via customElements.getName) or defaults to `a2ui-${api.name.toLowerCase()}`.
```
josemontespg, line 23: Also mention that calling this function registers the web component in the customElements registry

josemontespg: Updated the JSDoc description to state that calling this function registers the web component in the `customElements` registry if not already defined.

renderers/web_core/src/v0_9/universal/create_web_component_implementation.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  it('creates a WebComponentImplementation pairing ComponentApi with element and resolves tagName from getName', () => {
+    const originalCustomElements = (globalThis as any).customElements;
+    (globalThis as any).customElements = {
+      getName: (cls: any) => (cls === MockElement ? 'custom-sample-widget' : undefined),
+      get: (tag: string) => (tag === 'custom-sample-widget' ? MockElement : undefined),
+      define: () => {},
+    };
+
+    try {
```
josemontespg: Don't use try catch in tests. To handle cleanup, use beforeEach and afterEach. Same below.

josemontespg: Refactored test suite to manage `customElements` mocking and restoration via `beforeEach` and `afterEach` hooks, removing all `try/finally` blocks.

renderers/web_core/src/v0_9/basic_catalog/components/AudioPlayer.ts
```diff
@@ -63,7 +63,7 @@ export class A2uiAudioPlayerElement extends BasicCatalogA2uiLitElement<typeof Au
...
 }
 
-export const A2uiAudioPlayer: WebComponentImplementation = {
-  ...AudioPlayerApi,
-  tagName: 'a2ui-audioplayer',
-};
+export const A2uiAudioPlayer = createWebComponentImplementation(
+  AudioPlayerApi,
+  A2uiAudioPlayerElement,
+);
```
ditman: (Same comment as in the previous entry, maybe it's enough to export createWebComponentImplementation and isWebComponentImplementation to users of web_core (the catalog implementers) and they can create this glue over there? 

Or maybe all catalogs are going to use these A2uiAudioPlayer implementation instead of the Element? In that case, do we need to re-export the A2uiAudioPlayerElement custom element implementation?

My hunch is that we don't need to export both the WebComponentImplementation and the BasicCatalogA2uiLitElement classes?

renderers/web_core/src/v0_9/basic_catalog/components/Card.test.ts
```diff
@@ -25,14 +25,15 @@ import {
...
+import {type A2uiWebComponentElement, registerUniversalElement} from '../../universal/index.js';
 
 describe('Card Component', () => {
   let basicCatalog: Catalog<ComponentApi>;
 
   before(async () => {
     setupTestDom();
     basicCatalog = (await import('../index.js')).basicCatalog;
+    basicCatalog.components.forEach(c => registerUniversalElement(c as any));
     await import('./Card.js');
```
josemontespg: Now we can remove these async imports, because we removed the side effect of importing it. Apply to all tests for these component files.

renderers/web_core/src/v0_9/universal/register_universal_element.test.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+import {describe, it, beforeEach, afterEach} from 'node:test';
+import * as assert from 'node:assert';
+import {z} from 'zod';
+import {registerUniversalElement} from './register_universal_element.js';
+import type {WebComponentImplementation} from './web_component_implementation.js';
+
+describe('registerUniversalElement', () => {
+  class MockElementA extends (typeof HTMLElement !== 'undefined' ? HTMLElement : Object) {}
```
josemontespg: This is weird. Define MockElementA as a class extending HTMLElement directly.

renderers/web_core/src/v0_9/index.ts
```diff
@@ -24,6 +24,7 @@
 export * from './catalog/function_invoker.js';
 export * from './catalog/types.js';
 export * from './common/events.js';
+
```
josemontespg: Remove empty line

renderers/web_core/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## Unreleased
 
 - **BREAKING CHANGE**: (v0_9) Move Universal Web Component symbols (`A2uiLitElement`, `A2uiController`, `renderA2uiNode`, `WebComponentImplementation`, `ResolvedChildList`, `A2uiChildRef`, `ResolvedChildRef`) from root `@a2ui/web_core/v0_9` to `@a2ui/web_core/v0_9/universal`. Recommended migration: update import paths from `@a2ui/web_core/v0_9` to `@a2ui/web_core/v0_9/universal`. [#2488](https://github.com/a2ui-project/a2ui/pull/2488)
+- (v0_9) Add `A2uiWebComponentElement`, the DOM contract for an A2UI custom element: an `HTMLElement` that accepts a `context` property. [#2596](https://github.com/a2ui-project/a2ui/pull/2596)
```
josemontespg: These entries should go on top, because they are more recent.

josemontespg: Document the breaking change here. Now elements in the basic catalog are not registered statically in the customElements.

renderers/web_core/src/v0_9/basic_catalog/components/AudioPlayer.ts
josemontespg: Right on both counts, and that is what this revision does. The element classes and `BasicCatalogA2uiLitElement` are module-private now; `@a2ui/web_core/v0_9/basic_catalog` exports only the `A2uiAudioPlayer` implementation object and its `AudioPlayerApi` schema.

There is no `createWebComponentImplementation` either — an implementation is a plain object literal:

```ts
export const A2uiAudioPlayer: WebComponentImplementation<typeof AudioPlayerApi.schema> = {
  ...AudioPlayerApi,
  tagName: 'a2ui-audioplayer',
  element: A2uiAudioPlayerElement,
};
```

so catalog implementers have no glue to learn. What they do get is `registerUniversalElement`, which defines the element from that object.

renderers/web_core/src/v0_9/basic_catalog/components/Card.test.ts
josemontespg: Removed across all 18 component tests. `../index.js` already loads every component module, and registration is explicit on the line above, so these did nothing.

Typed `basicCatalog` as `Catalog<WebComponentImplementation>` in the same pass, which drops the `as any` on `registerUniversalElement`. `processor` and `surface` follow, since the generic is invariant through the event listener sets.

renderers/web_core/src/v0_9/universal/register_universal_element.test.ts
josemontespg: Done. `test/dom-setup.js` installs the JSDOM globals when it is imported, so the mocks extend `HTMLElement` directly and the `as CustomElementConstructor` casts are gone. The `customElements` stub stays, since each test re-registers the same tag name and a real registry rejects that.

renderers/web_core/src/v0_9/index.ts
josemontespg: Removed.

renderers/web_core/CHANGELOG.md
josemontespg: Moved the #2596 entries above the #2488 one, and dropped the stray blank line before the #2272 entry.

josemontespg: Documented:

> **BREAKING CHANGE**: (v0_9) Basic catalog components no longer define their custom element as a side effect of being imported. `renderA2uiNode` defines the element it is about to render, so surfaces rendered through it are unaffected; code that relied on the import alone must call `registerUniversalElement`.

renderers/web_core/src/v0_9/universal/render-a2ui-node.test.ts
```diff
@@ -34,16 +34,20 @@ describe('renderA2uiNode', () => {
   let surface: any;
   let testCatalog: Catalog<WebComponentImplementation>;
 
+  class MockButtonElement extends (typeof HTMLElement !== 'undefined' ? HTMLElement : Object) {}
```
josemontespg, line 37: extend HTMLElement directly here.

renderers/web_core/src/v0_9/universal/a2ui_web_component_element.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * The DOM contract an A2UI Custom Element fulfills.
+ *
+ * A host renderer creates the element for a `WebComponentImplementation` and
+ * assigns `context`; the element binds itself from there. Renderers use this
+ * type to set the property on an element they have only as an `HTMLElement`.
+ */
+export interface A2uiWebComponentElement extends HTMLElement {
+  context?: ComponentContext;
```
josemontespg, line 27: Add jsdoc to these properties, as this is part of the public API. Explain what they do and how they should be implemented.

renderers/lit/src/v0_9/surface/render-a2ui-node.ts
josemontespg, line 1: Let's keep this file, because Lit should not have to depend on web_core/universal just to render any component. You might need to register components here also

renderers/web_core/CHANGELOG.md
josemontespg: Done. Title and description updated. This PR was also split: it now only adds the universal element helpers (`element` field, `isWebComponentImplementation`, `registerUniversalElement`, `A2uiWebComponentElement`). The basic catalog lazy-registration refactor moved to #2698 (next in the stack).

renderers/web_core/CHANGELOG.md
josemontespg: Done. The CHANGELOG entries here link only #2596; the registration refactor entry lives in #2698.

renderers/web_core/src/v0_9/universal/a2ui_web_component_element.ts
josemontespg: Done. Added JSDoc for `context` and `updateComplete` on `A2uiWebComponentElement`.

renderers/web_core/src/v0_9/universal/render-a2ui-node.test.ts
josemontespg: Done. The mock is now `class MockButtonElement extends HTMLElement` with a module-level `setupTestDom()`.

renderers/lit/src/v0_9/surface/render-a2ui-node.ts
josemontespg: Done in #2698: kept `renderers/lit/src/v0_9/surface/render-a2ui-node.ts` and added the same `isWebComponentImplementation` → `registerUniversalElement` step there, so `a2ui-surface.ts` is no longer touched by this stack layer.

renderers/web_core/src/v0_9/universal/a2ui_web_component_element.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Resolves once the element has finished rendering its pending update.
+   *
+   * Lit elements provide this natively. Other implementations may expose it so
+   * hosts and tests can await a stable DOM after assigning `context`; hosts
+   * must not rely on it being present.
+   */
+  updateComplete?: Promise<boolean>;
```
josemontespg: Can this be removed?

josemontespg: Done. Removed it. Nothing reads `updateComplete` through this type; callers use the concrete Lit types or `as any`. The one test in #2698 that did (`ChoicePicker.test.ts`) now uses the `asyncUpdate` helper like the other component tests.

renderers/web_core/src/v0_9/universal/register_universal_element.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * This operation is idempotent: if the custom element is already registered with the
+ * identical constructor, this function safely no-ops. If a different constructor is
+ * already registered for the same tag name, this function throws an error.
+ *
+ * @param component The WebComponentImplementation defining the custom element.
+ */
+export function registerUniversalElement(component: WebComponentImplementation): void {
+  if (typeof customElements === 'undefined') {
+    return;
```
ditman, line 42: Consider emitting a console.error at some point in this return (or somewhere else that makes sense). This code hinges so hard on customElements that failing silently is probably not a great experience (nothing will work as expected, right?)

renderers/web_core/src/v0_9/basic_catalog/components/Tabs.ts
```diff
@@ -108,7 +108,8 @@ export class A2uiLitTabs extends BasicCatalogA2uiLitElement<typeof TabsApi> {
   }
 }
 
-export const A2uiTabs: WebComponentImplementation = {
+export const A2uiTabs: WebComponentImplementation<typeof TabsApi.schema> = {
```
ditman: I'm surprised that we're adding the schema type on this PR, it seems that should have been part of a prior PR? (the type was already defined like that)

ditman: Also, this typeof is a little bit sus, why do we need the type of the TabsApi **schema** here? (isn't that already available on the spread of ...TabsApi on the A2uiTabs object? (It should have an A2uiTabs.schema that matchs TabsApi's?)

ditman (approved):
Small question about the way we're passing schema types to the webcomponentimplementations (which maybe should be a separate PR). Anyway, this needs to land fast so we can reduce the churn of major versions!

renderers/web_core/src/v0_9/universal/register_universal_element.ts
josemontespg: Agreed. It now logs a `console.error` (once per process, so SSR does not spam) naming the tag it could not register, and still returns without throwing so `renderA2uiNode` keeps working in Node.

renderers/web_core/src/v0_9/basic_catalog/components/Tabs.ts
josemontespg: Fair, that was scope creep: the generic only narrows the declared type of `A2uiTabs.schema` and nothing consumes it. Reverted on all 18 components; they are back to `WebComponentImplementation` with just the `element` addition.

PR #2604, opened 2026-09-10
feat(angular): support markdown in universal basic catalog

josemontespg:
## Overview

Support markdown rendering in the Angular basic catalog when universal components are enabled.

## Details

- In `A2uiRendererService`, automatically resolve `MarkdownRenderer` from Angular DI during construction and wire it to `@a2ui/web_core`'s global `setMarkdownRenderer` function. This ensures all catalogs (standard, derived, or custom catalogs composing `BASIC_COMPONENTS`) automatically support markdown rendering without requiring catalog subclassing or custom lifecycle hooks.
- Safely coerce `text` values in Angular's `TextComponent` computed signal to string to prevent runtime errors when non-string values are bound.
- Align `MarkdownRendererOptions` with `@a2ui/web_core/v0_9` where `tagClassMap` is typed as `Record<string, string[]>`.

## Verification

- `yarn test:unit` passed (288 tests).
- `yarn test:integration` passed (208 tests including `Markdown Text Support`).
- Formatted with Prettier.


renderers/angular/src/v0_9/core/markdown.ts
```diff
@@ -15,10 +15,12 @@
...
 import {Injectable} from '@angular/core';
+import type {
+  MarkdownRenderer as MarkdownRendererFn,
+  MarkdownRendererOptions,
+} from '@a2ui/web_core/v0_9';
 
-export interface MarkdownRendererOptions {
-  tagClassMap?: Record<string, string>;
-}
+export type {MarkdownRendererOptions};
```
josemontespg: Don't reexport from web_core. Consumers can import from web_core directly

renderers/angular/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - (v0_9) Fix `ChoicePicker` radio groups colliding across surfaces: the radio group `name` now combines the surface id, component id, and data context path instead of using the surface-scoped component id alone, and checkboxes no longer receive a `name`. [#2447](https://github.com/a2ui-project/a2ui/issues/2447)
 - (v0_9) Implement `createComponentImplementation` helper and deprecate `extraComponents` and `functions` in `BasicCatalogOptions` to align with the core API. [#2060](https://github.com/a2ui-project/a2ui/pull/2060)
+- (v0_9) Align `MarkdownRendererOptions` with `@a2ui/web_core/v0_9`, typing `tagClassMap` as `Record<string, string[]>`.
```
josemontespg: Move this entry to the top of the unreleased section.

Also mark it as a breaking change, since we are now not going to be exporting MarkdownRendererOptions. Guide consumers to now import it from web_core

renderers/angular/src/v0_9/core/markdown.ts
josemontespg: Removed the export of `MarkdownRendererOptions` from `markdown.ts`. Consumers should import `MarkdownRendererOptions` directly from `@a2ui/web_core/v0_9`.

renderers/angular/CHANGELOG.md
josemontespg: Moved the entry to the top of the `## Unreleased` section and documented it as a breaking change guiding consumers to import `MarkdownRendererOptions` from `@a2ui/web_core/v0_9`.

renderers/angular/src/v0_9/core/markdown.ts
```diff
@@ -48,9 +50,7 @@ export class DefaultMarkdownRenderer extends MarkdownRenderer {
   }
 }
 
-export function provideMarkdownRenderer(
-  renderFn?: (markdown: string, options?: MarkdownRendererOptions) => Promise<string>,
-) {
+export function provideMarkdownRenderer(renderFn?: MarkdownRendererFn) {
```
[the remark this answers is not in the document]
josemontespg: Acknowledged. This PR is a minimal typing refactor and does not alter runtime behavior, so additional unit tests are omitted to keep the PR minimal.

renderers/angular/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- (v0_9) **BREAKING CHANGE**: Removed export of `MarkdownRendererOptions` from `@a2ui/angular/v0_9`. Consumers should now import `MarkdownRendererOptions` directly from `@a2ui/web_core/v0_9`. [#2604](https://github.com/a2ui-project/a2ui/pull/2604)
```
josemontespg: Also document the change in the function parameters

josemontespg: Updated the CHANGELOG entry to also document that `provideMarkdownRenderer` parameter type was aligned with `MarkdownRenderer` from `@a2ui/web_core/v0_9`.

renderers/angular/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+- (v0_9) Support markdown in universal basic catalog:
+  - Add optional `initialize?(injector: Injector): void` lifecycle hook to `AngularCatalog`, invoked automatically by `A2uiRendererService` upon initialization.
+  - Implement `initialize` in `BasicCatalogBase` to bridge Angular DI `MarkdownRenderer` to `@a2ui/web_core`'s global markdown renderer when universal components are rendered.
```
josemontespg: This line is an internal implementation detail. Remove.

josemontespg: Removed the internal implementation detail line from the changelog entry.

renderers/angular/src/v0_9/catalog/basic/text.component.ts
```diff
@@ -145,7 +145,12 @@ export class TextComponent extends BasicCatalogComponent<typeof TextApi> {
   ]);
 
   readonly variant = computed(() => this.props()['variant']?.value() || 'body');
-  readonly text = computed(() => this.props()['text']?.value() || '');
+  // Dynamic property bindings (e.g., binding { path: 'quantity' } to text where quantity is numeric)
+  // can resolve to numbers or non-string values. Safely coerce to string before markdown rendering.
+  readonly text = computed(() => {
+    const val = this.props()['text']?.value();
+    return typeof val === 'string' ? val : val != null ? String(val) : '';
```
ditman, line 152: No need to skip the coercion if `val` is already a String, let the browser optimizations do that. I'd just do:

```
return val !== null ? String(val) : ''
```

Or maybe even?

```
return this.props()['text']?.value()?.toString() ?? ''
```

renderers/angular/src/v0_9/core/a2ui-renderer.service.spec.ts
```diff
@@ -58,9 +60,54 @@ describe('A2uiRendererService', () => {
...
+      const result = await registeredFn!('# heading', {tagClassMap: {h1: ['custom-h1']}});
+      expect(result).toBe('<p>rendered</p>');
+      expect(mockRenderer.render).toHaveBeenCalledWith('# heading', {
+        tagClassMap: {h1: ['custom-h1']},
+      });
+    });
+
+    it('should do nothing to markdown renderer when MarkdownRenderer is not in the injector', () => {
+      expect(getMarkdownRenderer()).toBeUndefined();
+    });
```
ditman, line 110: This test is not very useful, because beforeEach is setting markdown renderer to undefined, and we're not exercising any code right before this expectation that might mutate the markdown renderer, right?

Either do something that might do things with the markdownRenderer (like actually attempting to render something) before the expectation, or remove the test?

renderers/angular/src/v0_9/core/a2ui-renderer.service.ts
```diff
@@ -74,6 +76,10 @@ export class A2uiRendererService implements OnDestroy {
   constructor() {
     const injector = inject(Injector);
     initializeAngularReactivity(injector.get(EnvironmentInjector));
+    const markdownRenderer = injector.get(MarkdownRenderer, null);
+    if (markdownRenderer) {
+      setMarkdownRenderer((markdown, options) => markdownRenderer.render(markdown, options));
+    }
```
ditman, line 82: We're setting the amrkdownRenderer regardless of us using universal components or not. Should we skip setting the web_core markdown renderer if the app is not using universal components?

ditman (approved):
Small comments, but in general this is very straightforward and uses the APIs we've defined as expected.

renderers/angular/src/v0_9/catalog/basic/text.component.ts
josemontespg: Done, switched to `this.props()['text']?.value()?.toString() ?? ''`.

renderers/angular/src/v0_9/core/a2ui-renderer.service.spec.ts
josemontespg: Agreed. Replaced it with a test that seeds a web_core renderer first, constructs a fresh service without `MarkdownRenderer` provided, and asserts the seeded renderer is untouched.

renderers/angular/src/v0_9/core/a2ui-renderer.service.ts
josemontespg: Good call. It is now gated on `useUniversalComponents`: native Angular `TextComponent` injects `MarkdownRenderer` directly, so web_core's global renderer only matters when universal elements render. Added a test for the disabled case and updated the changelog.

renderers/angular/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Unreleased
 
+- (v0_9) Support markdown in universal basic catalog:
+  - When `useUniversalComponents` is enabled, `A2uiRendererService` bridges `MarkdownRenderer` from Angular DI to `@a2ui/web_core`'s global markdown renderer.
+  - **BREAKING CHANGE**: Removed export of `MarkdownRendererOptions` from `@a2ui/angular/v0_9`. Consumers should now import `MarkdownRendererOptions` directly from `@a2ui/web_core/v0_9`. Additionally, `provideMarkdownRenderer` now accepts `MarkdownRenderer` from `@a2ui/web_core/v0_9` (where `tagClassMap` is typed as `Record<string, string[]>`). [#2604](https://github.com/a2ui-project/a2ui/pull/2604)
```
ditman, line 5: I think as part of landing these universal component changes we'll need to update the docsite, or prepare a migration guide, so users can see very clearly what changes they need to do to their apps (maybe we end up updating some sample app to accomodate all these changes first?)

Also, since this is a breaking change already, it'd be a *great* time to remove the automagic markdown renderer fallback to auto-import a2ui/markdown-it as an optional peer dependency, and ask users to bring their own?

ditman (approved):
I think this is still good! But I think we need to start thinking of preparing some sort of migration guide for all the breaking changes we're bringing to v0.12 or whatever it ends up being called!

renderers/angular/CHANGELOG.md
josemontespg: Regarding a migration guide: I agree we should do this, but currently there is not a defined way to organize these migration guides. I think it would be nice to have something like the Angular migration guides https://angular.dev/update-guide. 

On the other hand, now with agents I suspect that dependents won't be migrating by hand in most cases, and maybe these changelog notes on their own are good enough.

About removing the markdown automagic: I attempted this, but the PR got too large, and since it is unrelated to the universal refactor I kept this as is. After I'm done with the universal components I would like to move the basic catalog to its own npm package. I think then it will be the right time to rethink the markdown dependency.

PR #2283, opened 2026-08-14
feat(react): add support for rendering universal components

josemontespg:
The v0_9 React renderer dispatches on what a catalog entry is, with no configuration. A React implementation renders through React; a W3C Custom Element renders as an element with its `ComponentContext` attached. One catalog can hold both kinds, nested in either order.

## How it works

`NodeView` asks the implementation what it is:

```tsx
if (isReactComponentImplementation(impl)) {
  // render through `view`, or fall back to `render`
}
if (isWebComponentImplementation(impl)) {
  // render <tagName> and set .context
}
```

React implementations are checked first because they now carry a `tagName` too.

## Why React components get a custom element wrapper

web_core resolves a node's children by tag name: [render-a2ui-node.ts](https://github.com/a2ui-project/a2ui/blob/main/renderers/web_core/src/v0_9/universal/render-a2ui-node.ts) looks up `implementation.tagName` and creates that element. It never calls back into React. A React implementation without a tag name is therefore unreachable below a Web Component parent.

So `createComponentImplementation` and `createBinderlessComponentImplementation` register a custom element wrapper as they build the implementation, and return the implementation carrying its `tagName`. Registering at creation, rather than on first render, keeps the catalog complete before anything paints.

This PR refactors component rendering onto Web Components and depends on the basic catalog implementation in `@a2ui/web_core/v0_9/basic_catalog`.

## API

- Added: `ReactCatalogComponent`, the union of a React implementation and a Web Component implementation. `A2uiSurface` and the node layer are typed against it.
- Changed: both factories return a `tagName`.
- No new configuration. There is no rendering mode and no provider.

## Testing

`tests/v0_9/universal-components.test.tsx` builds a catalog holding a React `Badge`, a React `Panel`, and web_core's Lit `Column`, and covers: a React implementation rendering through its view, a Web Component rendering as its element with `.context` set, a Web Component under a React parent, and a React component under a Web Component parent.

PR #2274, opened 2026-08-14
feat(angular): showcase universal components in the angular explorer sample

josemontespg:
## Summary

Demonstrates how developers can build native Angular container components in A2UI that host both native Angular and universal Web Component children within the same A2UI surface, and adds dynamic query parameter toggling (`useUniversalComponent`) for testing both rendering modes.

### Key Changes
- **Native Container Component (`CustomGridComponent`)**: Implemented a 4-cell responsive grid container in `@a2ui/angular-explorer` using `@Component`, `CatalogComponentBase`, and `<a2ui-v09-component-host>` to instantiate child slots dynamically.
- **Example JSON (`37_native-grid.json`)**: Showcases `CustomGrid` hosting a mix of:
  - Native Angular components: `CustomSlider` with reactive two-way data binding mapped to independent paths (`/sliderValue1` and `/sliderValue2`).
  - Universal Web Components: `Card`, `Column`, `Text` (with dynamic `${/path}` string interpolation), and `Button` (with client action dispatching).
  - This example is placed in the core `specification/v0_9/catalogs/basic/examples` directory for cross-framework compatibility.
- **Dynamic Universal Mode QueryParam**:
  - Added `useUniversalComponent` URL query parameter support to both the **A2UI Angular Explorer** (`renderers/angular/a2ui_explorer`) and the **Restaurant Finder Sample App** (`samples/client/angular/projects/restaurant`).
  - Defaults to `false` (native Angular components) when absent, and enables W3C universal Web Components when `?useUniversalComponent=true`.
- **Test Suite & Verification**:
  - Unit tests for `CustomGridComponent`.
  - Integration test suite (`37_native-grid.spec.ts`) verifying rendering, mixed child instantiation, and reactive data updates.
  - **Manual Verification**: Manually verified that both sample applications (A2UI Angular Explorer and Restaurant Finder) function as intended in both native and universal modes.

## Pre-launch Checklist
- [x] I read the [Contributors Guide].
- [x] I read the [Style Guide].
- [x] My code changes have tests.

Related to #1270

PR #2615, opened 2026-09-11
refactor(react): extract NodeView and RenderFallback from A2uiSurface

josemontespg:
## Overview

Extracts `RenderFallback` and `NodeView` from `A2uiSurface.tsx` into `node-view.tsx`.

## Details

- Moves `RenderFallback` and `NodeView` into `node-view.tsx`, exporting `NodeView` for `A2uiSurface` while keeping `RenderFallback` module-private.
- Decouples the surface container/resolver lifecycle in `A2uiSurface.tsx` from the recursive node rendering logic in `NodeView`.
- Moves the module documentation with the code: the `A2uiSurface.tsx` header now covers only the resolver lifecycle and root rendering, and `node-view.tsx` gains a module doc describing the node rendering layer.
- Mechanical refactor with zero behavior or type changes; only comments and symbol locations move.


renderers/react/src/v0_9/node-view.tsx
```diff
@@ -307,3 +308,74 @@ export function useNodeView(
   }
   return {viewProps, context, viewBuildChild, rawBuildChild};
 }
+
+/** Renders an implementation that has no `view`: its wrapper binds itself. */
+export const RenderFallback: React.FC<{
+  node: ComponentNode<ReactComponentImplementation>;
+  impl: ReactComponentImplementation;
+  buildChild: NodeBuildChild;
+}> = ({node, impl, buildChild}) => {
```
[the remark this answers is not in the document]
josemontespg: Done. Removed the `export` keyword so `RenderFallback` remains private to `node-view.tsx`.

renderers/react/src/v0_9/A2uiSurface.tsx
```diff
@@ -26,95 +26,10 @@
...
-import {
-  type ComponentNode,
-  isComponentNode,
-  NodeResolver,
-  effect,
-  getValue,
-  peekValue,
-  type SurfaceModel,
-} from '@a2ui/web_core/v0_9';
+import React, {useCallback, useMemo, useSyncExternalStore} from 'react';
```
josemontespg, line 27: [PR Hygiene] Module doc still describes the dispatch logic that moved out

The header block above still says the surface "hands each `view` its node and a `buildChild` that renders resolved child nodes". That contract now lives in `node-view.tsx`, so the file that no longer implements it is the one documenting it.

Trim the header down to what `A2uiSurface` still owns (constructing and disposing the `NodeResolver`, subscribing to the root node, rendering `NodeView` under `NodeSurfaceContext`) and move the dispatch paragraph to a module doc on `node-view.tsx`. Docs that follow the code are the main reason a mechanical move like this is worth doing.

renderers/react/src/v0_9/A2uiSurface.tsx
```diff
@@ -26,95 +26,10 @@
...
-    }
-    const View = impl.view;
-    if (!View) {
-      return <RenderFallback node={node} impl={impl} buildChild={buildChild} />;
-    }
-    return <View node={node} buildChild={buildChild} />;
-  },
-);
-NodeView.displayName = 'NodeView';
+import {LoadingPlaceholder, NodeSurfaceContext, NodeView} from './node-view';
```
josemontespg, line 30: [PR Hygiene] React-only refactor is stacked on an Angular explorer sample

The base branch is `angular-custom-grid-sample` (#2274, an Angular explorer sample). Nothing in this diff depends on it, so the parent imposes an ordering constraint with no technical basis: this cannot land until an unrelated Angular change lands, and any rework or revert there drags this along. It also inverts the layer order, putting a renderer-internal refactor above a consumer-app change.

Reparenting onto `main`, or onto the nearest React layer if the stack has one, would let this merge on its own. Raising it as feedback only, no restack without your call.

renderers/react/src/v0_9/node-view.tsx
```diff
@@ -307,3 +308,74 @@ export function useNodeView(
...
+  // conversion puts in view props, so it resolves through the raw-id map.
+  const {context, rawBuildChild} = useNodeView(node, buildChild);
+  const Render = impl.render;
+  if (!context) {
+    return <LoadingPlaceholder componentId={node.componentId} />;
+  }
+  return <Render context={context} buildChild={rawBuildChild} />;
+};
+
+export const NodeView = memo(
```
josemontespg, line 343: [Public API Boundaries] `node-view.tsx` now carries five separate concerns

After the move this module holds the surface context, the loading and unresolved-reference placeholders, `useSignalValue`, the child-index machinery behind `useNodeView`, and now the recursive `NodeView`/`RenderFallback` renderer: 382 lines with no module doc stating what belongs here. The extraction cleans up `A2uiSurface.tsx`, but the receiving file drifts toward a grab-bag, which is how these modules become untestable in isolation.

Either split works: keep the component here and move `useNodeView` plus the child-index internals to their own module, or leave the layout and add a module doc at the top of the file defining its responsibility, so the next addition has a rule to follow.

renderers/react/src/v0_9/A2uiSurface.tsx
josemontespg: This is a valid comment but let's keep it in the PR stack for now because it's easier to work with a linear change for the universal components feature

renderers/react/src/v0_9/A2uiSurface.tsx
josemontespg: Done. The header now covers only what the surface still owns (resolver lifecycle, root subscription, rendering `NodeView` under the context) and the dispatch paragraph moved into a new module doc on `node-view.tsx`.

renderers/react/src/v0_9/node-view.tsx
josemontespg: Added a module doc stating that this file is the node rendering layer and that surface lifecycle concerns stay in `A2uiSurface.tsx`. Splitting `useNodeView` and the child-index helpers into their own module is a separate mechanical move; I would rather not grow this layer, since #2283 already rewrites both files.

PR #2633, opened 2026-09-11
Conformance tests to prepare for Dart agent SDK.

polina-c:
Contributes to https://github.com/a2ui-project/a2ui/issues/2356

Extracted from https://github.com/a2ui-project/a2ui/pull/2408

Splits the agent conformance data along the APIs that `a2ui_agent.blueprint.md` defines, so each suite covers one module instead of three sharing a file. Four new suites under `conformance/agent/`, two new case shapes in `conformance_schema.json`, and the Python and Kotlin harnesses wired to read them.

Conformance data and harness wiring only. No SDK source changes, in any language. Python and Kotlin run every case they ran before, plus the new pruning cases.

### DIFF
--- a/agent_sdks/python/a2ui_agent/tests/conformance/test_conformance.py
+++ b/agent_sdks/python/a2ui_agent/tests/conformance/test_conformance.py
@@ -276,7 +276,11 @@ def test_validator_conformance(name, test_case):
 
 
 # --- Catalog Conformance ---
-cases_catalog = get_conformance_cases("core/catalog.yaml")
+# `agent/catalog_transformer.yaml` holds the pruning cases that belong to the
+# agent SDK rather than to core; both suites drive the same `prune` action.
+cases_catalog = get_conformance_cases("core/catalog.yaml") + get_conformance_cases(
+    "agent/catalog_transformer.yaml"
+)
 
 
 @pytest.mark.parametrize(
@@ -329,7 +333,12 @@ def test_catalog_conformance(name, test_case):
 
 
 # --- Schema Manager Conformance ---
-cases_schema_manager = get_conformance_cases("agent/inference_format.yaml")
+# `agent/catalog_provider.yaml` holds the catalog loading cases, which used to
+# sit in `inference_format.yaml`; both suites drive the same `load_catalog`
+# action.
+cases_schema_manager = get_conformance_cases(
+    "agent/inference_format.yaml"
+) + get_conformance_cases("agent/catalog_provider.yaml")
 
 
 @pytest.mark.parametrize(
--- a/conformance/agent/catalog_provider.yaml
+++ b/conformance/agent/catalog_provider.yaml
@@ -1,0 +1,80 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Behaviour of the catalog providers described by
+# `blueprints/modules/a2ui_agent.blueprint.md` section 3.F: turning a catalog
+# document into a `Catalog`, whatever backing store it came from. Prompt
+# rendering and capability negotiation are covered by `inference_format.yaml`
+# and `catalog_resolver.yaml` respectively.
+#
+# These cases load the published basic catalog schema from
+# `specification/v0_9_1/catalogs/basic/catalog.json` rather than any SDK's own
+# catalog, so implementations are measured against the same document. Paths are
+# resolved relative to the `conformance/` directory.
+
+- name: test_load_basic_catalog_v0_9
+  description: Loads the published v0.9 basic catalog document from disk.
+  action: load_catalog
+  catalog_configs:
+    - name: basic
+      path: "../specification/v0_9_1/catalogs/basic/catalog.json"
+  expect:
+    supported_catalog_ids: ["https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"]
+
+# Relocated from `inference_format.yaml`: these exercise
+# `CatalogProvider.load`, not an inference format. See the API-to-suite table
+# in section 6 of `blueprints/modules/a2ui_agent.blueprint.md`.
+
+- name: test_manager_with_modifiers
+  description: Tests that A2uiSchemaManager applies modifiers during loading.
+  catalog_configs:
+    - name: test_strict
+      path: "test_data/load_examples/schema_with_strict/catalog.json"
+  modifiers: ["remove_strict_validation"]
+  action: load_catalog
+  expect:
+    catalog_schema:
+      catalogId: "test_strict"
+      components:
+        Text:
+          type: object
+          properties:
+            text: {type: string}
+
+- name: test_manager_no_modifiers
+  description: Tests that A2uiSchemaManager works fine without modifiers.
+  catalog_configs:
+    - name: test_strict
+      path: "test_data/load_examples/schema_with_strict/catalog.json"
+  action: load_catalog
+  expect:
+    catalog_schema:
+      catalogId: "test_strict"
+      components:
+        Text:
+          type: object
+          properties:
+            text: {type: string}
+          additionalProperties: false
+
+- name: test_schema_manager_init_custom_catalog
+  description: Tests that A2uiSchemaManager can load a custom catalog from path.
+  action: load_catalog
+  catalog_configs:
+    - name: standard
+      path: "test_data/schema_manager/catalog.json"
+    - name: Custom
+      path: "test_data/schema_manager/custom_catalog.json"
+  expect:
+    supported_catalog_ids: ["basic", "Custom"]
--- a/conformance/agent/catalog_resolver.yaml
+++ b/conformance/agent/catalog_resolver.yaml
@@ -1,0 +1,193 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Behaviour of `resolve_catalogs`, the capability negotiation helper described
+# by `blueprints/modules/a2ui_agent.blueprint.md` section 3.G. It matches an
+# `a2uiClientCapabilities` object against the catalogs an agent registered and
+# returns the active, transformed catalogs for the session.
+#
+# This replaces the legacy single-catalog `select_catalog` helper: negotiation
+# now yields a list, in agent preference order, and may include catalogs the
+# renderer supplied inline.
+#
+# `args.catalogs` are the registered `CatalogConfig`s in agent preference
+# order; each `catalog_schema` is either an inline catalog document or a path
+# resolved relative to the `conformance/` directory. `expect_active_catalog_ids`
+# asserts the ids of the resolved catalogs, in order.
+
+- name: test_resolve_catalogs_selects_the_declared_catalog
+  description: A renderer that declares one registered catalog id negotiates to it.
+  catalog:
+    version: "0.9"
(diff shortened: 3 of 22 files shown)


dart/a2ui_core/lib/src/processing/expressions.dart
```diff
@@ -235,7 +250,13 @@ class ExpressionParser {
...
       scanner.advance();
     }
-    return num.parse(scanner.input.substring(start, scanner.pos));
+    final String text = scanner.input.substring(start, scanner.pos);
+    // The grammar is spelled out here rather than delegated to the platform's
+    // number parser, so that every implementation accepts the same literals.
+    if (!_numberLiteral.hasMatch(text)) {
+      throw A2uiExpressionError("Invalid number literal: '$text'");
+    }
+    return num.parse(text);
```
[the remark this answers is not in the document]
polina-c: Not applying this one here. `dart/a2ui_core/lib/src/processing/expressions.dart` is not part of this PR — it was never in the diff, even before the Dart changes were reverted. The trailing-decimal case (`1.`) was addressed on main in #2491.

If `num.parse` still has the gap you describe after that fix, it is worth a separate issue against `a2ui_core` rather than a comment on a conformance-data PR.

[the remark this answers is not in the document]
polina-c: Correcting my previous reply: I cited #2491 as having fixed this, which is wrong — that PR made the recursion depth guard reachable and did not touch number parsing.

The reason not to apply this still stands, but it is a different one: the premise does not hold. `num.parse("1.")` returns `1.0` on the Dart VM, it does not throw.

```
Dart SDK 3.13.2
num.parse('1.') = 1.0
num.parse('12.') = 12.0
num.parse('1.5') = 1.5
```

That is also the documented intent at [expressions.dart:18-21](https://github.com/a2ui-project/a2ui/blob/main/dart/a2ui_core/lib/src/processing/expressions.dart#L18-L21) — `_numberLiteral` permits a trailing point on purpose, because every client implementation accepts `1.` today.

Either way this file is outside this PR: it is not in the diff and never was, so there is nothing to change here.

dart/a2ui_agent/lib/src/processor/catalog_providers.dart
```diff
@@ -0,0 +1,140 @@
...
+      throw A2uiCatalogError('Catalog file not found: $path');
+    }
+    final Object? decoded;
+    try {
+      decoded = jsonDecode(file.readAsStringSync());
+    } on FormatException catch (e) {
+      throw A2uiCatalogError(
+        'Catalog file $path is not valid JSON: ${e.message}',
+      );
+    }
```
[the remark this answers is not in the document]
polina-c: Nothing to apply here now. This PR no longer contains any Dart changes — `dart/a2ui_agent/lib/src/processor/catalog_providers.dart` was reverted along with the rest of the Dart agent SDK, so the file this comment points at does not exist on the branch.

The point itself is reasonable and worth carrying forward when the Dart agent SDK lands in its own PR: a provider parses a document it did not write, so an I/O failure should surface as `A2uiCatalogError` rather than a raw `FileSystemException`.

conformance/conformance_schema.json
```diff
@@ -572,6 +576,113 @@
...
+          "type": "array",
+          "items": {"type": "string"},
+          "description": "Catalog ids negotiated for the turn."
+        },
+        "expect_prompt_contains": {
+          "type": "array",
+          "items": {"type": "string"},
+          "description": "Substrings the generated system prompt snippet must contain."
+        },
+        "expect": {"type": "array", "description": "Expected response parts, in order."},
```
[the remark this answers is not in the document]
polina-c: Applied in f0fa5892. `ProcessRequestTest.expect` now carries the same items schema as `ProcessChunkTest`, so response parts validate as `{text, a2ui}` objects.

Confirmed the constraint actually bites rather than sitting inert:

```
expect item = bare string : REJECTED: 'oops' is not of type 'object'
expect item = {text,a2ui} : ACCEPTED
```

conformance/conformance_schema.json
```diff
@@ -572,6 +576,113 @@
...
+          "properties": {
+            "catalogs": {
+              "type": "array",
+              "description": "The catalog configurations the agent registered, in preference order.",
+              "items": {
+                "type": "object",
+                "properties": {
+                  "catalog_schema": {
+                    "description": "An inline catalog document, or a path to one relative to the conformance directory."
+                  },
```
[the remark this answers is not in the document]
polina-c: Applied in f0fa5892, matching the `oneOf` the root `catalog.catalog_schema` already uses.

```
catalog_schema = 42      : REJECTED: 42 is not valid under any of the given schemas
catalog_schema = path    : ACCEPTED
catalog_schema = inline  : ACCEPTED
```

All 16 conformance YAML files still validate.

conformance/agent/inference_format.yaml
```diff
@@ -12,7 +12,149 @@
...
+  description: Tests full prompt generation with capabilities and allowed components.
+  action: generate_prompt
+  args:
+    version: "0.8"
+    role_description: "Role"
+    include_schema: true
+    client_ui_capabilities:
+      supportedCatalogIds: ["https://a2ui.org/specification/v0_8/standard_catalog_definition.json"]
+
+    allowed_components: ["Text"]
```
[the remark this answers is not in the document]
polina-c: Removed in f0fa5892.

Worth noting for anyone reading the thread that the blank line predates this PR — `test_generate_system_prompt_full_with_caps` is unchanged here, it just moved position when the file was reorganised. Tidying it anyway since the file is already being restructured.

agent_sdks/python/a2ui_agent/tests/conformance/test_conformance.py
```diff
@@ -276,7 +276,11 @@ def test_validator_conformance(name, test_case):
 
 
 # --- Catalog Conformance ---
-cases_catalog = get_conformance_cases("core/catalog.yaml")
+# `agent/catalog_transformer.yaml` holds the pruning cases that belong to the
+# agent SDK rather than to core; both suites drive the same `prune` action.
+cases_catalog = get_conformance_cases("core/catalog.yaml") + get_conformance_cases(
```
nan-yu, line 281: Do we need to run test cases in `core/catalog.yaml` in the **agent** test runner? That should be covered by the **core** test runner.

conformance/agent/catalog_provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Behaviour of the catalog providers described by
+# `blueprints/modules/a2ui_agent.blueprint.md` section 3.F: turning a catalog
+# document into a `Catalog`, whatever backing store it came from. Prompt
+# rendering and capability negotiation are covered by `inference_format.yaml`
+# and `catalog_resolver.yaml` respectively.
+#
+# These cases load the published basic catalog schema from
```
nan-yu, line 21: Not sure if this was generated before https://github.com/a2ui-project/a2ui/pull/2559, but the basic catalog is no longer loaded from disk or bundled resources.

https://github.com/a2ui-project/a2ui/blob/v1_0/conformance/core/catalog.yaml#L679 sort of covers the round-trip conversions:
- Deserialization: loading from the BasicCatalog class
- Serialization: `catalog_schema` converts the Catalog object into JSON representation.

With that, I think we can remove this test case from the agent side.

conformance/agent/catalog_provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - name: basic
+      path: "../specification/v0_9_1/catalogs/basic/catalog.json"
+  expect:
+    supported_catalog_ids: ["https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"]
+
+# Relocated from `inference_format.yaml`: these exercise
+# `CatalogProvider.load`, not an inference format. See the API-to-suite table
+# in section 6 of `blueprints/modules/a2ui_agent.blueprint.md`.
+
+- name: test_manager_with_modifiers
```
nan-yu, line 39: The remove_strict_validation modifier feature has been removed from the new agent SDK because it breaks the agreement between the supported catalog on the agent and on the renderer side. Basically, it allows additional properties on the agent side but the renderer may not support it.

Let's remove the test case.

conformance/agent/catalog_provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+  expect:
+    catalog_schema:
+      catalogId: "test_strict"
+      components:
+        Text:
+          type: object
+          properties:
+            text: {type: string}
+
+- name: test_manager_no_modifiers
```
nan-yu, line 55: This can also be removed.

conformance/agent/catalog_provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+      catalogId: "test_strict"
+      components:
+        Text:
+          type: object
+          properties:
+            text: {type: string}
+          additionalProperties: false
+
+- name: test_schema_manager_init_custom_catalog
+  description: Tests that A2uiSchemaManager can load a custom catalog from path.
```
nan-yu, line 72: A2uiSchemaManager is deprecated and no longer available in the new agent SDK.

conformance/agent/catalog_resolver.yaml
```diff
@@ -0,0 +1,193 @@
...
+          catalogId: "https://example.com/small.json"
+          components:
+            Text: {type: object}
+    renderer_capabilities:
+      "v0.9":
+        supportedCatalogIds:
+          - "https://example.com/small.json"
+          - "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"
+  expect_active_catalog_ids:
+    - "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"
```
nan-yu, line 66: Does the v0.9 supportedCatalogId match the v0.9.1 catalogs?

conformance/agent/catalog_resolver.yaml
```diff
@@ -0,0 +1,193 @@
...
+    - "https://example.com/small.json"
+
+- name: test_resolve_catalogs_defaults_to_the_first_registered_catalog
+  description: A renderer that declares no catalog id gets the agent's first.
```
nan-yu, line 70: Per https://github.com/a2ui-project/a2ui/blob/main/specification/v1_0/docs/a2ui_protocol.md, `There is no fallback to the list of catalogs declared in rendererCapabilities (even if the renderer only advertises a single supported catalog).`.

conformance/agent/catalog_resolver.yaml
```diff
@@ -0,0 +1,193 @@
...
+# renderer supplied inline.
+#
+# `args.catalogs` are the registered `CatalogConfig`s in agent preference
+# order; each `catalog_schema` is either an inline catalog document or a path
+# resolved relative to the `conformance/` directory. `expect_active_catalog_ids`
+# asserts the ids of the resolved catalogs, in order.
+
+- name: test_resolve_catalogs_selects_the_declared_catalog
+  description: A renderer that declares one registered catalog id negotiates to it.
+  catalog:
```
nan-yu, line 31: Would it be possible to move the catalog protocol version into the `catalogs` list?

conformance/agent/inference_format.yaml
```diff
@@ -12,7 +12,148 @@
...
+# the end of this file stay put: that helper is frozen, not removed, and the
+# shipped Python and Kotlin agent SDKs still run against them. See the Legacy
+# Surfaces table in section 6.
+
+- name: test_generate_system_prompt_minimal
+  description: Tests prompt generation with minimal arguments.
+  action: generate_prompt
+  args:
+    version: "0.8"
+    role_description: "Just Role"
```
nan-yu, line 29: The new prompt generator API no longer needs the role description and workflow description sections. Before, it generates the full prompt with user provided role description and workflow description. It is not flexible enough. With the re-design, we've decided to let the generator only generate the prompt snippet, which only includes the catalog schema part.

kotlin/agent_sdk_legacy/src/test/kotlin/com/google/a2ui/conformance/ConformanceTest.kt
```diff
@@ -294,9 +294,14 @@ class ConformanceTest {
 
   @TestFactory
   fun testCatalogConformance(): List<DynamicTest> {
-    val conformanceFile = ConformanceTestHelper.getConformanceFile(CATALOG_YAML_FILE)
     val conformanceDir = ConformanceTestHelper.getConformanceDir()
-    val rawList = yamlMapper.readValue(conformanceFile, Any::class.java) as List<*>
+    // `catalog_transformer.yaml` holds the pruning cases that belong to the
+    // agent SDK rather than to core; both suites drive the same `prune` action.
+    val rawList =
+      listOf(CATALOG_YAML_FILE, CATALOG_TRANSFORMER_YAML_FILE).flatMap { file ->
```
nan-yu, line 301: This is the legacy kotlin SDK, and we're not planning to update this version to support the new APIs, e.g. catalog_transformer. Can we ignore test cases in `catalog_transformer.yaml`?

nan-yu (commented):
Thanks for adding test cases for the agent. 

Some of those test cases are generated with the legacy interfaces. We'd like to update them to follow the new API design.

PR #2634, opened 2026-09-11
docs(blueprints): align a2ui_core blueprint with implemented core SDK APIs

nan-yu:
Updates `blueprints/modules/a2ui_core.blueprint.md` so the prescribed API matches what the Python and TypeScript core SDKs actually implement. Several sections described types and methods that exist in neither SDK, which makes the blueprint unusable as a reference for new language implementations.

## Version adapters

The section declared `InboundOperation`, `OutboundOperation`, `extractInboundOperations()`, and `extractOutboundOperations()`. None of these exist. The real contract is a single inbound `extractOperations()` returning `InternalOperation[]`, in both `python/a2ui_core/src/a2ui/core/processing/adapters/base.py` and `typescript/web_core/src/processing/adapters/base.ts`.

The section now documents `version`, `validActions`, `compatibleCatalogVersions`, `isCatalogCompatible()`, and `extractOperations()`, plus the `VersionAdapterResolver` contract that replaces the single-method static factory. It also states that adapters are inbound-only and names where outbound envelopes are actually built, so the absence of an outbound path reads as a decision rather than an omission.

Two behaviors that both base classes implement but the blueprint never described are now written down: payload unwrapping across single messages, arrays, and `{messages: [...]}` wrappers; and the exactly-one-action rule, including the cross-version error path that reports an action as unsupported in this version rather than missing.

## New section for internal operations

`InternalOperation` is the vocabulary every downstream consumer is written against, so it now has its own section instead of being buried in the adapter contract. It documents the six-member union with per-operation fields, the discriminator list, and a table mapping each operation to the action key that produces it in each protocol version.

That table records something the previous text got wrong: v0.8 names three of its actions differently (`beginRendering`, `surfaceUpdate`, `dataModelUpdate`), so the discriminator does not universally match the wire action. Absorbing that renaming is the clearest example of what the operation layer is for.

## Other corrections

- `RpcHandler` was documented with a `registerOutboundListener(listener: (operation: OutboundOperation) => void)` method and a `catalogResolver` constructor option. Neither exists. Both SDKs take `catalogs` plus a constructor-injected `outboundListener` receiving a `RendererFunctionResponseMessage | CallAgentFunctionMessage`.
- The validation matrix referenced `VersionAdapter.extractInboundOperations()`; renamed to `extractOperations()`.
- Added the `operations` module to the directory tree in section 2.

## Verification

`python3 blueprints/validate_blueprints.py` passes. Documentation only; no source changes.


blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -474,71 +714,207 @@ agentProcessor.processMessages(generatedLlmPayloadMessages);
...
+```
+
+#### RPC Execution Lifecycle & Safeguards
+
+1. **Inbound Call Execution (`handleCallRendererFunction`)**:
+   - **Catalog Resolution Order**: Resolve the target catalog in this order, stopping at the first hit:
+     1. The explicit `catalogId` on the call, looked up in the surface's `availableCatalogs`.
+     2. The explicit `catalogId`, looked up in the processor's global catalog list.
+     3. The surface's `defaultCatalog`.
+     4. Failing all of the above, return `RpcErrorCode.INVALID_FUNCTION_CALL`.
```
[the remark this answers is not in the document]
nan-yu: fixed

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -720,9 +1128,14 @@ class DataModel {
...
+3.  **Traversing an Existing Primitive Raises**: If an intermediate segment already holds a primitive, `set` MUST raise `A2uiDataError` naming the path. Replacing that primitive with a fresh container to continue the write destroys committed application data with no error and no event, so the loss is only discovered later when a read returns the wrong shape.
+4.  **Forbidden Keys**: Reject any path containing `__proto__`, `constructor`, or `prototype` with `A2uiDataError`. The check is normative in every language, not only those with prototype chains, so that a payload behaves the same everywhere.
+5.  **Notification Strategy (Bubble & Cascade)**: Notify exact matches, bubble up to all parent paths, and cascade down to all nested descendant paths.
+6.  **Undefined Handling**: Setting an object key to `undefined` removes the key. Setting an array index to `undefined` preserves length but empties the index (sparse array).
+7.  **Null Write to a Missing Path**: If the path does not exist, writing `null` is a no-op. It creates no parents and fires no notification. Auto-vivifying parents only to delete the leaf leaves behind empty containers the payload never asked for.
```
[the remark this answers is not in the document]
nan-yu: rephrased

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -30,6 +30,8 @@ Its core responsibilities include:
 6. **Validation:** Performs structural JSON Schema checks, reference checks, loop/recursion analysis, and layout integrity checks.
 7. **Resolution:** Resolves bound context paths and binds state variables to components for local evaluation.
 8. **Multi-Version Protocol Branching:** Supports multiple versions of the protocol.
+9. **Multi-Catalog Surface Resolution (v1.0+):** Tracks every catalog a surface may draw from, and resolves per-component and per-function `catalogId` overrides against that set rather than against a single surface-wide default.
+10. **Bidirectional RPC Execution (v1.0+):** Coordinates asynchronous and synchronous remote function calls between renderers and server agents via `RpcHandler` (only available starting from version 1.0), managing correlation IDs, timeouts, cancellation callbacks, permission boundaries, and lifecycle disposal.
```
polina-c: Can we link descriptions for these features?

Add something like this at the end of each section:

See [Bidirectional RPC Execution](<link>).

polina-c (approved):
LGTM with one comment.

blueprints/modules/a2ui_core.blueprint.md
nan-yu: Thanks. Added links to all the features.

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -387,9 +485,20 @@ class MessageProcessor<T extends ComponentApi> {
...
 }
 ```
 
+##### RPC ownership
+
+The RPC lifecycle belongs to [`RpcHandler`](#d-bidirectional-remote-procedure-call-rpc-layer-a2uicorerpc), reachable as `processor.rpc`. `MessageProcessor` does not re-expose the inbound handlers.
+
+Inbound `callRendererFunction` and `agentFunctionResponse` envelopes arrive through `processMessages` like any other message. The processor recognizes them, builds the `DataContext` for the target surface (or a headless one if there is no surface), and forwards them to `RpcHandler`. Consumers never call the inbound handlers directly, so publishing them on the processor would advertise an entry point that bypasses envelope validation and context construction.
+
+`callAgentFunction` is the exception, and is deliberately mirrored on the processor. It is outbound, callers already hold the processor, and requiring them to reach through `processor.rpc` for the one call they routinely make buys nothing. It forwards its arguments unchanged.
```
gspencergoog: This might need a little expansion/clarification.  Why does it buy nothing?  Why might you try to do that?

blueprints/modules/a2ui_core.blueprint.md
```diff
@@ -474,71 +733,224 @@ agentProcessor.processMessages(generatedLlmPayloadMessages);
...
+2. **Relative Path Mapping**: The in-memory registry registers dynamic types under relative keys (e.g., `"common_types.json"`, `"catalog.json"`), resolving relative `$ref` targets (such as `"common_types.json#/$defs/DynamicString"`) without using external HTTP URIs.
+3. **No External Dependencies**: This eliminates runtime file system discovery and remote HTTP fetching, guaranteeing that core validation succeeds identically when packaged as a standalone library distribution (e.g. standalone pip or npm installs) or executed in offline evaluation runners.
+
+##### UAX #31 Identifier Validation
+
+To prevent malformed identifiers, injection vectors, and subtle rendering ambiguities across language platforms, v1.0 specifications mandate that all component names, property keys, and function names adhere to Unicode Standard Annex #31 (UAX #31: Default Identifier Syntax):
+
+- **Start Characters**: Must satisfy `is_xid_start(c)` or equal `'_'` (Unicode ID_Start plus underscores).
+- **Continue Characters**: Must satisfy `is_xid_continue(c)` (alphanumerics, connector punctuation, combining marks).
```
gspencergoog, line 808: It's probably good to indicate that this should be implemented using standard regex libraries so that it isn't implemented ad hoc. For example, in Python it should use the `regex` package, and then they all should use a Regex that is something like: `r'^\p{XID_Continue}$'`.

And also, the parser should make sure that it doesn't split surrogate pair characters incorrectly.  It should be splitting things on 32-bit code points.

(See my comments to Varun here, where he's fixing these bugs on main https://github.com/a2ui-project/a2ui/pull/2527 )

gspencergoog (approved):

blueprints/modules/a2ui_core.blueprint.md
nan-yu: Good catch, the original asserted the conclusion without the argument. Rewritten in 4764b2cc to answer both questions in order.

**Why you might try to hide it:** consistency. If RPC belongs to `RpcHandler`, it is reasonable to expect every RPC entry point to be reached the same way.

**Why that does not apply here:** the rule exists because the inbound handlers *can* be called wrongly. Invoking one directly skips the envelope validation and `DataContext` construction that `processMessages` performs first, so keeping them off the processor removes a real hazard. `callAgentFunction` has no such setup to skip. The processor method is a one-line pass-through (`message-processor.ts:199` is just `return this.rpc.callAgentFunction<TRes>(surfaceId, call, options);`), so `processor.callAgentFunction(...)` and `processor.rpc.callAgentFunction(...)` are the same call. Hiding it would oblige every caller to write `.rpc` and would rule out no incorrect usage.

blueprints/modules/a2ui_core.blueprint.md
nan-yu: Agreed, and thanks for the pointer to #2527 — I read your comments there and folded the specifics in rather than paraphrasing. Added in 4764b2cc:

- **Why not ad hoc.** `isalnum()` plus underscore is wrong in both directions: it admits characters UAX #31 excludes, and it rejects the combining marks and connector punctuation that XID_Continue allows. Hand-maintained code point ranges drift as the Unicode tables are revised, and they drift independently per SDK, which is how the same identifier ends up valid in one language and rejected in another.
- **Per-language patterns**, with the two traps you flagged: Python needs the `regex` package because stdlib `re` has no `\p{...}`, and TypeScript needs the `u` flag or `\p{...}` is not read as a Unicode property.

  | Language | Pattern |
  | :-- | :-- |
  | Python | `regex.compile(r"^\p{XID_Continue}$")` |
  | TypeScript | `/^\p{XID_Continue}$/u` |
  | Dart | `RegExp(r"^\p{XID_Continue}$", unicode: true)` |

- **Single code point guard** before matching, since the patterns are anchored single-character (`Array.from(s).length === 1`, `s.runes.length == 1`).
- **Scan by code point, not code unit.** A `peek` written with `charAt` or bare index access splits a surrogate pair into two halves, neither of which is a valid identifier character, so an identifier containing an astral character such as an emoji or a rare CJK ideograph is rejected even though UAX #31 admits it. `codePointAt` with `String.fromCodePoint` in TypeScript, runes in Dart; Python strings are already sequences of code points.

One deliberate omission: I did not link #2527 from the blueprint. It is still open, and a durable spec should not depend on a PR reference — the guidance stands on its own.

PR #2644, opened 2026-09-14
feat(catalogs/mcp): resolve UI resources and process A2UI messages inside callMcpTool

wrenj:
## What

`callMcpTool` previously handed the raw `CallToolResult` back to the host, which
had to decode it, fetch any UI template, and feed the messages to a
`MessageProcessor`. The recipe sample carried a 300-line `A2uiMcpEngine` to do
exactly that, and every other host would have had to write its own copy.

This moves that work into the catalog. The host now supplies one hook,
`getMcpClientForTool(toolName)`, plus the processor that should receive the
messages.

```diff
-createMcpCatalog(callMcpTool: McpToolCaller, onResult?: McpToolResultHandler)
+createCallMcpToolImplementation(getMcpClientForTool: McpClientResolver, processor: MessageProcessor)
```

On each call the function issues `tools/call` on the resolved client, then
collects A2UI messages from three routes:

1. A resource the result names, via `result._meta.ui.resourceUri`.
2. A resource the tool declares in `tools/list`, used when the result names
   none. Discovered once per client.
3. Messages inline in `result.content`, as an embedded resource block or a text
   block whose JSON decodes to a message or message list.

## Why

Decoding a tool result is protocol work, not application work. Keeping it in the
catalog means a host wires a resolver and a processor and is done, and the
`_meta.ui.resourceUri` convention has one implementation instead of one per
client.

## Changes

| Area | Change |
| :--- | :--- |
| `catalogs/mcp/v0_9/src/functions/callMcpTool.ts` | Resource resolution, caching, message extraction, and processing |
| `catalogs/mcp/v0_9/src/catalog.ts` → `index.ts` | `createMcpCatalog` dropped; module now re-exports the function and `MCP_CATALOG_ID` |
| `catalogs/mcp/README.md` | Rewritten wiring guide, including the lazy-catalog-array pattern and the bootstrap invocation |
| `samples/.../recipe/client/engine.ts` | Deleted, along with its tests; the app wires the catalog directly |
| `samples/.../recipe/client/app.ts` | Builds one composite Basic + MCP catalog, tracks `clientsByTool`, invokes the entrypoint tool against a scratch `DataContext` |
| `samples/.../recipe/client/app.test.ts` | New coverage for connection, routing, and entrypoint rendering |
| `vite.config.ts` | `vitest/config` with a `jsdom` environment |

## Testing

`callMcpTool.test.ts` was reworked around a fake MCP client and now asserts
against processor state: resource-created surfaces, data applied to a surface
the resource did not create, the caching path, declared-URI fallback, and
decode failures. The sample gains `app.test.ts`.


catalogs/mcp/v0_9/src/functions/callMcpTool.ts
```diff
@@ -14,69 +14,167 @@
...
+    }
+    return messages;
+  }
+
+  /**
+   * Returns the UI resource URI each tool declares, reading `tools/list` once
+   * per client.
+   */
+  function getDeclaredUiResourceUris(client: McpToolClient): Promise<Map<string, string>> {
+    let discovery = declaredUiResourceUris.get(client);
```
sugoi-yuzuru: WeakMap supports `.has`. 

Can we instead do 

```
if (declaredUiResourceUris.has(client)) {
  return declaredUiResourceUris.get(client);
}

const discovery = (async () => {
        const uris = new Map<string, string>();
        try {
          const {tools} = await client.listTools();
          for (const tool of tools ?? []) {
            const uri = readUiResourceUri(tool);
            if (uri) {
              uris.set(tool.name, uri);
            }
          }
        } catch (err) {
          console.warn('Could not query MCP tool UI resources:', err);
        }
        return uris;
      })();
      declaredUiResourceUris.set(client, discovery);
return discovery;
```

it reduces nesting and mildly improves the readability as discovery becomes `const`.

catalogs/mcp/v0_9/src/functions/callMcpTool.ts
```diff
@@ -14,69 +14,167 @@
  * limitations under the License.
  */
 
+/**
+ * `callMcpTool`: the A2UI catalog function that calls a Model Context Protocol
+ * tool and renders the A2UI messages the tool returns.
```
sugoi-yuzuru, line 19: Is there an inherent assumption that the response will include A2UI resource link? 

In my mind we expect 3 distinct MCP responses: 1) Tool response + resource pointer for A2UI payload, 2)  Tool response sans resource (e.g., A2UI datamodel update only), and 3) just a regular tool response such as text/number. 

Can we list the intended handling logic for these different types of responses in the docstring for ease-of-navigation?

sugoi-yuzuru (commented):
few comments about the scope of what utility `callMcpTool` catalog function is expected to cover.

catalogs/mcp/v0_9/src/functions/callMcpTool.ts
wrenj: Refactored 

getDeclaredUiResourceUris
 to an early return, so discovery is now a const and the body is un-nested.

One deviation from your snippet: I used get() + truthiness check instead of has() + get(). WeakMap.get() is typed Promise<...> | undefined, so the has() form would still need a non-null assertion (get(client)!) to typecheck. The single-lookup version avoids that while reading the same.

sugoi-yuzuru (approved):

PR #2646, opened 2026-09-14
fix(web_core): revert `A2uiLitElement.controller` to a writable property

josemontespg:
## Summary

Restores `A2uiLitElement.controller` as a writable public property, reverting the read-only getter introduced in version 0.11.0. This re-enables assigning mock controllers in unit tests and documents the controller's lifecycle binding invariant.

## Problem

In 0.11.0, `A2uiLitElement.controller` was converted from a property to a read-only getter over a private `_controller` field to prevent external reassignment.

However, external reassignment was already prevented by TypeScript type checking. In practice, the getter broke legitimate testing workflows:
1. **Unit tests could no longer inject mock controllers.** Tests that construct an element and assign `el.controller = mockController` failed with a runtime error:
   ```
   TypeError: Cannot set property controller of #<A2uiLitElement> which has only a getter
   ```
2. **Callers were forced to use private-field workarounds.** To test components, code had to bypass encapsulation with `(el as any)._controller = mockController`.
3. **Misleading type signature.** The getter returned `this._controller!`, returning `undefined` at runtime if accessed before the element was bound to a context, despite the type signature promising a non-null `A2uiController`.

## Solution

- **Restore writable property:** `A2uiLitElement.controller` is once again a public writable property (`public controller!: A2uiController<Api>;`), and the private `_controller` backing field is removed.
- **Document binding invariant:** The JSDoc explicitly documents when the controller is guaranteed:
  - Bound once `context` is assigned.
  - Guaranteed to be present in `render()` and any helper methods called from `render()`.
  - Not guaranteed in `willUpdate()` or `updated()` before the element is bound to a context.
- **Add unit test:** Adds a unit test verifying that tests can assign a controller double directly before mounting.

## Alternatives considered

Adding a protected setter or a test-only helper method was considered, but that would widen the API surface without providing any benefits over a simple property with a documented invariant.

## Verification

- `yarn --cwd renderers/web_core test` passes (524 tests).

## Pre-launch Checklist


renderers/web_core/src/v0_9/catalog/a2ui-lit-element.ts
```diff
@@ -56,14 +56,14 @@ export abstract class A2uiLitElement<Api extends ComponentApi = ComponentApi> ex
    */
   protected readonly api?: Api;
 
-  private _controller?: A2uiController<Api>;
-
   /**
    * The reactive controller instance managing property bindings and state subscriptions.
+   *
+   * Bound when `context` is first set. Because `update()` short-circuits while unbound, this is
+   * guaranteed to be present in `render()`, `updated()`, and any code they reach. It is NOT
```
josemontespg: [Component Lifecycle] The invariant documented here is not true at this commit.

`update()` returning early skips rendering, but Lit calls `updated()` from `_$didUpdate` after `update()` returns, regardless of whether `super.update()` ran. So an unbound element does reach `updated()` with no controller. `Row.updated()` and `Column.updated()` both read `this.controller.props` and throw on a controller-less mount, which contradicts the guarantee stated on this line.

Each layer has to stand on its own, so either narrow the claim here and let a follow-up widen it, or move the gate fix into this PR.

```suggestion
   * Bound when `context` is first set. Because `update()` short-circuits while unbound, this is
   * guaranteed to be present in `render()` and any code it reaches. It is NOT guaranteed inside
   * `willUpdate()` or `updated()` overrides, both of which run outside that gate.
```

renderers/web_core/src/v0_9/catalog/a2ui-lit-element.test.ts
```diff
@@ -248,4 +248,27 @@ describe('A2uiLitElement', () => {
...
+    class TestAssignableElement extends A2uiLitElement<typeof TextApi> {
+      protected override readonly api = TextApi;
+      override render() {
+        return this.controller.props.text;
+      }
+    }
+    customElements.define('test-assignable-element', TestAssignableElement);
+
+    const el = document.createElement('test-assignable-element') as TestAssignableElement;
+    el.controller = {props: {text: 'injected'}} as any;
```
josemontespg: [TypeScript Rigor] `as any` on the injected double.

The point of this test is that the seam is typed, so the cast should go through the real type rather than erase it. `as unknown as A2uiController<typeof TextApi>` keeps the mock honest: if the controller contract changes shape, this line stops compiling instead of silently drifting.

Requires adding `import type {A2uiController} from './a2ui-controller.js';`. There is existing `as any` in this file, so treat this as a nit rather than a blocker.

renderers/web_core/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- **BREAKING CHANGE**: (v0_9) `A2uiLitElement.controller` is a writable property again, reverting the read-only getter introduced in 0.11.0. The getter prevented nothing TypeScript did not already prevent, and it removed the only sanctioned seam for injecting a controller double. Recommended migration: tests that moved to `(el as any)._controller = mock` should assign `controller` directly; the private backing field no longer exists. [#2646](https://github.com/a2ui-project/a2ui/pull/2646)
```
josemontespg, line 3: [PR Hygiene & Scope] The PR description cross-references the sibling PR in this stack. Two places point at #2647: "[#2647], the next PR in this stack, fixes one of those" and "[#2647] depends on it".

`pr-hygiene` §4 and `stack-prs` §1 both require a description to stand on its own with no stack metadata or sibling PR numbers. A reviewer evaluates this PR in isolation and should not need to open another tab to judge whether the change is correct. The references also rot the moment the stack is reordered or a layer is split, and neither statement is load-bearing: the first is an example of a downstream bug that this change does not fix, and the second is a claim about a PR that does not exist yet from this PR's point of view.

The changelog links are a different case and should stay. Those point at the merging PR, which `update-changelog` requires.

renderers/web_core/src/v0_9/catalog/a2ui-lit-element.ts
josemontespg: Narrowed. The gate is in `update()` at this commit, so `updated()` is outside it. #2652 moves the gate to `shouldUpdate()` and widens the claim back.

renderers/web_core/src/v0_9/catalog/a2ui-lit-element.test.ts
josemontespg: Done. Cast goes through `A2uiController<typeof TextApi>` now.

renderers/web_core/CHANGELOG.md
josemontespg: Stripped both references. The description stands alone.

PR #2647, opened 2026-09-14
fix(web_core): derive the active tab index instead of writing it in willUpdate

josemontespg:
## Summary

Updates the `Tabs` component to derive its active tab index dynamically during `render()` rather than mutating `@state()` in `willUpdate()`, and removes defensive optional chaining on `this.controller` across the basic catalog.

## Problem

The `Tabs` component previously used `willUpdate()` to clamp `this.activeIndex` whenever the number of available tabs changed:

```ts
protected override willUpdate() {
  const props = this.controller.props;
  if (this.activeIndex >= props.tabs.length) {
    this.activeIndex = 0;
  }
}
```

This pattern caused several issues:

1. **Unnecessary extra render cycles:** Modifying an internal `@state()` property inside `willUpdate()` requests an additional update pass from Lit.
2. **Destructive state overwrite:** Overwriting `this.activeIndex` permanently erased the user's selection. If a tabs array temporarily shrunk (such as during dynamic data filtering) and later restored its items, the user's tab selection was lost.
3. **Crash before context attachment:** Because `willUpdate()` ran before `this.controller` was instantiated, reading `this.controller.props` threw a `TypeError` if `<a2ui-tabs>` was rendered before being bound to a context.
4. **Misleading optional chaining:** Multiple places in the basic catalog used `this.controller?.props`. Because `this.controller` is guaranteed once an element updates, optional chaining masked programming errors and allowed components to silently render blank UI instead of surfacing an issue.

## Solution

- **Derive active tab index in `render()`:**
  Remove `willUpdate()` entirely from `Tabs.ts`. In `render()`, compute the clamped index as a local derived value:
  ```ts
  const activeIndex = this.activeIndex < props.tabs.length ? this.activeIndex : 0;
  ```
  This eliminates the extra render pass, preserves the stored user selection when tab counts change, and ensures the controller is accessed only when guaranteed to exist.
- **Remove redundant optional chaining:**
  Replace `this.controller?.props` with `this.controller.props` in `BasicCatalogA2uiLitElement` and `ChoicePicker`. A missing controller now surfaces immediately at the access point rather than silently producing empty markup.

## Verification

- Added tests in `Tabs.test.ts`:
  - Verifies that `<a2ui-tabs>` can mount without a context without throwing an error.
  - Verifies that when the tabs list shrinks below `this.activeIndex`, the rendered content falls back to the first tab while preserving the user's selection in `this.activeIndex`.
- `yarn --cwd renderers/web_core test` passes (545 tests).

## Pre-launch Checklist


renderers/lit/README.md
```diff
@@ -114,7 +114,7 @@ export class MyProfileElement extends A2uiLitElement<typeof MyProfileApi> {
   protected readonly api = MyProfileApi;
 
   render() {
-    const props = this.controller?.props;
+    const props = this.controller.props;
```
josemontespg, line 117: [Public API & Changelogs] Missing a `renderers/lit/CHANGELOG.md` entry.

`renderers/lit/src/v0_9/a2ui-lit-element.ts` re-exports `A2uiLitElement` straight from `@a2ui/web_core/v0_9`, so the lifecycle change in this PR is visible to `@a2ui/lit` consumers, and this README edit is the recipe they follow. Changelog entries belong in every package whose published behavior changes, not only the package where the code lives.

Add an entry under `## Unreleased` in `renderers/lit/CHANGELOG.md` covering the gate move and the `this.controller?.` guidance change.

renderers/web_core/src/v0_9/basic_catalog/catalog.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as assert from 'node:assert';
+import {describe, it, before, after} from 'node:test';
+import {setupTestDom, teardownTestDom, asyncUpdate} from '../test/dom-setup.js';
+import {basicCatalog} from './catalog.js';
+
+describe('basicCatalog', () => {
+  before(async () => {
+    setupTestDom();
+    // Importing the catalog registers every component's custom element.
+    await import('./catalog.js');
```
josemontespg: [Testing Rigor] This dynamic import is a no-op, and the comment on it is misleading.

The static `import {basicCatalog} from './catalog.js'` on line 20 has already evaluated the module by the time `before()` runs, so registration happened at module load, not here. Re-importing an evaluated module returns the same instance.

The static import is load-bearing for a different reason worth stating: the harness enumerates `basicCatalog.components` to generate `it()` cases, which has to happen synchronously at describe time, and it only works because `dom-setup.js` is imported first and installs the DOM globals that the Lit class definitions need at evaluation time. That ordering constraint is the thing a future reader will break.

```suggestion
    // `dom-setup.js` must be imported before `./catalog.js` (see the imports above): the DOM
    // globals have to exist when the component modules evaluate their Lit class definitions,
    // which is also when they register their custom elements.
    setupTestDom();
```

renderers/web_core/src/v0_9/basic_catalog/catalog.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  for (const implementation of basicCatalog.components.values()) {
+    it(`should update <${implementation.tagName}> without a context`, async () => {
+      const el = document.createElement(implementation.tagName);
+      document.body.appendChild(el);
+
+      // An element mounted before a context is attached — standalone custom element usage,
+      // static hydration, or a direct test mount — must not throw on its first update.
+      await asyncUpdate(el);
+
+      assert.ok(el.isConnected);
```
josemontespg: [Testing Rigor] `assert.ok(el.isConnected)` restates the line above it rather than checking the invariant.

`appendChild` sets `isConnected`, so this passes even for a tag that is not registered at all, and it would also pass against the old `update()` gate that leaves the cycle pending forever. The invariant this PR establishes is that an unbound element closes its update cycle, so assert that directly:

```suggestion
      // The cycle must be closed, not abandoned: a pending update would mean later
      // `requestUpdate()` calls are dropped and the element never renders once bound.
      assert.strictEqual((el as LitElement).isUpdatePending, false);
```

That also fails loudly on an unregistered tag, where `isUpdatePending` is `undefined`.

renderers/web_core/src/v0_9/catalog/a2ui-lit-element.ts
```diff
@@ -258,10 +259,20 @@ export abstract class A2uiLitElement<Api extends ComponentApi = ComponentApi> ex
...
+   * Gates the update cycle on the element being bound to a controller.
+   *
+   * `willUpdate()` creates the controller from `context` and has not run yet on the cycle that
+   * binds it, so `context` stands in for the controller that is about to exist. Skipping here
+   * rather than inside `update()` lets Lit close the cycle: `updated()` is not reached without a
+   * controller, and later updates are still scheduled once the element is bound.
+   *
+   * @param changedProperties Map of changed properties with their previous values.
+   */
+  protected override shouldUpdate(changedProperties: PropertyValues) {
```
josemontespg, line 276: [PR Hygiene & Scope] Consider splitting the gate move out of the `Tabs` fix.

This PR carries two changes with different blast radii. The `Tabs` derivation is local and low risk. Moving the unbound gate from `update()` to `shouldUpdate()` changes the lifecycle contract for all 18 catalog components and for any external subclass, and it is what makes the `this.controller?.` removals safe. If it needs to be reverted later, it would be cleaner to revert on its own.

A defensible split is: gate move plus the catalog mount harness in one layer, then `Tabs` plus the `?.` cleanups on top. Raising this as a question rather than a request, since the two are genuinely coupled through the invariant and the split adds a stack layer.

One behavioral note to confirm either way: the gate accepts an element with a controller but no `context`, which keeps the injected-double case from the parent PR working. Worth a sentence in the JSDoc so the `&&` is not read as a typo.

renderers/web_core/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- **BREAKING CHANGE**: (v0_9) `A2uiLitElement` now skips the update cycle in `shouldUpdate()` while the element is unbound, instead of returning early from `update()`. The old gate left Lit's cycle pending forever, so an element mounted before its `context` was attached never rendered even once bound, and `updated()` still ran without a controller — which crashed `Row` and `Column`. The controller is now guaranteed in every hook that runs after `super.willUpdate()`. Recommended migration: subclasses that override `shouldUpdate()` must call `super.shouldUpdate()`; subclasses that overrode `update()` to work around the old gate can drop the workaround. [#2647](https://github.com/a2ui-project/a2ui/pull/2647)
```
josemontespg: [PR Hygiene & Scope] The PR description cross-references the sibling PR in this stack: "the JSDoc invariant from [#2646] is updated to say so".

`pr-hygiene` §4 and `stack-prs` §1 both require a description to stand on its own with no stack metadata or sibling PR numbers. A reviewer evaluates this PR in isolation, and the reference breaks if the stack is reordered or a layer is split. The sentence works without it: state that the JSDoc invariant on `controller` is updated to match the new gate, and let the diff show what the previous wording was.

The changelog links are a different case and should stay. Those point at the merging PR, which `update-changelog` requires.

renderers/lit/README.md
josemontespg: Added, on #2652, which is where the gate move and the README recipe now live. This PR no longer touches `renderers/lit`.

renderers/web_core/src/v0_9/basic_catalog/catalog.test.ts
josemontespg: Applied on #2652, which now owns this file. The import ordering is the real constraint, so it is what the comment says.

renderers/web_core/src/v0_9/basic_catalog/catalog.test.ts
josemontespg: Applied on #2652. `isUpdatePending === false` fails against the old `update()` gate, which the `isConnected` check did not.

renderers/web_core/src/v0_9/catalog/a2ui-lit-element.ts
josemontespg: Split. #2652 now holds the gate move, its JSDoc, the catalog mount harness, and the `renderers/lit` docs and changelog. This PR keeps the `Tabs` derivation and the `this.controller?.` removals, and touches `renderers/web_core` only.

The JSDoc on `shouldUpdate()` in #2652 now states that the `controller` check covers an element with an injected controller and no `context`.

renderers/web_core/CHANGELOG.md
josemontespg: Description rewritten for the narrowed scope, without the reference.

PR #2656, opened 2026-09-14
fix(swift): restrict swift-json-schema to 0.13.x minor releases

pinieb:
### Summary

Updates the `swift-json-schema` dependency constraint in `Package.swift` from `from: "0.13.1"` to `.upToNextMinor(from: "0.13.1")`.

### Rationale

`swift-json-schema` 0.14.0 introduced breaking API changes to `JSONValue`:
* The `.integer(Int)` and `.number(Double)` enum cases were replaced by `.numberLiteral(JSONNumberLiteral)`.
* `.integer(_: Int)` and `.number(_: Double)` became static factory methods, causing pattern matching on `.integer(let value)` and `.number(let value)` to fail compilation in `A2UICore`, `BasicCatalog`, and `BasicCatalogSwiftUI`.

Swift Package Manager treats `from: "0.13.1"` as `.upToNextMajor(from: "0.13.1")`, which resolves to `0.14.0` when dependencies are updated. Pinning to `.upToNextMinor(from: "0.13.1")` restricts resolution to `0.13.x` releases (currently `0.13.2`) until the codebase is migrated to the `0.14.0` API.

Note: Apple's guidance on SwiftPM version pinning generally suggests using `upToNextMajor(from:)` so that other packages using the same dependency have the best chance of finding a mutually acceptable version. To accommodate that behavior, Swift packages are supposed to follow semantic versioning where breaking changes are only introduced in major version updates. `upToNextMinor(from:)` is one step more restrictive and intended for packages that introduce breaking changes in minor versions (like what prompted this PR). If a breaking change is introduced to `swift-json-schema` in a patch version, we can pin an exact version of the package.

### Testing

* Cleared the Swift package cache (`swift package purge-cache && swift package reset`).
* Verified package resolution selects `swift-json-schema` `0.13.2`.
* Ran `swift build --build-tests` and `swift test` (214 tests in 23 suites passed).
* Built `A2UISampleClient` via `xcodebuild`.


ditman (approved):
Thanks for pinning the problem! :)

PR #2657, opened 2026-09-14
feat(web_core,a2ui_core): supply v1.0 catalog function bodies and align the two engines

gspencergoog:
## Summary

Supplies missing implementations for the v1.0 basic catalog functions in `web_core`, aligns function semantics and expression depth limits with the Python reference engine, and removes 11 undeclared operators from `a2ui_core` so both engines strictly match the published specification.

### `web_core`
- **v1.0 Basic Catalog Implementations**: Added bodies for 9 functions (`and`, `or`, `not`, `formatString`, `formatNumber`, `formatCurrency`, `formatDate`, `pluralize`, `openUrl`) in `basic_functions.ts`, exported alongside the existing 5 validation functions.
- **Deterministic Formatting**: Defaulted `formatNumber`, `formatCurrency`, and `pluralize` to `en-US`. `formatDate` resolves localized month and weekday names through `Intl` for the configured locale, parsing naive timestamps as UTC.
- **System Functions & Data Binding**: `@index` now raises `A2uiValidationError` when evaluated outside a collection template. `createSurface` assigns its `dataModel` directly at the root, preserving literal keys that contain `/` or `~`.

### `a2ui_core` & Parity
- **Non-Spec Operator Removal**: Removed 11 undeclared arithmetic and comparison operators (`add`, `subtract`, `multiply`, `divide`, etc.) from Python v0.9 and v1.0 basic catalogs and deleted unused `operator_apis.py`.
- **Expression Depth Limits**: Aligned maximum expression recursion depth to 100 across Python, Swift, and TypeScript, counting function arguments as nested depth levels.
- **Python Formatting Alignment**: `formatDate` ISO formatting now converts to UTC with 3 millisecond digits (`.000Z`). `pluralize` falls back on missing keys rather than falsy values.

### Conformance
- Removed operator cases from `conformance/core/functions.yaml` and added expression depth boundary checks to `conformance/core/expressions.yaml`.
- Regenerated `converted_basic_catalog_v09.json` and `converted_basic_catalog_v10.json`.

## Impact & Risks

- **Breaking (Python)**: Calling or importing the 11 removed operators (`AddApi`, `DivideApi`, etc.) in `a2ui.core.basic_catalog` now raises an `ImportError` or unknown-function error. These operators are not declared in the published specification and were never supported in `web_core`.
- **Breaking (TypeScript)**: Evaluating `@index` outside a template now throws `A2uiValidationError` rather than returning 0.

## Testing

1. Run TypeScript unit tests and linter:
   ```bash
   yarn workspace @a2ui/web_core test
   yarn workspace @a2ui/web_core lint
   ```
2. Run TypeScript conformance harness:
   ```bash
   node typescript/web_core/tests/conformance/conformance_test.mjs
   ```
3. Run Python unit tests and type checks:
   ```bash
   uv run pytest
   uv run --all-packages mypy .
   ```


typescript/web_core/src/v1_0/basic_catalog/functions/basic_functions.ts
```diff
@@ -0,0 +1,485 @@
...
+/**
+ * Creates the number formatting function implementation for a specific locale.
+ *
+ * @param locale Optional BCP 47 language tag. Defaults to `en-US`.
+ * @returns The function implementation.
+ */
+export function createFormatNumberImplementation(locale?: string): FunctionImplementation {
+  const resolvedLocale = locale ?? DEFAULT_LOCALE;
+  return createFunctionImplementation(FormatNumberApi, args => {
+    if (isNaN(args.value)) return '';
```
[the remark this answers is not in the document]
gspencergoog: I checked this against the invocation path, and `null` cannot reach `isNaN` here.

Arguments are resolved before they are validated, and validated before the body runs. `DataContext.resolveDynamicValue` resolves each argument (`rendering/data-context.ts:121-123`), then `Catalog.invoker` runs `fn.schema.parse(rawArgs)` before calling `execute` (`catalog/types.ts:346`). `value` is `DynamicNumberSchema`, a union of `z.number()`, `DataBindingSchema` and `FunctionCallSchema`, so a resolved `null` fails the union.

Measured through the invoker:

| input | result |
| --- | --- |
| `{value: null}` | throws `A2uiExpressionError: Validation failed for function 'formatNumber'` |
| `{}` | throws, `value` is required |
| `{value: 'abc'}` | throws |

The parity argument does not hold as stated either. Python does not reach the `ValueError` in `_to_float`: pydantic rejects `None` against `FormatNumberArgs` first, with the same four-way union failure. Both engines already reject a null at the schema layer, for the same reason, so the guard is not what keeps them in agreement.

One path does produce `"0"`: calling `FormatNumberImplementation.execute({value: null})` directly, bypassing the catalog. That works because Python's `create_function_implementation` validates inside `execute` while the TypeScript wrapper only casts. If that gap is worth closing, the fix belongs in `createFunctionImplementation` once, rather than as a `typeof` check in each of the fourteen function bodies.

Leaving the guard as it is.

typescript/web_core/src/v1_0/basic_catalog/functions/basic_functions.ts
```diff
@@ -0,0 +1,485 @@
...
+ * Creates the currency formatting function implementation for a specific
+ * locale.
+ *
+ * @param locale Optional BCP 47 language tag. Defaults to `en-US`.
+ * @returns The function implementation.
+ */
+export function createFormatCurrencyImplementation(locale?: string): FunctionImplementation {
+  const resolvedLocale = locale ?? DEFAULT_LOCALE;
+  return createFunctionImplementation(FormatCurrencyApi, args => {
+    if (isNaN(args.value)) return '';
```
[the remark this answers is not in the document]
gspencergoog: I checked this against the invocation path, and `null` cannot reach `isNaN` here.

Arguments are resolved before they are validated, and validated before the body runs. `DataContext.resolveDynamicValue` resolves each argument (`rendering/data-context.ts:121-123`), then `Catalog.invoker` runs `fn.schema.parse(rawArgs)` before calling `execute` (`catalog/types.ts:346`). `value` is `DynamicNumberSchema`, a union of `z.number()`, `DataBindingSchema` and `FunctionCallSchema`, so a resolved `null` fails the union.

Measured through the invoker:

| input | result |
| --- | --- |
| `{value: null}` | throws `A2uiExpressionError: Validation failed for function 'formatCurrency'` |
| `{}` | throws, `value` is required |
| `{value: 'abc'}` | throws |

The parity argument does not hold as stated either. Python does not reach the `ValueError` in `_to_float`: pydantic rejects `None` against `FormatCurrencyArgs` first, with the same four-way union failure. Both engines already reject a null at the schema layer, for the same reason, so the guard is not what keeps them in agreement.

One path does produce `"0"`: calling `FormatCurrencyImplementation.execute({value: null})` directly, bypassing the catalog. That works because Python's `create_function_implementation` validates inside `execute` while the TypeScript wrapper only casts. If that gap is worth closing, the fix belongs in `createFunctionImplementation` once, rather than as a `typeof` check in each of the fourteen function bodies.

Leaving the guard as it is.

typescript/web_core/src/processing/message-processor.ts
```diff
@@ -619,10 +619,15 @@ export class MessageProcessor<T extends ComponentApi = ComponentApi> {
...
-      }
+      // Assign the whole object at the root rather than building a pointer per
+      // key. Key names are literal property names, so a key containing '/' or
+      // '~' would otherwise be misread as a nested path.
+      this.processUpdateDataModelOp({
+        type: 'updateDataModel',
+        surfaceId,
+        path: '/',
+        value: dataModel,
+      });
```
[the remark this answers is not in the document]
gspencergoog: Applied in a2589026a.

`processUpdateDataModelOp` re-resolves the surface from the model map and then calls `surface.dataModel.set(path, value)`. Here `surface` is constructed and added two lines above, so the lookup cannot fail and the guard cannot fire, making the direct call behaviour-identical.

python/a2ui_core/src/a2ui/core/basic_catalog/operator_apis.py
nan-yu, line 1: https://github.com/a2ui-project/a2ui/blob/v1_0/typescript/web_core/src/v1_0/basic_catalog/functions/basic_functions_api.ts and https://github.com/a2ui-project/a2ui/blob/v1_0/typescript/web_core/src/v0_9/basic_catalog/functions/basic_functions_api.ts define these functions in the basic catalogs, which are not included in the specification.

Does Python node need to match the TypeScript version?

typescript/web_core/src/v1_0/basic_catalog/functions/basic_functions.ts
nan-yu, line 1: These basic functions don't seem to change across versions. Should we extract it to a shared package?

gspencergoog: We can, and probably should, although I think I'd like to still have stubs in the versioned directories that re-expoert the common implementations, so that when they diverge they can easily be overridden.

python/a2ui_core/src/a2ui/core/basic_catalog/operator_apis.py
gspencergoog: > [v1_0/typescript/web_core/src/v1_0/basic_catalog/functions/basic_functions_api.ts](https://github.com/a2ui-project/a2ui/blob/v1_0/typescript/web_core/src/v1_0/basic_catalog/functions/basic_functions_api.ts?rgh-link-date=2026-09-15T23%3A31%3A55Z) and 
> [v1_0/typescript/web_core/src/v0_9/basic_catalog/functions/basic_functions_api.ts](https://github.com/a2ui-project/a2ui/blob/v1_0/typescript/web_core/src/v0_9/basic_catalog/functions/basic_functions_api.ts?rgh-link-date=2026-09-15T23%3A31%3A55Z) define these functions in the basic catalogs, which are not included in the specification.

No, the TypeScript code doesn't define the expression functions, other than `and`, `or`, and `not`.  It doesn't define the ones in this file: `add`, `subtract`, `multiply`, `divide`, `equals`, `greaterThan`, `lessThan`, `contains`, `startsWith`, and `endsWith`, which aren't defined in the spec or the basic catalog.  Nothing uses these implementations.

> Does Python node need to match the TypeScript version?

Sure, I think it would be ideal if they were similar at least.

typescript/web_core/src/v1_0/basic_catalog/functions/basic_functions.ts
gspencergoog: I separated them out into a common file and just left some shims.  There are some differences in the implementations (e.g. 0.9 validation functions return bool, not ValidationResult).

python/a2ui_core/src/a2ui/core/basic_catalog/operator_apis.py
nan-yu: Gotcha. I was looking at the legacy TypeScript code on the main branch, https://github.com/a2ui-project/a2ui/blob/main/renderers/web_core/src/v0_9/basic_catalog/functions/basic_functions_api.ts.

nan-yu (approved):

PR #2669, opened 2026-09-15
[dart] feat(a2ui_core): add the node-resolution layer

andrewkolos:
# feat(a2ui_core): add the node-resolution layer

The Dart core half of [#1282](https://github.com/a2ui-project/a2ui/issues/1282), following the TypeScript node layer ([#2077](https://github.com/a2ui-project/a2ui/pull/2077)) and its renderer adoption ([#2393](https://github.com/a2ui-project/a2ui/pull/2393)).

## What it adds

`NodeResolver(surface)` turns a surface's flat component model into a reactive tree of read-only `ComponentNode`s. It:

- Resolves child references and expands `ChildList` templates in each item's data scope.
- Publishes dynamic values as `ResolvedBinding` snapshots and actions as callable closures.
- Represents missing components, unknown types, and cyclic references as placeholders.

A node's `props` emits when its own properties change, not when a descendant's properties change.

Initial surface payloads reject unresolved references under `ValidationConfig.strict`. Later updates can leave references pending until the child arrives. Component and batch validation still apply.

For smoke testing from Flutter, the genui renderer adapted to this layer lives on [node-layer-flutter](https://github.com/andrewkolos/A2UI/tree/node-layer-flutter) as `dart/a2ui_flutter` with a `simple_chat` example; it will be a separate PR.

## Breaking changes

Three changes to 0.1.1 API, each needed for the node layer. The [changelog](https://github.com/a2ui-project/a2ui/blob/fd0cc70c56701d4f6f56d3efae3aef1391e37a28/dart/a2ui_core/CHANGELOG.md?plain=1#L3-L148) has the full 0.2.0 notes; its other breaking entries are already on `main`.

- **`GenericBinder` publishes `ResolvedBinding` values instead of raw values plus generated `set<Property>` setters.** A generic renderer needs to know, per property, the current value and whether it can be written back, without knowing the component's schema. 0.1.1 split that across a raw value and a setter named by convention, so a renderer had to guess which properties were path-bound. A binding carries the value, the write path and `set` together. Callers that read a raw value now read `.value`; callers of `setValue(...)` call `binding.set(...)`.
- **`SurfaceModel.dispatchAction` emits only `event` payloads; `functionCall` payloads run when a binder action closure is invoked.** In 0.1.1 the closure resolved the payload, which invoked the function, then `dispatchAction` parsed the resolved value as a `FunctionCall` and invoked it again at root scope, which failed on the cast. A catalog function now runs once, in the scope of the node that declared it. Listeners on `onAction` see only events, the payloads meant for the agent.
- **With an error reporter, a missing or throwing catalog function resolves to null and reports `EXPRESSION_ERROR` instead of throwing.** Node trees rebuild reactively, so an exception from one binding would abort the whole surface. The blueprint's [rule](https://github.com/a2ui-project/a2ui/blob/0086493c40b119a4143fe15197006678467cad60/blueprints/modules/a2ui_core.blueprint.md?plain=1#L1280) is that expression errors are dispatched, not thrown. `ComponentContext` supplies the reporter by default; a bare `DataContext` without one still throws, so direct callers keep the exception.

<details>
<summary>Dart differences from TypeScript (compared at 1c45c809)</summary>

- Dart creates bindings in `GenericBinder`, which already classifies property schemas; TypeScript creates them in the resolver.
- Dart copies nested lists and maps and makes them unmodifiable. This prevents data-model writes from changing previously emitted values or hiding updates. TypeScript's [data-model copy is shallow](https://github.com/a2ui-project/a2ui/blob/1c45c809b655878d06e3afc6dda22100afecc0a4/renderers/web_core/src/v0_9/state/data-model.ts#L342-L352).
- Bindings use Dart's `==` for custom objects, preserving their value equality; TypeScript compares them by identity.
- Literal maps are resolved recursively, per the blueprint's [rule](https://github.com/a2ui-project/a2ui/blob/0086493c40b119a4143fe15197006678467cad60/blueprints/modules/a2ui_core.blueprint.md?plain=1#L1278); TypeScript's `resolveDynamicValue` does not recurse into plain objects.

</details>

## Tests

- `dart analyze --fatal-infos` and `dart format` are clean.
- `a2ui_core`: 513 passed, 25 skipped (the pre-existing v0.8 conformance skips). `a2ui_agent` still passes.

## Pre-launch Checklist

One time (author confirmations):

- [x] I signed the [CLA].

For this PR:

- [ ] If my branch is on a fork, I have verified that [scripts/e2e_test.sh](https://github.com/a2ui-project/a2ui/blob/1c45c809b655878d06e3afc6dda22100afecc0a4/scripts/e2e_test.sh) passes. Not run: it exercises the restaurant_finder Flutter sample with a Gemini key.

[Contributors Guide]: https://github.com/a2ui-project/a2ui/blob/1c45c809b655878d06e3afc6dda22100afecc0a4/CONTRIBUTING.md
[Style Guide]: https://github.com/a2ui-project/a2ui/blob/1c45c809b655878d06e3afc6dda22100afecc0a4/CONTRIBUTING.md#coding-style
[sample]: https://github.com/a2ui-project/a2ui/blob/1c45c809b655878d06e3afc6dda22100afecc0a4/samples/README.md#samples


polina-c:
Please, add central blueprint to this PR for 'node-resolution layer' in blueprints/features - It will help to:

 - iterate on the design, initilly specified in https://github.com/a2ui-project/a2ui/issues/1282
 - review this pull request

jacobsimionato (commented):
I still need to review the public APIs here, but just a quick initial comment:

Either in this PR, or in a followup, can we dedupe the tests here to be language agnostic conformance tests, with expected inputs/outputs represented in YAML?

Seeing as this is a critical and finicky piece of logic, I think it would be fine to create a new subfolder under conformance/core/node etc, with a separate YAML schema if necessary. Happy for this to be in a followup PR! Just want to avoid replicating these same tests across 3+ languages, and ensure we get the behavior consistent.

jacobsimionato (approved):
LGTM! The public APIs look great to me, and the tests are very complete, though let make them conformance tests in due course.

The exact ergonomics will be more clear when there's a basic catalog implementation, so let's merge this ASAP (let's see if Polina wants to take a pass first?) and then you can always refine while you're building out the basic catalog and Flutter framework adapter?

jacobsimionato:
BTW, now that we have this beautiful Node API, let's make as much other stuff private as possible. E.g. let's make the generic binder private now.

polina-c:
> LGTM! The public APIs look great to me, and the tests are very complete, though let make them conformance tests in due course.
> 
> The exact ergonomics will be more clear when there's a basic catalog implementation, so let's merge this ASAP (let's see if Polina wants to take a pass first?) and then you can always refine while you're building out the basic catalog and Flutter framework adapter?

I would love to see blueprint while reviewing. Feel free to merge this PR without my review, if you both think blueprint is optional and does not worth it.

polina-c:
This is some agent generated notes. Feel free to address in other PRs if they make sense.

## Summary

Everything the PR description claims checks out locally: `dart analyze --fatal-infos` reports no issues, `dart format --set-exit-if-changed` reports 65 files and 0 changed, `a2ui_core` is 513 passed / 25 skipped, and `a2ui_agent` passes.

The implementation is careful and the tests are dense: 141 tests and roughly 630 assertions across the new files, with real coverage of the hard parts. Reentrancy during diagnostic dispatch, placeholder upgrade and downgrade, parent-scoped ownership of shared children, and the unmodifiable-snapshot contract are all exercised. The comments explain why rather than what, which is what made the reentrancy logic reviewable at all.

One blocking issue, five worth addressing, four minor, two questions. Most of them sit at edges the tests do not reach.

## Bug

### One malformed binding blanks the surface and leaks nodes

> TL;DR: `_materialize` has no `try`/`finally`, so a cast error in child N orphans children 1..N-1 with live binders and leaves `rootNode` permanently null.

[`node_resolver.dart#L673-L811`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/node_resolver.dart#L673-L811)

`_materialize` fills `newEdges` incrementally in [`resolveChild`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/node_resolver.dart#L684-L702) and commits it to `record.childEdges` only after the whole loop, at [L796](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/node_resolver.dart#L796). Nothing guards the loop. When `_childNode` throws on the Nth child, children 1..N-1 are already in `_records` and `_nodesByEdge` with live `GenericBinder`s and data-model subscriptions, and no parent owns them: the `catch` in [`_createNode`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/node_resolver.dart#L506-L514) disposes only the node whose binder failed, and disposing the parent walks `record.childEdges`, which still holds the old map.

Reproduced on this branch with a `Column` whose `children` are `['ok', 'bad']`, where `bad` is `{"text": {"path": 42}}`:

```
thrown: type 'int' is not a subtype of type 'String' in type cast
activeNodeCount: 1
rootNode: null
```

That is two failures, not one:

* The leaked `ok` node keeps a subscribed binder alive until `resolver.dispose()`.
* `_rootRecord` stays null and `rootNode` stays null. Nothing re-arms `_buildRoot`, so the surface is blank until a new `root` create or delete event arrives. The `TypeError` also escapes `SurfaceComponentsModel.addComponent` into the `MessageProcessor` caller.

The second is exactly what [a2ui_core.blueprint.md#L1280](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/blueprints/modules/a2ui_core.blueprint.md?plain=1#L1280) forbids: "Throwing out of the resolution pass aborts the whole tree, so one malformed binding blanks an otherwise renderable surface." This PR fixes the catalog-function half of that rule. Malformed payload shapes that reach a cast still take the old path.

Payload validation catches most such payloads, but `ValidationConfig` is configurable and the resolver is reachable without `MessageProcessor`. The conformance fixtures in `node_resolver_test.dart` write to `componentsModel` directly, so the resolver should not assume validation ran.

Smallest fix: wrap the child loop in a `try`, and on the way out dispose any node in `newEdges` that did not make it into `record.childEdges`. Dispatching a diagnostic for the offending property instead of rethrowing would follow the blueprint rule more closely.

TypeScript has the same gap in [`node-resolver.ts#L525-L550`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/renderers/web_core/src/v0_9/nodes/node-resolver.ts#L525-L550). Fixing both together is reasonable; landing the Dart side with neither a fix nor a tracking issue is not.

## Worth addressing

### An action listener can mutate the live component model

> TL;DR: `resolveSync` stopped copying, so `A2uiClientAction.context` now aliases `componentModel.properties` below the first level. This is a regression against main.

[`contexts.dart#L86-L99`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/core/contexts.dart#L86-L99)

The `if (!_containsDynamicValue(value)) return value;` short-circuits are a good optimization and the doc comment states the change. The consequence is not local, though. The binder's action closure passes the result straight through: [`binder.dart#L197-L209`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/rendering/binder.dart#L197-L209) shallow-copies with `Map<String, dynamic>.from(resolved)`, then [`surface_model.dart#L55-L67`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/core/surface_model.dart#L55-L67) shallow-copies again into `A2uiClientAction.context`. Everything below the first level is the live component-model map.

Reproduced: an `onAction` listener writing `action.context['nested']['k']` changed `componentModel.properties['action']['event']['context']['nested']['k']`. On main, `resolveSync` always rebuilt maps and lists, so the action payload was detached.

Node props go to real trouble to publish recursively unmodifiable snapshots. The action payload is the structure that actually leaves the process, so it is the odd one to leave aliased. Three ways out: keep the copy on the action path only, snapshot at the `dispatchAction` boundary, or document the aliasing so embedders know not to mutate what they receive.

The PR description notes that TypeScript's `resolveDynamicValue` does not recurse into plain objects either, so this matches TS. But [blueprint L1278](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/blueprints/modules/a2ui_core.blueprint.md?plain=1#L1278) treats the recursing Dart behavior as the correct one, which makes Dart the reference implementation here rather than the follower.

### `dispatchAction` with a `functionCall` payload is now a silent no-op

> TL;DR: the 0.1.1 way of invoking a function through public API now produces no action, no error and no log.

[`surface_model.dart#L51-L70`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/core/surface_model.dart#L51-L70)

The rationale in the PR description is convincing and the binder path is clearly better. But `SurfaceModel.dispatchAction` is public API, and an embedder calling it directly gets nothing back:

```
await surface.dispatchAction({'functionCall': {'call': 'nope', 'args': {}}}, 'root');
// actions=0 errors=0
```

A dispatched `A2uiClientError`, or at minimum a logged warning since `logging` is being added anyway, turns a silent migration failure into a visible one. The comment at L68-L69 explains the decision to the next reader of the source, not to the caller at runtime.

### `extractRefFields` walks the full schema once per node

> TL;DR: a maxed `ChildList` does 10,000 schema walks that a resolver-scoped cache would eliminate safely.

[`ref_fields.dart#L22-L28`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/ref_fields.dart#L22-L28), called from [`node_resolver.dart#L475-L478`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/node_resolver.dart#L475-L478)

The doc comment explains why memoizing by schema identity globally is unsafe: "local aliases can resolve differently when the same schema is used in different catalog documents." That is right, and it is a real hazard that TypeScript's [`refFieldsCache` WeakMap](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/renderers/web_core/src/v0_9/nodes/ref-fields.ts#L54) does not guard against.

But the hazard is only about sharing a cache across documents. `_surface.catalog.catalogSchema` is fixed for a resolver's lifetime, so a cache keyed by schema identity and scoped to the resolver or the catalog is safe. Today a `ChildList` expanding to 10,000 items does 10,000 full schema walks on top of 10,000 `GenericBinder` constructions, each of which runs its own walk in `_scrapeSchemaBehavior`. Worth measuring before the Flutter adoption PR lands on top of it.

### The unresolved-reference warning only reaches a logger

> TL;DR: a payload diagnostic is logged instead of dispatched, is discarded when no `Logger.root` listener is attached, and is the sole reason for a new dependency.

[`node_resolver.dart#L816-L842`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/node_resolver.dart#L816-L842), drained at [L269-L275](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/node_resolver.dart#L269-L275)

Two questions.

First, this is the reason for the new [`logging: ^1.3.0`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/pubspec.yaml#L28) dependency, the first in `a2ui_core`. Every other diagnostic in the package goes through `dispatchError` as a structured `A2uiClientError`. A payload that references components the resolver will not mount looks like exactly that kind of diagnostic, and a renderer would want to show it.

Second, TypeScript emits nothing here, so this is Dart-only. If the warning is worth having, it is worth a spec'd error code and a matching TS change. If it is only a development aid, `dispatchError` still beats a new dependency.

### `structuralChildLists` makes validation and resolution disagree about one schema

> TL;DR: an unmarked `{componentId, path}` schema is a child list to the resolver and not a reference to the validator, so strict mode never checks those references.

[`reference_schema.dart#L60-L68`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/primitives/reference_schema.dart#L60-L68) and [`component_refs.dart#L70-L75`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/validation/component_refs.dart#L70-L75)

Sharing one reader between validation and resolution is a clear improvement over the two duplicated walkers it replaces. The flag creates a gap, though: for a catalog that declares a child list structurally, strict-mode reference checking never sees those references, so the PR's "initial surface payloads reject unresolved references under `ValidationConfig.strict`" guarantee holds everywhere except where the catalog author forgot the marker.

The reasoning in the doc comment, that validation rejects a whole batch and so keeps to what the catalog states, is sound as far as it goes. What is missing is what a catalog author should expect: a structurally-declared child list renders but is not reference-checked, and adding the `$ref` or `REF:` marker is what turns checking on.

## Minor

* [`binder.dart#L510-L525`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/rendering/binder.dart#L510-L525). TL;DR: reads as copy-paste. The `Map<String, Object?>` and `Map` branches of `_snapshotBindingValue` have identical bodies and differ only in the static type they produce. One line saying the typed branch exists to preserve the key type would settle it.

* [`binder.dart#L494-L505`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/rendering/binder.dart#L494-L505). TL;DR: a value that survives a snapshot can throw on write-back. `_mutableCopy` does `entry.key as String` while `_snapshotBindingValue` does not cast. Unreachable from JSON payloads, and easy to make symmetric.

* [`binder.dart#L245-L249`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/rendering/binder.dart#L245-L249). TL;DR: static id arrays allocate past the cap. They build one `ChildNode` per entry uncapped, while the template path caps at [L226-L228](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/rendering/binder.dart#L226-L228). The resolver caps both at [L720-L722](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/node_resolver.dart#L720-L722), so nothing over-mounts, but the full list is allocated first.

* [`component_refs.dart#L83-L87`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/validation/component_refs.dart#L83-L87). TL;DR: a deliberate divergence reads like a bug. `if (field.value case NestedRef(:final fields))` shadows the outer `fields`, and the inner `fields.keys` is intentionally all sub-fields rather than `NestedRef.keys`, which is `SingleRef`-only and is what [`_materialize`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/resolution/node_resolver.dart#L737-L766) iterates. The divergence is correct, since validation should check nested `ChildList` sub-properties that the resolver leaves as descriptors. Renaming the pattern variable and adding a comment saves the next reader the trip.

## Questions

### Package placement against the normative layout

> TL;DR: a new file lands in `rendering/`, the directory the blueprint names as a deviation.

[Blueprint L200](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/blueprints/modules/a2ui_core.blueprint.md?plain=1#L200) says: "Package names are normative. The validation package is `validation` and the resolution package is `resolution`. Naming them `validating` or `rendering` is a deviation, since `rendering` in particular suggests UI work that this layer does not do." The layout at [L174-L180](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/blueprints/modules/a2ui_core.blueprint.md?plain=1#L174-L180) puts `component_node`, `component_context`, `node_resolver`, `generic_binder` and `data_context` together under `resolution/`.

This PR puts `component_node.dart`, `node_resolver.dart` and `ref_fields.dart` in `resolution/`, which is right. It then adds the new [`resolved_binding.dart`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/rendering/resolved_binding.dart) to the pre-existing `rendering/`, leaves `binder.dart` there, keeps `contexts.dart` under `core/`, and puts the new `reference_schema.dart` under `primitives/`.

Moving `binder.dart` and `contexts.dart` is out of scope here. Putting a brand new file into the directory the blueprint calls a deviation is a choice worth stating. Either move `resolved_binding.dart` to `resolution/` next to its only consumers, or say in the PR description that the `rendering/` to `resolution/` move is deferred, so the audit does not read it as fresh drift.

### `checks` name matching

> TL;DR: the blueprint names this exact line as a violation, and this PR rewrites the function containing it.

[`binder.dart#L413`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/dart/a2ui_core/lib/src/rendering/binder.dart#L413) still opens with `if (propertyName == 'checks')`. [Blueprint L1290](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/blueprints/modules/a2ui_core.blueprint.md?plain=1#L1290) calls that a catalog-agnostic violation under `AGENTS.md` §8. It is pre-existing and fixing it is its own change, but `_scrapeSchemaBehavior` is rewritten wholesale here, so it is fair to ask whether this was the moment, or to link the tracking issue so a reader of the diff knows it is known.

## Checked, no problem found

Recorded so the next reviewer does not repeat it:

* Diagnostic deduplication and the reservation and release protocol across `_reportOnce`, `_discardPendingErrors` and `_flushErrors`, including reentrant dispatch. `_dispatchingErrors` correctly defers the inner flush to the outer drain.
* Dispose during update. `_disposeRequested` is honoured at the outer `_runUpdate` boundary and pending diagnostics are dropped, matching the class doc.
* Component type changes. `ComponentModel.type` is final and `addComponent` throws on a duplicate id, so a type change is always delete-then-create, which `_onComponentDeleted` and `_onComponentCreated` handle.
* `_stabilize` reference-identity preservation against the shallow comparison in `MutableComponentNode.setProps`. The invariant holds, including for a list that changes length.
* `_escapeIdPart` covers every character used as a separator in composed instance ids.
* Dart's `Behavior.object` rebuilds `result` from `value.entries`, so a removed property drops its key. The Dart resolver therefore does not need the `delete next[key]` reconciliation TypeScript does at [`node-resolver.ts#L529-L537`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/renderers/web_core/src/v0_9/nodes/node-resolver.ts#L529-L537). The omission is deliberate, not missed.
* The `_actionClosures` cache key (`path.join('/')`) and its lack of eviction match TypeScript's [`generic-binder.ts#L306-L324`](https://github.com/a2ui-project/a2ui/blob/b3909ffd1dd8b19619a443e0fc413e7fe09290b0/renderers/web_core/src/v0_9/rendering/generic-binder.ts#L306-L324) exactly, so no Dart-side comment.
* Nested `ChildList` values do reach `_queueUnresolvedReferenceWarnings` in both the template and static-array forms, so the documented limitation is observable rather than silent.

PR #2693, opened 2026-09-18
refactor(spec): move the v1.0 basic catalog to the top-level catalogs/ directory

josemontespg:
## Summary

Catalogs are versioned independently of the A2UI protocol, so their canonical schemas now live under a top-level `catalogs/` directory instead of inside `specification/<version>/`. This PR moves the v1.0 basic catalog to `catalogs/v1_0/basic/` (it keeps its version directory to avoid breaking existing consumers) and updates everything that read it from the old path.

Basic catalogs for protocol versions before v1.0 stay under `specification/<version>/catalogs/`. New catalogs are authored directly in flat, unversioned directories (`catalogs/<catalog>/catalog.json`), as described in the new `catalogs/README.md`.

## Changes

- Move `specification/v1_0/catalogs/basic/` (catalog and examples) to `catalogs/v1_0/basic/` and add `catalogs/README.md` describing the layout.
- Point the spec validator (`specification/scripts/validate.py`), the v1.0 spec test runner, the `web_core` `copy-spec` script, the conformance suite, the Python tests, the eval scripts and the docs at the new path.
- Publish `catalogs/v1_0` under `docs/public/specification/v1_0/catalogs/` in the docs workflow so the catalog `$id` URL (`https://a2ui.org/specification/v1_0/catalogs/basic/catalog.json`) keeps resolving.
- Trigger the spec, web and Python CI workflows on `catalogs/**` changes.

The catalog `$id` and `catalogId` values are unchanged; only the file location moved.

## Verification

- `python3 specification/scripts/validate.py`: v0.8, v0.9 and v1.0 pass.
- `specification/v1_0/test/run_tests.py`: 167 passed, 0 failed.
- `yarn workspace @a2ui/web_core build`: `copy-spec` picks the catalog up from `catalogs/v1_0/`.
- Prettier and pyink checks pass on the modified files.

## Pre-launch Checklist

- [x] My code changes have tests (existing spec and conformance suites cover the moved catalog).


eval/CONTRIBUTING_USE_CASES.md
```diff
@@ -67,7 +67,7 @@ Every data point is a YAML object inside a list in `eval/datasets/<dataset_name>
 
 - `name` (required, string): Unique identifier for the sample (lowercase with underscores).
 - `description` (required, string): A concise summary of what the scenario tests.
-- `catalog` (required, string): Relative path to the component catalog. Use `'specification/{version}/catalogs/basic/catalog.json'` for standard components.
+- `catalog` (required, string): Relative path to the component catalog. Use `'catalogs/v1_0/basic/catalog.json'` for standard v1.0 components; older versions live at `'specification/{version}/catalogs/basic/catalog.json'`.
```
josemontespg, line 70: Is this true? Why are we keeping older catalogs under specification?

specification/v1_0/test/run_tests.py
```diff
@@ -40,14 +43,16 @@
...
+    the specified file, resolved relative to the test directory) with the $id
+    modified to match what agent_to_renderer.json expects.
     """
-    basic_catalog_path = os.path.join(SPEC_DIR, catalog_file)
-    if not os.path.exists(basic_catalog_path):
+    if catalog_file is None:
+        catalog_file = BASIC_CATALOG_FILE
+        basic_catalog_path = BASIC_CATALOG_FILE
+    else:
         basic_catalog_path = os.path.join(TEST_DIR, catalog_file)
```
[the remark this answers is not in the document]
josemontespg: Done. Suites now resolve `catalog` against `test/` first and the repository root second, matching the two-location lookup that existed before the move.

eval/CONTRIBUTING_USE_CASES.md
josemontespg: Yes — this PR only moves v1.0. v0.9 and v0.9.1 are frozen, and both the Python (`BundledCatalogProvider.get_catalog_id`) and Kotlin (`BasicCatalogProvider.kt`) SDKs derive the catalog `$id` from the repo-relative path, so moving them means decoupling ID derivation first (~40 more files across Python/Kotlin/Dart/Swift/TS). Tracking that in #2700 rather than growing this PR.

catalogs/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+catalogs/
+├── v1_0/
+│   └── basic/          # Basic catalog for protocol v1.0 (with examples/)
+│       └── catalog.json
+└── mcp/                # MCP catalog
+```
+
+- `v1_0/basic/` keeps its protocol-version directory because existing consumers
+  depend on that path. Do not add new versioned directories here.
```
jacobsimionato: But why do existing customers require the path to have a `v1.0` in it, even though we are able to move it from one place to another? The catalog should only be identified by the catalog ID, which is independent of the path. We have chosen to host the catalog definition at the URI which we use for it's path (even though the spec doesn't require this!), but we could continue to do that with any new path by using some kind of build step or mapping in the doc site right?

I would actually vote to still put this under a v1 folder, because for v1, we have a different catalog format to v0.9, so perhaps it makes sense for this to be at:

`catalogs/basic/v1/`

Ideally, we will make it so that this same catalog can also be used for v1.1, v1.2 etc.

catalogs/README.md
jacobsimionato, line 1: Can we move https://github.com/a2ui-project/a2ui/blob/main/specification/v1_0/docs/basic_catalog_implementation_guide.md to be alongside this file?

jacobsimionato (commented):
Some extra thoughts:

The direction I'm hoping will happen is that starting from v1.0, we figure out how to evolve the API with minor updates in v1.1, v1.2 etc, which have the property that a v1.x catalog can be used with the v1.y protocol, as long as x <= y. We will then likely prefer to keep the basic catalog expressed in the v1.0 format, to ensure broad compatibility across minor versions. However, we may want to evolve the basic catalog independently, with its own versioning scheme. This may require us to have paths with some combination of protocol version and catalog version in them, or else invent a catalog format which allows multiple versions to be expressed in one file (e.g. `@available_from_version(1.4)` annotation on APIs etc). But I think we don't have enough clarity on this yet, so very happy to at least break the catalog definition out of the spec for now.

catalogs/README.md
josemontespg: Agreed on all points — nothing depends on the repo path, only on the `$id`, which `docs.yml` keeps serving from the old URL. Moved to `catalogs/basic/v1/` (catalog-first, catalog-format major version) and reworded the README accordingly. `$id` is unchanged.

catalogs/README.md
josemontespg: Done — moved to `catalogs/basic/v1/basic_catalog_implementation_guide.md` next to the catalog it documents, and repointed the docs-site include so the published page is unchanged.

josemontespg:
Thanks @jacobsimionato for the review. Ready for another pass when you can

jacobsimionato (approved):
Awesome!

PR #2698, opened 2026-09-18
refactor(web_core)!: register basic catalog elements on render instead of on import

josemontespg:
## Overview

Basic catalog components no longer call `customElements.define` as a side effect of being imported. Registration moves to render time: `renderA2uiNode` (in `@a2ui/web_core` and in the Lit renderer's own copy) calls `registerUniversalElement` for the implementation it is about to render.

## Key changes

- **`@a2ui/web_core/v0_9/basic_catalog`**
  - **BREAKING CHANGE**: the 18 component modules drop `@customElement(...)`. The element classes and `BasicCatalogA2uiLitElement` are module-private; only the `WebComponentImplementation` constants (`A2uiCard`, `A2uiButton`, ...) are exported from `components/index.ts`.
  - Code that relied on the import alone to define the elements must call `registerUniversalElement` (added in #2596).
- **`@a2ui/web_core/v0_9/universal`**
  - `renderA2uiNode` registers web component implementations (`isWebComponentImplementation` → `registerUniversalElement`) before emitting the tag.
- **`@a2ui/lit`**
  - `renderers/lit/src/v0_9/surface/render-a2ui-node.ts` keeps its own `renderA2uiNode` and performs the same on-demand registration, so `A2uiSurface` is unchanged.
- **Tests**
  - The 18 component suites register the basic catalog explicitly in `before` instead of relying on import side effects, and drop the now-dead dynamic imports.
  - `render-a2ui-node.test.ts` asserts the element is defined on first render.


renderers/web_core/src/v0_9/basic_catalog/components/index.ts
```diff
@@ -35,23 +35,21 @@ export {
...
-export {A2uiLitModal, A2uiModal} from './Modal.js';
-export {A2uiBasicRowElement, A2uiRow} from './Row.js';
-export {A2uiSliderElement, A2uiSlider} from './Slider.js';
-export {A2uiLitTabs, A2uiTabs} from './Tabs.js';
-export {A2uiBasicTextElement, A2uiText} from './Text.js';
-export {A2uiBasicTextFieldElement, A2uiTextField} from './TextField.js';
-export {A2uiVideoElement, A2uiVideo} from './Video.js';
+export {A2uiAudioPlayer} from './AudioPlayer.js';
+export {A2uiButton} from './Button.js';
+export {A2uiCard} from './Card.js';
```
ditman, line 40: Awesome!

ditman (approved):
We lose some of the magic of Lit, but... this is much better! Have you verified that the demo app still runs as expected after this change?

josemontespg:
Thanks for the review! Yes, both the Lit explorer app and restaurant sample apps are working as expected, and all tests are passing.

